PR #10, opened 2020-12-18
Import ocw-to-hugo output into database

### LINKED ISSUE #8: Create functions/tasks to import ocw-to-hugo output into the database
Import the output of ocw2hugo into the database (`Website` and `WebsiteContent` models)

Assumption: ocw2hugo output will be available in an S3 bucket

mbertrand:

Closes #8

#### What's this PR do?

Adds a management command, plus related celery tasks, to import ocw-to-hugo output from S3 into database models.

I added some s3 & celery related helper functions into `main/utils.py` on the assumption that they may be useful for something other than ocw2hugo import in the future.

There is a [code block](https://github.com/mitodl/ocw-studio/pull/10/files#diff-99dc8a77ec10391867c362180e6db632fe1e43cfcab696949ad1de969bff5540R60-R76) in this PR that deals with missing uids in some pages, this can be removed after https://github.com/mitodl/ocw-to-hugo/pull/155 merges and the S3 bucket is updated with the latest output.

Ask me for AWS keys and the S3 bucket name. Then run the following:
```
docker-compose run web python manage.py backpopulate_ocw_courses -b <bucket_name>
```
It should complete after about 45 minutes.


app.json
```diff
@@ -95,6 +95,10 @@
       "description": null,
       "required": false
     },
+    "MAX_S3_GET_ITERATIONS": {
+      "description": null,
```
rhysyngsun, line 99: Could you give this a description? We'll need them anyway once we upgrade to the newer config stuff

websites/api.py
```diff
@@ -0,0 +1,121 @@
...
+                    content_type = CONTENT_TYPE_FILE
+                if parent_uuid:
+                    parent, _ = WebsiteContent.objects.get_or_create(
+                        website=website, uuid=parent_uuid
+                    )
+                if not uuid:
+                    # This code block is a temporary hack until every hugo2ocw output file has a uuid
+                    WebsiteContent.objects.update_or_create(
+                        website=website,
+                        hugo_filepath=obj["Key"].replace(prefix, ""),
```
rhysyngsun, line 63: Pull out `obj["Key"].replace(prefix, "")` into a variable and     use it here and in the `else`

websites/api.py
```diff
@@ -0,0 +1,121 @@
...
+                                s3_content_parts[1]
+                                if len(s3_content_parts) == 2
+                                else None
+                            ),
+                            "uuid": uuid4(),
+                            "parent": parent,
+                            "title": content_json.get("title"),
+                            "type": content_type,
+                        },
+                    )
```
rhysyngsun, line 76: Define `defaults` once as a var and use it here and in the `else`

websites/tasks.py
```diff
@@ -0,0 +1,68 @@
...
+from django.conf import settings
+
+from main.celery import app
+from main.utils import chunks
+from websites.api import import_ocw2hugo_course
+
+log = logging.getLogger(__name__)
+
+
+@app.task(bind=True)
```
rhysyngsun, line 15: `self` isn't used so this doesn't need `bind=True`, which will also let you remove `# pylint:disable=unused-argument`

websites/tasks_test.py
```diff
@@ -0,0 +1,47 @@
...
+from websites.tasks import import_ocw2hugo_course_paths, import_ocw2hugo_courses
+
+
+@pytest.mark.parametrize(
+    "paths", [["1-050-mechanical-engineering", "3-34-transportation-systems"], [], None]
+)
+def test_import_ocw2hugo_course_paths(mocker, paths):
+    """ mock_import_course should be called from task with correct kwargs """
+    mock_import_course = mocker.patch("websites.tasks.import_ocw2hugo_course")
+    import_ocw2hugo_course_paths(paths, MOCK_BUCKET_NAME, TEST_OCW2HUGO_PREFIX)
```
rhysyngsun, line 15: This and the other task calls should use `delay()` to ensure we're calling the tasks via celery

websites/tasks.py
```diff
@@ -0,0 +1,68 @@
...
+        prefix (str): S3 prefix before start of course_id path
+        chunk_size (int): Number of courses to process per task
+    """
+    if not bucket:
+        raise TypeError("Bucket name must be specified")
+    s3 = boto3.resource(
+        "s3",
+        aws_access_key_id=settings.AWS_ACCESS_KEY_ID,
+        aws_secret_access_key=settings.AWS_SECRET_ACCESS_KEY,
+    )
```
rhysyngsun, line 51: Seeing similar code to this in a few places, it'd probably be good to pull this out into a utility method. [`boto3.session.Session`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html#boto3.session.Session) would probably be useful here too.

rhysyngsun (changes requested):
First pass on code review

main/utils.py
```diff
@@ -36,3 +39,53 @@ def webpack_dev_server_url(request):
     return "http://{}:{}".format(
         webpack_dev_server_host(request), settings.WEBPACK_DEV_SERVER_PORT
     )
+
+
+def now_in_utc():
```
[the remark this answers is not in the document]
mbertrand: `from mitol-django-common` requires `Django<3.0.0,>=2.2.12` but django is already at 3.1 in this repo's requirements.txt file

```
There are incompatible versions in the resolved dependencies:
  django==3.1 from https://files.pythonhosted.org/packages/2b/5a/4bd5624546912082a1bd2709d0edc0685f5c7827a278d806a20cf6adea28/Django-3.1-py3-none-any.whl#sha256=1a63f5bb6ff4d7c42f62a519edc2adbb37f9b78068a5a862beff858b68e3dc8b (from -r requirements.in (line 1))
  Django<3.0.0,>=2.2.12 (from mitol-django-common==0.3.0->-r requirements.in (line 12))
  Django>=2.2 (from django-redis==4.12.1->-r requirements.in (line 5))
```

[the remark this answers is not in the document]
rhysyngsun: Hmmm, I can look into fixing that, I don't think there's anything in that repo that will break

websites/api.py
mbertrand: The defaults are slightly different in each code block (uuid, hugo_filepath), and once the hack becomes unnecessary there will be only usage, but I'll create a `base_defaults` variable with the common attributes and update as necessary for each block.

rhysyngsun: ok, yeah, i missed that

main/utils.py
```diff
@@ -36,3 +39,53 @@ def webpack_dev_server_url(request):
...
+    Returns:
+        bytes: The contents of a json file read from S3
+    """
+    try:
+        return obj.get()["Body"].read()
+    except Exception:  # pylint: disable=broad-except
+        if iteration < settings.MAX_S3_GET_ITERATIONS:
+            return get_s3_object_and_read(obj, iteration + 1)
+        else:
+            raise
```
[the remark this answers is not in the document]
Ferdi: Does it make sense to use something like https://github.com/invl/retry ?

mbertrand:
@rhysyngsun comments addressed (except adding mitol-django-common due to the dependency mismatch).

rhysyngsun (approved):

PR #31, opened 2021-01-21
Fix black formatting check

noisecapella:
None

#### What's this PR do?
CI hasn't been running black formatting checks correctly. It used `black .` which did the reformatting rather than `black --check .` which errors when the formatting is not correct. I also made a quick fix to the pull request template.

Tests should pass


mbertrand (approved):
👍

PR #36, opened 2021-01-29
Add mitol-django-common

### LINKED ISSUE #9: Utilize mitol-django-common
We should pull in https://github.com/mitodl/ol-django/tree/main/mitol-django-common (https://pypi.org/project/mitol-django-common/) so we can start using the common data models and utilities from there.

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
- [x] Settings

Closes #9 

#### What's this PR do?
- Add `isort` check to github actions + ran it on the code
- Incorporates `mitol-django-common`
- Removes a number of functionality that is now in that library:
  - `main.envs`
  - `main.templatetags.render_bundle`
  - `main.utils`
- Ran `./manage.py generate_app_json`, which means everything in `app.json` gets a description now.

The app should start and you should be able to login to django-admin


main/apps.py
```diff
@@ -1,26 +1,18 @@
...
-            raise ImproperlyConfigured(
-                "The following settings are missing: {}".format(
-                    ", ".join(missing_settings)
-                )
-            )
+    required_settings: List[str] = [
+        "MAILGUN_KEY",
+        "MAILGUN_SENDER_DOMAIN",
+        "SECRET_KEY",
+    ]
```
mbertrand, line 18: I removed MAILGUN_* from my .env and no warning/errors showed up, should they?

mbertrand (commented):
Looks good, 1 question about required settings

main/apps.py
rhysyngsun: They should but the don't appear to be - this also seems to duplicate the behavior of `envs.validate()`, so i just removed this and put in a call that that instead. I'll need to figure out later why I seemingly built in 2 similar mechanisms into that library...

rhysyngsun:
@mbertrand all set finally

rhysyngsun:
Rebased, but i had to squash to do so

ocw_import/tasks.py
```diff
@@ -2,6 +2,7 @@
 import logging
 
 import celery
+from mitol.common.utils.collections import chunks
 
 from main.celery import app
 from main.utils import chunks
```
mbertrand, line 8: Need to remove this import

mbertrand (changes requested):
Works great, just one stray import

mbertrand (approved):
👍

Issue #37, opened 2021-02-01
Support local site config


gsidebo:
Our course authoring experience is going to be driven by configuration. In production, we intend to have any number of "starter" Github projects that authors can select when they add a new site. Those starter projects will contain a config which defines the types of pages for that new site and the fields that will be supported in those pages. In order to make development easier, we should allow developers to define that config locally.

**Notes:**
- The final structure of this config is still TBD as of the time this is being written. For now, we can loosely base the config off of the [Netlify CMS config examples](https://www.netlifycms.org/docs/collection-types/). `collections` items will represent different types of pages that can be created for the site, and `fields` will define the types of inputs that an author can use to build that page.
  - In the initial set of wireframes, [this screen](https://lucid.app/lucidchart/6be33dd8-f7db-47f8-8af8-338255f23c3c/edit?shared=true&page=ksBT.K4_qGrz#?folder_id=home&browser=icon) shows "Resources" and "Pages" in the left nav under "Content". In that example, the corresponding config would define "Resources" and "Pages" as items in the `collections` list.
  - For now, we can assume that we want to support at least these field types: `markdown`, `string`/`text`
- It probably makes sense to implement this with yaml file(s) in the `localdev` directory. One approach would be to add a subdirectory called something like `starters`. Every yaml file within that directory would be considered a separate site config which could then be chosen in the eventual "Add Site/Course" UI
- We just need site configs for now, but eventually this could be built out to define full starter projects in the local repo, hence the suggestion of `starters` for the subdirectory above, instead of just `siteconfigs` et al
- It would be helpful to make sure that updating local site configs triggers a server restart, and that updated configs are registered right away in the API results (see below). It would be unfortunate if changes to local configs required a restart of a docker container
- We don't want to make a request to Github or the filesystem every time we want the contents of the site config. To get around that, we can add a model that stores the config at the time that the site is created, and contains fields that reference the config source (`local` or `github`), and the Github repo and commit (which would be blank if the config was sourced from the local repo), and the contents of the config. The `Website` model can be updated to include a foreign key to one of these config records.

**Acceptance criteria:**
- Developers can define site/course configs locally, i.e.: the config can be fetched from some local file/setting/database record, as opposed to fetching it from Github
- Django setting added which allows developers to flip the switch between local site configs and Github-based site configs. This can be a feature flag.
[...]

PR #41, opened 2021-02-02
Added WebsiteStarter model with local development support

gsidebo:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [ ] Code is tested

Partially addresses #37 

#### What's this PR do?
- Adds the `WebsiteStarter` model to (a) track the available starter projects that an author can use to create their site and (b) store their site config
- Allows developers to define starter projects in the local filesystem (instead of fetching them from Github) and load them into the database via management command

Run the `populate_local_starters` management command and check out the Django admin class

The structure of the site config is still TBD, so the contents of `localdev/starters/example-starter/ocw-studio.yml` are likely to change in the future. This is just a rough draft.


localdev/starters/example-starter/ocw-studio.yml
```diff
@@ -0,0 +1,6 @@
+collections:
+  - label: "Page"
+    name: "page"
+    fields:
+      - {label: "Title", name: "title", widget: "string"}
+      - {label: "Content", name: "content", widget: "markdown"}
```
gumaerc, line 6: We should remove this line as every markdown file is assumed to have markdown formatted content, and every page will be represented by a markdown file.

localdev/management/commands/populate_local_starters.py
```diff
@@ -0,0 +1,67 @@
...
+from django.conf import settings
+from django.core.management import BaseCommand
+
+from websites.constants import WEBSITE_CONFIG_FILENAME, STARTER_SOURCE_LOCAL
+from websites.models import WebsiteStarter
+
+LOCAL_STARTERS_DIR = "localdev/starters/"
+
+
+class Command(BaseCommand):
```
gumaerc, line 14: Correct me if I'm wrong because I'm not super familiar with our Python development practices, but should we have some unit tests for this file that try and initialize the example config and read some properties of it?  `backpopulate_ocw_courses.py` doesn't have any tests either, maybe it's not necessary?

gumaerc (changes requested):
Just a few questions.  I fired it up locally and was able to import the starter and access it in the shell, so that part works.

localdev/management/commands/populate_local_starters.py
gsidebo: We rarely add unit tests to management commands. That has bitten us once or twice but in most cases I don't see the benefit outweighing the added work

gumaerc: That makes sense

gumaerc (approved):
:+1:

PR #43, opened 2021-02-02
get frontend stuff up to speed

### LINKED ISSUE #4: Install typescript
As a developer, I'd like to install and use typescript for type checking because JS static analysis has been useful in other projects, and flow does not look viable for the long term

#### Acceptance Criteria:

 - [ ] typescript is installed
 - [ ] type check is run on travis
 - [ ] there is an option to run a type check locally via npm

#### Out of scope of this issue
Type checking existing JS code. At this point there isn't much of that.

### LINKED ISSUE #42: set up Jest for frontend JS tests
It's quite a bit nicer and more 'batteries included' that the mocha+chai setup we've been using for a few years. We're using Jest over in course-search-utils and it's quite nice.

aliceriot:

closes #42
closes #4 

#### What's this PR do?

does some cleanup and updating of our frontend stuff, adding jest, typescript, cleaning out a few old things, etc. should give us a clean base to build on.

I think if the build passes and everything looks legit we're good.


scripts/test/js_test.sh
```diff
@@ -31,8 +28,8 @@ if [[ ! -z "$2" ]]; then
     CMD_ARGS+=" -g \"$2\""
 fi
 
-echo "Running: $CMD $CMD_ARGS"
-eval "$CMD $CMD_ARGS" 2> >(tee "$TMP_FILE")
+echo "Running: $CMD"
+eval "$CMD" 2> >(tee "$TMP_FILE")
```
gsidebo, line 32: Is it not possible to target test cases by name with Jest? That's an unfortunate drawback. If that's true, we should also get rid of the section above that parses the extra command args (line 17-29) since we're not doing anything with it

gsidebo (changes requested):

scripts/test/js_test.sh
aliceriot: oh probably it is - I'll look into that. I also sort of eviscerated the `js_test.sh` file here and forgot about coverage too - I'll see about making that work with it

static/js/index.tsx
```diff
@@ -1,14 +1,16 @@
...
-import configureStore from "../store/configureStore"
-import Router, { routes } from "../Router"
+import configureStore from "./store/configureStore"
+import Router, { RootProps } from "./Router"
 
 import * as Sentry from "@sentry/browser"
-// Object.entries polyfill
-import entries from "object.entries"
+
+// polyfill for ES5 support :/
```
noisecapella, line 11: Do we need this? It seems like modern browsers support ES6, though I don't know if that means every part https://caniuse.com/?search=es6

static/js/Router.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { ReactElement } from "react"
```
noisecapella, line 1: What's the convention for using `tsx` vs `ts`? I see both used in this PR

aliceriot: afaik `.tsx` tells typescript to use the JSX parser, and `.ts` is just the normal one. If you change a file with JSX in it to `.tsx` you'll get all kinds of syntax errors. If we wanted to standardize on one thing I suppose we could just use `.tsx` everywhere (there may also be a compiler option to always use the JSX mode regardless of the file extension)

static/js/index.tsx
aliceriot: maybe it is finally time - we talked about dropping support for IE11. I guess for this project support for old browsers is even less important, since people are going to be using it on new computers at work.

aliceriot: I'll go ahead and remove the polyfills, we can add them back in the future if need be

static/js/Router.tsx
noisecapella: I'm fine with both, just curious when to use each one. It makes sense to leave it as is

scripts/test/js_test.sh
aliceriot: ok, so a question for you about this. I changed it over to run jest directly in `package.json` scripts, rather than using this `js_test.sh` script because it was causing the test output to be printed twice.

so running `npm test` will run `jest`, `npm run test:watch` will run `jest --watch`, and so on.

with this setup, you can run a test by matching on its name by doing `npm run -- -t "test name"`. we only have one test right now :joy: but I tried it and it seems to work to me.

does that work for your use-case for specifying test names?

aliceriot: also adjusted the coverage setup and I think got that working.

aliceriot:
@noisecapella I made a few tweaks to the codecov setup to get it working with Jest - would you mind taking a look and seeing if it's wired up correctly?

aliceriot:
also removed the polyfills and made some adjustments to the JS testing setup.

scripts/test/js_test.sh
gsidebo: @aliceriot That sounds great. Could you update the README to include these details? (the new command for running tests, "watching", running test cases by description, etc.)

aliceriot: definitely!

aliceriot:
@noisecapella @gsidebo ok I think everything should be addressed now, just updated the README as well.

noisecapella:
I'm seeing this when I run `docker-compose up`: 

    watch_1   | ERROR in ./static/js/index.tsx
    watch_1   | Module not found: Error: Can't resolve '../Router' in '/src/static/js'

aliceriot:
ahh I fixed the top level import but forgot about the thing at the bottom for the hot-module-reloading. I think it should be good now.

noisecapella (approved):

PR #44, opened 2021-02-02
Added website starter API endpoints and feature flag

### LINKED ISSUE #37: Support local site config

gsidebo:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
- [x] Settings

Closes #37

#### What's this PR do?
- Adds list and detail endpoints for fetching `WebsiteStarter` records (see #41)
- Adds a feature flag which can optionally include `WebsiteStarter` records that were loaded from the local fs

- Set `FEATURES_USE_LOCAL_STARTERS=True` in your .env
- Run the management command to load the local starter project (`manage.py populate_local_starters`)
- Run a shell and make an API call to fetch the serialized starter info:
    ```python
    from django.test.client import Client
    client = Client()
    client.force_login(user)
    resp = client.get("/api/starters/", follow=True)
    print(resp.json())
    ```

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -6,3 +6,27 @@ import pytest
 def default_settings(settings):
     """Set default settings for all tests"""
     settings.DISABLE_WEBPACK_LOADER_STATS = True
+
+
+def pytest_addoption(parser):
+    """Pytest hook that adds command line parameters"""
+    parser.addoption(
+        "--simple",
+        action="store_true",
+        help="Run tests only (no cov, warning output silenced)",
+    )
+
+
+def pytest_configure(config):
+    """Pytest hook to perform some initial configuration"""
+    if getattr(config.option, "simple") is True:
+        # NOTE: These plugins are already configured by the time the pytest_cmdline_main hook is run, so we can't
+        #       simply add/alter the command line options in that hook. This hook is being used to
+        #       reconfigure/unregister plugins that we can't change via the pytest_cmdline_main hook.
+        # Switch off coverage plugin
+        cov = config.pluginmanager.get_plugin("_cov")
+        cov.options.no_cov = True
+        # Remove warnings plugin to suppress warnings
+        if config.pluginmanager.has_plugin("warnings"):
+            warnings_plugin = config.pluginmanager.get_plugin("warnings")
+            config.pluginmanager.unregister(warnings_plugin)
--- a/websites/views.py
+++ b/websites/views.py
@@ -2,10 +2,16 @@
 from rest_framework import viewsets, mixins
 from rest_framework.pagination import LimitOffsetPagination
 
+from main import features
 from main.permissions import ReadonlyPermission
 from main.utils import now_in_utc
-from websites.models import Website
-from websites.serializers import WebsiteSerializer
+from websites.constants import STARTER_SOURCE_GITHUB
+from websites.models import Website, WebsiteStarter
+from websites.serializers import (
+    WebsiteSerializer,
+    WebsiteStarterSerializer,
+    WebsiteStarterDetailSerializer,
+)
 
 
 class DefaultPagination(LimitOffsetPagination):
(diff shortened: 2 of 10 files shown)


conftest.py
```diff
@@ -6,3 +6,34 @@
...
     """Set default settings for all tests"""
     settings.DISABLE_WEBPACK_LOADER_STATS = True
+
+
+def pytest_addoption(parser):
+    """Pytest hook that adds command line parameters"""
+    parser.addoption(
+        "--simple",
+        action="store_true",
+        help="Run tests only (no cov, no pylint, warning output silenced)",
```
noisecapella, line 16: There isn't pytest-pylint anymore FYI, we are just running pylint directly

websites/views.py
```diff
@@ -40,3 +46,27 @@ def get_queryset(self):
...
+class WebsiteStarterViewSet(
+    mixins.ListModelMixin,
+    mixins.RetrieveModelMixin,
+    viewsets.GenericViewSet,
+):
+    """
+    Viewset for WebsiteStarters
+    """
+
+    permission_classes = (ReadonlyPermission,)
```
noisecapella, line 60: We don't have pagination classes here like we do for `WebsiteViewSet`. Is that intentional? I can see reasons for and against so no preference here

noisecapella (changes requested):

websites/views.py
gsidebo: Just an oversight. I'll add

gsidebo:
@noisecapella Ready for review again

noisecapella (approved):

PR #49, opened 2021-02-04
Website CRUD permissions

### LINKED ISSUE #33: Permissions for site creation and editing
- Only Django users of certain group should be able to create `Website`s of a given type (ie `course`)
- A user can be a member of 1 or more `Website`s , with a different role/privilege on each. For example, a user can be an “admin” on site1 but just an “editor” on site2

Implement using Django's built-in permission system, Rest framework permissions, django-guardian object-level permissions as necessary.

Screen Shot 2021-02-05 at 2 57 20 PM

mbertrand:
#### Pre-Flight checklist
- [x] Migrations
- [x] Testing
  - [x] Code is tested

Closes #33 

#### What's this PR do?
- Sets up groups and django-guardian object-specific permissions to determine who should be able to CRUD websites and website content, publish websites, and assign collaborators (admins, editors).

- Migrate, then run `manage.py backpopulate_groups` if you've already imported/created some websites.
Tests should pass.  There are no views yet that use these permissions but you could experiment with them in a shell following the unit tests as examples.  If you go to the django admin edit page for.a Website, there will be a "Object Permissions" button near the top right.  Clicking it should show you this:

Screen Shot 2021-02-05 at 3 25 44 PM

- For creating and deleting WebsiteContent, I made the assumption that we will use [nested routes/views](https://chibisov.github.io/drf-extensions/docs/#nested-routes) provided by drf-extensions - this is how a user's course list items are handled in open-discussions.

gandalf


websites/permissions.py
```diff
@@ -0,0 +1,163 @@
...
+    Args:
+        group_or_user (Group or User): the group/user to add permissions to
+        website (Website): The website that the permissions are for
+        perms (list of str): The permissions to be assigned
+    """
+    for perm in perms:
+        try:
+            assign_perm(perm, group_or_user, website)
+        except Permission.DoesNotExist as err:
+            raise Permission.DoesNotExist(f"Perm {perm} not found") from err
```
noisecapella, line 25: Is the other exception message not useful?

websites/permissions.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+class HasWebsitePermission(BasePermission):
+    """Permission to view/modify/create Website objects"""
+
+    def has_permission(self, request, view):
+        if request.method == "POST":
+            # Only global editors and admins can create new Websites
+            return request.user.has_perm(constants.PERMISSION_ADD)
+        return True
```
noisecapella, line 109: I think we need to return true if `SAFE_METHODS` and return false otherwise, in case there is a `PATCH` or `DELETE` method for the view. Does that make sense? I'm not too familiar with guardian so I might be missing something

noisecapella (changes requested):

websites/permissions.py
mbertrand: The default message doesn't indicate what the permission is:
`Permission matching query does not exist.` vs `Perm fakePermission not found`

websites/permissions.py
mbertrand: I thought `PATCH` and `DELETE` would always end up going through the `has_object_permission` function below, because `has_permission` is used for viewing lists and creating (`POST`) new objects.  But I'm not 100% sure of that, and it wouldn't hurt to add the `SAFE_METHODS` conditional.

websites/permissions.py
```diff
@@ -0,0 +1,165 @@
...
+
+@transaction.atomic
+def create_global_groups():
+    """Create the global groups for website permissions"""
+    admin_group, _ = Group.objects.get_or_create(name=constants.GLOBAL_ADMIN)
+    author_group, _ = Group.objects.get_or_create(name=constants.GLOBAL_AUTHOR)
+    for perm in constants.PERMISSIONS_ADMIN:
+        assign_perm(perm, admin_group)
+    assign_perm(constants.PERMISSION_ADD, admin_group)
+    assign_perm(constants.PERMISSION_ADD, author_group)
```
gsidebo, line 54: General comment about this function, `create_website_groups`, and the management command: I think it would be helpful to return the number of groups created (maybe created _and_ updated) in these functions, and you could probably add some logic to exit out of them if the groups already exist with the right permissions. The management command could then report on the number of groups that were actually affected

gsidebo (changes requested):

websites/permissions.py
```diff
@@ -0,0 +1,165 @@
...
+        website (Website): The website that the permissions are for
+    """
+    admin_group, _ = Group.objects.get_or_create(
+        name=f"{constants.ADMIN_GROUP}{website.uuid.hex}"
+    )
+    editor_group, _ = Group.objects.get_or_create(
+        name=f"{constants.EDITOR_GROUP}{website.uuid.hex}"
+    )
+    assign_object_permissions(admin_group, website, constants.PERMISSIONS_ADMIN)
+    assign_object_permissions(editor_group, website, constants.PERMISSIONS_EDITOR)
```
gsidebo, line 43: It seems like this method should be doing what the signal does: check if the website has an owner and add the permissions for that user if there is one. If we run that backpopulate management command, I think we would want to ensure that any existing websites with an owner have the correct permissions assigned for that owner.

gsidebo (changes requested):

websites/permissions.py
mbertrand: I went ahead and did this, but personally I feel the usefulness it provides is marginal, especially considering the extra code complexity it introduces and the expected low frequency (probably just once per environment) that this command will run.

noisecapella (approved):

websites/permissions.py
```diff
@@ -0,0 +1,245 @@
...
+    elif isinstance(group_or_user, Group):
+        current_perms = {perm.codename for perm in group_or_user.permissions.all()}
+    elif isinstance(group_or_user, User):
+        current_perms = {
+            perm.codename for perm in Permission.objects.filter(user=group_or_user)
+        }
+    else:
+        raise TypeError("Permissions must be assigned to a user or group")
+    if not {perm.replace("websites.", "") for perm in perms}.difference(current_perms):
+        return False
```
gsidebo, line 36: This will help a little with readability I think

```suggestion
    if website:
        current_perm_codenames = set(get_perms(group_or_user, website))
    elif isinstance(group_or_user, Group):
        current_perm_codenames = {perm.codename for perm in group_or_user.permissions.all()}
    elif isinstance(group_or_user, User):
        current_perm_codenames = {
            perm.codename for perm in Permission.objects.filter(user=group_or_user)
        }
    else:
        raise TypeError("Permissions must be assigned to a user or group")
    added_perm_codenames = {perm.replace("websites.", "") for perm in perms}
    if not added_perm_codenames.difference(current_perm_codenames):
        return False
```

websites/permissions.py
```diff
@@ -0,0 +1,245 @@
...
+    )
+    if admin_created:
+        groups_created += 1
+    author_group, author_created = Group.objects.get_or_create(
+        name=constants.GLOBAL_AUTHOR
+    )
+    if author_created:
+        groups_created += 1
+
+    global_perms = [constants.PERMISSION_ADD]
```
gsidebo, line 123: What's the significance of this? Should it be represented in constants.py? We have `PERMISSIONS_EDITOR` and `PERMISSIONS_ADMIN`. Should we also have `PERMISSIONS_AUTHOR` or something like that?

websites/models.py
```diff
@@ -41,20 +42,46 @@ class Website(TimestampedModel):
...
+
+    class Meta:
+        permissions = (
+            ("publish_website", "Publish or unpublish a website"),
+            ("preview_website", "Create preview markdowm"),
+            (
+                "add_collaborators_website",
+                "Add or remove collaborators (admins, editors, etc)",
+            ),
+            ("edit_content_website", "edit website content"),
```
gsidebo, line 74: Nit of all nits:

```suggestion
            ("edit_content_website", "Edit website content"),
```

websites/management/commands/backpopulate_groups.py
```diff
@@ -0,0 +1,36 @@
...
+        created, updated = create_global_groups()
+        self.stdout.write(
+            f"Global groups: created {created} groups, updated {updated} groups"
+        )
+
+        for website in Website.objects.iterator():
+            created, updated, owner_updated = create_website_groups(website)
+            self.stdout.write(
+                f"{website.name} groups: created {created}, updated {updated}, owner updated: {str(owner_updated)}"
+            )
```
gsidebo, line 29: If the entire set of OCW course sites has been imported into the database and this command is run, this output is going to be written that many times. That's a potentially huge number. I think it would be better to keep running totals then report them below in the last write.
```suggestion
        website_groups_created, website_groups_updated, owners_updated = 0, 0, 0  
        for website in Website.objects.iterator():
            created, updated, owner_updated = create_website_groups(website)
            website_groups_created += created
            website_groups_updated += updated
            owners_updated += 1 if owner_updated else 0
```

websites/management/commands/backpopulate_groups.py
```diff
@@ -0,0 +1,36 @@
+""" Backpopulate website groups and permissions"""
+from django.core.management import BaseCommand
+
+from main.utils import now_in_utc
+from websites.models import Website
+from websites.permissions import create_global_groups, create_website_groups
+
+
+class Command(BaseCommand):
```
gsidebo, line 9: Related to my comment below: there can be a potentially huge amount of sites in the db. We should add optional parameters to this command to target specific websites based on `name` or `title`.

gsidebo (changes requested):

websites/permissions.py
mbertrand: This is the only permission a global author has - creating new websites (but then will have all the admin permissions of any site created, as the owner of that site).  I didn't create a constant for it because I felt it would be a little redundant, since it's just another constant in a list (`constants.PERMISSION_ADD`)

websites/management/commands/backpopulate_groups.py
mbertrand: If all that's reported is a list of created/updated owners and groups at then end, and it doesn't seem to match up with the number of websites, then you know something went wrong somewhere but no idea where.   So seeing even less value for this.  If something really goes wrong there should be an exception raised.  How about adding a `--verbose` option, so that if the numbers don't seem to match up at the end, you can run it again and see the counts per site to help narrow down where the problem might be occurring?

websites/management/commands/backpopulate_groups.py
mbertrand: Okay, though the intended purpose of this command is to create permission groups for all existing websites, or conceivably to add new groups/permissions to all websites if we add any in the future.  I was originally going to do it as a migration script but didn't want a migration to run that long so chose to go this route.

websites/management/commands/backpopulate_groups.py
```diff
@@ -0,0 +1,70 @@
...
+                )
+                total_websites += 1
+                total_created += created
+                total_updated += updated
+                total_owners += 1 if owner_updated else 0
+
+                if is_verbose:
+                    self.stdout.write(
+                        f"{website.name} groups: created {created}, updated {updated}, owner updated: {str(owner_updated)}"
+                    )
```
gsidebo, line 60: Pretty sure we can just change the queryset before iterating instead of using the `if` statement

```suggestion
        if filter_str:
            website_qset = Website.objects.filter(Q(name__icontains=filter_str) | Q(title__icontains=filter_str))
        else:
            website_qset = Website.objects.all()
        for website in website_qset.iterator():
                created, updated, owner_updated = setup_website_groups_permissions(
                    website
                )
                total_websites += 1
                total_created += created
                total_updated += updated
                total_owners += 1 if owner_updated else 0

                if is_verbose:
                    self.stdout.write(
                        f"{website.name} groups: created {created}, updated {updated}, owner updated: {str(owner_updated)}"
                    )
```

gsidebo (changes requested):
One more suggestion

gsidebo (approved):
LGTM! 👍

PR #50, opened 2021-02-04
Decoupled 'websites' app from import code

### LINKED ISSUE #47: Decouple 'websites' Django app from OCW course site import
The `websites` Django app should be general enough that it supports all workflows for creating and updating websites and content. There are a few ways that it is tightly coupled to imported OCW course sites:

1. The `url_path` field. This is essentially just an OCW course id, and we only use it as a key to make sure we don't import course sites more than once. We can change this field to something more generally applicable, since `url_path` is not relevant to newly-authored sites.
2. The `type` field is not relevant to newly-authored sites.
3. OCW course site importing code exists in the `websites` app, but it's not relevant to newly-authored sites.

**Acceptance criteria:**
- `url_path` field renamed to something like `name`, set to nullable, not unique, and indexed
- `type` field removed
- Migrations work forward and backward
- OCW import-specific API functionality, tasks, commands, etc. moved into separate Django app (something like `ocw_import`)

gsidebo:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Closes #47

#### What's this PR do?
- Moves OCW import code to its own Django app
- Renames `Website.url_path` to `name`
- Adds a data migration to create a `WebsiteStarter` record for our OCW starter repo
- Removes `Website.type`, opting instead to get the website's de facto "type" from the associated `WebsiteStarter.slug`
- Adds the `Website.source` field to identify which sites were imported and which ones were authored in the app
- Adds a bunch of data migrations to ensure that imported Websites are handled correctly and can still be imported

- Run the OCW import command to test that it still works (ex: `backpopulate_ocw_courses -b <bucket> --filter "esd-7`). It would be best to delete a `Website` and re-import it.
- Check through imported `Website` objects in Django admin to make sure the data looks correct
- Make sure that requests to `/api/websites?type=course` are unchanged


websites/factories.py
```diff
@@ -16,39 +16,37 @@
...
+
 class WebsiteFactory(DjangoModelFactory):
     """Factory for Website"""
 
-    title = factory.Sequence(lambda n: "OCW Course %s" % n)
-    url_path = factory.Sequence(lambda n: "/ocw_site_x/%s" % n)
-    type = FuzzyChoice((WEBSITE_TYPE_COURSE, "other"))
-
+    title = factory.Sequence(lambda n: "Site %s" % n)
+    name = factory.Sequence(lambda n: "/site_x/%s" % n)
```
noisecapella, line 37: Just curious, what does the `x` mean in `site_x`?

gsidebo: No meaning. I'll fix

noisecapella (approved):

Issue #40, opened 2021-02-01
Basic site detail UI


gsidebo:
[Example wireframe](https://lucid.app/lucidchart/6be33dd8-f7db-47f8-8af8-338255f23c3c/edit?shared=true&page=ksBT.K4_qGrz#?folder_id=home&browser=icon)

This will be the initial implementation of the site/course authoring detail page, i.e.: the top-level page that an author will see when they create a new site or select a site to edit.

**Acceptance criteria:**
- [ ] API endpoint added that returns serialized site data
  - Note: It might make sense to return the site config as part of the serialized site data
- [ ] Route added to access the site detail page (`/sites/<id>`)
  - TBD: We might want to use a slug instead of an id, in which case we'll need to generate that slug somehow
- [ ] Site title is shown above main content window _without_ "Import"/"Add" buttons
- [ ] Side nav included with "Content" section heading, and nested links to the types described in the `collections` section of the config ("Resource" or "Pages" in the wireframe example)
  - `localdev/starters/example-starter/ocw-studio.yml` was added in #41. Using the example of the config as it exists in that PR, the "Content" section would just have 1 option: "Page"
- [ ] Side nav links navigate to an empty page (component can be called `SiteContent` or something like that). This page only needs to show the page type as the title ("Resource" or "Pages" in the wireframe example)

PR #54, opened 2021-02-08
Detail view for websites

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Part of #40 

#### What's this PR do?
Adds a detail view for `/api/website`

Go to `/api/website/<uuid>/` and make sure it looks right. There should be a config as well if there is a starter, and if not config should be null.

### DIFF
--- a/websites/serializers.py
+++ b/websites/serializers.py
@@ -34,4 +34,32 @@ class WebsiteSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = Website
-        fields = "__all__"
+        fields = [
+            "uuid",
+            "created_on",
+            "updated_on",
+            "name",
+            "title",
+            "source",
+            "publish_date",
+            "metadata",
+            "starter",
+        ]
+
+
+class WebsiteDetailSerializer(serializers.ModelSerializer):
+    """ Serializer for websites with serialized config """
+
+    config = serializers.SerializerMethodField()
+    starter = WebsiteStarterSerializer(read_only=True)
+
+    def get_config(self, instance):
+        """Returns parsed YAML config"""
+        if instance.starter is not None:
+            return yaml.load(instance.starter.config, Loader=yaml.Loader)
+        else:
+            return None
+
+    class Meta:
+        model = Website
+        fields = WebsiteSerializer.Meta.fields + ["config"]
--- a/websites/serializers.py
+++ b/websites/serializers.py
@@ -50,16 +50,8 @@ class WebsiteSerializer(serializers.ModelSerializer):
 class WebsiteDetailSerializer(serializers.ModelSerializer):
     """ Serializer for websites with serialized config """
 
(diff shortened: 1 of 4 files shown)


websites/views.py
```diff
@@ -25,6 +26,7 @@ class DefaultPagination(LimitOffsetPagination):
 
 class WebsiteViewSet(
     mixins.ListModelMixin,
+    mixins.RetrieveModelMixin,
```
mbertrand, line 29: I think we might need 2 separate viewsets for the Website model.  One read-only viewset, accessible to unauthenticated users, currently used for listing newly published websites on the front page.  And another requiring authentication, to list and view/edit details only for the websites the user has view/edit access to (`WebsiteEditorViewSet`), that does not exclude unpublished websites?  For example, right now if I try to bring up the detail API view for an unpublished website I get a 404 response - the queryset excludes them:

```python
        queryset = Website.objects.filter(
            publish_date__lte=now_in_utc(),
        ).order_by(ordering)
```

mbertrand (changes requested):

websites/views.py
gsidebo: We shouldn't need a separate viewset. We should be able to manipulate the queryset based on the request user

websites/serializers.py
```diff
@@ -34,4 +34,32 @@ class WebsiteSerializer(serializers.ModelSerializer):
...
+            "publish_date",
+            "metadata",
+            "starter",
+        ]
+
+
+class WebsiteDetailSerializer(serializers.ModelSerializer):
+    """ Serializer for websites with serialized config """
+
+    config = serializers.SerializerMethodField()
```
gsidebo, line 53: Is this just for convenience? I guess having `config` as a top-level property is nice to have but it seems more straightforward just to use the detail serializer for `starter` (`starter = WebsiteStarterDetailSerializer(read_only=True)`)

noisecapella: That's true, I'll do that

websites/views.py
noisecapella: @mbertrand Should we filter unpublished websites in this PR? I can make the change now if so

mbertrand: Either way is fine with me (I guess it would be changed to filtering by published only for anonymous users?). If you prefer I can do that as part of the work on integrating the permission classes with the views, once this PR and the permissions PR are merged.

noisecapella: Sure, I'll leave it to you since it might be easier to deal with the views all at once

mbertrand: Okay, it looks like the API is already broken anyway on CI - https://github.com/mitodl/ocw-studio/issues/59

mbertrand (approved):

gsidebo (approved):
👍

PR #55, opened 2021-02-08
Changed WebsiteStarter.config to store JSON instead of YAML

gsidebo:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

N/A

#### What's this PR do?
Changes `WebsiteStarter.config` property so that it stores JSON instead of raw YML

Migrate, check the `config` values for some starters in Django admin, and try updating them

- Our own APIs are going to be passing this config around as JSON, so it doesn't make much sense to store it as YML. That was an oversight on my part during the first implementation
- I added an admin class for pretty-printing JSONField-type fields

ss 2021-02-05 at 14 12 36


localdev/starters/example-starter/ocw-studio.yml
```diff
@@ -2,4 +2,5 @@ collections:
   - label: "Page"
```
noisecapella, line 2: Should we convert this file to JSON if it's only going to be converted to JSON anyway? No strong feelings either way

main/admin.py
```diff
@@ -60,3 +65,21 @@ def get_readonly_fields(self, request, obj=None):
...
+    def format_value(self, value):
+        try:
+            value = json.dumps(json.loads(value), indent=2, sort_keys=True)
+            # these lines will try to adjust size of TextArea to fit to content
+            row_lengths = [len(r) for r in value.split("\n")]
+            self.attrs["rows"] = min(max(len(row_lengths) + 2, 10), 30)
+            self.attrs["cols"] = min(max(max(row_lengths) + 2, 40), 120)
+            self.attrs["style"] = "font-family: monospace"
+            return value
+        except Exception as e:  # pylint: disable=broad-except
```
noisecapella, line 83: What kind of exceptions are we expecting here? I'm also wondering if we should `log.exception` below instead of `log.warning`

websites/factories.py
```diff
@@ -24,7 +31,7 @@ class WebsiteStarterFactory(DjangoModelFactory):
     slug = factory.Sequence(lambda n: "starter-%x" % n)
     source = FuzzyChoice(STARTER_SOURCES)
     commit = factory.Faker("md5")
-    config = EXAMPLE_SITE_CONFIG
+    config = json.dumps(EXAMPLE_SITE_CONFIG)
```
noisecapella, line 34: I think this should be just `EXAMPLE_SITE_CONFIG`. If you try to create a `WebsiteStarter` you can see that `config` contains a JSON string rather than the unserialized contents of it

noisecapella (changes requested):

localdev/starters/example-starter/ocw-studio.yml
gsidebo: We're expecting the starter repos to define the config as yaml, so I think it makes sense to keep that pattern here

websites/factories.py
gsidebo: I think you're right. I'll change

main/admin.py
gsidebo: I want to avoid sending errors to Sentry if something just goes wrong in these admin forms. As far as exceptions go, I guess the obvious one is if someone sets the underlying value in the db to a non-JSON value. Personally I think it's OK just to default to a normal textarea if something in the `try` block fails for any reason

noisecapella: That's fine with me

noisecapella (approved):

PR #65, opened 2021-02-11
Integrate permissions with WebsiteViewSet

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
Closes #63 

#### What's this PR do?
- replaces `ReadOnlyPermission` with `HasWebsitePermission` for `WebsiteViewSet` and adds `UpdateModelMixin`
- tweaks `WebsiteDetailSerializer` to set the owner to the request user if present
- adds django-hijack

#### How should this be manually tested?
In a shell, create a test website and users with various permissions:
```
from websites.models import Website
from users.models import User
from django.contrib.auth.models import Group

for i in range(1,7):
     User.objects.update_or_create(username=f"fakeuser{i}", email=f"fakeuser{i}.mit.edu", is_active=True)
owner = User.objects.get(username="fakeuser3")
website = Website.objects.create(name="9999_fake_course", title="Fake Course", owner=owner)

User.objects.get(username="fakeuser1").groups.add(website.admin_group)
User.objects.get(username="fakeuser2").groups.add(website.editor_group)
User.objects.get(username="fakeuser4").groups.add(Group.objects.get(name="global_admin"))
User.objects.get(username="fakeuser5").groups.add(Group.objects.get(name="global_author"))
```

Go to `../api/websites/` in a browser without logging in.  You should get a list of published websites sorted by publishing date.

Go to `../api/websites/<website_uuid>` for any particular website.  You should not be able to view it.  

From django admin, hijack fakeuser1 through fakeuser6.

Go to `../api/websites/`:
  - The website created above should be the only one listed for fakeuser1,2,3,4
  - fakeuser5 and 6 should not be able to view any websites (initially, see below)
  - Only fakeuser4 and fakeuser5 (global admin, global author) should have the option to create a new website at the bottom of the page.  Give it a try (name is mandatory), a new website should be created with that user as the owner.  If you refresh the page the new website should be the first one in the list.
  
Go to `../api/websites/<website_uuid>`:
  - fakeuser1,2,3,4 should be able to view the details for this website
  - fakeuser1,3,4 (site admin, site owner, global admin) should be able to edit (patch) the website details at the bottom, give it a try.  fakeuser2 should not be able to.
  - fakeuser5,6 should get a 404
  - fakeuser5 should be able to view the details for the other website created above

Issue #63, opened 2021-02-10
Integrate permissions with current views for websites


mbertrand:
Integrate permission classes from #49 with current API views for websites

- Anonymous users should be able to view a list of published websites.
- Authenticated users should be able to view a list of the websites, and details of each website, that they are collaborators on (editor, admin, owner).

PR #69, opened 2021-02-16
Fix Website API

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Related to #63

#### What's this PR do?
Reverts a late change to PR #65 which broke API functionality

See instructions for #65

### DIFF
--- a/websites/serializers.py
+++ b/websites/serializers.py
@@ -1,6 +1,8 @@
 """ Serializers for websites """
+from django.db import transaction
 from rest_framework import serializers
 
+from users.models import User
 from websites.models import Website, WebsiteStarter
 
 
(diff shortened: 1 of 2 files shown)


mbertrand:
@gsidebo even though the tests passed using the DRF test client, that suggestion on PR #65 actually broke the API if you try it in a browser ("This QueryDict instance is immutable"), so I've reverted back to having that logic in the serializer instead.  I really prefer this approach anyway.

gsidebo (approved):
LGTM. Minor issue either way. My general concern is that it increases the responsibilities of the serializer a little, and I think we've had problems in the past from making our serializers too "smart". Another option here would be to use `get_serializer` and add `owner` to the kwargs before passing it in to `super`. The `hasattr` and `isinstance` wouldn't be necessary in that case. You can look into that if you want, but this is good to merge otherwise 👍

mbertrand:
Thanks, I merged this but will look into the other approach as part of the current PR I'm working on

PR #71, opened 2021-02-17
Site detail page

### LINKED ISSUE #40: Basic site detail UI

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #40 

#### What's this PR do?
Implements the site detail page, and adds some scaffolding 

Find the name of a website you have permission to access in `/api/websites/`, then go to `/sites/website-name-here/`. You should see 2 or 3 links in the sidebar. You should be able to click on all of them and have the link change without a full page refresh. The "Content" page shouldn't have anything but the other links should go to component pages where you see a link at the top, similar to the mockup in the issue.


static/js/Router.tsx
```diff
@@ -1,25 +1,26 @@
 import React from "react"
-import { Route, Router as ReactRouter } from "react-router"
+import { BrowserRouter } from "react-router-dom"
 import { Provider } from "react-redux"
-import { History } from "history"
+import { Provider as ReduxQueryProvider } from "redux-query-react"
 import { Store } from "redux"
 
 import App from "./pages/App"
+import { getQueries } from "./store/configureStore"
```
aliceriot, line 8: there's a `getQueries` function already set up in `lib/redux_query` which has the typing all set up already

static/js/selectors/websites.ts
```diff
@@ -0,0 +1,4 @@
+// TODO: how to type state?
+
+export const getWebsite = (state: any, uuid: string) =>
```
aliceriot, line 3: we might want to grab a library like [reselect](https://github.com/reduxjs/reselect) to get these selectors to be memoized and all that

static/js/index.tsx
```diff
@@ -18,15 +18,14 @@ Sentry.init({
   environment: SETTINGS.environment
 })
 
-const store = configureStore()
+const store: any = configureStore()
```
aliceriot, line 21: hum, I hoped that we could avoid `any` for redux stuff, did this not work at all without it?

aliceriot (commented):
just a few things

static/js/Router.tsx
aliceriot: (`getEntities` too)

package.json
```diff
@@ -91,8 +92,8 @@
     "repl": "node --require ./scripts/repl.js",
     "fmt": "LOG_LEVEL= prettier-eslint --write --no-semi --ignore $PWD/'static/js/types/*.d.ts' $PWD/'static/js/**/*.ts' $PWD/'static/js/**/*.tsx'",
     "fmt:check": "LOG_LEVEL= prettier-eslint --list-different --no-semi --ignore $PWD/'static/js/types/*.d.ts' $PWD/'static/js/**/*.ts' $PWD/'static/js/**/*.tsx'",
-    "typescript": "rm -rf dist && rm -f .tsbuildinfo && tsc",
-    "typecheck": "tsc --noEmit",
+    "typescript": "rm -rf dist && rm -f .tsbuildinfo && tsc --strictNullChecks",
+    "typecheck": "tsc --noEmit --strictNullChecks",
```
aliceriot, line 96: might be preferable to add this option to `tsconfig.json`, that way the config for typescript only lives in one place (also jest reads that config when it runs the tests)

aliceriot (commented):
one other small thing

static/js/selectors/websites.ts
noisecapella: Sure but do you think it would be useful for this particular selector? In the examples they only use `createSelector` for more complicated selectors: https://github.com/reduxjs/reselect#selectorsindexjs

noisecapella: Also, any thoughts on what to use instead of `any` for a type for `state`?

static/js/index.tsx
noisecapella: I'm seeing this error message:

    ➜  ocw-studio git:(gs/site-detail-ui) ✗ dwatch npm run typecheck
    Creating ocw-studio_watch_run ... done
    
    > ocw_studio@0.0.0 typecheck /src
    > tsc --noEmit --strictNullChecks
    
    static/js/components/SiteComponent.tsx:14:68 - error TS2345: Argument of type 'DefaultRootState' is not assignable to parameter of type 'ReduxState'.
      Type 'DefaultRootState' is missing the following properties from type 'ReduxState': queries, entities
    
    14   const website = useSelector(state => websiteSelectors.getWebsite(state, name))
                                                                          ~~~~~
    
    static/js/index.tsx:27:18 - error TS2322: Type 'Store<CombinedState<{ entities: Entities; queries: QueriesState; }>, ReduxQueryAction<Entities>> & { ...; }' is not assignable to type 'Store<any, AnyAction>'.
      Types of property 'dispatch' are incompatible.
        Type 'Dispatch<ReduxQueryAction<Entities>>' is not assignable to type 'Dispatch<AnyAction>'.
          Type 'AnyAction' is not assignable to type 'ReduxQueryAction<Entities>'.
            Type 'AnyAction' is not assignable to type 'MutateAsyncAction<Entities>'.
              Property 'url' is missing in type 'AnyAction' but required in type 'QueryConfig<Entities>'.
    
    27       <Component store={store} />
                        ~~~~~
    
      node_modules/redux-query/index.d.ts:167:5
        167     url: Url;
                ~~~
        'url' is declared here.
      static/js/Router.tsx:11:3
        11   store: Store
             ~~~~~
        The expected type comes from property 'store' which is declared here on type 'IntrinsicAttributes & RootProps & { children?: ReactNode; }'
    
    static/js/pages/SitePage.tsx:17:68 - error TS2345: Argument of type 'DefaultRootState' is not assignable to parameter of type 'ReduxState'.
    
    17   const website = useSelector(state => websiteSelectors.getWebsite(state, name))
                                                                          ~~~~~
    
    
    Found 3 errors.
    
    npm ERR! code ELIFECYCLE
    npm ERR! errno 2
    npm ERR! ocw_studio@0.0.0 typecheck: `tsc --noEmit --strictNullChecks`
    npm ERR! Exit status 2
    npm ERR! 
    npm ERR! Failed at the ocw_studio@0.0.0 typecheck script.
    npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
    
    npm ERR! A complete log of this run can be found in:
    npm ERR!     /home/node/.npm/_logs/2021-02-18T19_33_27_790Z-debug.log
    ERROR: 2

package.json
noisecapella: Looks like `strict` is already set to true so this option should be redundant

static/js/index.tsx
aliceriot: oh ok yeah, I know what the issue is - the `store: Store` type on `App.tsx` isn't sufficiently specific. I fix that here: https://github.com/mitodl/ocw-studio/pull/66/files#diff-0dab17b074318efbcc682ae6d96b4c08de1c795a1b83f549ccac4e6f49d35937 you should be able to use the same fix, and then get specific, full inferred types for the store

aliceriot: needs to be `store: ReturnType<typeof configureStore>`

static/js/selectors/websites.ts
aliceriot: I think you can use `ReturnType<typeof configureStore>` for it. We should actually create a type alias for that, like

```js
type Store = ReturnType<typeof configureStore>
```

aliceriot: in the `configureStore.ts` file, and then we can import that elsewhere

static/js/components/SiteComponent.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from "react"
```
aliceriot, line 1: did you run into an issue with doing `import React from 'react'` ? just curious

static/js/selectors/websites.ts
noisecapella: Thanks, that fixed it

aliceriot: I think the advantage to using reselect is the memoization, which then in turn reduces the number of times that react will needlessly re-render (because for objects react only does a shallow compare).

for this one I think you could do something like:

```js
const getWebsite = createSelector(
  state => state.entities?.websites,
  websites => (
    lodash.memoize(name => websites[name])
  )
)
```

this is a pattern we use in open-discussions in a few places. then basically you get two things 1) the return value of the selector (the generated fucnction) will be memoized, and regenerated whenever the value in the store changes and 2) the memoized function has the websites within a closure so it will be usable within the component.

then you can do in the component:

```js
const memoizedWebsites = useSelector(getWebsite)
const website = memoizedWebsites(name)
```

one example of this pattern in open-discussions in the `learningResourceSelector` (although that one's more complicated because the accessor function needs to take two params)

aliceriot: if you do the above probably good ot name the selector `getWebsiteCursor` or something that captures that it returns an accessor function instead of a straight up value, then you'd do something like

```js
const websiteCursor = useSelector(getWebsiteCursor)
const website = websiteCursor(name)
```

static/js/components/SiteComponent.tsx
noisecapella: I'm not sure why I added it but it doesn't appear to break anything to import only the default export

static/js/query-configs/websites.ts
```diff
@@ -0,0 +1,20 @@
+import { Website } from "../types/websites"
+
+interface Websites {
+  string: Website
+}
```
aliceriot, line 5: I believe this type as written is typing an object shaped like this:

```ts
{ "string": { ... } }
```

I think we want to do this instead:

```ts
interface Websites {
  [key: string]: Website
}
```

static/js/components/SiteComponent.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+jest.mock("react-router-dom", () => ({
+  // @ts-ignore
+  ...jest.requireActual("react-router-dom"),
+  useRouteMatch: mockUseRouteMatch
+}))
+
+describe("SiteComponent", () => {
+  let helper: IntegrationTestHelper,
+    render: ReturnType<IntegrationTestHelper["configureRenderer"]>,
```
aliceriot, line 21: we could add a type alias `TestRenderer` in `IntegrationTestHelper` to hold this type

static/js/pages/SitePage.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { Route, Switch, useRouteMatch } from "react-router-dom"
+
+import SiteSidebar from "../components/SiteSidebar"
+import SiteComponent from "../components/SiteComponent"
+
+import { websitesRequest } from "../query-configs/websites"
+import { getWebsiteCursor } from "../selectors/websites"
+
+export default function SitePage(): JSX.Element | null {
+  const match = useRouteMatch()
```
aliceriot, line 13: I think you should be able to add a param for this `useRouteMatch` too

aliceriot (changes requested):
a few comments and questions

static/js/query-configs/websites.ts
noisecapella: Ah good catch

static/scss/layout.scss
```diff
@@ -1,8 +1,34 @@
...
+@import "node_modules/bootstrap/scss/bootstrap";
 
-// Fake style to confirm that webpack dev server works
-h1 {
-  border-radius: 25px;
-  border: 2px solid blue;
-  padding: 2px;
+$max-width: 1447px;  // arbitrary value, copied from ocw-next
+
+.site-wide-header {
```
gsidebo, line 6: I think it's almost certain we will want a separate page of styles for the site detail page rather than putting them all in a stylesheet with styles for the entire site. I'd be in favor of doing that now rather than later

gsidebo (changes requested):

gsidebo (changes requested):

static/js/types/websites.ts
```diff
@@ -0,0 +1,36 @@
...
+  source: string
+  commit: string | null
+  slug: string
+  config: WebsiteStarterConfig | null
+}
+
+export interface Website {
+  uuid: string
+  created_on: string // eslint-disable-line
+  updated_on: string // eslint-disable-line
```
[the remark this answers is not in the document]
noisecapella: I'm fine with that, though I think the change needs to be made in the `eslint-config-mitodl` library https://github.com/mitodl/eslint-config/blob/master/index.js#L46

static/js/selectors/websites.ts
```diff
@@ -0,0 +1,9 @@
+import { createSelector } from "reselect"
+import { memoize } from "lodash"
+
+import { ReduxState } from "../reducers"
+
+export const getWebsiteCursor = createSelector(
+  (state: ReduxState) => state.entities?.websites ?? [],
```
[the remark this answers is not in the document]
noisecapella: Yes you're right, I was confusing it with the response from the list view API

static/js/components/SiteSidebar.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  const configItems: ConfigItem[] = website.starter?.config?.collections ?? []
+
+  return (
+    <div className="sidebar">
+      <ul>
+        <li>
+          <NavLink exact to={siteUrl(website.name)} activeClassName={"active"}>
+            Content
+          </NavLink>
```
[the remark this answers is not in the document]
noisecapella: Sure, but which wireframes? The one linked in #40 still shows "Content"

static/js/types/websites.ts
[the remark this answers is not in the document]
aliceriot: yeah, we can just leave this as-is for now

aliceriot (approved):
LGTM! :+1:

static/js/components/SiteSidebar.tsx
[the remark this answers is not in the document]
gsidebo: @noisecapella Ah, these must have changed since I looked at them last. I still think it's a good idea to get rid of this since we'll probably have child menu items for each content type

static/js/components/SiteSidebar.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  it("renders some links", () => {
+    const wrapper = shallow(<SiteSidebar website={website} />)
+
+    const links = wrapper
+      .find("NavLink")
+      .map(link => [link.text(), link.prop("to")])
+
+    const expected = [
+      ["Content", siteUrl(website.name)],
```
gsidebo, line 26: Since you removed the "Content" item, shouldn't this cause the test to fail?

static/scss/site-page.scss
```diff
@@ -0,0 +1,17 @@
...
+  max-width: $max-width;
+  margin: 0 auto;
+
+  .sidebar {
+    background-color: #eeeeee;
+    border: 1px solid black;
+
+    ul {
+      list-style: none;
+    }
```
gsidebo, line 11: Small nit: can you introduce a ~15px vertical margin between each immediate child of this `ul`? The top-level menu items are kind of cramped.

gsidebo (approved):
LGTM! The comment about the test and the menu spacing are take-or-leave. Feel free to merge whenever 👍

static/js/components/SiteSidebar.test.tsx
noisecapella: Yes, good catch. It seems like I had the `links` and `expected in the wrong order. `arrayContaining` should be used on the array with all of the items to be matched while the part on the left should be the subset

Issue #60, opened 2021-02-09
Add collaborator screens


gsidebo:
**Depends on permissions PR: https://github.com/mitodl/ocw-studio/pull/49** 
Wireframes: [list page](https://lucid.app/lucidchart/6be33dd8-f7db-47f8-8af8-338255f23c3c/edit?shared=true&page=BYpVt0I~zBD_#?folder_id=home&browser=icon), [create page](https://lucid.app/lucidchart/6be33dd8-f7db-47f8-8af8-338255f23c3c/edit?shared=true&page=o3pVIT4811zL#?folder_id=home&browser=icon)

The REST APIs to support these screens can be added in a separate PR or as part of this one

**Acceptance criteria:**
- Listing and create/edit pages added
- Create/edit page has simple text input (required) and dropdown for user and permission group respectively
  - Text input will be for user email (case-insensitive). Non-existent emails should result in a validation error message under the text input. We may want a more sophisticated input for finding users in the future, but that can be added later
Saving this form should add the appropriate groups/permissions for the given user
- Listing page shows all users with website-specific permissions/groups
  - Show the current permission level/group for each user in the list as simple text
  - If the logged-in user can change website-specific permissions, place an edit icon next at the end of the line. Clicking on that icon should take you to the create/edit page to edit that user's permission level
  - Garbage can/large X icon should be shown if the current user has the permission level to delete users. We should ask for confirmation before going through with the delete

mbertrand:
> "If the logged-in user can change website-specific permissions, place an edit icon next at the end of the line"

Should an editor, who is not allowed to add/change collaborators, even be able to view this page?  Any benefit to an editor knowing who admins & other editors are?

> "If the logged-in user can change website-specific permissions, place an edit icon next at the end of the line. Clicking on that icon should take you to the create/edit page to edit that user's permission level"

The [list page design](https://lucid.app/lucidchart/6be33dd8-f7db-47f8-8af8-338255f23c3c/edit?shared=true&page=BYpVt0I~zBD_#?folder_id=home&browser=icon) shows a dropdown instead next to the user's name, presumably changing this will instantly change the user's permission level, without the need to go to the create/edit page?  Which approach is preferred?

gsidebo:
@mbertrand 

> Should an editor, who is not allowed to add/change collaborators, even be able to view this page? Any benefit to an editor knowing who admins & other editors are?

[...]

PR #72, opened 2021-02-17
Website collaboration API

mbertrand:
~Temporarily blocked by #73, this PR includes the changes from that one.~

#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Partially addresses #60 

#### What's this PR do?
- Adds an API for collaborator CRUD

Create a `Website` and some users with various permissions in a shell:
```python
from websites.models import Website
from users.models import User
from django.contrib.auth.models import Group

for i in range(1,7):
     User.objects.update_or_create(username=f"fakeuser{i}", email=f"fakeuser{i}@mit.edu", is_active=True)
owner = User.objects.get(username="fakeuser3")
website = Website.objects.create(name="9999_fake_course", title="Fake Course", owner=owner)

User.objects.get(username="fakeuser1").groups.add(website.admin_group)
User.objects.get(username="fakeuser2").groups.add(website.editor_group)
User.objects.get(username="fakeuser4").groups.add(Group.objects.get(name="global_admin"))
User.objects.get(username="fakeuser5").groups.add(Group.objects.get(name="global_author"))
```

If you log in as a superuser or admin for that website (fakuser1, fakeuser4, owner), you should be able to view the URL `http://localhost:8043/api/websites/<website_name>/collaborators/`.  Other users should get a 403.  The page should display a list of users who have permission to view/edit the website, and show the user's username, email, name, plus the group and role they are a member of ('owner' if they are the website owner).

From that api page you should be able to add a new collaborator by sending an email and a role, for example:
```json
{
    "role": "admin",
    "email": "fakeuser1.mit.edu"
}
```

If you try adding a user that already has permission for that website, or use an email that doesn't exist for a user, or try to assign a non-existent group or group for another website, you should get a validation error.

Go to the detail URL for a specific collaborator:  `http://localhost:8043/api/websites/<website_name>/collaborators/<user_username>/`

From here, try POSTing a change from an admin_group to an editor_group or vice versa:
```json
{
    "role": "editor"
}
```

As long as the user is not a global admin or website owner, it should work, otherwise you should get a permission denied error.  You should also get a validation error if you enter an invalid role name.

Click the DELETE button on one of these detail pages.  You should get a 403 if you try it on the site owner or a global admin.  Otherwise, the user should be removed from all permission groups associated with the website.  The user should not be deleted.


noisecapella:
What should happen if you add a collaborator without any role? I'm seeing a `KeyError` in that case:

    web_1     | [2021-02-19 20:20:40] ERROR 12 [django.request] log.py:224 - [810a490615db] - Internal Server Error: /api/websites/site-24/collaborators/
    web_1     |   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    web_1     |     response = get_response(request)
    web_1     |   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 179, in _get_response
    web_1     |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
    web_1     |   File "/usr/local/lib/python3.8/site-packages/sentry_sdk/integrations/django/views.py", line 52, in callback
    web_1     |     return old_callback(*args, **kwargs)
    web_1     |   File "/usr/local/lib/python3.8/site-packages/sentry_sdk/integrations/django/views.py", line 52, in callback
    web_1     |     return old_callback(*args, **kwargs)
    web_1     |   File "/usr/local/lib/python3.8/site-packages/sentry_sdk/integrations/django/views.py", line 52, in callback
    web_1     |     return old_callback(*args, **kwargs)
    web_1     |   File "/usr/local/lib/python3.8/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    web_1     |     return view_func(*args, **kwargs)
    web_1     |   File "/usr/local/lib/python3.8/site-packages/rest_framework/viewsets.py", line 125, in view
    web_1     |     return self.dispatch(request, *args, **kwargs)
    web_1     |   File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 509, in dispatch
    web_1     |     response = self.handle_exception(exc)
    web_1     |   File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 469, in handle_exception
    web_1     |     self.raise_uncaught_exception(exc)
    web_1     |   File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    web_1     |     raise exc
    web_1     |   File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 506, in dispatch
    web_1     |     response = handler(request, *args, **kwargs)
    web_1     |   File "./websites/views.py", line 184, in create
    web_1     |     group_name = f"{ROLE_GROUP_MAPPING[serializer.validated_data.get('role')]}{website.uuid.hex}"
    web_1     | KeyError: None

noisecapella:
I'm seeing a different error when POST-ing to the collaborators API because the `Group` is missing. Do you know where the `Group`s are meant to be created?

mbertrand:
Groups should be created automatically on website creation via signals.  You might need to create groups for any old websites predating the signal with `manage.py backpopulate_groups`

noisecapella (changes requested):

websites/constants.py
```diff
@@ -32,6 +32,12 @@
 PERMISSION_EDIT_CONTENT = "websites.edit_content_website"
 PERMISSION_COLLABORATE = "websites.add_collaborators_website"
 
+ROLE_ADMINISTRATOR = "admin"
+ROLE_EDITOR = "editor"
+ROLE_GLOBAL = "global_admin"
```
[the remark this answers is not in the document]
mbertrand: The `global_author` group is actually irrelevant in terms of collaboration - someone in that group just has permission to create new websites, and becomes the owner of the websites they create (so should show up in the collaborators list as the owner).  They have no permission to do anything for websites they did not create.

websites/permissions.py
```diff
@@ -237,8 +237,22 @@ def has_object_permission(self, request, view, obj):
...
+        )
+        return check_perm(request.user, constants.PERMISSION_COLLABORATE, website)
+
     def has_object_permission(self, request, view, obj):
-        return check_perm(request.user, constants.PERMISSION_COLLABORATE, obj)
+        website = get_object_or_404(
+            Website, name=view.kwargs.get("parent_lookup_website", None)
+        )
+        if request.method not in SAFE_METHODS and (
+            website.owner == obj or is_global_admin(obj)
```
noisecapella, line 251: I might be missing something but wouldn't we want to return `True` in this expression since it's checking if the user is a owner or admin of the object?

noisecapella:
@mbertrand Do you know where we assign the `PERMISSION_COLLABORATE` permission?

websites/permissions.py
mbertrand: `obj` here is the user for which permissions are trying to be changed, not the user making the request, and permissions should never be changed for global admins or website owners.

mbertrand:
`PERMISSION_COLLABORATE` and other administrator permissions are assigned to the `admin_group` for the website [in this function](https://github.com/mitodl/ocw-studio/blob/e5bfe024505e23c946175896177dd31bd7b8bce1/websites/permissions.py#L78-L80).

websites/views.py
```diff
@@ -89,10 +107,101 @@ def get_queryset(self):
...
+        for group in get_groups_with_perms(website):
+            instance.groups.remove(group)
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+    def create(self, request, *args, **kwargs):
+        """ Add a user to the website as a collaborator"""
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        website = Website.objects.get(name=self.kwargs.get("parent_lookup_website"))
+        group_name = f"{ROLE_GROUP_MAPPING[serializer.validated_data.get('role')]}{website.uuid.hex}"
```
noisecapella, line 182: Can we refactor this into a utility function? I see this pattern in `update` below and also a couple places in `websites/permissions.py`

noisecapella (changes requested):

noisecapella (approved):
Functionality works great :+1:

websites/views.py
```diff
@@ -89,10 +107,105 @@ def get_queryset(self):
...
+            if user_name == owner_username:
+                return User.objects.filter(username=user_name).annotate(
+                    group=Value(constants.ROLE_OWNER, CharField())
+                )
+            group_subquery = Group.objects.filter(
+                Q(user__username=OuterRef("username")) & Q(name__in=website_groups)
+            )
+            return User.objects.filter(
+                Q(username=user_name) & Q(groups__name__in=website_groups)
+            ).annotate(group=Subquery(group_subquery.values("name")[:1]))
```
gsidebo, line 155: To me it seems awkward to re-query for the same user we've already fetched for the purpose of adding the annotated group. This group_subquery logic is also kind of difficult to follow. I think we should (a) figure this out in the view and pass it into the serializer via context, or (b) let the serializer itself figure it out. If we go with (b) for eample, we could pass `owner_username` to the serializer via context, and add `get_group` to the serializer. If the `group` attribute exists on the object, just return that. If not, run this logic to compare the owner username, then fetch the relevant group name separately if the username doesn't match. That method could probably be cached as well so we avoid doubling queries when serializing the `group` field and using the group value in `get_role`. This adds some responsibilities to the serializer but I think it's worth it.

Once we move that logic, we can get rid of the multiple `Group.objects.get` queries. You could just fetch the records in their entirety right away instead of just fetching the `name`.

gsidebo (changes requested):

PR #73, opened 2021-02-18
Lookup websites by name instead of uuid

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Related to #71 

#### What's this PR do?
Changes the website detail API to return websites by `name` instead of `uuid`. This means we can have `/sites/name-goes-here` and look up the website info by the same name in the URL path.

Go to `/api/websites`, pick a site which you have permission to view, then go to `/api/websites/<name>/` where the name comes from the `name` field in the website.


gsidebo (approved):
LGTM! 👍

PR #76, opened 2021-02-19
Remove settings regarding reloading worker processes

noisecapella:
#### Pre-Flight checklist

- [x] Testing

Related issues: https://github.com/mitodl/uwsgi-formula/pull/2, https://github.com/mitodl/cookiecutter-djangoapp/pull/163
Possibly related discussion: https://github.com/unbit/uwsgi/issues/1804

#### What's this PR do?
This removes a few settings from `uwsgi.ini` which kill worker processes regularly to prevent memory leaks. Though this should work, it seems like the worker processes are being killed and then not being restarted until after the website is accessed. If the website is not being accessed over an hour, a 502 is returned. The settings removed are the ones listed listed [here](https://www.techatbloomberg.com/blog/configuring-uwsgi-production-deployment/) under "Worker Recycling".

You can reproduce this issue locally by setting `max-worker-lifetime` to `15` and watching the logs. You should see this:

    web_1     | worker 1 lifetime reached, it was running for 16 second(s)
    web_1     | worker 2 lifetime reached, it was running for 16 second(s)
    web_1     | Respawned uWSGI worker 1 (new pid: 49)
    web_1     | spawned 2 offload threads for uWSGI worker 1
    web_1     | Respawned uWSGI worker 2 (new pid: 52)
    web_1     | running "exec:touch /tmp/app-initialized" (accepting1)...
    web_1     | spawned 2 offload threads for uWSGI worker 2
    web_1     | Python auto-reloader enabled

repeated every 15 seconds or so. So far so good. At some point, go to `/api/websites/`. You should see some kind of successful result. However, if you look at the logs you should see this:

    web_1     | worker 1 lifetime reached, it was running for 16 second(s)
    web_1     | worker 2 lifetime reached, it was running for 16 second(s)

It will stay like this indefinitely. Go to `/api/websites/` and you should see a 502, and this will trigger a respawn of the worker processes in the logs. If you go to `/api/websites/` a second time it should work fine, until 15 seconds elapse and the problem occurs again.

Note that you won't be able to verify that this is fixed since this removes the `max-worker-lifetime` option.


shaidar:
👍

Issue #52, opened 2021-02-04
Implement page content UI and evaluate Netlify


gsidebo:
[Wireframe](https://lucid.app/lucidchart/6be33dd8-f7db-47f8-8af8-338255f23c3c/edit?shared=true&page=hHG8TGWUOjR0#?folder_id=home&browser=icon)

This will be the initial implementation of the page where authors will add/edit site content.

Here's an example of the initial mapping we have in mind from the site config to the database record:

```yml
  - label: "Page"
    name: "page"
    fields:
      - {label: "Title", name: "title", widget: "string"}
      - {label: "Body", name: "body", widget: "markdown"}
      - {label: "Field 1", name: "field-1", widget: "string"}
      - {label: "Field 2", name: "field-2", widget: "string"}
```
```python
WebsiteContent.title = "My Title"
WebsiteContent.markdown = "## My page\n\n### Subsection 1\n\n..."
WebsiteContent.metadata = {
  "field-1": "Value 1",
  "field-2": "Value 2",
}
```

In other words, a config field named `title` maps to the `WebsiteContent.title` field, a config field named `body` maps to the `WebsiteContent.markdown` field, and all other fields are saved in `metadata` as JSON as `{[field_name]: [value]}`. The `populate_local_starters` command can be used to load in local site configs.

We will be evaluating [Netlify widgets](https://www.netlifycms.org/docs/widgets/#default-widgets) ([Github](https://github.com/netlify/netlify-cms/tree/master/packages)) for this project. For this issue, we just need widgets for simple text and for markdown. We're still evaluating CKEditor (#39) for use as our WYSIWYG editor. If we decide that CKEditor suits our needs, we will replace the Netlify markdown editor with that. It would help to also start evaluating the other widgets they offer.

**Acceptance criteria:**
- This page is rendered when the "Content" submenu items are clicked (see #40)
- Routes added for adding a new piece of content and editing an existing piece of content
- Content editing form has widgets that match the `fields` element of the site config
  - The ordering of these fields should match the config, and they can be rendered vertically in a single column
- Saving the form creates/updates a WebsiteContent record with the values in the form fields
- If page content listing UI is done before this, the "Add" buttons and content type list items link to this page
- Evaluate Netlify text widget and markdown widget for use in OCWS

**Not needed:**
- Widget types other than text/string and markdown
- Field validation

PR #85, opened 2021-02-24
Use resource for content type instead of file

noisecapella:
#### Pre-Flight checklist

- [x] Migrations

Part of #52 

#### What's this PR do?
Updates `file` to `resource` in `WebsiteContent.type` and adds a migration to make that change for backpopulated content in the database.

Migration should work, and you should be able to reverse it. There is no user facing code for `WebsiteContent` yet (that's in #84)

### DIFF
--- a/ocw_import/api.py
+++ b/ocw_import/api.py
@@ -9,7 +9,7 @@ import yaml
 from dateutil import parser as dateparser
 
 from main.s3_utils import get_s3_object_and_read, get_s3_resource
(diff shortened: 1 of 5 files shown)


gsidebo:
@noisecapella Looks great. While we're at it, though, what do you think of adding an index to the `type` field? We'll be filtering on that field for sure.

noisecapella:
The `type` field has low cardinality since it's just `page` and `resource` in roughly equal parts, so an index wouldn't do much to speed up queries there.

gsidebo (approved):
LGTM! 👍

Issue #39, opened 2021-02-01
Evaluate CKEditor and produce live example


gsidebo:
We use CKEditor as our WYSIWYG editor in Open. We want to evaluate this library to see if it's suitable for OCW Studio. 

Goals:
- Make sure we can have editable hugo shortcodes that editor understands
- Make sure we can support embedded youtube links (et al)
- Make sure that the editor can be initialized with markdown content and display it correctly

It may be helpful to evaluate this by using content from a `Website` that has been imported from our actual OCW courses (the `backpopulate_ocw_courses` command can be used). We know that course content has hugo shortcodes et. al.

**Acceptance criteria:**
- Comment on this issue with a 👍 /👎   for the viability of CKEditor for our OCW Studio use cases and the goals described above
- If 👍 , PR created that satisfies the following:
  - npm dependency added for CKEditor
  - Any relevant tooling added (editor tools for adding shortcodes, youtube links, etc)
  - Instructions included in the PR to render an editor and use those tools
- If 👎 , another issue(s) created to evaluate other WYSIWYG libraries

gumaerc:
Something we're going to want to support here are custom buttons for adding [Hugo Shortcodes](https://gohugo.io/templates/shortcode-templates/).  They're basically like template blocks that can be embedded in Markdown like:

```md
These video lectures of Professor Gilbert Strang teaching 18.06 were recorded live in the Fall of 1999. Support for the video production was provided by the Lord Foundation of Massachusetts under a grant to the MIT Center for Advanced Educational Services.

{{< video-gallery-item href="/courses/18-06-linear-algebra-spring-2010/sections/video-lectures/lecture-1-the-geometry-of-linear-equations" section="Video Lectures" title="Lecture 1: The geometry of linear equations" description="These  video lectures of Professor Gilbert Strang teaching 18.06 were recorded  in Fall 1999 and do not correspond precisely to the current   edition  of the textbook. However, this book is  still the best  reference  for  more information on the topics covered in  each lecture. Strang, Gilbert. Introduction to Linear Algebra. 5th ed. Wellesley-Cambridge Press, 2016. ISBN: 9780980232776. Instructor/speaker: Prof. Gilbert Strang" thumbnail="https://img.youtube.com/vi/J7DzL2_Na80/default.jpg" >}}
```

So ideally there would be buttons for things like "Insert Video Gallery Item" that would take the different parameters defined here (`href`, `description`, `thumbnail`) and write out the shortcode based on a template.

[...]

PR #87, opened 2021-02-24
documentation of adding CKEditor plugin

aliceriot:

part of #39 

#### What's this PR do?

I wrote up documentation of implementing a CKEditor plugin to add YouTube embeds from scratch. We don't actually want to use such a plugin ourselves (we'd want to just use CKEditor's media embed plugin) but it is helpful as an example of how to build a plugin in general.

Read it and make sure it all makes sense.

If you wanted to be really adventurous you could try to follow along!


Ferdi:
Amazing ! 

Should we put this in its own repo ? It could serve as a cookie cutter for us and at the same time it's a great open source contribution imo.

Ferdi:
I just noticed this is a doc file in markdown. Initially thought it was a cookie cutter plugin heavily documented,  which would be even more helpful

aliceriot:
I could look into re-writing this as a cookiecutter. I'm not sure how helpful that will be though, since as far as I can tell (I'm not super familiar with how cookiecutter works) it doesn't have built-in facility for adding code to an existing project. It seems it can be done, but requires a decent bit of fiddling. Still, perhaps the cookiecutter could put the code in the top-level or something and then the developer would be responsible for adding on their own? I'm wondering if @noisecapella you have thoughts on whether cookiecutter is helpful for something like this.

docs/ckeditor/plugin.md
```diff
@@ -0,0 +1,476 @@
...
+          class: YOUTUBE_EMBED_CLASS,
+          src:   youtubeEmbedUrl(videoId),
+          ...YOUTUBE_EMBED_PARAMS
+        })
+        return toWidget(iframe, viewWriter, { label: "Youtube Embed" })
+      }
+    })
+  }
+}
+```
```
gsidebo, line 257: This looks like a great example to me, especially with all of the comments. I think this section could be briefer and more straightforward if it was rewritten like this:

```
### Define converters

(same intro up to the iframe example)

(full code example above)

(Notes explaining upcast/downcast etc. in more detail, 
links to relevant CKEditor docs, `toWidget`, anything else 
that isn't covered in code comments) 
```

To me it seems like that format could get rid of some redundant details, but you can take or leave that suggestion.

docs/ckeditor/plugin.md
```diff
@@ -0,0 +1,476 @@
...
+
+So let's get started!
+
+### Define a new plugin
+We'll do so by creating a new class that extends `@ckeditor/ckeditor5-core/src/plugin`:
+
+```ts
+class YoutubeEmbedEditing extends Plugin {
+  init() {}
+}
```
gsidebo, line 46: Nit: any reason we wouldn't just include the import in the example?
```suggestion
import Plugin from "@ckeditor/ckeditor5-core/src/plugin"
class YoutubeEmbedEditing extends Plugin {
  init() {}
}
```

docs/ckeditor/plugin.md
```diff
@@ -0,0 +1,476 @@
...
+```
+
+The `init` method will be called when the plugin is instantiated, so we'll call our methods for defining the schema and the converters from there.
+
+### Define the schema
+We need to write a bit of code to let CKEditor know about what we're trying to have it do. Because we're trying to add an embedded video player we want this to be a 'root-level' node, meaning the user shouldn't be able to insert one in the middle of a paragraph. We also need to store a bit of text inside of the node, so we can store the video ID of the video which the user is embedding.
```
gsidebo, line 52: Nit, take or leave: this seems like an important concept to emphasize
```suggestion
We need to write a bit of code to let CKEditor know about what we're trying to have it do. Because we're trying to add an embedded video player we want this to be a 'root-level' node. **A 'root-level' node means a user shouldn't be able to insert one in the middle of a paragraph**. We also need to store a bit of text inside of the node, so we can store the video ID of the video which the user is embedding.
```

gsidebo (changes requested):
Adding to the inline comments...

- If I'm remembering right, you implemented this plugin. What do you think of pushing a branch with the actual implementation, then referring to that branch in this doc to give the reader a better idea of all the files that need to be added/updated?
- We might want to include a note near the top saying explicitly that CKEditor stores its contents as HTML, and that we're using a plugin to convert to/from markdown. I can imagine a developer reading this doc and being confused that these examples are in HTML

docs/ckeditor/plugin.md
```diff
@@ -0,0 +1,476 @@
...
...zUOh7+DadHTWJSwiPvfu5/Y/rHMGD6DlJgUhuUN4/4b7m87WNvPvx/rfrGOhz54iAfffxC3x03yD5L5YdgPvbm7Xmfy+OjQze12s3HfmevX3L2mmsq5A7xY0dn27dvH3Xffzb///W+sVqu3y+k2unaNdEXEMwd5MfnMmW23RF5eY8ji4mIGDx5McHAwvXr16vCDlJcjY+zF+PHw3HMtj5977pIsn168mJycHJ4KDcU6ceIl779Hl0XEMIwR8hs3gtvdElZud48vF//pT7ywfDkPvPMOo2+88ZL33+PLImIYmq4RkW6h6RrfZIy9EBGRDinkRUQMTCEvImJgCnkREQNTyIuIGJhPh7zNcuZxeF8L75U3eK8YETmn98obCO97Jk6+/X9XvMunz3EK9PPgOt1yyYE5P+5H9ge1/G7tMS9XJSLfNbCfhYd+fOb6Tb1tvnnv6HOpqanBZrO1u82mUfjsefIej4evTrgoPaohgcjlxlbzOeZTJ7xdRqfZbDYcDge9e/cmKCgIf39/w4S9T4/kgwNMhNpcHHF17m5JIuJ9/dy1OBwBQEC79R6Ph4S8ajbdc5V3CjuH1tG71WrFZrMZ5kNQrXw25E0mExaLhWFBzYQ4T/JVox8nT1tp8mhkL+JrrKbTBJpOEWxpwEYT0OusbTweDyarf7ub73hb62jdYrFgtVoJCAjAYrEYZhQPPhzy0PKD9/f3J9TPj5A+bjweNx7PaW+XJSId8gP6nPNZj8eDOcBFcHDwpSupE1pH8mazGYvFgsVirIGkT4c8tAR969snHz18ICKd4PF4MPn51ki+VevI3Ugj+FY+H/Jg7F+AyJXC4/FgMpsNN+ft6/TTFhExMIW8iIiBKeRFRAxMIS8iYmAKeRERA1PIi4gYmEJeRMTAFPIiIgamkBcRMTCFvIiIgSnkRUQMTCEvImJgCnkREQNTyIuIGJhCXkTEwBTyIiIGppAXETEwhbyIiIEp5EVEDEwhLyJiYAp5EREDU8iLiBiYQl5ExMAU8iIiBqaQFxExMIW8iIiBKeRFRAxMIS8iYmAKeRERA1PIi4gYmEJeRMTAFPIiIgamkBcRMTCFvIiIgSnkRUQMTCEvImJgCnkREQNTyIuIGJhCXkTEwBTyIiIGppAXETEwhbyIiIEp5EVEDEwhLyJiYP8PsVWb5OUPUKIAAAAASUVORK5CYII=)
```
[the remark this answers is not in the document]
aliceriot: it worked in my local markdown editor so I was hoping it would work here, oh well

docs/ckeditor/plugin.md
aliceriot: I like having a separate section for each key part, and then a 'putting it together' section which is consistent across the things, so I'm inclined to leave it as-is

aliceriot:
I'm planning to expand the documentation with info about markdown support once that is added - after the first markdown PR I'm going to open another to add support for [ckeditors media embed plugin](https://ckeditor.com/docs/ckeditor5/latest/features/media-embed.html) which will involve adding custom markdown extensions, so that will be a natural time to add documentation for it I think.

gsidebo:
@aliceriot Yeah I think a link to the branch would be really helpful, probably near the top. I think we'll get a lot of use out of a link to a diff that has the full implementation. You could squash the commits down and just link directly to the single commit (the "compare" link will start getting confusing as we push more commits to master)

aliceriot:
@gsidebo updated the example branch showing the full implementation of this plugin, and added a link to the doc file.

docs/ckeditor/plugin.md
```diff
@@ -0,0 +1,504 @@
...
+    1. [Adding a glue plugin](#Adding-a-glue-plugin)
+    1. [Configuring CKEditor](#Configuring-CKEditor)
+        1. [Adding the plugin](#Adding-the-plugin)
+        1. [Configure the toolbar](#configure-the-toolbar)
+
+---
+
+This document describes how to add a custom plugin to CKEditor, using the example of embedding a Youtube video as a simple example.
+
+A complete example show all the code from this example can be found here: <https://github.com/mitodl/ocw-studio/commit/25f57c11540fd001a985b38fdd50a321fba67a4e>
```
gsidebo, line 25: Nit: 
```suggestion
A complete example with the full implementation described in this doc can be found here: <https://github.com/mitodl/ocw-studio/commit/25f57c11540fd001a985b38fdd50a321fba67a4e>
```

gsidebo (approved):
Just one [phrasing nit](https://github.com/mitodl/ocw-studio/pull/87/files#r586836226). You can take or leave & merge whenever. This looks great! 👍

docs/ckeditor/plugin.md
aliceriot: not gonna lie I didn't re-read that after typing :joy:

PR #89, opened 2021-02-25
Add markdown support to CKEditor

aliceriot:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

part of #39 

#### What's this PR do?

This adds support for Markdown to our CKEditor component. How?

- a markdown library (`lib/markdown.ts`) exporting two functions:
    - `md2html`: supports translating Markdown to HTML. Implemented with [showdown](https://github.com/showdownjs/showdown/)
    - `html2md`: translates HTML to Markdown. This is implemented with the customized turndown from the `@ckeditor/ckeditor5-markdown-gfm`
- a `DataProcessor` plugin for CKEditor which subclasses `GFMDataProcessor` from `@ckeditor/ckeditor5-markdown-gfm`. This basically provides for transforming the input and output of the editor from / to Markdown.

I also added a fix for some strange default behavior in Turndown (using ` *` to denote lists instead of `- `).

test out the editor on the `/markdown-editor` page. Fiddle with the stuff in the editor and check out the output below, and just generally make sure nothing too weird is going on. Should be pretty straightforward I think!

Screenshot from 2021-02-25 16-16-31


static/js/test_util.ts
```diff
@@ -9,3 +12,43 @@ export const defaultFormikChildProps: FormikState<any> = {
   status:       null,
   submitCount:  0
 }
+
+export const TEST_MARKDOWN = `## A heading
```
gsidebo, line 16: What do you think about implementing this as a separate `.md` file? `static/js/resources/example-markdown.md` or something like that. I'm thinking we would keep this constant but set the value by importing those file contents. You'd know better than me regarding what that would take in webpack, and there might be some drawbacks I'm not seeing

static/js/lib/markdown.ts
```diff
@@ -0,0 +1,46 @@
...
+      // @ts-ignore
+      const start: string = parent.getAttribute("start")
+      const index = Array.prototype.indexOf.call(parent.children, node)
+      prefix = `${start ? Number(start) + index : index + 1}. `
+    }
+    return (
+      prefix + content + (node.nextSibling && !/\n$/.test(content) ? "\n" : "")
+    )
+  }
+})
```
gsidebo, line 40: I'm guessing we need to add some custom markdown conversion if we add a new plugin (shortcode link, for example). Is that right? If so, will you be adding some documentation for that like you did for adding a new plugin? That can come later, just wanted to make sure we have a plan for it

gsidebo (changes requested):

static/js/lib/markdown.ts
aliceriot: yeah, it's coming later (working on it now but don't want to hold up basic support) and I'll expand the plugin documentation to show how the implementation works.

static/js/test_util.ts
aliceriot: this file isn't included in webpack so we should be good to do that, I'll give it a try.

aliceriot: ahhh this is more difficult than it should be, so if it's alright with you I think it's best to keep this all-typescript

static/js/lib/markdown.ts
```diff
@@ -0,0 +1,46 @@
...
+  ) => {
+    content = content
+      .replace(/^\n+/, "") // remove leading newlines
+      .replace(/\n+$/, "\n") // replace trailing newlines with just a single one
+      .replace(/\n/gm, "\n    ") // indent
+
+    let prefix = `${options.bulletListMarker} `
+    const parent = node.parentNode
+    if (parent && parent.nodeName === "OL") {
+      // @ts-ignore
```
noisecapella, line 31: Does typescript understand `getAttribute`?

noisecapella:
I tried it out and LGTM. I'll leave it for @gsidebo to approve

static/js/lib/markdown.ts
aliceriot: it gives this error without the `// @ts-ignore`: `Property 'getAttribute' does not exist on type 'Node & ParentNode'`

gsidebo (changes requested):
ss 2021-03-01 at 16 49 43

When I nest list items, the output seems to show them all at the same depth. Is that just a quirk with the output? If not, that seems like a bug

aliceriot:
just looked into this and I think the conversion is actually working right, but the issue is actually the display in the 'output' section. I made a quick tweak to prevent it from swallowing leading spaces and it looks like this:

Screenshot from 2021-03-01 17-26-25

I had `white-space: pre-line` set on it which squashes leading whitespace :grimacing:

gsidebo (approved):
Looks great, and thanks for adding those nested bullets to the test markdown! 👍  👍

PR #91, opened 2021-02-26
Turn off pagination for website content API

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Part of #52 

#### What's this PR do?
Removes pagination so all website content results will be in one request. This will simplify work in #52. I'm not 100% sure but I doubt there will be too much data being sent here since it's only returning the `uuid`, `title`, and `type` for each item, not the content itself.

Go to `/api/websites/website-name-here/content/`. You should not see a `count` or `results` field, just the results themselves.

### DIFF
--- a/websites/views_test.py
+++ b/websites/views_test.py
@@ -670,14 +670,14 @@ def test_websites_content_list(drf_client, filter_type, permission_groups):
         ),
         {"type": filter_type},
     )
-    assert resp.data["count"] == (num_pages if filter_type else num_pages + 1)
+    assert len(resp.data) == (num_pages if filter_type else num_pages + 1)
 
     for idx, content in enumerate(
         reversed(sorted(contents, key=lambda _content: _content.updated_on))
     ):
-        assert content.title == resp.data["results"][idx]["title"]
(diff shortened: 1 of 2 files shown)


websites/views_test.py
```diff
@@ -650,14 +650,14 @@ def test_websites_content_list(drf_client, filter_type, permission_groups):
         ),
         {"type": filter_type},
     )
-    assert resp.data["count"] == (num_pages if filter_type else num_pages + 1)
+    assert len(resp.data) == (num_pages if filter_type else num_pages + 1)
```
mbertrand, line 653: Minor nitpick: This isn't really a number of pages, maybe change the variable name to something like `num_results`?

mbertrand (changes requested):
Works great.  The website with the largest number of WebsiteContent objects I had on my local instance is `16-01-unified-engineering-i-ii-iii-iv-fall-2005-spring-2006` with 1121, but the response size still seemed reasonable, about 230KB.

websites/views_test.py
noisecapella: Ah, makes sense

mbertrand (approved):
👍

PR #95, opened 2021-03-01
Add metadata to list of editable fields

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Part of #52 

#### What's this PR do?
Adds `metadata` field to website content serializers

Go to the website content API and try adding or editing with content in the metadata field

### DIFF
--- a/websites/serializers.py
+++ b/websites/serializers.py
@@ -139,8 +139,9 @@ class WebsiteContentSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = WebsiteContent
-        fields = ["uuid", "title", "type"]
-        read_only_fields = ["uuid", "type"]
+        read_only_fields = ["uuid", "title", "type"]
+        # See WebsiteContentCreateSerializer below for creating new WebsiteContent objects
+        fields = read_only_fields
 
 
 class WebsiteContentDetailSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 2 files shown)


websites/serializers.py
```diff
@@ -163,6 +164,6 @@ def create(self, validated_data):
 
     class Meta:
         model = WebsiteContent
-        fields = ["type", "title", "markdown", "uuid"]
         # we want uuid in the returned result but we also want to ignore any attempt to set it on create
         read_only_fields = ["uuid"]
+        fields = read_only_fields + ["type", "title", "markdown", "metadata"]
```
mbertrand, line 169: Are there separate`WebsiteContentCreateSerializer` and `WebsiteContentDetailSerializer` serializers just to make sure that once a `WebsiteContent` object is created, it's type can't be changed?  Maybe could use the same serializer but [adjust its `update` method to ignore `type` if changed](https://github.com/mitodl/ocw-studio/blob/master/websites/serializers.py#L66-L67)?  Either way works though.

noisecapella: Yes, that's the intent. Adjusting the update method works but I'd rather not make the change since it would be extra work at this point

mbertrand (approved):
👍

PR #97, opened 2021-03-01
Use 'string' for description instead of 'markdown'

noisecapella:
#### Pre-Flight checklist

- [x] Migrations

None

#### What's this PR do?
Updates the `description` field to be a `string` type


gsidebo (changes requested):
We only want to update the "description" field in the "Resource" section of the course site starter specifically (which we can target with the constant, something like `COURSE_STARTER_SLUG` IIRC). Nothing else needs to be touched

gsidebo (approved):
👍

Issue #92, opened 2021-02-26
Implement basic file upload


gsidebo:
[Wireframe](https://lucid.app/lucidchart/6be33dd8-f7db-47f8-8af8-338255f23c3c/edit?shared=true&page=gw.UNA5605io#?folder_id=home&browser=icon)

The goal of this issue is to enable file upload for resources in a way that ensures the file will be included in the generated static site.

In terms of UI, we will go with the simplest option to start out. We should be able to use a basic HTML input with type="file", and a simple text caption indicating the uploaded file if there is one. If there's an even simpler approach, we can go with that. A more UX-friendly file uploader can be worked on in a separate issue when the need arises.

**Acceptance criteria:**
- A site config field with type "file" should result in the file input being rendered in the relevant authoring UI
  - If a file has already been uploaded for the resource, the file name (with no path or uid) should be shown. Note that if we use the basic HTML input, we can't populate the value, so this can just be simple text underneath the input.
- When a file is uploaded and the resource record is saved, the file is saved to the configured Django storages backend 
  - To start out, we can namespace the files by site uuid and file type, and we should make sure to avoid name collisions. One way to do that would be to prefix each file with a separate uuid. Example: uploading `syllabus.pdf` --> file saved at `<storages destination>/<site uuid>/pdf/<uuid generated for file>_syllabus.pdf`.
- If a file was already uploaded for the particular resource, the newly-uploaded file should be saved and the old one deleted.
- If the file fails to upload, an error message should be shown under the input in the same way that a validation error would be shown.
- In the WebsiteContent record, the "file" value in the metadata should be set to the uploaded file path relative to the Storages root destination, or relative to the site folder within that storages destination (`<site uuid>/pdf/<file uuid>_syllabus.pdf` or `pdf/<file uuid>_syllabus.pdf` in the example above)
- S3 buckets and credentials created by devops, and credentials added to Vault
- Salt-ops PR(s) created and merged so that CI/RC/Prod are all ready to upload to the relevant buckets

**Not needed yet:**
- External URLs allowed instead of uploaded files

mbertrand:
@shaidar for the S3 buckets, looks like [they were already created](https://github.com/mitodl/salt-ops/issues/1343) but do you know what their names are?  There is an `AWS_STORAGE_BUCKET_NAME` env variable for the app but no values for it yet on heroku.

shaidar:
@mbertrand The bucket and creds should be all set on CI and RC

[...]

PR #100, opened 2021-03-04
File uploads API for WebsiteContent 

mbertrand:
#### Pre-Flight checklist
- [x] Migrations
- [x] Testing
  - [x] Code is tested

Related to #92 

#### What's this PR do?
- Installs `django-storages`
- Adds a `FileField` to the `WebsiteContent` model
- Adds the file field to the `WebsiteContent` serializers

-Go to `http://localhost:8043/api/websites/<WEBSITE_NAME>/content/`
-Use the form at the bottom to create a new object of type `resource` and choose a file to upload.
-The response should contain a link to the uploaded file, click on it and make sure the file loads.
-Set `OCW_STUDIO_USE_S3=True`, copy `AWS_` settings from CI, and restart docker containers
-Upload a file again.  This time the link should be an S3 URL.  Currently you won't have access to that URL but you can confirm the file exists in a shell:

```python
import boto3
s3 = boto3.resource("s3") 
obj = s3.Object(settings.AWS_STORAGE_BUCKET_NAME, key="<WEBSITE_NAME>/<CONTENT_OBJECT_UUID>_<FILENAME>"
print(obj.content_type, obj.last_modified, obj.content_length)
```
    
Some things which this PR doesn't address:
  - File type validation (not sure what if any file type restrictions should be in place)
  - Syncing uploads to OCW preview/live buckets
  - Viewing the file when `OCW_STUDIO_USE_S3=True` (The S3 URLs are currently not accessible).
  - Importing legacy OCW files


noisecapella:
Should we restrict file uploads to `resource` typed content?

websites/views_test.py
```diff
@@ -722,6 +723,36 @@ def test_websites_content_create(drf_client, permission_groups):
     assert resp.data["uuid"] == str(content.uuid)
 
 
+def test_websites_content_create_with_upload(drf_client, permission_groups):
```
noisecapella, line 726: Can you add another test for updating an existing resource with a new file?

noisecapella (changes requested):
Functionality works great, just a comment and the question above

noisecapella (approved):
Works great!

PR #104, opened 2021-03-09
update ocw import to match new ocw-to-hugo output structure

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Related to https://github.com/mitodl/ocw-to-hugo/issues/214

#### What's this PR do?
This PR alters the way `ocw-studio` parses `ocw-to-hugo` output to match the new output structure as detailed in [this](https://github.com/mitodl/ocw-to-hugo/pull/216) PR.  Some issues were also corrected with the way the import script determines the UID of a page.  Rather than trying to grab the UID from the menu, it's now taken from the `uid` property on the front matter of any given page.  The "page or resource" question is answered by checking the `layout` property and comparing it to known layouts for each type.  Root level pages also now correctly set the course home page as their parent.

I'm not exactly sure how this could be manually tested aside from the unit tests, as the import script is set up to pull directly from an s3 bucket and cannot be redirected at a local folder.  Either way, the unit tests should be checked locally against the new test data.


mbertrand:
Would it be possible to reduce the number of test files?

gumaerc:
@mbertrand Hypothetically the new content structure should be in the bucket.  I just checked the build server's deploy script and it runs the s3 sync with the `--delete` flag so the new structure should be there.  I wanted to get this PR up preemptively so that import isn't broken for too long.  I think we can reduce the amount of test data, I'll work on that tomorrow.

mbertrand (changes requested):
I'm getting the error `botocore.errorfactory.NoSuchKey: An error occurred (NoSuchKey) when calling the GetObject operation: The specified key does not exist.` on this line: 

```
celery_1  |   File "/src/ocw_import/api.py", line 143, in import_ocw2hugo_course
celery_1  |     get_s3_object_and_read(bucket.Object(f"{path}/data/course.json")).decode()
```

I added some logging statements to see what the paths were:

```
8-972-exoplanets-and-brown-dwarfs-spring-2007/data/course.json/data/course.json
16-30-feedback-control-systems-fall-2010/data/course.json/data/course.json
1-00-introduction-to-computers-and-engineering-problem-solving-spring-2012/data/course.json/data/course.json
12-517-dynamics-of-complex-systems-complexity-in-ecology-spring-2000/data/course.json/data/course.json
17-478-great-power-military-intervention-fall-2013/data/course.json/data/course.json
```

These files, without the extra `/data/course.json` appended to them, do exist in the bucket.

ocw_import/api.py
```diff
@@ -67,19 +81,19 @@ def import_ocw2hugo_content(bucket, prefix, website):  # pylint:disable=too-many
             parent = None
             if len(s3_content_parts) >= 1:
                 content_json = yaml.load(s3_content_parts[0], Loader=yaml.Loader)
+                layout = content_json.get("layout", None)
                 menu = content_json.get("menu", None)
+                uuid = content_json.get("uid", None)
```
mbertrand, line 86: I should be able answer this myself once I can complete a successful run, but do you know if `menu_values.get("identifier")` will ever differ from `content_json.get("uid")`?  If so, we might end up with dupes unless the old imports from previous runs of this command are deleted first.

gumaerc:
@mbertrand should be ready for another look

ocw_import/api.py
gumaerc: They are pulled from the same property, the only difference I can think of might possibly be course home pages with no uid property, they would maybe be undefined but I'll have to check on that.

gumaerc:
@mbertrand as far as the test data goes, I think the major difference is in the amount of PDF's recognized.  The files themselves are all fairly small.  What I'm wary about is that if we remove some of the files representing the PDF's that there will be links referencing them that will need to be edited out.  If it's a problem maybe we should just find better test courses?

mbertrand (approved):
👍

PR #108, opened 2021-03-11
Add site listing at site dashboard

### LINKED ISSUE #53: Update factory values for WebsiteFactory.metadata
Not a high priority but we should make sure to update this factory to produce useful metadata, once we have a better idea what the schema is for that value

### LINKED ISSUE #106: Site listing page
As a user of ocw-studio, I'd like to see a simple listing of sites with links to view each site

#### Designs and Mockups
https://xd.adobe.com/view/6074f9dd-fa3f-4a6c-a1fe-4d70e7d0ca78-3527/screen/4a33bacb-50b2-49d4-bd38-b5eb9bbf0cff/specs/

#### Acceptance Criteria:

- [ ] Site listing page
- [ ] Handle pagination gracefully using the existing paginated API at /api/websites (ie, we should not crawl the API and paginate entirely on the frontend)

#### Out of Scope

- Add site page (#107)

### LINKED ISSUE #107: Move create site form to a new page
As a user of ocw-studio, I'd like to add a new site to ocw-studio. The functionality to create a site is already implemented, it just needs to be located on its own page.

#### Designs and Mockups
https://xd.adobe.com/view/6074f9dd-fa3f-4a6c-a1fe-4d70e7d0ca78-3527/screen/43fefb77-710d-4123-92a9-4d4e9646303f/specs/

#### Acceptance Criteria:

- [ ] Link to add site page from the site listing page (to be implemented in #106)
- [ ] Move the existing `SiteForm` to a new URL (TBD: what URL?)
- Do we have information on a user's permission to create a new site? If so, we should not show a button to add the site. If we don't have that information, we should add it but that may be outside the scope of this issue

noisecapella:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Fixes #106 
Fixes #107 
Fixes #53 

#### What's this PR do?
Implements the site listing with pagination, and moves the existing add new site UI to its own page at `/new-site/`.

 - Go to `/sites/`. You should see a list of courses with a `>` link at the bottom to go to next page. Click that link and you should see content show "Loading" and then show the next page of results. On the second page you should see a `<` link to go back to the previous page. The page should not do a full refresh but you should see the URL change to include `offset=10` for whatever the offset is expected to be. 
 - There should be an "Add new" link which goes to a form to create a new site at `/new-site/`
 - Links to the site detail pages should work as well.

#### Reviewer notes
`SitesDashboard.tsx` and its test file were renamed to `SiteCreationPage.tsx` with only certain variables renamed like `Website` to `WebsiteDetail`. Then I made a new file also called `SitesDashboard.tsx` which has most of the new functionality for this PR. This means the diff looks strange but in general you can ignore `SiteCreationPage.tsx` and treat `SitesDashboard.tsx` as a brand new file instead of an updated one (and same for the test files).

Screenshot_2021-03-11 ocw_studio


static/js/pages/SiteCreationPage.tsx
```diff
@@ -0,0 +1,88 @@
+import * as React from "react"
```
noisecapella, line 1: This is just what used to be SiteDashboard.tsx. No changes were made here except for the renamed function

static/js/pages/SiteCreationPage.test.tsx
```diff
@@ -0,0 +1,167 @@
+const mockUseRouteMatch = jest.fn()
```
noisecapella, line 1: This is just what used to be `SitesDashboard.test.tsx` but renamed and with other variables renamed like `Website` => `WebsiteDetail`

static/js/types/websites.ts
```diff
@@ -40,7 +40,10 @@ export interface Website {
   source: string | null
   starter: WebsiteStarter | null
   metadata: any
-  is_admin?: boolean // eslint-disable-line
+}
+
+export interface WebsiteDetail extends Website {
+  is_admin: boolean // eslint-disable-line
```
aliceriot, line 46: is this key not present on the `Website` objects always for performance reasons or something? If possible it might be more straightforward to stick with just the one `Website` type unless we really need the separation or need to specifically keep track of when this value is present (rather than just duck-typing when it's truthy).

aliceriot: I did a little experiment locally and it doesn't seem to cause any issues to put `is_admin?` back on `Website` and switch `WebsiteDetail` things over to just `Websites`

static/js/query-configs/websites.ts
```diff
@@ -36,9 +38,32 @@ export const getTransformedWebsiteName = (
...
+type WebsitesListing = Record<string, WebsiteListingResponse>
+export const websiteListingRequest = (offset: number): QueryConfig => ({
+  url:       siteApiListingUrl(offset),
+  transform: (body: WebsiteListingResponse) => ({
+    websitesListing: {
+      [`${offset}`]: body
+    }
+  }),
+  update: {
+    websitesListing: (prev: WebsitesListing, next: WebsitesListing) => ({
```
aliceriot, line 56: would it be worthwhile to put the website objects themselves into the `websites` entity for consistency? the pages could be stored here as arrays of website IDs or something along those lines. I think we did that somewhere in Open.

aliceriot (changes requested):
two questions

static/js/types/websites.ts
noisecapella: Sure, I'll make the change with `is_admin?`. I do think it is left off the listing for performance though there is probably a way to workaround those issues if we need to

static/js/query-configs/websites.ts
noisecapella: Sure, I'll look into that

noisecapella: Actually I'm not sure this is feasible since the website objects are slightly different between the listing and detail view (the `is_admin` field). If we populated the detail view with websites fetched in the list view, they would all implicitly have a falsy value for `is_admin`.

aliceriot: won't the detail view make the request to that endpoint and fetch that detail record though?

noisecapella: Right, I guess we can set `force: true` to make sure the new data is always up to date. I'll try that out

noisecapella: We could also add `is_admin` to the list view and just fix the performance issues that we find. But I don't know if there are plans to add more data to the website serializer down the line

noisecapella:
@aliceriot I made the change to have the listing stored as a series of site names. What do you think of using the selector in this way?

static/js/selectors/websites.ts
```diff
@@ -8,8 +8,21 @@ import { ReduxState } from "../reducers"
...
+    memoize((offset: number) => {
+      const response = lookup.listing[offset] ?? {}
+      const names = response?.results ?? []
+      const sites = names.map((name: string) => lookup.details[name])
+
+      return {
+        ...response,
+        results: sites
+      }
+    })
```
aliceriot, line 25: this looks fine to me I think, you could also use the `getWebsiteDetailCursor` as an input selector here if you wanted, but that wouldn't make it all *that* much more simple (just perhaps a bit more elegant, you could then do something like `names.map(cursor)` I think)

noisecapella: I made an attempt though I'm uncertain that it's an improvement. What do you think? Here's the diff: https://github.com/mitodl/ocw-studio/pull/108/commits/ac1c55c3841619205a149133dffae73235b569c8

static/js/selectors/websites.ts
```diff
@@ -12,6 +12,24 @@ export const getWebsiteDetailCursor = createSelector(
...
+        getWebsiteDetailCursor(state)(name)
+      )
+
+      return {
+        ...response,
+        results: sites
+      }
+    })
+)
+
```
aliceriot, line 32:
```suggestion
export const getWebsiteListingCursor = createSelector(
  (state: ReduxState) => state.entities?.websitesListing ?? {},
  getWebsiteDetailCursor,
  listing, websiteDetailCursor => {
    memoize((offset: number) => {
      const response = listing[offset] ?? {}
      const names = response?.results ?? []
      const sites = names.map(websiteDetailCursor)

      return {
        ...response,
        results: sites
      }
    })
  )
)
```

this is how I'd suggest doing it - reselect lets us compose selectors in a pretty fluid way like this

noisecapella: Oh neat

aliceriot (approved):
LGTM! :+1:

PR #114, opened 2021-03-15
fix bug relating to turndown 'li' rule

### LINKED ISSUE #113: Markdown editor fails to add new list items
### Steps to Reproduce

Go to `/markdown-editor`, and click on the text next to one of the list items. You should be able to bold text or make other changes. However when you press enter, it will create a second CKEditor window instead of adding a new `li` element

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #113

#### What's this PR do?

this fixes an issue with the patch I made to the turndown rule for list items. Basically, I should have patched it by modifying the rule in place, instead of adding another rule. I updated it to modify the rule in place by changing the `replacement` function for `li` tags instead of adding a new rule, and it fixes the issue.

You should be able to add new list items to the editor without error.

### DIFF
--- a/static/js/lib/ckeditor/turndown.test.ts
+++ b/static/js/lib/ckeditor/turndown.test.ts
@@ -25,4 +25,17 @@ describe("turndown service", () => {
     })
     turndownTest("<figure></figure>", "This rule is being used")
   })
+
+  it("supports a two-item list", () => {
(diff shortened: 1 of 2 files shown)


noisecapella (approved):
Works great :+1:

PR #134, opened 2021-03-22
Add name to site creation form, but autopopulate based on title input

### LINKED ISSUE #128: misleading error message when trying to create a site with non-latin characters
### Steps to Reproduce

try a string like "这是我的“ or "русский язык"

### Expected Behavior

If we disallow certain characters there should be an error that says something like "Sorry, you've got to stick to latin characters here"

### Actual Behavior

Screenshot from 2021-03-19 12-35-38

it incorrectly says the issue is that the field isn't filled in.

Strangely, if you add some text in latin characters, so the title is somehting like "это мой website" or "这是我的website" then it seems to work alright, and the site name works correctly. Only the latin characters show up in the URL slug though:

Screenshot from 2021-03-19 12-40-12

mbertrand:
Blocked until a final decision is made on whether or not to allow non-ascii urls, not a high priority yet.

#### Pre-Flight checklist

  - [ ] Desktop screenshots
  - [ ] Tag @ferdi or @pdpinch for review
- [ ] Testing
  - [ ] Code is tested

Closes #128 

#### What's this PR do?
Adds a `name` input field to the site creation form, autopopulates it with a slug based on the `title` field input, and prevents non-ascii characters from being used.

Go to the form for creating a new site.  As you enter a title, the name field below should start to populate.  It should contain only lowercase a-z, 0-9, and - characters.  If you try to add other characters you should see a validation error.  Submit the form, the site should be created with that name/slug.

Initially thought about allowing non-ascii characters in the name by passing `allow_unicode=True` to `django.utils.text.slugify`, but after talking with Ferdi & Nathan decided to disallow it and use this approach instead.

#### Screenshots
Screen Shot 2021-06-14 at 3 22 36 PM


mbertrand:
This is really out of date so I'm going to close it until the related issue becomes a priority.

PR #144, opened 2021-03-23
fix issue with turndown when editor is instantiated twice

### LINKED ISSUE #142: issue w/ figure turndown rule when editor is instantiated, destroyed, and then instantiated again
### Steps to Reproduce

1. Visit a websites 'page' content page
2. Edit a page object (opening it in the drawer)
3. Close the drawer
4. Open it again and try to add a youtube video to the editor (either by pasting a youtube link in directly or using the little widget in the toolbar)

### Expected Behavior

You should be able to add a youtube video to the editor.

### Actual Behavior

There is an error relating to multiple rules having been defined for translating `<figure>` tags to Markdown. This is occuring because right now our custom turndown rules are always added to the `turndownService` every time the editor is instantiated. We need to modify the CKEditor turndown in order to re-use some of the code that they wrote, so we're basically modifying that exported object. If we add some sort of guard to ensure that this is only done once we should be good.

### Stacktrace

```
Uncaught Error: should only be a single matching rule
    at Object.Dk.rules.blankRule.replacement (turndown.ts:21)
    at pk.bk (turndown.es.js:875)
    at turndown.es.js:834
    at NodeList.reduce (<anonymous>)
    at pk.mk (turndown.es.js:827)
    at pk.turndown (turndown.es.js:739)
    at Lk.html2md (Markdown.ts:68)
    at Lk.toData (Markdown.ts:46)
    at Yh.stringify (datacontroller.js:216)
    at Yh.get (datacontroller.js:198)
```

aliceriot:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

closes #142 
 
#### What's this PR do?

this fixes an issue relating to adding turndown rules - basically, when the editor is instantiated and destroyed more than once while the application is running we need to have a way to ensure that we only set turndown rules once, lest we run into issues. I added a little check to make sure this isn't done.

Follow the repro steps on the issue, and confirm that this fixes the problem.

### DIFF
--- a/static/js/lib/ckeditor/plugins/Markdown.test.ts
+++ b/static/js/lib/ckeditor/plugins/Markdown.test.ts
@@ -1,6 +1,7 @@
 import Markdown, { MarkdownDataProcessor } from "./Markdown"
 
 import { createTestEditor, markdownTest } from "./test_util"
+import { turndownService } from "../turndown"
 
 import { TEST_MARKDOWN, TEST_HTML } from "../../../test_constants"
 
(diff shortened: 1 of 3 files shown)


noisecapella (approved):
Confirmed this fixes the issue

PR #161, opened 2021-03-29
Implement conditional fields

### LINKED ISSUE #159: Site config: Dependent field
As a cours/website dev, I would like to tell studio to show / hide fields based on the value of some other fields

Add a `condition` attribute to the field definition.  `condition` accepts 2 options : 
`field` : the name of the field this field depends on 
`equals`:  the value of the field this filed depends on

For example, in this config file

```
collections:
  - name: fields
    label: Fields
    files:
      - label: Dependent
        name: dependent
        file: config/dependent.yml
        fields:
         - { label: What kind, name: kind, widget: select,
             options: [internal, external]
           }
         - {
             label: Link to,
             name: linkto-internal,
             widget: select,
             condition: {field: kind, equals: internal}
           }
         - {
             label: Link to,
             name: linkto-external,
             widget: string,
             condition: {field: kind, equals: external}
           }

```

Only of  `linkto-internal` or `linkto-external` form element  will show based on the value selected from “kind” dropdown

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #159

#### What's this PR do?
Adds support for conditional fields. If the condition property is set for in the field config, 

 - Run `./manage.py override_site_configs`
 - Go to the form to add a new piece of site metadata for some course. Type "extra" in the title field and you should see an extra markdown field appear. If you click save and the field exists, the content should be saved and retrieved correctly.
 - Change the title field so that it's not "extra" anymore, then click save. Click "Edit" to see the content you saved, and type "extra" in the title field. You should see the markdown field appear but there should be no text.
 - Adding and editing content should otherwise work the same as it did before. There shouldn't be any regressions.

### DIFF
--- a/localdev/starters/site-config-override.yml
+++ b/localdev/starters/site-config-override.yml
@@ -38,3 +38,8 @@ course:
           "widget": "boolean",
           "help": "Whether the course includes fries or not."
           }
+        - { "label": "Tags", "default": ["Design"], "max": 3, "min": 1, "multiple": true, "name": "tags", "options": ["Design", "UX", "Dev"], "widget": "select" }
+        - { label: "Align Content", name: "align", widget: "select", options: ["left", "center", "right"]}
+        - {
+          label: "Extra information",
+          name: "extra", widget: "markdown", default: "Extra details go here", "condition": { field: title, equals: extra } }
--- a/localdev/starters/site-config-override.yml
+++ b/localdev/starters/site-config-override.yml
@@ -38,8 +38,6 @@ course:
           "widget": "boolean",
           "help": "Whether the course includes fries or not."
           }
-        - { "label": "Tags", "default": ["Design"], "max": 3, "min": 1, "multiple": true, "name": "tags", "options": ["Design", "UX", "Dev"], "widget": "select" }
-        - { label: "Align Content", name: "align", widget: "select", options: ["left", "center", "right"]}
         - {
           label: "Extra information",
           name: "extra", widget: "markdown", default: "Extra details go here", "condition": { field: title, equals: extra } }
--- a/static/js/types/websites.ts
+++ b/static/js/types/websites.ts
@@ -8,7 +8,7 @@ export enum WidgetVariant {
   String = "string",
   Select = "select"
 }
-export interface FieldCondition {
+export interface FieldValueCondition {
   field: string
   equals: string
 }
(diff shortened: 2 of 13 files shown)


noisecapella:
@aliceriot open to suggestions on a more useful example config for this

aliceriot:
it might be good to have a new conditional field and keep the current markdown field in place, that would be my only suggestion for the override config

noisecapella:
@aliceriot I updated it so that typing `extra` in the title pops up a different field

localdev/starters/site-config-override.yml
```diff
@@ -38,3 +38,8 @@ course:
           "widget": "boolean",
           "help": "Whether the course includes fries or not."
           }
+        - { "label": "Tags", "default": ["Design"], "max": 3, "min": 1, "multiple": true, "name": "tags", "options": ["Design", "UX", "Dev"], "widget": "select" }
+        - { label: "Align Content", name: "align", widget: "select", options: ["left", "center", "right"]}
```
aliceriot, line 42: did you mean to duplicate the 'Tags' and 'Align Content' fields here?

static/js/types/websites.ts
```diff
@@ -8,6 +8,10 @@ export enum WidgetVariant {
   String = "string",
   Select = "select"
 }
+export interface FieldCondition {
```
aliceriot, line 11: is `FieldCondition` the right name for this? I'm not sure if there's been discussion about making it more general, but as I'm reading it now it looks to me like this limited to checking that a field equals a given value. Maybe the name should be `FieldValueCondition` ? I'm not sure, what do you think?

aliceriot (changes requested):
works great, just had a few questions

localdev/starters/site-config-override.yml
noisecapella: Probably just a merge error

static/js/types/websites.ts
noisecapella: I'll rename it to `FieldValueCondition` I don't know what the longer term plans are for conditional fields, if it's just checking for equality or if there's other stuff down the line. I imagine it's probably for select fields, something like if country == "US" then show the state dropdown, in which case I doubt it will get more complicated.

aliceriot (approved):
LGTM! :+1:

PR #168, opened 2021-04-01
add Boolean widget

### LINKED ISSUE #145: boolean widget
As a person editing the site metadata, I'd like to have a boolean input for toggling true / false values.

#### Acceptance Criteria:

- [ ] Implement widget a radio input
- [ ] Validation on backend
- [ ] Schema validation should support the new widget. See schema here: https://docs.google.com/document/d/14diQGCrhlIRhemY7OAl__0uPCocGcbp5X64ZUF0BEIA/edit#

aliceriot:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #145 

#### What's this PR do?

This adds a new widget type for setting a Boolean value in the site config. I made a few other changes to parts of the code that I touched when implementing it too:

- I created a new `enum` type called `WidgetVariant` that captures the different possible widgets. The enum will be transformed into string constants at compile time by typescript.
- I slightly refactored how props get from the `ConfigField` down to the `SiteContentField`, creating a `widgetExtraProps` function in `lib/site_content.ts` which switches on the `WidgetVariant` and grab the extra props we want out of the `ConfigField` object for each type.

Anyhow, I think these changes increase clarity a little bit :)

You can use the `override_site_configs` management command to add a Boolean field to the site config (I checked in a change to add one to the `localdev/starters/site-config-override.yml` file). Then check out the metadata config section and make sure you can edit the boolean field.

Screenshot from 2021-04-01 12-38-43

### DIFF
--- a/static/js/components/SiteContentListing.test.tsx
+++ b/static/js/components/SiteContentListing.test.tsx
@@ -1,5 +1,6 @@
 const mockUseRouteMatch = jest.fn()
 
+import React from "react"
 import { act } from "react-dom/test-utils"
 
 import SiteContentListing from "./SiteContentListing"
@@ -33,7 +34,14 @@ jest.mock("react-router-dom", () => ({
 }))
 
 // ckeditor is not working properly in tests, but we don't need to test it here so just mock it away
-jest.mock("./widgets/MarkdownEditor", () => "div")
+function mocko() {
+  return <div>mock</div>
+}
+
+jest.mock("./widgets/MarkdownEditor", () => ({
+  __esModule: true,
+  default:    mocko
+}))
 
 describe("SiteContentListing", () => {
   let helper: IntegrationTestHelper,
--- a/static/js/components/forms/SiteContentField.tsx
+++ b/static/js/components/forms/SiteContentField.tsx
@@ -2,33 +2,29 @@ import { ErrorMessage, Field } from "formik"
 import React from "react"
 
 import { FormError } from "./FormError"
-import { componentFromWidget } from "../../lib/site_content"
+import { componentFromWidget, widgetExtraProps } from "../../lib/site_content"
 
-import { ConfigField } from "../../types/websites"
+import { ConfigField, WidgetVariant } from "../../types/websites"
 
 interface Props {
   field: ConfigField
   setFieldValue?: (key: string, value: File | null) => void
 }
 
+/**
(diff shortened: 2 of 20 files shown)


static/js/components/forms/SiteContentField.tsx
```diff
@@ -2,33 +2,29 @@ import { ErrorMessage, Field } from "formik"
...
-  default:
-    extraProps = {}
-    break
+  const extraProps = widgetExtraProps(field)
+
+  if (
+    field.widget === WidgetVariant.File ||
+    field.widget === WidgetVariant.Boolean
+  ) {
+    extraProps.setFieldValue = setFieldValue
```
noisecapella, line 27: Should we just set this prop for every field? I assume it would just be ignored for the ones where it isn't relevant

noisecapella (commented):
Functionality looks good, just a couple questions

static/js/components/SiteContentListing.test.tsx
```diff
@@ -33,7 +34,14 @@ jest.mock("react-router-dom", () => ({
 }))
 
 // ckeditor is not working properly in tests, but we don't need to test it here so just mock it away
-jest.mock("./widgets/MarkdownEditor", () => "div")
+function mocko() {
```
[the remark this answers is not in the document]
aliceriot: It was an issue related to the `minimal` prop we use for the MarkdownEditor. Basically, React throws an error if you pass a random unexpected prop to an HTML element (like doing `<div foobar={bizBaz} />`), and we were mocking the MarkdownEditor component with a simple `div`. So then the extra props like `setFieldValue` and `minimal` were, because of the way that Formik works, being passed through to the `div` and causing an error. So to fix it I just defined a little custom component to use as a mock instead, since a custom component can receive any prop at runtime and won't throw any error.

Does that make sense? It's sort of hard to explain :thinking:

static/js/components/forms/SiteContentField.tsx
aliceriot: that was my thought too, but I ran into an issue related to the issue I mentioned above about passing unexpected props to HTML elements. The `input` tag doesn't expect to receive a `setFieldValue` prop, so passing that in to every `<Field>` that we instantiate causes some runtime errors.

noisecapella (approved):

PR #172, opened 2021-04-02
Changed config to user 'folder'/'files' collections types

gsidebo:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
  - [ ] Tag @ferdi or @pdpinch for review
- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested
- [ ] Settings

(Required)

#### What's this PR do?
(Required)

(Required)

(Optional)

(Optional)

(Optional)

(Optional)


noisecapella:
This looks a little strange to me. Do we have a mockup for this yet? If not I'm fine with adjusting this in a separate PR 

Singleton:
Screenshot from 2021-04-05 14-09-36

Collection:
Screenshot from 2021-04-05 14-09-04

static/js/components/SiteAddContent.tsx
```diff
@@ -16,28 +16,65 @@ import {
...
 
+function findEditableConfigItem(
+  config: WebsiteStarterConfig,
+  type: string,
+  subtype: string | null
+): EditableConfigItem | null {
+  const topLevelConfigItem: TopLevelConfigItem | null =
+    config.collections.find(
+      (config: TopLevelConfigItem) => config.name === type
+    ) || null
```
noisecapella, line 40: Can we use `??` instead of `||` here and below? It wouldn't make a difference in functionality here but it would be more explicit

gsidebo:
@noisecapella No mockup for this yet. If we had a similarly-styled "Add" button for the site metadata in your screenshot, it would need to be inline with the title. Our example config only has a single item in that "files" config section, so there's only one line, but remember that it can have multiple items defined, and they would all be rendered in that listing. If we had a fully-styled "Add" button on each of those lines, and those were mixed in with plain text "Edit" links, I'd argue that would be less consistent and less desirable. Consistency was my main goal with this UI

@Ferdi any thoughts on George's comment above?

static/js/components/SiteAddContent.tsx
```diff
@@ -16,28 +16,65 @@ import {
...
+  if (!topLevelConfigItem) {
+    return null
+  }
+  if (!subtype) {
+    if ("folder" in topLevelConfigItem) {
```
noisecapella, line 45: Can we use `isRepeatableCollectionItem` here?

static/js/query-configs/websites.ts
```diff
@@ -235,57 +236,71 @@ export const createWebsiteCollaboratorMutation = (
...
   WebsiteContentListItem
 >
 type WebsiteContentListing = Record<string, string[]> // website name to list of uuids
 export const contentListingKey = (
-  name: string,
-  type: string,
-  offset: number
-): string => JSON.stringify([name, type, offset])
+  listingParams: ContentListingParams
+): string => JSON.stringify(Object.values(listingParams))
```
noisecapella, line 245: I think we need to specify the values to be stringified explicitly so that the key doesn't change with a different insertion order for `listingParams`. Otherwise you could have something like `[ name, type, offset ]` and `[type, name, offset]` which are two different keys for the same object.

static/js/util/factories/websites.ts
```diff
@@ -38,25 +41,60 @@ export const makeWebsiteConfigField = (
...
+  label:  casual.word,
+  name:   name || casual.word
+})
+
+export const makeTopLevelConfigItem = (name?: string): TopLevelConfigItem => ({
+  fields:   cloneDeep(exampleFields),
+  name:     name || casual.word,
+  label:    casual.word,
+  category: casual.word,
+  ...(Math.random() >= 0.5 ?
```
noisecapella, line 82: Can we add a parameter which says which kind of config item this is, and use `Math.random` if the parameter is not set?  It would be useful to be help avoid flaky tests

noisecapella (changes requested):

Ferdi:
If the collection is of type "files" and there is only one file in that collection , it should show the form for that file 

Side-nav – add-external – 1

Every other file shows as an extra tab 

Course-metadata-multi

gsidebo:
@Ferdi I'll add a new issue for those UI changes. This PR is pretty huge and we'd benefit from having it merged sooner rather than later

gsidebo:
@noisecapella Ready for review again

gsidebo:
@Ferdi Sorry, [this comment](https://github.com/mitodl/ocw-studio/pull/172#issuecomment-814134460) was way too brief and completely lacked context. I'm thinking that it would be best to merge this with the current UI and implement those suggestions in a separate PR mostly because this introduces a fundamental change in the config structure, and it's a big PR that touches a lot of the codebase so we risk merge conflicts the longer it stays open. 

Re: those suggestions, there are some implementation details we need to sort out:

> If the collection is of type "files" and there is only one file in that collection , it should show the form for that file

The wrinkle with this is that have a dedicated page for adding a new item, and a sidebar for editing an item. If we wanted to keep this same approach, we would need the "Site Metadata" link in our sidebar to either link to a new page or open the editing UI directly. To support this, we'd need to add an API call to fetch WebsiteContent records at the point when we're building the sidebar menu. We haven't had to do that up to this point. We just build the sidebar menu from the config.

> Every other file shows as an extra tab

Since a "files" config item can include an unlimited number of items, we may run into horizontal spacing issues if we use tabs.

websites/views.py
```diff
@@ -256,10 +247,13 @@ class WebsiteContentViewSet(
     def get_queryset(self):
         parent_lookup_website = self.kwargs.get("parent_lookup_website")
         filter_type = self.request.query_params.get("type")
+        filter_subtype = self.request.query_params.get("subtype")
 
         queryset = WebsiteContent.objects.filter(website__name=parent_lookup_website)
         if filter_type:
             queryset = queryset.filter(type=filter_type)
+        if filter_subtype:
+            queryset = queryset.filter(subtype=filter_subtype)
```
noisecapella, line 256: Can you add a test for this? It looks like it's not covered by code coverage https://app.codecov.io/gh/mitodl/ocw-studio/compare/172/diff#D3L256

static/js/query-configs/websites.ts
```diff
@@ -235,57 +236,82 @@ export const createWebsiteCollaboratorMutation = (
...
+  listingParams: ContentListingParams
+): string => {
+  const key = [
+    listingParams.name,
+    listingParams.type,
+    listingParams.subtype,
+    listingParams.offset
+  ]
+  if (listingParams.limit) {
+    key.push(listingParams.limit)
```
noisecapella, line 253: Can we add or modify a test so that it covers this line?

static/js/query-configs/websites.ts
```diff
@@ -235,57 +236,82 @@ export const createWebsiteCollaboratorMutation = (
...
-    })
-  },
-  force: true // try to prevent stale information
-})
+  listingParams: ContentListingParams
+): QueryConfig => {
+  const { name, type, subtype, offset, limit } = listingParams
+  let url = siteApiContentListingUrl.param({ name }).query({ type, offset })
+  if (subtype) {
+    url = url.param({ subtype })
```
noisecapella, line 263: Can we add or modify a test so it covers this line?

static/js/query-configs/websites.ts
```diff
@@ -235,57 +236,82 @@ export const createWebsiteCollaboratorMutation = (
...
+  }
+  if (limit) {
+    url = url.query({ limit })
```
noisecapella, line 266: Can we add or modify a test so it covers this line?

static/js/query-configs/websites.ts
```diff
@@ -235,57 +236,82 @@ export const createWebsiteCollaboratorMutation = (
...
+      body: WebsiteContentListingResponse | Array<WebsiteContentListItem>
+    ) => {
+      const details = {}
+      let pagingParams, results
+      if ("results" in body) {
+        pagingParams = body
+        results = body.results
+      } else {
+        // If results are not paged, use dummy paging data.
+        pagingParams = { count: null, next: null, previous: null }
```
noisecapella, line 280: Can we add a test covering this case too?

static/js/components/SiteAddContent.tsx
```diff
@@ -16,28 +19,66 @@ import {
...
+    ) ?? null
+  if (!topLevelConfigItem) {
+    return null
+  }
+  if (!subtype) {
+    if (isRepeatableCollectionItem(topLevelConfigItem)) {
+      return topLevelConfigItem as RepeatableConfigItem
+    }
+  } else if ("files" in topLevelConfigItem) {
+    return (
```
noisecapella, line 53: Can we add a test covering this condition?

noisecapella (changes requested):
Some missing tests. Functionality works great

websites/migrations/0017_content_subtype.py
```diff
@@ -0,0 +1,18 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("websites", "0016_remove_site_content_type_constraint"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="websitecontent",
+            name="subtype",
```
Ferdi, line 15: What is the need for this field ?

static/js/query-configs/websites.ts
gsidebo: This and the other lines you commented on in this file are tested by `static/js/components/SiteContentListing.test.tsx`

websites/migrations/0017_content_subtype.py
gsidebo: Given a config that looks like this:

```yml
collections
  - name: navigation
    label: Navigation
    tag: Settings
    files:
      - file: config/menu.sidenav.yml
        label: Side Navigation
        name: sidenav
        fields: [...]
```

If you save a record for side navigation, that record would have "navigation" as the type and "sidenav" as the subtype

static/js/query-configs/websites.ts
noisecapella: I'm not sure, if I add a `throw new Error` at this line it doesn't change any test output. Same for the other lines

websites/migrations/0017_content_subtype.py
Ferdi: Why is that field necessary?

noisecapella:
Just a thought, in a couple of those lines the condition tests for `limit`, but `limit` is always either `0` or `undefined`. Is the intention to test if `limit !== undefined || limit !== null` instead? I don't think we are sending `limit=0` to the server

websites/migrations/0017_content_subtype.py
gsidebo: It's the field that relates the database record to the actual item in the config. In the authoring UI, we need a way to know if there is an existing record for one of those "file" items. In the example above, we would query the db for a record with type="navigation", subtype="sidenav". We could just set "sidenav" as the name, but then we wouldn't be able to query for all objects under "navigation", which is useful to determine which of the "file" items have an object created for them.

static/js/query-configs/websites.ts
gsidebo: That's odd... we're mocking the API request to return certain data given the URL with those extra parameters. I'll look into it

websites/migrations/0017_content_subtype.py
gsidebo: We could potentially flip that and add a "parent_type" or something like that instead of "subtype". Then that example would have type="sidenav", parent_type="navigation"

gsidebo:
@noisecapella Great catch, you're absolutely right. I forgot about `if (0)`. Still works but we could run into issues with more than 1 page of results

gsidebo:
@noisecapella Just pushed a pretty big update to this PR and removed a bunch of the changes it introduced. I updated the PR description to reflect what this PR is doing now. Biggest change is that I removed the "subtype" field and there's no more setting the "limit" param in the API calls. This UI just lists the titles/labels for "file"-type items with no add/edit links, and you can't use the Add form for those item types either. That's just temporary and will be resolved in #192. 

Make sure you migrate down before pulling these changes (`manage.py migrate websites 0016`). This is ready for review again now. Sorry for the late pivot!

noisecapella:
FYI, more rebases incoming

gsidebo:
@noisecapella Should be ready now

noisecapella (approved):

PR #175, opened 2021-04-06
Added management command to ensure single source of truth for example config files

### LINKED ISSUE #141: Define example site config files for use in front and back end
We currently have some site config examples scattered around the codebase ([this one in JS](https://github.com/mitodl/ocw-studio/blob/09a364f12e112c38c456836af3f1ed0e4919e353/static/js/constants.ts#L28), [this one in Python](https://github.com/mitodl/ocw-studio/blob/09a364f12e112c38c456836af3f1ed0e4919e353/localdev/starters/site-config-override.yml#L2)). We know we want an up-to-date OCW course site config defined in the codebase for testing the application in a consistent way across all envs, and we may want to define some other configs to exercise certain options.

It would be really nice to have a single version of each one of these site configs defined somewhere in the codebase instead of these separate configs that can easily drift and leave us open to bugs. 

In terms of implementation, one possible approach would be to borrow a technique from [`generate_app_json`](https://github.com/mitodl/bootcamp-ecommerce/blob/aa652dda49a1dfe015f39045f0c050cef126fa68/main/envs.py#L280). We have a single source of truth for our Django settings in some of our other apps (`settings.py`), and from that single source we have a command that generates `app.json`. To make sure that they are synced, we have a Python test case that asserts that the `app.json` matches the current contents of `settings.py`. We could potentially do the same with defining one site config, then using a command to generate a copy wherever else it's needed (in static/js, for example)

gsidebo:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #141 

#### What's this PR do?
Adds a command to generate equivalent JSON example configs for the frontend based on the YML config files in the localdev directory (i.e.: the single source of truth)

- Run `manage.py override_site_config`. It should work as it did before, and should take params to target a different config/starter. 
- Run `manage.py generate_example_configs`. This should take every yml file in `localdev/configs` and create an equivalent JSON file in `static/js/resources`


static/js/types/websites.ts
```diff
@@ -12,7 +12,7 @@ export enum WidgetVariant {
 export interface ConfigField {
   name: string
   label: string
-  widget: WidgetVariant
+  widget: string
```
gsidebo, line 15: Made this change b/c we're now pulling example configs from JSON files, and when typescript analyzes the contents, it just sees a string in this "widget" field. Even though those strings are members of the WidgetVariant enum, it fails. Since our site config schema is defined in the backend and we validate the schema as its created/edited, we can safely remove the type constraint here.

aliceriot: sometimes a string literal may need to be casted (i.e. `widgetString as WidgetVariant`). if possible I think it would be nice to keep the typing on the `widget` field more specific than `string` so we can encode the knowledge we have about the possible values it can take in the type system. if it's totally unfeasible to do so then I suppose that's fine, but if we can keep that I think it would be good to do so.

aliceriot: I was able to get this to typecheck (leaving `widget` typed as `WidgetVariant`) by just adding a cast to `WebsiteStarterConfig` to the `makeWebsiteStarterConfig` factory function:

```diff
diff --git a/static/js/types/websites.ts b/static/js/types/websites.ts
index 36e129f..e72311f 100644
--- a/static/js/types/websites.ts
+++ b/static/js/types/websites.ts
@@ -12,7 +12,7 @@ export enum WidgetVariant {
 export interface ConfigField {
   name: string
   label: string
-  widget: string
+  widget: WidgetVariant,
   minimal?: boolean
   help?: string
   required?: boolean
diff --git a/static/js/util/factories/websites.ts b/static/js/util/factories/websites.ts
index 15b4cd1..fc5e0d3 100644
--- a/static/js/util/factories/websites.ts
+++ b/static/js/util/factories/websites.ts
@@ -96,7 +96,7 @@ export const makeEditableConfigItem = (name?: string): EditableConfigItem => ({
 })
 
 export const makeWebsiteStarterConfig = (): WebsiteStarterConfig =>
-  cloneDeep(exampleSiteConfig)
+  cloneDeep(exampleSiteConfig as WebsiteStarterConfig)
 
 export const makeWebsiteStarter = (type = "course"): WebsiteStarter => ({
   id:     incr.next().value,
```

that could be an option for keeping the typing, if we decide we want to.

gsidebo: Nice! Forgot about `as`. Will fix

aliceriot (approved):
lgtm! 👍

PR #183, opened 2021-04-06
use the WidgetVariant constants instead of string literals

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

none

#### What's this PR do?

nothing much, I added a `WidgetVariant` enum in #168 which defines the constants that we use for defining which widget a field needs to use when it's been edited. This just updates some code to use the constants declared in that enum rather than string literals.

there shouldn't be any functional changes. would be good to read through the changes and make sure I didn't make a typo!

### DIFF
--- a/static/js/components/forms/SiteAddContentForm.test.tsx
+++ b/static/js/components/forms/SiteAddContentForm.test.tsx
@@ -70,8 +70,10 @@ describe("SiteAddContentForm", () => {
     componentFromWidget.mockImplementation(() => widget)
 
     const fieldGroups = [
-      configItem.fields.filter(field => field.widget === "markdown"),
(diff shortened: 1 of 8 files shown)


noisecapella:
Any thoughts on this comment? Not sure if this interferes with this PR or addresses something mostly separate https://github.com/mitodl/ocw-studio/pull/175#discussion_r607992563

aliceriot:
wrote a comment there. sort of a related concern, but I think it is generally a good practice to encode more information into the type system if possible, and especially with something like this where we're dealing with a constained vocabulary (i.e. "file", "markdown", "string", etc)

noisecapella (approved):

Issue #187, opened 2021-04-06
Add support for forwarded header usage to be enabled


rhysyngsun:
These two settings are necessary in some deployment scenarios, the latter in particular is needed for SAML auth on Heroku.
- USE_X_FORWARDED_PORT
- USE_X_FORWARDED_HOST

PR #188, opened 2021-04-06
Add settings for using X-Forwarded-* headers

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
- [x] Settings

Fixes https://github.com/mitodl/ocw-studio/issues/187

#### What's this PR do?
Adds `USE_X_FORWARDED_HOST` and `USE_X_FORWARDED_PORT` as configurable settings.

App should run, functionally this is built-in django settings and we'll test in CI/RC

### DIFF
--- a/main/settings.py
+++ b/main/settings.py
@@ -88,6 +88,18 @@ SECURE_SSL_REDIRECT = get_bool(
     description="Application-level SSL redirect setting.",
 )
 
+
+USE_X_FORWARDED_HOST = get_bool(
+    name="USE_X_FORWARDED_HOST",
+    default=False,
+    description="Set HOST header to original domain accessed by user",
+)
+USE_X_FORWARDED_PORT = get_bool(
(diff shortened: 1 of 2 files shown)


main/settings.py
```diff
@@ -88,6 +88,18 @@
     description="Application-level SSL redirect setting.",
 )
 
+
+USE_X_FORWARDED_HOST = get_bool(
+    name="USE_X_FORWARDED_HOST",
+    default=False,
```
mbertrand, line 94: Minor nitpick - any harm in setting the default to True instead?  Otherwise will need to make sure they're set to True in saltops.

mbertrand (approved):
Looks great, 1 minor question that shouldn't hold this up.

main/settings.py
rhysyngsun: This will break environments where that header isn't set so leaving it off (django's default) is a safer default.

PR #201, opened 2021-04-07
Use SelectField widget for all select fields site-wide

### LINKED ISSUE #194: Use react-select sitewide
Use `react-select` for all select form fields site-wide.

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #194

#### What's this PR do?
- Replaces the standard select field with the `SelectField` widget on `SiteForm`, `SiteCollaboratorForm`
- Modifies the `SelectField` widget to handle option elements where the label and value differ.

Try out the following, the select field should work as expected on each form:
- Create a new site
- Create/edit site metadata
- Create/edit a collaborator

### DIFF
--- a/static/js/components/widgets/SelectField.tsx
+++ b/static/js/components/widgets/SelectField.tsx
@@ -1,6 +1,6 @@
 import React, { useCallback } from "react"
 import Select from "react-select"
-import { isNil } from "ramda"
+import { is, isNil } from "ramda"
 
 export interface Option {
   label: string
@@ -14,11 +14,14 @@ interface Props {
   multiple?: boolean
   min?: number
   max?: number
-  options: string[]
+  options: Array<string | Option>
 }
 export default function SelectField(props: Props): JSX.Element {
   const { value, onChange, name, options } = props
   const multiple = Boolean(props.multiple)
+  const selectOptions = options.map((option: any) =>
+    is(String, option) ? { label: option, value: option } : option
+  )
 
   const changeHandler = useCallback(
     (newValue: any) => {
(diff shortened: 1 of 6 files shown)


static/js/components/widgets/SelectField.tsx
```diff
@@ -31,23 +34,30 @@ export default function SelectField(props: Props): JSX.Element {
...
+    return { label: selectOption ? selectOption.label : value, value: value }
+  }
+
   let selected
   if (multiple) {
     if (!Array.isArray(value)) {
       selected = []
     } else {
-      selected = value.map(option => ({ label: option, value: option }))
+      selected = value.map(option => getSelectOption(option))
```
aliceriot, line 49: ok so if I'm understanding this right, we need to get options that have a `value` prop in the `value` array - would it maybe make sense to do something like this instead of the `getSelectedOption` thing?

```ts
selected = selectOptions.filter(option => {
  value.includes(option.value)
})
```

I'm not sure if I'm understanding all of what this is doing though! so maybe it's more complicated than that. In any case, might be good to throw in a comment to explain what's up.

aliceriot (changes requested):
works great, just had one question

static/js/components/widgets/SelectField.tsx
mbertrand: The original code will return a `{label, value}` `Option` object even if that option is not included in the select form field.  I'm not sure if that's something we need to continue supporting but felt I shouldn't take it away in this PR if that's the way it's working now.   That's why I wrote the `getSelectOption` function to return the existing `Option` if it exists, otherwise generate a new one based on the value (`return { label: selectOption ? selectOption.label : value, value: value }`).

mbertrand: Added a comment and simplified the function a bit.

mbertrand: @aliceriot ready for another look

aliceriot (approved):
LGTM! 👍

PR #202, opened 2021-04-08
Cleaning up collaborator api

### LINKED ISSUE #81: Collaborator API improvements
I added a review to the[ initial collaborators API PR](https://github.com/mitodl/ocw-studio/pull/72) and unfortunately it had already been merged. [Here's a link](https://github.com/mitodl/ocw-studio/pull/72#pullrequestreview-595584484) to the suggested changes. This can be considered tech debt for now

rhysyngsun:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Fixes #81 

#### What's this PR do?
- Moves data mutations into `WebsiteCollaboratorSerializer`
- Simplified `WebsiteCollaboratorViewSet.get_queryset` down to a single query definition
- Removed `update` / `create` methods on `WebsiteCollaboratorViewSet` which were now unnecessary
- Simplified validation in `WebsiteCollaboratorSerializer` to lean on DRF as much as possible
- Moved `permissions_group_name_for_role` into a new `website/utils.py` to avoid circular imports
- Removed `group` from `WebsiteCollaboratorSerializer` since it's unused and is a backend detail that the frontend doesn't need to know about

Go to the collaborators UI and verify you can view/add/edit/delete editors and administrators.


gsidebo (changes requested):
Still reviewing the code & functionality right now, but one question/suggestion in the meantime you can take or leave: What do you think about splitting the collaborator view tests from `websites/views_test.py` to a separate file (`collaborator_views_test.py` et al)? That `views_test.py` file is pretty big, and I expect it will only get bigger. Separating the collaborator tests from the tests for API views that just deal with Website/WebsiteContent data might be helpful

rhysyngsun:
> Still reviewing the code & functionality right now, but one question/suggestion in the meantime you can take or leave: What do you think about splitting the collaborator view tests from `websites/views_test.py` to a separate file (`collaborator_views_test.py` et al)? That `views_test.py` file is pretty big, and I expect it will only get bigger. Separating the collaborator tests from the tests for API views that just deal with Website/WebsiteContent data might be helpful

Could we do that in a separate PR? I think splitting up the views themselves (and the tests along with them) is a good goal but that definitely explodes the scope of this work.

websites/serializers.py
```diff
@@ -138,11 +133,46 @@ def validate_email(self, email):
...
 
+    def create(self, validated_data):
+        """Creating a contributor adds the user to the group corresponding to the specified role"""
+        website = self.website
+        role = validated_data.get("role")
+        user = User.objects.get(email=validated_data.get("email"))
+        group_name = permissions_group_name_for_role(role, website)
+        if user in get_users_with_perms(website) or user == website.owner:
+            raise ValidationError(
+                {"errors": ["User is already a collaborator for this site"]}
```
gsidebo, line 149: I think we should namespace this error under "email". If we do that, the error message will show under the email input rather than under the form, which I think makes more sense 

ss 2021-04-13 at 14 04 07

gsidebo (changes requested):
Great improvements here! Just a couple minor comments to go with the suggestion above

gsidebo (approved):
Looks great! 👍

PR #205, opened 2021-04-08
Merge add and edit content forms

### LINKED ISSUE #154: Combine site content forms
`SiteAddContentForm` and `SiteEditContentForm` are implemented separately right now. This may have been to support the multi-column layout from the mockups in the "add" view, and the single column layout in the right panel "edit" view. All the same, the functionality of these forms is largely identical, and we should not be in the practice of implementing add/edit forms more than once unless there is a significant functional difference.

### LINKED ISSUE #191: Forms layout
As a content author , I would like a distraction free environment when authoring longer posts. 

Treat the "body" field as a special case in the UI.  When present, it should occupy its own (left) column. 

#### Designs and Mockups

* A collection with 'body' field
Screen Shot 2021-04-06 at 8 36 55 PM

* A collection with no 'body' field
Screen Shot 2021-04-06 at 8 38 09 PM

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #191 
Fixes #154 

#### What's this PR do?
The forms and pages for editing and adding content are merged. Both will use a drawer according to the mockup in #191

 - Editing content should work exactly like before.
 - When you click the "Add Resource" or "Add Page" button, you should see a drawer pop out instead of being directed to a new page like before.
   - If you leave the title blank it should fail validation.
   - When you click 'Save' you should see the new item show up on the listing.

#### Screenshots
Screenshot from 2021-04-09 09-12-28
Screenshot from 2021-04-09 09-12-48

Screenshot from 2021-04-09 09-14-16
Screenshot from 2021-04-09 09-14-30


static/js/components/SiteContent.tsx
```diff
@@ -0,0 +1,140 @@
...
+} from "./forms/SiteContentForm"
+import { contentFormValuesToPayload } from "../lib/site_content"
+import { getResponseBodyError, isErrorResponse } from "../lib/util"
+
+import { EditableConfigItem, Website } from "../types/websites"
+import { ContentFormType } from "../types/forms"
+
+type SiteFormValues = {
+  [key: string]: string
+}
```
aliceriot, line 26: should we type this is `Record<string, string>` ?

static/js/components/SiteContent.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const [
+    { isPending: addIsPending },
+    addWebsiteContent
+  ] = useMutation((payload: NewWebsiteContentPayload) =>
+    createWebsiteContentMutation(site.name, payload)
+  )
+  const [{ isPending: editIsPending }, editWebsiteContent] = useMutation(
+    (payload: EditWebsiteContentPayload | FormData) =>
+      // @ts-ignore: uuid and configItem should never be null if editing
```
aliceriot, line 59: can we get around the null error with the bang operator (`!`) instead of `@ts-ignore`?

static/js/components/forms/SiteContentForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+                <>
+                  <div className="form-group d-flex justify-content-end">
+                    <button
+                      type="submit"
+                      disabled={isSubmitting}
+                      className="px-5 btn blue-button"
+                    >
+                      Save
+                    </button>
+                  </div>
```
aliceriot, line 86: why don't we just render these buttons and the `.form-error` status below in a separate full-width div after we've closed out the columns? it seems a bit confusing that this is in the `fieldsByColumn.map` call

aliceriot (changes requested):
had a couple of questions and comments. I wasn't sure if some of the code was just being copied over from the existing implementation - if so, possibly some of these comments are out of date!

static/js/components/SiteContent.tsx
```diff
@@ -0,0 +1,140 @@
...
+  visibility: boolean
+  toggleVisibility: () => void
+  site: Website
+  configItem: EditableConfigItem
+  formType: ContentFormType
+  contentType: string
+  repeatListingRequest: () => void
+}
+
+export default function SiteContent(props: Props): JSX.Element | null {
```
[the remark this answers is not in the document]
gsidebo: +1 for this

static/js/components/SiteContentEditor.tsx
```diff
@@ -4,60 +4,90 @@ import { useMutation, useRequest } from "redux-query-react"
...
 interface Props {
-  uuid: string
+  uuid: string | null
   visibility: boolean
   toggleVisibility: () => void
   site: Website
   configItem: EditableConfigItem
+  formType: ContentFormType
+  contentType: string
+  repeatListingRequest: () => void
```
gsidebo, line 34: Can we change this name to `contentListingRequest` or `websiteContentListingRequest` here and in `SiteContentListing.tsx`?

static/js/components/SiteContentEditor.tsx
```diff
@@ -73,26 +103,35 @@ export default function SiteEditContent(props: Props): JSX.Element | null {
     }
     setSubmitting(false)
 
+    // refresh to have the new item show up in the listing
+    repeatListingRequest()
```
gsidebo, line 107: Is it necessary to hit this endpoint again if we're just editing an item? I was under the impression that we update the relevant redux query entities after an edit, so the listing page should automatically have up-to-date information for the edited item since it uses the same entities. If that's true, the API call wouldn't be necessary

static/js/components/SiteContentListing.tsx
```diff
@@ -124,16 +121,21 @@ export default function SiteContentListing(): JSX.Element | null {
...
+  const [
+    { isPending: contentListingPending },
+    repeatListingRequest
+  ] = useRequest(websiteContentListingRequest(listingParams))
   const listing: WebsiteContentListingResponse = useSelector(
     getWebsiteContentListingCursor
   )(listingParams)
+  const [addConfigItem, setAddConfigItem] = useState<EditableConfigItem | null>(
+    null
+  )
```
gsidebo, line 133: A bit of a nit, but we can probably change up the names for these vars and the editing equivalents to make them more consistent:

```typescript 
  const [configItemToAdd, setConfigItemToAdd] = useState<EditableConfigItem | null>(
    null
  )
  const [configItemToEdit, setConfigItemToEdit] = useState<{
    uuid: string
    configItem: EditableConfigItem
  } | null>(null)
```

We could also unify these and just set uuid: false for the 'add' case

```typescript
  const [activeConfigItem, setActiveConfigItem] = useState<{
    uuid: string
    configItem: EditableConfigItem
  } | null>(null)
```

Just some ideas. You can pick/choose

static/js/components/forms/SiteContentForm.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+} from "../../lib/site_content"
+import { getContentSchema } from "./validation"
+
+import { EditableConfigItem, WebsiteContent } from "../../types/websites"
+import { ContentFormType } from "../../types/forms"
+
+jest.mock("../../lib/site_content")
+jest.mock("./validation")
+
+describe("SiteContentForm", () => {
```
gsidebo, line 26: Seems like we should have a test for the `initialValues` prop

static/js/lib/site_content.test.ts
```diff
@@ -263,6 +265,43 @@ describe("site_content", () => {
...
+        },
+        makeWebsiteConfigField({ widget: WidgetVariant.Boolean })
+      ]
+      expect(splitFieldsIntoColumns(fields)).toEqual([
+        [
+          {
+            label:  "Body",
+            name:   MAIN_PAGE_CONTENT_FIELD,
+            widget: WidgetVariant.Markdown
+          }
```
gsidebo, line 298: Nit: could you just use `fields[1]` here?

static/js/lib/site_content.ts
```diff
@@ -55,10 +54,29 @@ export function widgetExtraProps(field: ConfigField): Record<string, any> {
   }
 }
 
-const isMainContentField = (field: ConfigField) =>
+/**
+ * determine whether a ConfigField is a main content field or not.
+ * this means, basically, whether it's the main Markdown field for
+ * a site.
```
gsidebo, line 60:
```suggestion
 * a piece of site content.
```

Thanks for adding these docs!

static/js/lib/site_content.ts
```diff
@@ -55,10 +54,29 @@ export function widgetExtraProps(field: ConfigField): Record<string, any> {
...
+ **/
+export const isMainContentField = (field: ConfigField): boolean =>
   field.name === MAIN_PAGE_CONTENT_FIELD &&
   field.widget === WidgetVariant.Markdown
 
+/**
+ * split an array of ConfigField into two arrays, with the first containing
+ * 'main content' fields and the second container all the other fields.
```
gsidebo, line 68: Small typo
```suggestion
 * 'main content' fields and the second containing all the other fields.
```

static/scss/layout.scss
```diff
@@ -87,9 +94,15 @@ $modal-sidebar-width: 540px;
   padding: 15px 15px 80px;
 }
 
-.modal.right.fade .modal-dialog {
-  right: -$modal-sidebar-width;
-  transition: opacity 0.3s linear, right 0.3s ease-out;
+.modal.right.fade {
```
gsidebo, line 97: Great idea nesting the `.modal-dialog` styles above. Could we do the same for the `.modal-content` styles at line 88 and these `.modal.right.fade` styles here?

gsidebo (changes requested):
Looks great so far! Checking functionality now

static/js/components/SiteContentEditor.tsx
aliceriot: yeah I think it's not strictly necessary, because if you check the sequence of updates to the store the new version of the site data is propagated to the store when the edit request comes back successfully. So we could just lean on that to ensure that the view on the content list remains consistent.

thinking about it more, though, just dropping a refresh on the listing API does protect us from some degree of inconsistency that could arise if multiple people are working on content at the same time (not really on the piece of content just edited, but on others). So I'm somewhat inclined to leave it around as 'insurance', if you will...idk, what do you think? It's a (very) small increase in load on the server, but my understanding is that ~5-10 users are going to be ever using this at the same time, so we're really not going to be dealing with scaling headaches here.

static/js/components/SiteContentListing.tsx
aliceriot: I like the latter - fewer lines == clearer code imo!

static/js/components/SiteContentEditor.tsx
gsidebo: Sure, we can leave it

static/js/lib/site_content.ts
aliceriot: My editor is a bit smarter with typescript than it was with JS, so for functions that have docstrings they pop up when I'm writing code and it's very nice to have a brief little explanation! I'm trying to add it to most library code that I touch.

static/js/lib/site_content.test.ts
aliceriot: yes! haha

static/js/components/SiteContentListing.tsx
aliceriot: Digging into that i found a number of things that seemed redundant or unnecessary, pushing code shortly but basically:

- I made the panel visibility a three-possibility state machine ("none", "edit", "add")
- there doesn't seem to be any need to set the focused `configItem` because when the function is passed down it is, afaict, always called with the `configItem` from the `website`, so I removed this
- then, with just a three-possibility state machine and another piece of state for storing the UUID a few things can be simplified

anyhow, take a look at the most recent commits and lmk what you think

gsidebo (changes requested):
I'm able to submit a blank form when adding new content

ss 2021-04-15 at 14 00 14

aliceriot:
I can do that on `master` as well, I think it's because in the localdev site config none of the fields are marked `required` for a page

aliceriot:
I'll make an update to at least require a title!

static/scss/layout.scss
aliceriot: I don't see why not!

gsidebo (approved):
Looks great! 👍  
Feel free to merge once the build is fixed

PR #206, opened 2021-04-09
upgrade ckeditor packages

### LINKED ISSUE #195: upgrade ckeditor packages to 27.0.0
they've all been bumped, we should do a PR to upgrade them. I think if we try to stay apace with updates to these packages our lives will be easier in the long run.

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #195 

#### What's this PR do?

This upgrades all of our ckeditor packages to `@latest`. I anticipate we'll need to do this again, and it's a pain to ugprade all 20 packages, so I wrote a little nodejs script to do that for us automatically.

The markdown editor should work as before with no errors, in both the minimal and the 'full fat' config. You can test those out on the `/markdown-editor` page.

You can also test out the little upgrade script. If you checkout the `package.json` and `yarn.lock` from `HEAD` on `main` you can then test out upgrading:

```sh
git checkout origin/master -- package.json yarn.lock
node ./scripts/upgrade_ckeditor.js
```

I used the program `commander`, so there is a little help text (`--help`).


static/js/lib/ckeditor/CKEditor.ts
```diff
@@ -22,19 +19,16 @@ import MarkdownMediaEmbed from "./plugins/MarkdownMediaEmbed"
...
     ItalicPlugin,
     UnderlinePlugin,
     BlockQuotePlugin,
-    EasyImagePlugin,
     HeadingPlugin,
     ImagePlugin,
     ImageCaptionPlugin,
     ImageStylePlugin,
     ImageToolbarPlugin,
-    ImageUploadPlugin,
```
mbertrand, line 37: Don't need/want these plugins?

mbertrand (approved):
LGTM, assuming there's no need for the deleted plugins.

static/js/lib/ckeditor/CKEditor.ts
aliceriot: we will sometime soon, but in the update they added a required dependency on a configuration plugin which we don't want to set up yet, which meant that having them enabled caused a runtime error. When we go to add image uploading to the editor we'll likely re-enable these.

PR #211, opened 2021-04-14
add validation for min and max on select field

### LINKED ISSUE #185: Min/max validation for select widget not working
### Steps to Reproduce

Edit the site metadata of a site. You should see one tag by default, `Design`. If you remove the tag and click "Save" there should be a validation error on save since the site config has `min: 1` for this field.

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #185 

#### What's this PR do?

this adds validation for the min and max property on a `Select` field (this is only enforced when the `multiple` prop is also set). I also added some general testing for the select validation, as there weren't yet any tests for it.

I made a little change to the `localdev/configs/ocw-course-site-config.yml` file, so you should be able to just run the `override_site_config` management command to add a multiple-select 'tag' field to the page content type. Then you can check in the UI that it's correctly enforcing the `min` and `max` (we may want to remove this again from `localdev/configs/ocw-course-site-config.yml` before we merge so it's not too annoying).

Screen Shot 2021-04-14 at 2 36 49 PM
Screen Shot 2021-04-14 at 2 37 02 PM

### DIFF
--- a/localdev/configs/ocw-course-site-config.yml
+++ b/localdev/configs/ocw-course-site-config.yml
@@ -6,6 +6,16 @@ collections:
     fields:
       - {"label": "Title", "name": "title", "widget": "string", "required": true}
       - {"label": "Body", "name": "body", "widget": "markdown"}
+      - {
+        "label": "Tags",
+        "default": [ "Design" ],
+        "max": 2,
+        "min": 1,
+        "multiple": true,
+        "name": "tags",
+        "options": [ "Design", "UX", "Dev" ],
+        "widget": "select"
+        }
 
   - name: resource
     label: Resource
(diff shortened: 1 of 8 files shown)


noisecapella:
If I start off with an empty tags list and click Save without modifying the field, it will save without a validation error. This probably won't happen with new pages but it could happen with existing pages that get a new field

aliceriot:
ok I fixed the issue, it related to not having a multiple-select specific default value, so an empty string was being put there instead of an empty list. I fixed it, so editing a page which does not have a value saved yet for `tags` should work (i.e. if you open the editor you should see a validation message saying 'tags must have at least one entry'

noisecapella (approved):

Issue #209, opened 2021-04-13
Epic: GitHub Sync Integration


rhysyngsun:
RFC: https://docs.google.com/document/d/1HCK1t-7L9poR4OHcL_tZuAeRB5VANSlQj5jXnsZTDA0/edit?usp=sharing

- [x] Create `ContentSyncState` [data model](https://docs.google.com/document/d/1HCK1t-7L9poR4OHcL_tZuAeRB5VANSlQj5jXnsZTDA0/edit#heading=h.w0avgftibuhn) (@rhysyngsun)
  - Migrate `WebsiteContent.hugo_filepath` to `ContentSyncState.content_filepath`
- [x] Create [Abstract Sync Backend](https://docs.google.com/document/d/1HCK1t-7L9poR4OHcL_tZuAeRB5VANSlQj5jXnsZTDA0/edit#heading=h.qe7jbq2m1mij) (@rhysyngsun)
- [x] Create [GitHub API wrapper](https://docs.google.com/document/d/1HCK1t-7L9poR4OHcL_tZuAeRB5VANSlQj5jXnsZTDA0/edit#heading=h.an0ws5n65p37) (@mbertrand)
- [x] Create sync backend for GitHub
- [x] Create [celery tasks](https://docs.google.com/document/d/1HCK1t-7L9poR4OHcL_tZuAeRB5VANSlQj5jXnsZTDA0/edit#heading=h.57df8kdcbsos)
  - [x] New `Website` -> new [github repo + branches](https://docs.google.com/document/d/1HCK1t-7L9poR4OHcL_tZuAeRB5VANSlQj5jXnsZTDA0/edit#heading=h.2232hzzes5x7)
  - [x] New `WebsiteContent`
  - [x] Update `WebsiteContent`
  - [x] Delete `WebsiteContent`
- [x] Add model signals to create/update `ContentSyncState` records and trigger appropriate celery tasks
- [x] Create [management command](https://docs.google.com/document/d/1HCK1t-7L9poR4OHcL_tZuAeRB5VANSlQj5jXnsZTDA0/edit#heading=h.92556rw1hgvz) to push all content to GitHub
- [ ] Create [management command](https://docs.google.com/document/d/1HCK1t-7L9poR4OHcL_tZuAeRB5VANSlQj5jXnsZTDA0/edit#heading=h.92556rw1hgvz) to pull all content from GitHub
[...]

PR #215, opened 2021-04-14
Added data model for ContentSyncState

rhysyngsun:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Part of #209 

#### What's this PR do?
Adds the `ContentSyncState` data model and signals around `WebsiteContent` to ensure records are created.

The migrations should run and the app should start


content_sync/apps.py
```diff
@@ -1,8 +1,12 @@
...
 
 class ContentSyncApp(AppConfig):
-    """App for content_sync"""
+    """ App for content_sync """
 
     name = "content_sync"
+
+    # def ready(self):
+    #     """ Application is ready """
+    #     import content_sync.signals  # pylint:disable=unused-import, import-outside-toplevel
```
mbertrand, line 12: Uncomment the above

content_sync/models.py
```diff
@@ -1,4 +1,32 @@
-"""Content sync models"""
-# from django.db import models
+""" Content sync models """
+from django.db import models
 
-# Create your models here.
+
+class ContentSyncState(models.Model):
```
mbertrand, line 5: Do you think it's worth adding this to Django admin?

content_sync/signals.py
```diff
@@ -0,0 +1,18 @@
...
+@receiver(
+    post_save,
+    sender=WebsiteContent,
+    dispatch_uid="website_content_create_sync_state",
+)
+def upsert_content_sync_state(
+    sender, instance, created, **kwargs
+):  # pylint: disable=unused-argument
+    """ Create/update the sync state """
+    api.upsert_content_sync_state(instance)
```
mbertrand, line 18: Do you think `ContentSyncState` objects should be backpopulated for all existing `WebsiteContent` objects via a management command or migration script?

mbertrand (changes requested):

mbertrand:
Also needs a formatting run to pass tests

websites/models.py
```diff
@@ -99,11 +100,15 @@ def upload_file_to(self, filename):
...
     metadata = models.JSONField(null=True, blank=True)
-    hugo_filepath = models.CharField(max_length=2048, null=True, blank=True)
+    content_filepath = models.CharField(max_length=2048, null=True, blank=True)
     file = models.FileField(
         upload_to=upload_file_to, editable=True, null=True, blank=True, max_length=2048
     )
 
+    def calculate_checksum(self) -> str:
+        """ Returns a calculated checksum of the content """
+        return sha256(str(self.markdown).encode("utf-8")).hexdigest()
```
mbertrand, line 110: Should include `metadata` too,  in case the markdown stays the same but the metadata changes.  For example, a `WebsiteContent` object of type `resource` will not have any markdown, just metadata.  Maybe something like this:

```python
sha256(json.dumps(self.metadata) + str(self.markdown).encode("utf-8")).hexdigest()
```

mbertrand (changes requested):

content_sync/models.py
rhysyngsun: I don't think there should be a need to work with this model in django-admin, since all of its fields shouldn't be modified outside of code.

rhysyngsun:
@mbertrand I addressed your feedback, I also switched the model to `TimestampedModel` since that will probably be helpful, so you'll want to `manage.py migrate content_sync zero` before you pull the latest.

content_sync/migrations/0002_backpopulate_content_sync_state.py
```diff
@@ -0,0 +1,36 @@
...
+
+from django.db import migrations
+
+
+def backpopulate_content_sync_states(apps, schema_editor):
+    """ Backpopulate ContentSyncState for existing WebsiteContent records """
+    WebsiteContent = apps.get_model("websites", "WebsiteContent")
+    ContentSyncState = apps.get_model("content_sync", "ContentSyncState")
+
+    for content in WebsiteContent.objects.filter(content_sync_state=None):
```
mbertrand, line 13: There's about 88K of these on RC, I've got about the same amount in my local instance and docker memory usage went pretty high (~2.5 GB), wondering if that might cause problems on heroku.  One thing I did in a similar migration for open was use a chunked iterator:

```
def chunked_iterator(queryset):
    paginator = Paginator(queryset, 100)
    for page in range(1, paginator.num_pages + 1):
        for obj in paginator.page(page).object_list:
            yield obj
            
 for content in chunked_iterator(WebsiteContent.objects.filter(content_sync_state=None)):
 ```
 
Or can leave as is and see how it goes, maybe it's not necessary.

mbertrand (commented):
Looks great, one concern about memory usage in the migration.

mbertrand (approved):
👍

PR #217, opened 2021-04-16
Implemented correct UI for 'files' config items

### LINKED ISSUE #192: UI/UX for  collections of type "Files"
If the collection is of type "files" and there is only one file in that collection , it should show the form for that file 

Side-nav – add-external – 1

Every other file shows as an extra tab 

Course-metadata-multi

* These files are singleton so there is no need for add/edit links
* We need to think some more whether these forms show in a sidebar or not.  Whatever is the easiest for now
* See https://github.com/mitodl/ocw-studio/issues/191 as it should help with consolidating add/edit forms

gsidebo:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested
 
Closes #192 

#### What's this PR do?
- Fixes the UI so "file"-type config items so they can be added/edited again
- Changes the `WebsiteContent.uuid` field name to `text_id` and changes it to a charfield, which allows us to use the config item `name` property as the id for a singleton `WebsiteContent` record
- Adds a requirement to the config schema that names must all be unique

- Add an item to "files" in the "metadata" config item for the "course" starter, then load the "Metadata" section. You should see a UI like the screenshot above
- You should be able to fill out the forms and save them. If the object did not exist in the db before, the form should be blank, and submitting should add a new object. If the object did exist, the form should be pre-populated, and submitting should update it.

You can also test that saving a site config with duplicate "name" values should cause a validation error.

- We'll probably want a more informative UX when the user submits these forms. I'll open up an issue to implement that for these forms and any other relevant forms

ss 2021-04-16 at 22 28 45
ss 2021-04-16 at 22 29 01
ss 2021-04-09 at 13 03 09


noisecapella:
Should the metadata form redirect or do anything after clicking "Save"?

gsidebo:
@noisecapella I mentioned this in the PR description, but the post-save UX can be taken care of in a separate issue. There isn't anything in the designs yet. It doesn't make much sense to redirect anywhere.

noisecapella:
I'm seeing some strange behavior switching tabs. I made a "Other metadata" entry in the yaml and updated the config, so I see that and "Site metadata". I updated site metadata, refreshed the browser, and changes are presisted correctly. However when I click the tab "Other metadata" and then "Site metadata", the data disappears from the form

gsidebo:
@noisecapella Probably a redux-query entity state thing. I'll take a look

gsidebo:
@noisecapella I was missing a prop, which turns off a default Formik behavior that I wasn't aware of. Great catch! Ready for review again 👍

noisecapella (approved):

PR #218, opened 2021-04-16
Add abstract base class for syncing backends

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Part of #209 

#### What's this PR do?
Adds an abstract base class for the interface around syncing to backends like GitHub.

Tests should pass and the app should start


content_sync/backends/base.py
```diff
@@ -0,0 +1,132 @@
...
+        An example would be making a commit to delete content in the VCS repository.
+        """
+        ...
+
+    # NOTE: this will be fired for every piece of content that is created, changed, or deleted in the database
+    def sync_content_to_backend(self, sync_state: ContentSyncState):
+        """ Sync a given piece of content given its ContentSyncState """
+        if not sync_state.synced_checksum:
+            self.create_content_in_backend(sync_state)
+        elif sync_state.content is None:
```
mbertrand, line 84: Does this mean no `ContentSyncState` objects will ever be deleted from the database?  Or will there be a series of steps in deleting content:

1. `ContentSyncState.content` set to `None`
2. `sync_content_to_backend -> delete_content_in_backend` is triggered
3. The `ContentSyncState` object is deleted
4. The `WebsiteContent` object is deleted

Currently, it is not possible to delete a `WebsiteContent` object if a ContentSyncState has a relation to it:
```
"Cannot delete some instances of model 'WebsiteContent' because they are referenced 
through protected foreign keys: 'ContentSyncState.content'."
```

content_sync/backends/base.py
```diff
@@ -0,0 +1,132 @@
...
+        self, sync_state: ContentSyncState
+    ):  # pragma: no cover
+        """
+        Called to delete a piece of content in the backend.
+
+        An example would be making a commit to delete content in the VCS repository.
+        """
+        ...
+
+    # NOTE: this will be fired for every piece of content that is created, changed, or deleted in the database
```
mbertrand, line 79: Fired how/where? A django signal if a particular backend has been specified in settings?

mbertrand (commented):
Code looks good, have some implementation questions.

content_sync/backends/base.py
rhysyngsun: Hmmm, yeah, I think I meant to have that relation be a `SET_NULL`, then I changed my mind in the data model PR, forgetting this code, so I'll switch it back over.

rhysyngsun: I set this to `SET_NULL` for now, but need to figure out in more depth how to handle this

content_sync/backends/base.py
rhysyngsun: I implemented a signal for create/update, will deal with delete separately since it's a little more challenging with the async behavior

rhysyngsun:
@mbertrand all set

content_sync/models.py
```diff
@@ -10,7 +10,7 @@ class ContentSyncState(TimestampedModel):
 
     content = models.OneToOneField(
         WebsiteContent,
-        on_delete=models.PROTECT,
+        on_delete=models.SET_NULL,
```
mbertrand, line 13: In a shell I noticed that after deleting a `WebsiteContent` object, retrieving its `ContentSyncState` would throw an exception:

```
ContentSyncState.objects.order_by("created_on").last()
>AttributeError: 'NoneType' object has no attribute 'title'
```

Can be avoided by modifying the `__str__` method to something like this:
```
return f"Sync State for content: {self.content.title if self.content else None}"
```

mbertrand (approved):
Looks good, noticed one error, feel free to merge after fixing.

PR #227, opened 2021-04-22
Differentiate between types of user-entered data

### LINKED ISSUE #221: Differentiate between types of user-entered data
[RFC](https://docs.google.com/document/d/1_7X9PsVSRONXtJ5SFQFivWmoHvYd6tR-hfrciw6TnXY/edit?usp=sharing)

We need to ensure that (1) authored data can be correctly identified by type ("content" or "configs/data"), and (2) when a piece of authored data is saved, it is assigned the correct type.

**Identifying data by type**
- For this we can add a new field to the `WebsiteContent` model. This can be a boolean field. "is_content" might be an awkward name given the model name – we might end up with code on the backend that looks like `content.is_content`. Something like "is_page_content" might work.

**Assigning the correct type upon creation**
- We can add a new top-level item to the site config to help identify "content"-type objects: `"content-dir"` (ex: `{"content-dir": "content/"}`). In the site config, if the destination for a config item (the "folder" or "file" value) begins with the "content-dir" value, we know that the object being created is of the "content" type. If not, we know the object being created is of the "settings/configs" type.
- When a new `WebsiteContent` record is being created, we can fetch the relevant site config item based on the `type`/`subtype` values, check its folder/file target, and assign it the "content" type if that folder/file target starts with the "content-dir" value

**Acceptance criteria:**
- "content-dir" added to site config schema and example files
  - That key should be optional, and if it's not provided, our app should use "content" as the default value
- Field added to `WebsiteContent` to represent the type
- `WebsiteContent` records are given the correct type when they're created

gsidebo:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Closes #221 

#### What's this PR do?
- Adds the "content-dir" key to the site config
- Makes it so that if a new WebsiteContent object is for page content (i.e.: the object is being created for a config item that has a file/folder that begins with the "content-dir" value), we set the `is_page_content` flag to `True` for that newly-created object
- Adds a management command for "syncing" the `is_page_content` flag for existing WebsiteContent records

- Create a page content object in the UI (in the OCW course site config, "Page" and "Resource" are configured as page content). The resulting `WebsiteContent` record should have `is_page_content=True`
- Create a non-page content object in the UI (in the OCW course site config, "Metadata" > "Site Metadata" is configured as non-page content). The resulting `WebsiteContent` record should have `is_page_content=False`

You can also test the `sync_contents_with_config` management command to make sure it accurately updates your WebsiteContent records


Ferdi:
is_content works better as a field name as we are going to add content in the future that is not "page" .

noisecapella:
I'm seeing this error when saving a new page:

      File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
        response = get_response(request)
      File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
        response = wrapped_callback(request, *callback_args, **callback_kwargs)
      File "/usr/local/lib/python3.8/site-packages/sentry_sdk/integrations/django/views.py", line 52, in callback
        return old_callback(*args, **kwargs)
      File "/usr/local/lib/python3.8/site-packages/sentry_sdk/integrations/django/views.py", line 52, in callback
        return old_callback(*args, **kwargs)
      File "/usr/local/lib/python3.8/site-packages/sentry_sdk/integrations/django/views.py", line 52, in callback
        return old_callback(*args, **kwargs)
      File "/usr/local/lib/python3.8/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
        return view_func(*args, **kwargs)
      File "/usr/local/lib/python3.8/site-packages/rest_framework/viewsets.py", line 125, in view
        return self.dispatch(request, *args, **kwargs)
      File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 509, in dispatch
        response = self.handle_exception(exc)
      File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 469, in handle_exception
        self.raise_uncaught_exception(exc)
      File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
        raise exc
      File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 506, in dispatch
        response = handler(request, *args, **kwargs)
      File "/usr/local/lib/python3.8/site-packages/rest_framework/mixins.py", line 17, in create
        serializer = self.get_serializer(data=request.data)
      File "/usr/local/lib/python3.8/site-packages/rest_framework/generics.py", line 109, in get_serializer
        kwargs.setdefault('context', self.get_serializer_context())
      File "./websites/views.py", line 254, in get_serializer_context
        site_config, config_item.item
    
    Exception Type: AttributeError at /api/websites/lets-try-a-new-course/content/
    Exception Value: 'dict' object has no attribute 'item'

gsidebo:
@noisecapella My fault, late change that I didn't test properly. Fixed now

websites/management/commands/sync_contents_with_config.py
```diff
@@ -0,0 +1,76 @@
...
+            "-s",
+            "--starter-slug",
+            dest="starter_slug",
+            default="",
+            help="If specified, only sync contents for websites that user the given starter.",
+        )
+
+    def handle(self, *args, **options):
+        starter_qset = WebsiteStarter.objects.all()
+        if "starter_slug" in options:
```
noisecapella, line 25: I think we need to check if `options["starter_slug"]` is empty, since the value will always be set. If it's missing the default value of `""` is used but it's still present in the dict

noisecapella (changes requested):

noisecapella (approved):

gsidebo:
@Ferdi Can we discuss that name change a little further? I'm worried that `is_content` is going to be confusing considering the model name `WebsiteContent`. We would end up having code like `content.is_content`, essentially saying "this content is content, but this other content is not content". It seems like the model name is the real problem - it makes sense for the term "content" to refer to the authored data that will yield a markdown page, and to use another term for the authored data that yields metadata/configuration/etc. Like we discussed in a previous RFC, though, it still makes technical sense to use a single model to store both of those. 

So I think at this point we should (a) come up with a more general model name (`WebsiteItem`?), which would involve a good amount of renaming in the front and back end, or (b) come up with a different name for this flag that is more specific than `is_content`, but still general enough for whatever other content types we're thinking about in the future. When you say 'content in the future that is not "page"', what do you have in mind?

PR #240, opened 2021-04-28
Added support for soft/hard deletes of content

rhysyngsun:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Part of #209 

#### What's this PR do?
This enables deletion of `WebsiteContent` and integrates updates to GitHub by performing deletes as soft deletes, following by an asynchronous hard delete after the deletion has been synced to GitHub.

Find a piece of content for a website and go to the browable API for that (e.g. `http://ocw-studio.odl.local:8043/api/websites/9-012-the-brain-and-cognitive-sciences-ii-spring-2002/content/3ba4040cbae19e58b113c7fd89033f53/`). Click the delete option, and confirm that soft deletes the record and then triggers a task that makes a commit to delete the content.


content_sync/backends/base.py
```diff
@@ -2,6 +2,8 @@
 import abc
 from typing import Any
 
+from safedelete.models import HARD_DELETE
```
mbertrand, line 5: Not used in `base.py`

websites/views.py
```diff
@@ -252,3 +249,9 @@ def get_serializer_context(self):
         if site_config and config_item is not None:
             added_context["is_page_content"] = is_page_content(site_config, config_item)
         return {**super().get_serializer_context(), **added_context}
+
+    def perform_destroy(self, instance: WebsiteContent):
+        """ (soft) deletes a WebsiteContent record """
+        super().perform_destroy(instance)
+        update_website_backend(instance.website)
```
mbertrand, line 256: `api.update_website_backend` calls `backends.sync_website_content` which currently doesn't handle deletes.  Should it?  Or change this to `api.sync_content` -> `backend.sync_content_to_backend` -> `backend.delete_content_in_backend`?

```python
sync_content(instance.content_sync_state)
```

mbertrand (changes requested):

websites/views.py
rhysyngsun: I'm going to let you merge your changes first, since you modified `upsert_content_files_for_user`

rhysyngsun:
@mbertrand all set

content_sync/apis/github.py
```diff
@@ -159,23 +171,25 @@ def upsert_content_files_for_user(self, user_id=None) -> Commit:
         unsynced_states = ContentSyncState.objects.filter(
             Q(content__website=self.website) & Q(content__updated_by=user_id)
         ).exclude(
-            Q(current_checksum=F("synced_checksum")) & Q(synced_checksum__isnull=False)
+            Q(current_checksum=F("synced_checksum"))
```
mbertrand, line 174: I noticed that if I log in as a particular user and delete one file, `upsert_content_files_for_user` is never called for that user, unless there are some other unsynced modified content objects.  This seems to be because the query used in `upsert_content_files` above (`website.websitecontent_set`) does not pick up the deleted content.

mbertrand (changes requested):

content_sync/apis/github.py
rhysyngsun: Huh, I swear I wrote a test to cover this, i'll double check

mbertrand (approved):
Works great 👍

Issue #233, opened 2021-04-26
Handle changes in `WebsiteContent.content_filepath` when syncing with git


mbertrand:
If the `content_filepath` for a `WebsiteContent` object changes, we need to mark its corresponding `ContentSyncState` as unsynced, and modify the git update function to properly move it from its old filepath to the new one.

PR #242, opened 2021-04-28
Handle filepath changes when syncing with github

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/233

#### What's this PR do?
- Adds `content.content_filepath` to the fields used in calculating sync state checksums.  This will later be changed to a dynamic file path based on site config.
- Stores the git filepath to the JSON field `ContentSyncState.data` after a commit to git.  If the current filepath and the sync state filepath don't match on the next sync, marks the git filepath for removal.
- Moves some logic (and related tests) out of `GithubBackend.create_website_in_backend` to `GithubApiWrapper.create_repo`

Tests should pass.
In a shell, you can do the following for an existing website:
```python
from websites.models import *
from content_sync.apis.github import GithubApiWrapper

website = Website.objects.get(id=<website_id>)
content = website.websitecontent_set.first()
content.content_filepath = "newfilepath1.md"
content.save()

api = GithubApiWrapper(website)
api.upsert_content_files()
```

Check the repo on github, there should be a "newfilepath1.md" there.

Then:
```python
content.content_filepath = "newfilepath2.md"
content.save()
api.upsert_content_files()
```

Check github again, "newfilepath1.md" should be gone and "newfilepath2.md" should be there instead.

Eventually the filepath will be determined dynamically from the site_config, and the `WebsiteContent.content_filepath` will be used as an override (and possibly renamed to `filepath_override`).  I think that should probably be in a separate PR.  I can mark this PR as blocked if you'd prefer that change to be done first.

This PR continues to use `WebsiteContent.content_filepath` for now, I tried to make it easier to change later by minimizing referencing it directly by name (ie `filepath = content.content_filepath` and then using `filepath` thereafter).

### DIFF
--- a/content_sync/apis/github.py
+++ b/content_sync/apis/github.py
@@ -152,9 +152,7 @@ class GithubApiWrapper:
             self.upsert_content_files_for_user(user_id)
 
     @retry_on_failure
-    def upsert_content_files_for_user(  # pylint:disable=too-many-locals
-        self, user_id=None, **kwargs
-    ) -> Commit:
+    def upsert_content_files_for_user(self, user_id=None) -> Commit:
         """
         Upsert multiple WebsiteContent objects to github in one commit
         """
@@ -171,7 +169,14 @@ class GithubApiWrapper:
             filepath = content.content_filepath  # TO DO: Use dynamic filepath
             # Add any modified files
             if filepath:
-                synced_filepaths.update({sync_state.id: filepath})
+                synced_filepaths.update(
+                    {
+                        sync_state.id: {
+                            "filepath": filepath,
+                            "checksum": content.calculate_checksum(),
+                        }
+                    }
+                )
                 data = self.format_content_to_file(content)
                 modified_element_list.extend(
                     self.get_tree_elements(sync_state, data, filepath)
@@ -182,12 +187,13 @@ class GithubApiWrapper:
                 modified_element_list, User.objects.filter(id=user_id).first()
             )
 
-            # Mark all as synced
-            for synced_state_id, filepath in synced_filepaths.items():
+            # Save last git filepath and checksum to sync state
+            for synced_state_id, sync_info in synced_filepaths.items():
                 sync_state = ContentSyncState.objects.get(id=synced_state_id)
-                sync_state.current_checksum = sync_state.content.calculate_checksum()
-                sync_state.data = {GIT_DATA_FILEPATH: filepath}
-                sync_state.mark_synced()
+                sync_state.data = {GIT_DATA_FILEPATH: sync_info["filepath"]}
+                sync_state.synced_checksum = sync_info["checksum"]
+                sync_state.save()
+
             return commit
 
     @retry_on_failure
--- a/content_sync/apis/github.py
+++ b/content_sync/apis/github.py
@@ -1,7 +1,7 @@
 """ Github API wrapper"""
 import logging
 from base64 import b64decode
-from typing import Iterator
+from typing import Iterator, List
 
 import yaml
 from django.conf import settings
@@ -163,10 +163,6 @@ class GithubApiWrapper:
         ).exclude(
             Q(current_checksum=F("synced_checksum")) & Q(synced_checksum__isnull=False)
         )
-        repo = self.get_repo()
-        main_ref = repo.get_git_ref(f"heads/{settings.GIT_BRANCH_MAIN}")
-        main_sha = main_ref.object.sha
-        base_tree = repo.get_git_tree(main_sha)
         modified_element_list = []
         synced_filepaths = {}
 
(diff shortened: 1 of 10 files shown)


content_sync/apis/github.py
```diff
@@ -134,22 +154,44 @@ def upsert_content_files_for_user(  # pylint:disable=too-many-locals
...
+                    sync_state.data
+                    and sync_state.data.get(GIT_DATA_FILEPATH, None)
+                    != content.content_filepath
+                ):
+                    modified_element_list.append(
+                        InputGitTreeElement(
+                            sync_state.data["filepath"], "100644", "blob", sha=None
+                        )
+                    )
+                # TODO: Remove deleted content
```
mbertrand, line 191: @rhysyngsun is this part of what you are working on or should I deal with it in this PR?

rhysyngsun: Yeah, I'm working on that.

content_sync/apis/github.py
```diff
@@ -141,22 +158,45 @@ def upsert_content_files_for_user(  # pylint:disable=too-many-locals
...
+                ):
+                    modified_element_list.append(
+                        InputGitTreeElement(
+                            sync_state.data[GIT_DATA_FILEPATH],
+                            "100644",
+                            "blob",
+                            sha=None,
+                        )
+                    )
+
```
rhysyngsun, line 197: Could you extract this into its own function?

content_sync/apis/github.py
```diff
@@ -170,10 +210,11 @@ def upsert_content_files_for_user(  # pylint:disable=too-many-locals
             main_ref.edit(commit.sha)
 
             # Mark all as synced
-            for content in content_set:
-                syncstate = content.content_sync_state
-                syncstate.current_checksum = content.calculate_checksum()
-                syncstate.mark_synced()
+            for synced_state_id, filepath in synced_filepaths.items():
+                sync_state = ContentSyncState.objects.get(id=synced_state_id)
+                sync_state.current_checksum = content.calculate_checksum()
```
rhysyngsun, line 215: I think there's a bug here, because `content` is always the content of the last item iterated from `unsynced_states`

rhysyngsun (changes requested):

content_sync/apis/github.py
mbertrand: oops, should be `sync_state.content.calculate_checksum()`

content_sync/apis/github.py
```diff
@@ -177,42 +173,19 @@ def upsert_content_files_for_user(  # pylint:disable=too-many-locals
...
+            commit = self.commit_tree(
+                modified_element_list, User.objects.filter(id=user_id).first()
             )
-            main_ref.edit(commit.sha)
 
             # Mark all as synced
             for synced_state_id, filepath in synced_filepaths.items():
                 sync_state = ContentSyncState.objects.get(id=synced_state_id)
-                sync_state.current_checksum = content.calculate_checksum()
+                sync_state.current_checksum = sync_state.content.calculate_checksum()
```
rhysyngsun, line 188: Actually, this now introduces a race condition where the content could change between when we read it for the git tree and when we read it here. You could probably add it to the data `synced_filepaths`.

rhysyngsun (changes requested):

content_sync/apis/github.py
mbertrand: Good point, how about something like this?
```python
                synced_filepaths.update(
                    {
                        sync_state.id: {
                            "filepath": filepath, 
                            "checksum": content.calculate_checksum()
                        }
                    }
                )
``` 

and then:

```python
            for synced_state_id, sync_info in synced_filepaths.items():
                sync_state = ContentSyncState.objects.get(id=synced_state_id)
                sync_state.data = {GIT_DATA_FILEPATH: sync_info["filepath"]}
                if sync_state.content.calculate_checksum() == sync_info["checksum"]:
                    sync_state.current_checksum = sync_info["checksum"]
                    sync_state.mark_synced()
                else:
                    sync_state.save()
```

rhysyngsun: I think we need to move away from `mark_synced()`, in fact remove it, cause I don't think there's any benefit and it can in fact cause issues. I think instead you want this to be:


```python
for synced_state_id, sync_info in synced_filepaths.items():
                sync_state = ContentSyncState.objects.get(id=synced_state_id)
                sync_state.data = {GIT_DATA_FILEPATH: sync_info["filepath"]}
                sync_state.synced_checksum = sync_info["checksum"]
                sync_state.save()
```

mbertrand:
@rhysyngsun ready for another look

rhysyngsun (approved):

Issue #241, opened 2021-04-28
add 'relation' field type


aliceriot:
We want to be able to allow content authors to save relations between the content. This should be a bit similar to [Netlify's](https://www.netlifycms.org/docs/widgets/#relation) relation widget but we want to have our own spin on it in a few ways.

PR #252, opened 2021-05-05
remove potential bug source

aliceriot:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

none

#### What's this PR do?

Just removes a potential bug. I found a few lines of code while working on #241 that violate the [rules of hooks](https://reactjs.org/docs/hooks-rules.html), in particular that we must always call hooks at the top level of a function and nowhere else (i.e. never in an if / else block, after an early return, etc).

Everything should be working normally on the `SiteContentEditor` component, and the code changes should make sense.

### DIFF
--- a/static/js/components/SiteContentEditor.test.tsx
+++ b/static/js/components/SiteContentEditor.test.tsx
@@ -260,8 +260,8 @@ describe("SiteContent", () => {
 
   //
   ;[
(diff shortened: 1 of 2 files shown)


noisecapella (approved):
Thanks for catching this. I should remember to look out for rule of hook violations during reviews

PR #263, opened 2021-05-11
add relation field type and widget

### LINKED ISSUE #241: add 'relation' field type

aliceriot:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #241 

#### What's this PR do?

This PR adds a new 'relation' field which can capture a relationship between one piece of content and one or more others. It is basically a wrapper around our existing `SelectField` component which hits the APIs for website content in order to get the possible values that can be selected.

the site override yaml file now has a collection called 'authors' and a relation field from 'post' called 'author.' So you should be able to add an author, and then set an author on a post. The validation for the field should work just like a normal select field, including min and max values and whatnot.

If you save it and then re-open it everything should work as you expect.

Screen Shot 2021-05-11 at 12 50 38 PM
Screen Shot 2021-05-11 at 12 50 48 PM

### DIFF
--- a/localdev/configs/ocw-course-site-config.yml
+++ b/localdev/configs/ocw-course-site-config.yml
@@ -20,11 +20,31 @@ collections:
       - name: address
         label: Address
         widget: object
+        required: false
         fields:
           - { label: "Street Address", name: "street_address", widget: "string", required: true }
           - { label: "City", name: "city", widget: "string", required: true }
           - { label: "State", name: "state", widget: "string", required: true }
           - { label: "Zip Code", name: "zip", widget: "string", required: true }
+      - {
+        "label": "Authors",
+        "name": "authors",
+        "widget": "relation",
+        "collection": "author",
+        "display_field": "title",
+        "multiple": true,
+        "max": 10,
+        "min": 1,
+        "required": false,
+        }
+
+  - name: author
+    label: Author
+    category: Content
+    folder: content
+    fields:
+      - {"label": "Name", "name": "title", "widget": "string", "required": true}
+      - {"label": "Profession", "name": "profession", "widget": "string", "required": false}
 
   - name: resource
     label: Resource
--- a/static/js/types/websites.ts
+++ b/static/js/types/websites.ts
@@ -11,7 +11,8 @@ export enum WidgetVariant {
   String = "string",
   Select = "select",
   Hidden = "hidden",
-  Object = "object"
+  Object = "object",
+  Relation = "relation"
 }
 
 export interface FieldValueCondition {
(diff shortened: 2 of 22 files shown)


localdev/configs/ocw-course-site-config.yml
```diff
@@ -25,6 +25,23 @@ collections:
...
+      - {
+        "label": "Authors",
+        "name": "authors",
+        "widget": "relation",
+        "collection": "author",
+        "display_field": "title",
+        "multiple": true,
+        "max": 10,
+        "min": 1,
+        }
```
gsidebo, line 37: Does this need to be a required field? I haven't seen author data in the front matter in any of the markdown pages for existing OCW course sites. If it doesn't need to be required, let's add `"required": false` here. If you want to do the same for the `address` field above while you're at it, I wouldn't complain 😁

localdev/configs/ocw-course-site-config.yml
```diff
@@ -25,6 +25,23 @@ collections:
...
+
+  - name: author
+    label: Author
+    category: Content
+    folder: content
```
gsidebo, line 42: This might point to the need for a different type of data collection. As this is written, each author you enter will create a `WebsiteContent` record. Right now our concept of `WebsiteContent` is that it represents user-entered data that will be used to create a file for the SSG, whether that's markdown, json, etc. As this is written, each individual author you enter will create a markdown file in the `content` folder in Github. That's definitely not something we want. All we want to do with this content type is to fill in the valid entries that can be used by a different field. I don't have any great ideas off the top of my head to implement that, but it seems like we'd at least want another model, because the current concept of `WebsiteContent` doesn't really fit here. It's probably worth discussing w/ Ferdi and the rest of the team

gsidebo (changes requested):

static/js/types/websites.ts
```diff
@@ -23,6 +24,11 @@ export interface FieldValueCondition {
  * A configuration for a field for site content. This type basically
  * contains the information needed to render the field in the UI, to edit it,
  * validate it, etc.
+ *
+ * TODO: this is getting a bit much, can we declare a ConfigFieldBaseProps interface
+ * and then declare variants, with their WidgetVariant set properly, that set exactly
+ * which props they have? Then ConfigField could be a union of these types - might get
+ * better typechecking that having all these optional properties.
```
gsidebo, line 31: Completely agree. I'll create a tech debt issue for this

localdev/configs/ocw-course-site-config.yml
aliceriot: Sure, we could change this to be a video or something like that where we'd want to have it have it's own page. The idea is just to have the relation field capture relation between two pieces of content. My understanding was that this `ocw-course-site-config.yml` was basically like a placeholder anyhow - for the purposes of development is there something I should be editing instead?

gsidebo: OK I'm on the same page now. I think it would be helpful to follow the example in your RFC more closely and add a "Profession" field or something like that to the "author" content type. Your example is fine to exercise the widget, but the intent is a little confusing with just this single field. We can worry about moving non-OCW-course-site items out of this config in a separate issue. I think it's fine to leave it here

aliceriot:
@gsidebo I updated the `ocw-course-site-config.yml` in line with your comments

gsidebo (approved):

gsidebo:
Looks great! 👍

PR #270, opened 2021-05-12
upgrade eslint config

aliceriot:

none, just a follow-up after #252 

#### What's this PR do?

This upgrades our eslint config to a version that I just pushed today which now includes proper configuration for linting react hooks usage.

build should pass, changes should make sense.

You could try introduced a change that violates the rules of hooks and confirming that you get a linting error too if you wanted.


mbertrand (approved):
👍

PR #280, opened 2021-05-20
split ConfigField type into specific sub-types

### LINKED ISSUE #278: Create specific ConfigField types
_Discussed in #263_

In `static/js/types/websites.ts`, we have a `ConfigField` type that has a bunch of optional properties. We do this because we have a lot of different fields, each of which has its own set of relevant properties. Representing those different fields with a single type and a lot of optional properties is way too permissive and bug-prone. We should look into defining a base type and separate types for each field we support, and replace the uses of the `ConfigField` type with a more specific type wherever possible.

NOTE: We do basically the same thing in the YAML schema we use on the backend. If it's possible to do declaratively, we should look into making the same kind of improvement there. That can be for a separate issue, though.

aliceriot:

closes #278 

#### What's this PR do?

just a typing change, basically creating a `MarkdownConfigField`, `BooleanConfigField`, etc so we can record in the typesystem when various properties should be present on the configuration for a field.

`ConfigField` is redefined to be a union of all these particular types, and some edits are made to introduce the specific types into the codebase where it's possible to be more specific than `ConfigField` (e.g. we pass `field: ObjectConfigField` into the `ObjectField` component).

Exercise the code related to editing and creating site content (and validation and whatnot) and make sure there are no regressions.

### DIFF
--- a/static/js/components/forms/validation.test.ts
+++ b/static/js/components/forms/validation.test.ts
@@ -3,7 +3,14 @@ import { getContentSchema } from "./validation"
 
 import { makeWebsiteConfigField } from "../../util/factories/websites"
 
-import { ConfigItem, WidgetVariant } from "../../types/websites"
+import {
+  ConfigItem,
+  FileConfigField,
+  MarkdownConfigField,
+  StringConfigField,
+  TextConfigField,
+  WidgetVariant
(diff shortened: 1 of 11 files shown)


mbertrand (approved):
👍

PR #299, opened 2021-06-02
Restrict routes not used to login or view home page

### LINKED ISSUE #130: Add login requirement to appropriate routes
See title. This will probably involve redefining some of our front-end routes as private routes. Developer's discretion as to what needs to be private

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #130 

#### What's this PR do?
Restricts UI routes to logged in user, except those used to login or visit the home page

As an anonymous user:
 - I don't know who has SAML set up to login, but you should still be able to do that and view the home page.
 - If you go to `/sites/` you should be redirected to `?next=...` which is just the home page at the moment

As a logged in user, the user experience should feel the same as before. You should be able to logout and end up on the home page.

### DIFF
--- a/main/views_test.py
+++ b/main/views_test.py
@@ -5,6 +5,7 @@ import json
 
 import pytest
 from django.urls import reverse
+from rest_framework.status import HTTP_200_OK, HTTP_302_FOUND
 
 from users.factories import UserFactory
 
@@ -12,12 +13,6 @@ from users.factories import UserFactory
 pytestmark = [
     pytest.mark.django_db,
 ]
-
-
-def test_index_view(client):
-    """Verify the index view is as expected"""
-    response = client.get(reverse("main-index"))
-    assert response.status_code == 200
 
 
 def test_webpack_url(mocker, settings, client):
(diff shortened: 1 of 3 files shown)


main/views_test.py
```diff
@@ -50,20 +59,24 @@ def test_js_settings(settings, client, is_authenticated):
...
+            }
+            if is_authenticated
+            else None,
         }
-        if is_authenticated
-        else None,
-    }
+    else:
+        assert response.status_code == HTTP_302_FOUND
+        assert response.url.startswith("/?next=")
```
aliceriot, line 82: should we have an assert for `js_settings` here too, or is that not important?

aliceriot (changes requested):
just had one question

main/views_test.py
noisecapella: I don't think there would be anything there since it's a redirect. The url being redirected to already has `js_settings` asserted as part of the `main-index` case here, though we may want to update that if we create a specific login page that we redirect to

aliceriot (approved):
👍

PR #305, opened 2021-06-04
Made 'title' field default for 'folder'-type items

### LINKED ISSUE #279: Make 'title' a default form field for "folder"-type config items
Since filename is a field that is relevant to all "folder"-type WebsiteContent records, and the filename will usually be auto-generated from a title, it probably makes sense to require that users enter a title for all "folder"-type WebsiteContent objects. Right now it's possible to have a "folder"-type config item with no title, which would leave us with nothing to generate a filename from. Also, it seems like a good idea for us to force users to enter a title for each of these records so we can guarantee there's something human-readable to use for a text search.

We still want to allow users to add a `title` field in their config as long as it has the same type (string) and is configured to be required. The main goal here is to guarantee that we collect a title for all "folder"-type (repeatable) config items.

**Acceptance criteria:**
- Website content form always shows a required `title` field for a "folder"-type (repeatable) config item, whether the field is in the config item or not
- Site config schema prevents a `title` field for a "folder"-type config item if the field is not "string" type or not required
- If the config item includes a `title` field, the website content form incorporates the properties from that config item (e.g.: the form uses the "label" value from the config item rather than the default of "Title")

gsidebo:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #279 

#### What's this PR do?
Makes the 'title' field a default for "folder"-type config items. If a "title" field is not included in the config item, it will be added in the front end. If a "title" field _is_ included, it will be validated to make sure it is set to required and has the correct widget type

- Edit a content object for a "file"-type config item with no "title" field. You should see no "title" form field.
- Edit a content object for a "folder"-type config item with no "title" field. You should see the "title" field in the form, and is should save/load correctly
- Edit a content object for a "folder"-type config item with a "title" field. You should see the "title" field in the form just like the previous example
- Save a site config with a "folder"-type config item that includes a "title" field with `required: false` or `widget: text` and try to save. You should see a validation error.

### DIFF
--- a/localdev/configs/ocw-course-site-config.yml
+++ b/localdev/configs/ocw-course-site-config.yml
@@ -5,7 +5,6 @@ collections:
     category: Content
     folder: content
     fields:
-      - {"label": "Title", "name": "title", "widget": "string", "required": true}
       - {"label": "Body", "name": "body", "widget": "markdown"}
       - {
         "label": "Tags",
@@ -43,7 +42,6 @@ collections:
     category: Content
     folder: content
     fields:
-      - {"label": "Title", "name": "title", "widget": "string", "required": true}
       - {"label": "Description", "name": "description", "widget": "text"}
       - {"label": "File", "name": "file", "widget": "file"}
       - {
@@ -62,7 +60,7 @@ collections:
         name: sitemetadata
         label: Site Metadata
         fields:
-          - { "label": "Course Title", "name": "title", "widget": "text", "required": true }
+          - { "label": "Course Title", "name": "title", "widget": "string", "required": true }
           - {
             "label": "Course Description",
             "name": "description",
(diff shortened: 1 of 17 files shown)


mbertrand (changes requested):
Needs a rebase, also getting this error when creating a new page:

Resource:
web_1     | [2021-06-14 15:03:56] ERROR 13 [django.request] log.py:224 - [e5eb71abca8d] - Internal Server Error: /api/websites/optionaltitles1/content/
web_1     |   File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
web_1     |     return self.cursor.execute(sql, params)
web_1     | psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "unique_page_content_destination"
web_1     | DETAIL:  Key (website_id, dirpath, filename)=(6d9632cc-fdee-4fed-9782-90f54a618158, content, ) already exists.

PR #306, opened 2021-06-04
link to content in other websites

### LINKED ISSUE #288: add support for linking to collections in other websites using Relation widget
As a site configuration author, I'd like to be free and unconstrained as to which collections I might link. It matters not to me whether they be here or there, in this site or that, I want simply that they be present and linkable.

For this field we essentially want three ways to use it:

1. Linking to a specified collection located in the same website. This is the use-case supported by what we've already implemented.
2. Linking to a specified collection in a specified website. This use-case is supported by the code I wrote here, and would support the use-case you mentioned Ferdi of having a list of departments in ocw-www and then wanting to reference them from within a course.
3. Linking to content within a user-specified website and collection. This would require two dropdowns: one to pick the website from a list of websites, and a second to pick the collection _within_ that website to which we want to be able to link. For example, consider the `resource` collection that we have in our example course config. If we wanted to support linking from one course to resources in another course, a course author would need to be able to select the course, then select linking to the `resource` objects within that course, and then finally get a list of all those resources to be able to select the relevant one.

For these three scenarios we want to support, we can set a `website` value on the ConfigField for the Relation field. Then in the UI we can switch on the `website` value to present relevant functionality to the user.

- if not present, then fetch the items in the `collection` within the same website.
- if specified, then fetch the items in the `collection` within `website`
- if `"*"`, then show the user a drop-down for picking a website. Once they pick a website, show another dropdown to pick which collection within that website. Then, once they've picked a collection, show the entries in that collection so that the user can pick from them.

One slight complication is that to support the "global" / "wildcard" config I think we'd want to store not just the `text_id` of the linked-to content but also the website and collection that it is from, so that we can then repopulate the UI correctly when editing later, etc.

#### Acceptance Criteria:

- [x] Optional configuration on the Relation field for which website the collection is within (if it's not set, assume that it's in the same website)
- [x] make the API call to the content listing API based on the website configuration (if present)
- [x] implement the simple, directly specified case (#306)
- [x] #813 
- [x] update Relation field documentation to reflect these changes

aliceriot:

part of #288 

#### What's this PR do?

This PR adds support to the Relation widget for linking to entries in collections within other websites. This is done by setting a 'website' prop on the Relation field, which is the `name` prop on the website in question.

I am wondering if there is a better approach for doing this, as this would mean that the name of the other website has to be known at config-writing time, and changing it would require re-writing the config.

However, doing it this way is fairly simple in the implementation. If we wanted to make it more complicated, i.e. allowing the user to pick which website they want to link to in the UI, we'd need an additional UI setup for doing that. This would make it more flexible though, and allow the user to link to content in any other website of their choosing.

So anyhow, interested in hearing some feedback about this. I am sort of leaning towards wanting to do the more complicated thing, but not sure.

As I understand it btw, the potential use-case for this is for linking from something like an OCW course to material which is contained in a different course (using the same config) or potentially even to content in a totally different site, for instance in ocw-www (once that is managed by ocw-studio I mean).

If you create two websites with the OCW course example config you should be able to then set a `website` prop on to the 'PDF File' relation field of one of them to the site name of the other (via the django admin) and then link to PDF files in the other website.

This all works and is relatively straightforward, but I'm wondering if this is the right thing to do.

### DIFF
--- a/static/js/components/forms/SiteContentField.tsx
+++ b/static/js/components/forms/SiteContentField.tsx
@@ -22,7 +22,8 @@ export default function SiteContentField({
 
   if (
     field.widget === WidgetVariant.File ||
-    field.widget === WidgetVariant.Boolean
+    field.widget === WidgetVariant.Boolean ||
+    field.widget === WidgetVariant.Relation
   ) {
     extraProps.setFieldValue = setFieldValue
   }
--- a/static/js/components/forms/validation.test.ts
+++ b/static/js/components/forms/validation.test.ts
@@ -167,103 +167,217 @@ describe("form validation util", () => {
     )
   })
 
-  //
-  ;[WidgetVariant.Select, WidgetVariant.Relation].forEach(selectLikeVariant => {
-    describe(`${selectLikeVariant} validation`, () => {
-      const makeSelectLikeConfigItem = (props = {}): [ConfigItem, string] => {
-        const configItem = {
-          ...repeatableConfigItem,
-          fields: [
-            makeWebsiteConfigField({ widget: selectLikeVariant, ...props })
-          ]
-        }
-
-        return [configItem, configItem.fields[0].name]
+  describe("select validation", () => {
+    const makeSelectConfigItem = (props = {}): [ConfigItem, string] => {
+      const configItem = {
+        ...repeatableConfigItem,
+        fields: [
+          makeWebsiteConfigField({ widget: WidgetVariant.Select, ...props })
+        ]
       }
 
-      it(`should validate for a required multiple ${selectLikeVariant} field`, () => {
-        const [configItem, name] = makeSelectLikeConfigItem({
-          multiple: true,
-          required: true
-        })
-        const schema = getContentSchema(configItem)
-        expect(() =>
-          schema.validateSync({
-            ...defaultFormValues,
-            [name]: null
-          })
-        ).toThrow(new yup.ValidationError(`${name} is a required field.`))
-      })
+      return [configItem, configItem.fields[0].name]
+    }
 
-      it(`should pass validation for valid multiple ${selectLikeVariant} values`, async () => {
-        const [configItem, name] = makeSelectLikeConfigItem({ multiple: true })
-        const schema = getContentSchema(configItem)
-        await Promise.all(
-          [[], ["some value"], ["some value", "another value"]].map(
-            async value => {
-              await expect(
-                schema.isValid({
-                  ...defaultFormValues,
-                  [name]: value
-                })
-              ).resolves.toBeTruthy()
-            }
-          )
+    it("should validate for a required multiple select field", () => {
+      const [configItem, name] = makeSelectConfigItem({
+        multiple: true,
+        required: true
+      })
+      const schema = getContentSchema(configItem)
+      expect(() =>
+        schema.validateSync({
+          ...defaultFormValues,
+          [name]: null
+        })
+      ).toThrow(new yup.ValidationError(`${name} is a required field.`))
+    })
+
+    it("should pass validation for valid multiple select values", async () => {
+      const [configItem, name] = makeSelectConfigItem({ multiple: true })
+      const schema = getContentSchema(configItem)
+      await Promise.all(
+        [[], ["some value"], ["some value", "another value"]].map(
+          async value => {
+            await expect(
+              schema.isValid({
+                ...defaultFormValues,
+                [name]: value
+              })
+            ).resolves.toBeTruthy()
+          }
         )
+      )
+    })
+
+    it("should validate a multiple select field with max and min set", async () => {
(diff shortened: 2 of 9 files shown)


noisecapella:
I've been thinking about this and I think it makes more sense to do something where the user can pick a website. I imagine that the use case is something like, a professor is authoring content and wants to link to a pdf in a previous course. If we are doing many websites per website starter, I don't see how you can feasibily hardcore that value. The UI could be as simple as a textbox for the website id plus dropdown for the list of pdfs

Ferdi:
Great, I think there is a need for both cases.  For example: I need to keep a list of all departments  (or all instructors)  as a collection in ocw-www site.  Then , when I author a course, I'd reference a department/instructor.  In this case, as a site config author I know exactly what website course authors need to reference. 

Maybe we do 
`website: webiste_id`

or 
`website: any`

for each c ase

aliceriot:
ok cool, so it sounds like for this field we essentially want three ways to use it:

1. Linking to a specified collection located in the same website. This is the use-case supported by what we've already implemented.
2. Linking to a specified collection in a specified website. This use-case is supported by the code I wrote here, and would support the use-case you mentioned Ferdi of having a list of departments in ocw-www and then wanting to reference them from within a course.
3. Linking to content within a user-specified website and collection. This would require two dropdowns: one to pick the website from a list of websites, and a second to pick the collection _within_ that website to which we want to be able to link. For example, consider the `resource` collection that we have in our example course config. If we wanted to support linking from one course to resources in another course, a course author would need to be able to select the course, then select linking to the `resource` objects within that course, and then finally get a list of all those resources to be able to select the relevant one.

If that makes sense in terms of the three scenarios we want to support, then I'd propose that we add the following things to our existing implementation to get that behavior:

- add a `scope` value to the config. This can take on three values: `Local`, `Specified`, and `Global`, which correspond to the three scenarios / use-cases above.
- keep the `collection` field in the config. If the `scope` is set to `Local` then `collection` is the name of the collection, in the same website to which we can link. If the `scope` is set to `Specified`, then `collection` is the name of the collection we're going to link to in the specified website.
- have a `website` field in the config. This is only relevant if the `scope` is set to `Specified`, in which case `website` defines the name (URL slug) of the website to which we are going to link.

Then in the UI we can switch on the `scope` value to present relevant functionality to the user.

- if `Local`, then fetch the items in the `collection` within the same website.
- if `Specified`, then fetch the items in the `collection` within `website`
- if `Global`, then show the user a drop-down for picking a website. Once they pick a website, show another dropdown to pick which collection within that website. Then, once they've picked a collection, show the entries in that collection so that the user can pick from them.

One slight complication is that to support the `Global` config I think we'd want to store not just the `text_id` of the linked-to content but also the website and collection that it is from, so that we can then repopulate the UI correctly when editing later, etc.

Anyhow, does that all make sense @Ferdi @noisecapella ? What do y'all think?

Ferdi:
Sounds good . Two comments 
not sure you need "scope" but feel free to do it that way.  I think  "website" could be enough.  
website:      # missing or not specified is equivalent to local scope
website: website_id   # specified
website:  *   # or something like that --> global

In the interest of moving a bit faster , and if it makes sense, can we do this in 2 stages / PRs ,  one for "specified" and later for "global"

aliceriot:
mm that's a good point, I think I will do it that way (i.e. use the `website` value as a signal for how it should behave)

yeah multiple PRs sounds like a good plan. @noisecapella what do you think? This PR as is will pretty much implement the 'specified' functionality.

noisecapella:
Sounds good, I'll start the review

noisecapella:
Functionality looks good. One question I have is what happens if the `website` field within `WebsiteStarter` is changed. Maybe we should make the change to have `text_id` or a related field include `website` in this PR instead of waiting for `Global` scope?

aliceriot:
mmm so you mean like so that we save the data the same way regardless of where it's local, in a specified collection, or global?

noisecapella:
I think that's a good idea. It would also make it easier to change the scope in the config without disrupting anything in the websites. Would it be hard to implement as part of this PR?

aliceriot:
I think not super difficult, I'll go ahead and do so

aliceriot:
@noisecapella ok that was a little more involved than I thought it would be! but I updated things to store a `RelationField` value always as an object like

```
{
  website: 'UUID',
  content: ['uuid_of_linked_content']
}
```

I think based on my testing it should all be working correctly, but it was a little tricky to get it working so definitely a good idea to test thoroughly.

noisecapella:
Functionality works great. I'm seeing that if I remove `website` from the config and refresh the page, the pdf isn't listed in the UI properly anymore, but the data is stored correctly on the backend so that can just be a frontend change in some future PR

noisecapella (approved):

PR #310, opened 2021-06-09
don't fail build if coverage upload doesn't work

aliceriot:

none

#### What's this PR do?

build should pass!

should make sense


gumaerc (approved):
👍

PR #313, opened 2021-06-09
Do not use git token for reading from the public site config repo

mbertrand:
N/A

#### What's this PR do?
Does not instantiate the git API with a token from the `content_sync.apis.github.sync_starter_configs` function because it just reads from a public repo.  If the token is set, and is not valid for that organization, a 401 authentication error will be returned from git.

This should work:

```python
from content_sync.apis.github import *
sync_starter_configs(
    "https://github.com/mitodl/ocw-hugo-projects", 
    ["ocw-www/ocw-studio.yaml"], 
    "123729069c5e28c4ff9ae281d3ae2754c42c5549"
)
```


gumaerc (approved):
:+1: LGTM

PR #321, opened 2021-06-15
Add info to backend repo required for publish/preview pipelines

### LINKED ISSUE #319: Add info to backend repo required for publish/preview pipelines
When creating a preview/publish pipeline for a new site, it will be required to know both the site slug and config repo path (can be derived from the WebsiteStarter path).  This information should be made available in 1 or more repo files.  

Since not all website starters are necessarily github-based, maybe the starter config itself should be saved to a file in the site repo?

mbertrand:
This is probably no longer necessary since the pipelines will be created from ocw-studio (#366)

#### Pre-Flight checklist
- [x] Migrations
- [x] Testing
  - [x] Code is tested

Closes #319 

#### What's this PR do?
- Changes the initial commit for new repos to be a `site_config_vars.yml` file containing the following info, formatted to be easily parsed by a bash script in a concourse publishing pipeline:

    ```
    CONFIG_PATH={website.starter.path}
    CONFIG_SLUG={website.starter.slug}
    SITE_SLUG={website.name}"
    ```
- Adds a `backpopulate_site_configs` management command to add the above file to one website or all websites with existing repos.
- Adds a migration to the `WebsiteStarter` model so that `slug` and `path` must be unique together, not `slug` alone (different repos might have the same slug).

- Set up github integration by following the README
- Create a new `Website` via the UI.  
- Check the website's git repo and verify that there is a `site_config_vars.yml` file there with the correct content.
- Run the `backpopulate_site_configs` command and verify that all the `Website` objects for which repos already exist have a `site_config_varsyml` file afterward.

The concourse pipeline that will be generated for each `Website` will be running the following:
```
hugo --config ../ocw-hugo-projects/((site-config))/config.yaml --themesDir ../ocw-hugo-themes/theme
```
and then uploading files to an S3 bucket with a path dependent on the `Website.name` value:
```
destination:           
   - dir: s3-remote:((ocw-bucket))/courses/((course-id))
```

This file is meant to provide the info needed to replace `((site-config))` and `((course-id))` values above.

This is assuming that all the websites we want to publish via concourse will be associated with a `WebsiteStarter` of `source=github` and that the repo will have 1 or more yaml files under its path/slug containing content that looks something like this:

```
---
baseUrl: "/"
enableRobotsTXT: true
languageCode: en-usrelative
URLs: false
title: MIT OpenCourseWare
theme: ["base-theme", "course"]
```

Pipelines should not be generated if such a file cannot be found at the specified github url/path (but that is a separate issue).

I originally thought about making this file a required `WebsiteContent` object and handling it that way, but ultimately felt like it didn't quite fit that use case.  The content of this file is info that currently cannot be modified after a  `Website` is created.


blarghmatey:
It might actually be better to format the values as YAML or JSON and use the [`load-vars`](https://concourse-ci.org/jobs.html#load-var-step) functionality in Concourse. This way we can use the `((.:site_config.foo))` syntax which expands beyond only being usable via Bash.

websites/utils.py
```diff
@@ -1,13 +1,17 @@
...
         return constants.GLOBAL_ADMIN
     elif role in constants.ROLE_GROUP_MAPPING.keys():
         return f"{constants.ROLE_GROUP_MAPPING[role]}{website.uuid.hex}"
     else:
         raise ValueError(f"Invalid role for a website group: {role}")
+
+
+def format_site_config_env(website: "Website") -> str:
+    """Return a string containing info for hugo publishing"""
+    return f"CONFIG_PATH={website.starter.path}\nCONFIG_SLUG={website.starter.slug}\nSITE_SLUG={website.name}"
```
blarghmatey, line 17: By making the site_config a YAML or JSON object we can elminate the hard-coding of these keys and just serialize the data structure that we want, making it easier to evolve the data structure.

mbertrand:
This is no longer relevant, closing.

PR #322, opened 2021-06-16
Added omnibus site config and pared down course config

### LINKED ISSUE #298: Add omnibus site config for local development
We have been adding config items to the OCW course site config just to act as a proof of concept for widgets that we want to be able to test, but some of them are not relevant to OCW course sites. The address fields, the author config item, and the relation field that refers to authors are examples of this. We should remove the fields that are not directly relevant to OCW course sites, and create a separate example config that exercises all of the widgets that we care about.

As a bonus, once we remove the unnecessary config items/fields from the OCW course site config, it would be nice to manually update that config in production since the production app does not include `localdev`, meaning we can't use the management commands to update that site config.

gsidebo:
#### Pre-Flight checklist

- [x] Migrations

Closes #298 

#### What's this PR do?
- Adds a site config that is intended to exercise all of our widget types
- Removes fields that aren't OCW course site-related from the OCW course site config
- Updates the `override_site_config` command to update the omnibus config by default

Run the migrations, and run the `override_site_config` command. You should be able to use the omnibus config for new sites, and that config should exercise every available widget


aliceriot:
checked out the branch, ran the migrations, and ran `override_site_config` but I only see the 'OCW Course Hugo Starter' in my dropdown for picking a starter:

Screen Shot 2021-06-16 at 9 28 17 AM

is there something else I missed that I need to do?

aliceriot:
it is there in the admin though...so not sure what's up?

gsidebo:
@aliceriot Hmmm I'll take a look

gsidebo:
@aliceriot It's working for me. Dumb question: have you tried refreshing the page or visiting it from a private browser?

aliceriot:
I'm not sure what's up, although it seems it's only returning the Hugo Course starter in the API:

Screen Shot 2021-06-16 at 3 37 48 PM

they're both there in the admin:

Screen Shot 2021-06-16 at 3 38 42 PM

should it matter if the source is set to 'github' for the one and 'local' for the other?

aliceriot:
it's consistent in incognito, logging in and out, restarting docker, etc too

aliceriot:
nvm this is my bad, I didn't know about `FEATURE_USE_LOCAL_STARTERS`

aliceriot (approved):
LGTM!

PR #330, opened 2021-06-17
Add MarkdownEditor tests

### LINKED ISSUE #138: Add widget component tests
Our custom widget components (right now that's file upload and markdown) are only tested indirectly. We have an example starter in JS which happens to exercise all of the widget types, and then a test case which makes sure they render. We should add some widget component-specific test cases

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #138 

#### What's this PR do?
Adds a couple tests for `MarkdownEditor`, mocking out ckeditor to avoid complications with it. We already have tests for the other custom components so this should complete the issue.

Tests should pass

### DIFF
--- a/static/js/components/widgets/MarkdownEditor.test.tsx
+++ b/static/js/components/widgets/MarkdownEditor.test.tsx
@@ -30,7 +30,7 @@ describe("MarkdownEditor", () => {
   ;[
     [true, MinimalEditorConfig],
     [false, FullEditorConfig]
-  ].forEach(([minimal, expectedMinimalComponent]) => {
+  ].forEach(([minimal, expectedComponent]) => {
     [
       ["value", "value"],
       [null, ""]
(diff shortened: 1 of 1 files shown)


static/js/components/widgets/MarkdownEditor.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  afterEach(() => {
+    sandbox.restore()
+  })
+
+  //
+  ;[
+    [true, MinimalEditorConfig],
+    [false, FullEditorConfig]
+  ].forEach(([minimal, expectedMinimalComponent]) => {
```
aliceriot, line 33: should maybe `expectedMinimalComponent` be called `expectedConfig` here?

aliceriot (approved):
just had a variable name question, but otherwise looks good! nice to have a few tests around this.

PR #333, opened 2021-06-21
Include uploaded filepath in backend (github) metadata

mbertrand:
Blocked by / based on #329 

#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #333 

#### What's this PR do?
- Modifies the content_sync serializer to include a `WebsiteContent.file` url in the metadata if it exists.
- Sets uploaded file ACL permissions to `public-read`

- Create a new `WebsiteContent` object with a file upload field and save
- Check that the uploaded file url is saved to github
- Check that you can view the S3 URL without getting an "Access Denied" message
- Edit the content above, and change just the file to another file, then save.
- Check that the uploaded file url is also updated in github
- Check that you can view the S3 URL without getting an "Access Denied" message


gumaerc (approved):
:+1: Code looks good, and I was able to successfully follow the testing procedures outlined.  However, there seems to be a possibly unrelated issue worth noting.  When I set `OCW_STUDIO_USE_S3` to true in my `.env` file and run `docker-compose run web pytest` locally, the tests fail on `test_s3_settings`.  Shouldn't your local development env settings not affect unit testing?  Environment variables like this could be better documented in general as well.

PR #335, opened 2021-06-21
added serializers and tests for the website collection related objects

### LINKED ISSUE #334: add serializers for updating WebsiteCollection and WebsiteCollectionItem models
We need serializers!

these are just a little more complicated than normal because of the `position` property on the `WebsiteCollectionItems`. I'm going to base these on the code we already wrote for handling very similar things in Open Discussions.

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #334

#### What's this PR do?

This adds serializers for the `WebsiteCollection` and `WebsiteCollectionItem` models.

hm, we'll I suppose it's not so easy to test this manually - I think you could do so in the django shell if you wanted. I mainly just wanted to keep my PR from getting too big, so I pulled this out into a separate PR. Will be doing another shortly to add the views to actually get APIs going with these.

But definitely read through the code carefully and make sure it makes sense!

### DIFF
--- a/websites/serializers.py
+++ b/websites/serializers.py
@@ -1,6 +1,7 @@
 """ Serializers for websites """
 from django.contrib.auth.models import Group
 from django.db import transaction
+from django.db.models import F, Max
 from guardian.shortcuts import get_groups_with_perms, get_users_with_perms
 from rest_framework import serializers
 from rest_framework.exceptions import ValidationError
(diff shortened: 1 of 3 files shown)


websites/serializers.py
```diff
@@ -261,3 +268,72 @@ class Meta:
...
+
+
+class WebsiteCollectionSerializer(serializers.ModelSerializer):
+    """Serializer for WebsiteCollections"""
+
+    item_count = serializers.SerializerMethodField(read_only=True)
+
+    def get_item_count(self, instance):
+        """get the number of items in this collection"""
+        return WebsiteCollectionItem.objects.filter(website_collection=instance).count()
```
noisecapella, line 280: This could be an n+1 query but it may not matter. Are there likely to be a lot of `WebsiteCollection`s?

aliceriot: I'm not sure, I think at least a few

I wasn't sure actually if we even needed this field, or if it maybe would be better to set it as a value on the model whenever we add or remove a `WebsiteCollectionItem`. I was also wondering if maybe this is something where I would need to do something with the `get_queryset` method on the associated view in order to eliminate possible issues - what do you think?

aliceriot: I think I'll just take out the 'num_items' thing for now to keep it simple

aliceriot:
@noisecapella I updated it to remove that thing for now

noisecapella (approved):

Issue #224, opened 2021-04-21
Implement UX for successful "file"-type form submissions


gsidebo:
In #192 (PR #217), we implemented the UI for "file"-type config items, which lets users add/update content for those config items. There should be some feedback to the user when those forms successfully submit.

gsidebo:
Blocked until we decide on the look and feel

pdpinch:
We removed this UI. Files are only added to a course through syncing with google drive now.

PR #340, opened 2021-06-23
Added front-end for configuring navigation menus

gsidebo:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Migrations
- [x] Testing
  - [x] Code is tested

Closes #340

#### What's this PR do?
- Adds the front-end portion for configuring navigation menus 
- Adds the [`react-nestable`](https://github.com/primetwig/react-nestable) package to support drag and drop

- Update your local configs (`manage.py override_site_config`)
- Create a new site or select an existing site with the OCW course site starter
- Select Settings > Menu, and click "Add New" to add pages/resources to the navigation menu

From here, you'll want to test several things:
- You can add pages/resources _or_ external links as menu items
- You can reorder and nest these items freely
- You can update individual items using the gear icon
- You can remove items using the trash bin icon
- Changes are only permanently applied if you click "Save"
- The exact same menu is shown if you refresh the page (if the menu was saved)
- The data that is saved for the menu is in a format that is compatible with the [Hugo menu format](https://gohugo.io/content-management/menus/#add-non-content-entries-to-a-menu), minus the `url` property for internal content links as those will be filled in when site content is pushed to Github 

`static/js/components/widgets/MenuField.tsx`

- To speed things up, this is being done in a Hugo-specific way. By that I mean the data structure that is being used to save the menus is in the format that Hugo expects ([doc](https://gohugo.io/content-management/menus/#add-non-content-entries-to-a-menu))

ss 2021-06-23 at 17 53 49
ss 2021-06-24 at 16 31 32
ss 2021-06-24 at 16 32 37
ss 2021-06-23 at 18 05 42


noisecapella:
Some things I've noticed:
 - If I'm on the Menu page, click on Metadata in the sidebar, then click on Menu again, it shows no items listed in the menu. If I do a full refresh the items return

UX items. I'm not sure if these is in the scope of this issue:
 - It's not obvious when I need to click the Save button on the menu page. It might be helpful to have some indicator like a `*` which says that there are unsaved changes, as well as a confirmation if you try to navigate away with unsaved changes.
 - Clicking away from the dialog means you lose your changes. However this is the same behavior for editing other content so it's likely a larger UX decision

noisecapella:
I'm wondering what happens if you delete content which has an internal link pointing to it. We don't support deleting content yet in the UI but it's supported in the REST API which makes me think we will soon

noisecapella:
If you add multiple external links with the same url, they have the same `identifier` which causes problems when moving the items around in the menu. This is an edge case to be fair but we should do something to forbid duplicate links to avoid bugs with menu placement, or maybe create unique ids for `identifier` instead of generating them from the url string

websites/config_schema/site-config-schema.yml
```diff
@@ -31,6 +31,7 @@ field:
     condition: include('field_condition', required=False)
     display_field: str(required=False)
     collection: str(required=False)
+    collections: list(str(), required=False)
```
noisecapella, line 34: What is this for?

static/js/components/widgets/MenuField.tsx
```diff
@@ -0,0 +1,366 @@
...
+
+const hugoItemToInternal = (item: HugoItem): InternalSortableMenuItem => {
+  const partialHugoItem: {
+    targetContentId: string | null
+    targetUrl: string | null
+  } = {
+    targetContentId: null,
+    targetUrl:       null
+  }
+  if (isUuid4(item.identifier)) {
```
noisecapella, line 59: Is it possible to add another field `menuItemType` to the `HugoItem`? Or if that's not possible, can we change the identifier to have a prefix like `external-...` and `internal-...` so that this is more explicit?

noisecapella (changes requested):

gsidebo:
@noisecapella 

> It's not obvious when I need to click the Save button on the menu page
We can roll that into this UX issue: https://github.com/mitodl/ocw-studio/issues/224

> Clicking away from the dialog means you lose your changes
I think that's a won't-fix until we hear different

> I'm wondering what happens if you delete content which has an internal link pointing to it
Automatically handling broken page links would be a pretty big undertaking with lots of UX decisions. Not sure that's something we'll tackle at any point. Definitely not something to include as part of this issue.

> If I'm on the Menu page, click on Metadata in the sidebar, then click on Menu again
Great catch, I'll look into it

> multiple external links with the same url
Think I should be able to solve this in a straightforward way

Ferdi:
> Clicking away from the dialog means you lose your changes

can we open an issue to track this ?

static/js/components/widgets/MenuField.tsx
gsidebo: Not a bad idea, but that prefix would need to be accounted for in a bunch of different places (here, in the form component, in the back end). We do get a slight advantage of brevity with the existing approach. Given the time crunch I think this is something we could address later

gsidebo:
@noisecapella Ready for review again. Issue to track that sidebar form behavior: https://github.com/mitodl/ocw-studio/issues/367

static/js/components/widgets/MenuField.tsx
noisecapella: I'm fine leaving it but IMO it is very unintuitive and we won't have time to fix this later either

websites/config_schema/site-config-schema.yml
noisecapella: @gsidebo is this required for the PR? I'm confused that there's a `collection` and `collections`

static/js/components/widgets/MenuField.tsx
gsidebo: @noisecapella Thought of a good compromise that doesn't involve a huge change: I'll prefix the external links. The absence of the prefix on internal links should be a helpful indicator.

websites/config_schema/site-config-schema.yml
gsidebo: I initially implemented this in a way that required the site config to point directly to the types that would be used for internal links. For example, the config would have to include `"collections": ["page", "resource"]` to indicate that only those items should be included in the list of available internal links. Instead of that we're just allowing all page content in that list, but the logic was implemented and could be useful in the future. You can test it by adding `"collections": ["page"]` et al to a menu field in the site config

gsidebo:
@noisecapella ready

static/js/components/widgets/MenuField.tsx
noisecapella: Thanks. Will we need to do something in the future to support internal links to other websites within ocw-studio? Not sure if there is a use case or not

noisecapella:
Do you have any thoughts on this? https://github.com/mitodl/ocw-studio/pull/340#discussion_r665573608
I was thinking about this in the context of this other comment on linking to other content in websites via `RelationField`: https://github.com/mitodl/ocw-studio/pull/306#issuecomment-868796708

websites/migrations/0029_blank_filename_dirpath.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 3.1.6 on 2021-06-22 14:35
```
noisecapella, line 1: There is a conflicting migration 0029_websitecollection_owner so I think this will need to be adjusted

noisecapella (changes requested):

static/js/components/widgets/MenuField.tsx
gsidebo: I would consider that an external link. We haven't discussed anything about linking to other sites within Studio so I don't think that needs to be part of this PR

noisecapella (changes requested):

static/js/components/widgets/MenuField.tsx
```diff
@@ -0,0 +1,434 @@
...
+
+  const startEditMenuItem = (menuItem: InternalSortableMenuItem) => {
+    setPanelState({
+      activeItemPath: getItemPath(menuData.internalItems, menuItem.id),
+      activeItem:     menuItem,
+      isVisible:      true
+    })
+  }
+
+  const removeMenuItem = () => {
```
noisecapella, line 298: It looks like this isn't covered by tests. Can we add a test for removing a menu item?

noisecapella (changes requested):

static/js/components/forms/MenuItemForm.tsx
```diff
@@ -0,0 +1,182 @@
...
+            </div>
+            <div className="form-group w-100">
+              <input
+                type="radio"
+                id="menuItemTypeInternal"
+                name="menuItemType"
+                value={LinkType.Internal}
+                checked={values.menuItemType === LinkType.Internal}
+                onChange={() => {
+                  setFieldValue("menuItemType", LinkType.Internal)
```
noisecapella, line 108: It looks like this function and the similar one for external links aren't covered. Can we add a test that shows a change of state when these two radio buttons are clicked?

noisecapella (changes requested):

noisecapella:
Do we need to add a case for defaultForValue for `WidgetVariant.Menu` or is an empty string a good default value? https://github.com/mitodl/ocw-studio/blob/master/static/js/lib/site_content.ts#L207

static/js/components/widgets/MenuField.test.tsx
```diff
@@ -0,0 +1,218 @@
...
+          internalLink:  internalLinkUuid
+        }
+        expectedMenuItem = {
+          name: title
+        }
+        expPartialIdentifier = internalLinkUuid
+      } else {
+        submitData = {
+          menuItemTitle: title,
+          menuItemType:  "external",
```
noisecapella, line 182: Can we use `LinkType.External` here?

static/js/components/widgets/MenuField.tsx
```diff
@@ -0,0 +1,434 @@
...
+    values,
+    hideModal
+  }: {
+    values: MenuItemFormValues
+    hideModal: () => void
+  }) => {
+    let updatedItems: InternalSortableMenuItem[]
+    const updatedItem = {
+      text: values.menuItemTitle,
+      ...(values.menuItemType === "external" ?
```
noisecapella, line 343: Can we use `LinkType.External` here?

noisecapella (changes requested):

gsidebo:
> Do we need to add a case for defaultForValue for WidgetVariant.Menu or is an empty string a good default value?

Empty string seems to be working since the form initializes without any issues when there is no active item/value

noisecapella (approved):
Looks great :+1:

PR #351, opened 2021-06-29
Skip validation if fields are not visible

### LINKED ISSUE #337: validation gets blocked by required fields in hidden objects
### Steps to Reproduce

 - Create a `WebsiteStarter` configuration that has the following configuration:
```
  - category: Content
    folder: content/resources
    label: Resources
    name: resource
    fields:
      - label: Title
        name: title
        required: true
        widget: string
      - label: Description
        name: description
        widget: markdown
        minimal: true
      - label: File Type
        name: filetype
        required: true
        widget: select
        options:
          - Image
          - Video
          - Document
          - Other
      - label: File
        name: file
        widget: file

      # show the field below only if the type of resource is "image"
      - label: Image Metadata
        name: metadata
        widget: object
        condition: { field: filetype, equals: Image }
        fields:
          - label: Text alternative
            name: image-alt
            widget: string
            required: true
          - label: Caption
            name: caption
            widget: string
          - label: Credit
            name: credit
            widget: text
```
 - Create a new `Website` using the `WebsiteStarter` you just created
 - Browse to the UI to edit this site, and bring up the form to add a new Resource
 - Select a `filetype` other than Image
 - Try and save the resource

### Expected Behavior

The resource saves using the supplied settings

### Actual Behavior

It does not save and does not indicate why it did not save

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #337 

#### What's this PR do?
This disables validation for fields which aren't visible in the UI due to being hidden fields or hidden because of a condition filter. 

There are some edge cases that aren't handled:
 - Collapse/expand behavior of `ObjectField` is not affected by this PR. I think it makes sense for validation to still happen for fields which are inside a collapsed `ObjectField` since it matches the behavior of most HTML forms I've seen. However I think we need some way to highlight the validation error outside the `ObjectField` so that there's some indication of a problem when the `ObjectField` is collapsed.
 - It seems like the `condition` filter can only point to top-level fields. Do we have a standard for referencing nested fields like `Formik` does using periods as separators? In any case our code only handles filtering top-level fields at the moment so this validation PR does that too.
 - Similarly, our hidden field code doesn't currently handle conditionally hiding nested fields. In the example schema in #337 the condition filter is a part of the `ObjectField`, not any of the nested fields. The validation is on a nested field which isn't handled by this PR, but this PR removes all validation for the `ObjectField` so all nested fields are included in that.

Copy the schema in the description of #337 into a starter config for a website. Try to add a new Resource. Enter something for the title, choose a file type which is not Image, and upload a small file. Click 'Save'. On master the Save button should not do anything, but on this PR you should be able to create a resource with the new file correctly.

### DIFF
--- a/static/js/components/forms/validation.test.ts
+++ b/static/js/components/forms/validation.test.ts
@@ -1,5 +1,7 @@
 import * as yup from "yup"
 import { getContentSchema } from "./validation"
+import * as siteContentFuncs from "../../lib/site_content"
+import sinon, { SinonSandbox } from "sinon"
 
 import {
   makeRepeatableConfigItem,
@@ -18,7 +20,6 @@ import {
 } from "../../types/websites"
 
 describe("form validation util", () => {
-  const yupFileFieldSchema = yup.mixed()
   const repeatableConfigItem = makeRepeatableConfigItem()
   const singletonConfigItem = makeSingletonConfigItem()
   const partialField = {
@@ -28,7 +29,15 @@ describe("form validation util", () => {
   const defaultFormValues = {
     title: "My Title"
   }
-  let configItem: ConfigItem
+  let configItem: ConfigItem, sandbox: SinonSandbox
+
+  beforeEach(() => {
+    sandbox = sinon.createSandbox()
+  })
+
+  afterEach(() => {
+    sandbox.restore()
+  })
 
   describe("for 'title' field", () => {
     const titleField: StringConfigField = {
@@ -52,7 +61,7 @@ describe("form validation util", () => {
           // @ts-ignore
           fields: field ? [field] : []
         }
-        const schema = getContentSchema(configItem)
+        const schema = getContentSchema(configItem, {})
         if (expAddedTitleField) {
           expect(() => schema.validateSync({})).toThrow(
             new yup.ValidationError("Title is a required field")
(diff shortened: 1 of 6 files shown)


static/js/components/forms/validation.test.ts
```diff
@@ -69,37 +78,74 @@ describe("form validation util", () => {
     })
   })
 
-  it("produces a validation schema for fields regardless of whether they're required or not", () => {
```
noisecapella, line 72: I removed this test because the comparison of `schema.toString()` just resulted in `[object Object]` being compared. This particular test didn't seem to do much. Other tests below I adjusted to run a validation

aliceriot: hmm yeah definitely a weird test...I did modify it a bit and get it to pass by doing `JSON.stringify` on the two objects instead, but I had to modify the compared-to schema bunch to get that to pass too.

aliceriot:
functionality works! reading code now.

static/js/components/forms/validation.test.ts
```diff
@@ -115,18 +161,13 @@ describe("form validation util", () => {
...
-          .object()
-          .shape({
-            [partialField.name]: expectedYupField
-              // @ts-ignore
-              .label(partialField.label)
-              .required()
-          })
-          .toString()
-      )
+      const schema = getContentSchema(configItem, {})
```
aliceriot, line 164: I think if we wanted to preserve the schema equality checks we probably could do so with `JSON.stringify` or `_.deepEqual` or someting, I'm a little torn on how valuable those are though

aliceriot (changes requested):
all looks good to me, just one question really

static/js/components/forms/validation.test.ts
noisecapella: It gets kind of tricky since there are many internal fields as well as some anonymous functions which don't compare well. I tried trimming away some of the internal fields but I feel like some of the relevant state is in a closure where we can't serialize it

aliceriot: word, let's just not worry about it then 😄

aliceriot (approved):

PR #358, opened 2021-07-01
add documentation comments to two types

aliceriot:

none, these things were introduced in #296 though

#### What's this PR do?

Ran across these two types in reading the code again, and just thought it would be a good idea to add a little documentation of what's up with them.

this is a non-code change, so just read the comments and make sure you think they make sense.


mbertrand (approved):
👍

PR #389, opened 2021-07-15
Revert "Generate concourse pipelines on website creation (#366)"

mbertrand:
This reverts commit 52dd75fdce957cedea31d00acb5e914d373183aa because `fly` is not getting installed on heroku as epxected.  Need an alternative approach to do that before this will work.

N/A

#### What's this PR do?
Reverts the pipeline commit because the fly executable is not being installed as expected.

Tests should pass


gumaerc (approved):
:+1:

PR #402, opened 2021-07-19
Use remote search for RelationField widget

### LINKED ISSUE #392: Use async search on type for RelationField
As a course author, I'd like to have a dropdown list of content items and pick several items from that list. However, in some cases I want to pick from thousands of content items, for example to add instructors to my course. In that case I want it to update with results as I type to keep the user experience quick and to reduce load on the server

#### Acceptance Criteria:

- [ ] Add optional `search` parameter on `/api/websites/.../content/` to filter content by title
- [ ] Update `RelationField` to use `react-select`'s async functionality to fetch results as a user types

#### Nice to have
- [ ] Implement an infinite scroll to fetch results as the user scrolls https://github.com/vtaits/react-select-async-paginate

noisecapella:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Fixes #392 

#### What's this PR do?
Use async react-select widget to do searches via RelationField. This means that searches will happen as the user types via API requests to the backend, rather than fetching all data at once and doing the filtering on the frontend.

TBD


mbertrand (approved):
LGTM 👍

PR #405, opened 2021-07-20
Website.short_id for repo names

mbertrand:
#### Pre-Flight checklist

- [x] Screenshots and design review for any changes that affect layout or styling
  - [x] Desktop screenshots
  - [ ] Mobile width screenshots
- [x] Migrations
  - [x] Migration is backwards-compatible with current production code
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
Closes #404 

#### What's this PR do?
- Adds `short_id` as a required, unique field to `Website`
- Adds the new field to the Website creation form
- Populates the field based on metadata during the OCW import process
- Updates the concourse pipeline definition to use this field as the git repo name
- Updates the `GithubApiWrapper` to use this field as the git repo name
- Backpopulates the field for existing websites

#### How should this be manually tested?
On the master branch, import some ocw sites if you don't have any yet (you will need some .env variables set: ):
  ```
  manage.py import_ocw_course_sites --bucket ocw-to-hugo-output-qa  --filter 15-00
  ```
Then switch to this branch and restart containers.  In django admin, check that all websites have a populated short_id field, and that the imported ocw sites have a short_id == `<course_number>-<semester>-<year>` in lowercase.

In particular, filter by "15-005" - there should be two ocw sites, one should have a `short_id=res.15-005-spring-2019` and the other should have `short_id=res.15-005-spring-2019-2`

Import some more sites on this branch, and make sure their short_id values are populated as above:
  ```
  manage.py import_ocw_course_sites --bucket ocw-to-hugo-output-qa  --filter 16-00
  ```

Create a new Website using the UI, make sure the short_id field appears on the form, validates correctly, and that the site is saved with the correct values.

#### Any background context you want to provide?
Since there are no production repos yet and just a handful of RC repos, decided to not create a management command for renaming them.

#### Screenshots (if appropriate)
Screen Shot 2021-07-21 at 12 29 37 PM

Screen Shot 2021-07-21 at 12 31 02 PM

PR #407, opened 2021-07-21
Async search backend

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Part of #392 

#### What's this PR do?
Adds backend parameters, extra context and tests to enable async relation fields in the frontend. There are three additions:
 - `text_id` parameter to allow filtering on text_id. This is useful for relation fields in menus, so that the serialized `WebsiteContent` for a selected menu item can be retrieved, so that we have the information to display the label
 - `content_context` is included in the `WebsiteContent` detailed listing. It should include any serialized `WebsiteContent` for `text_id`s which appear in the relation field value, so that we have this information to display the label on the frontend
 - `search` query parameter is added which does a case insensitive filter on `title`. **Note**: `title` can be overridden by specifying another value for `display_field`, so searches may not necessarily match the text that's displayed on the frontend. I don't know of a way around this issue since we would need to look up the starter config for each `WebsiteConfig` in order to know which field to search on.

Assume you start with some content in a website: 
 - Create a `RelationField` on some page on your site and select a piece of content.
 - Go to this URL: `/api/websites/website-slug-here/content/?detailed_list=true`, replacing your website name in the URL. You should see a new list called `content_context` in the results. 
 - Look for the content you updated with the relation field. It should be a list with one serialized `WebsiteContent` inside it. For other items without a relationfield you should see an empty list. If you have a relation field with many=true then you may see multiple serialized objects in the list.
 - Add `text_id=...` to the query parameters. You should see only the content with that text_id

### DIFF
--- a/websites/serializers.py
+++ b/websites/serializers.py
@@ -230,7 +230,8 @@ class WebsiteContentDetailSerializer(
         metadata = instance.metadata or {}
         site_config = SiteConfig(instance.website.starter.config)
         for field in site_config.iter_fields():
-            if field.field.get("widget") == "relation":
+            widget = field.field.get("widget")
+            if widget in ("relation", "menu"):
                 try:
                     if field.parent_field is None:
                         value = metadata.get(field.field["name"])
@@ -239,12 +240,22 @@ class WebsiteContentDetailSerializer(
                             field.field["name"]
                         )
 
-                    content = value["content"]
-                    website_name = value["website"]
-                    if isinstance(content, str):
-                        text_ids.append(content)
-                    else:
-                        text_ids.extend(content)
+                    if widget == "relation":
+                        content = value["content"]
+                        website_name = value["website"]
+                        if isinstance(content, str):
+                            text_ids.append(content)
+                        else:
+                            text_ids.extend(content)
+                    elif widget == "menu":
+                        website_name = instance.website.name
+                        text_ids.extend(
+                            [
+                                item["identifier"]
+                                for item in value
+                                if not item["identifier"].startswith("external-")
+                            ]
+                        )
 
                 except (AttributeError, KeyError, TypeError):
                     # Either missing or malformed relation field value
--- a/websites/serializers.py
+++ b/websites/serializers.py
@@ -1,4 +1,6 @@
 """ Serializers for websites """
+import logging
+
 from django.contrib.auth.models import Group
 from django.db import transaction
 from django.db.models import F, Max
@@ -22,7 +24,11 @@ from websites.models import (
     WebsiteStarter,
 )
 from websites.permissions import is_global_admin, is_site_admin
+from websites.site_config_api import SiteConfig
 from websites.utils import permissions_group_name_for_role
+
+
+log = logging.getLogger(__name__)
 
 
 ROLE_ERROR_MESSAGES = {"invalid_choice": "Invalid role", "required": "Role is required"}
@@ -202,6 +208,8 @@ class WebsiteContentDetailSerializer(
 ):
     """Serializes more parts of WebsiteContent, including content or other things which are too big for the list view"""
 
+    content_context = serializers.SerializerMethodField()
+
     def update(self, instance, validated_data):
         """Add updated_by to the data"""
         instance = super().update(
@@ -209,6 +217,45 @@ class WebsiteContentDetailSerializer(
         )
         update_website_backend(instance.website)
         return instance
+
+    def get_content_context(self, instance):
+        """
+        Create mapping of uuid to a display name for any values in the metadata
+        """
+        if not self.context or not self.context.get("content_context"):
+            return None
+
+        text_ids = []
+        website_name = None
+        metadata = instance.metadata or {}
+        site_config = SiteConfig(instance.website.starter.config)
+        for field in site_config.iter_fields():
+            if field.field.get("widget") == "relation":
+                try:
+                    if field.parent_field is None:
+                        value = metadata.get(field.field["name"])
+                    else:
+                        value = metadata.get(field.parent_field["name"], {}).get(
+                            field.field["name"]
+                        )
+
+                    content = value["content"]
+                    website_name = value["website"]
+                    if isinstance(content, str):
+                        text_ids.append(content)
+                    else:
+                        text_ids.extend(content)
+
+                except (AttributeError, KeyError, TypeError):
+                    # Either missing or malformed relation field value
+                    continue
+
+        contents = WebsiteContent.objects.filter(
+            text_id__in=text_ids, website__name=website_name
+        )
+        return WebsiteContentDetailSerializer(
+            contents, many=True, context={"content_context": False}
+        ).data
 
     def to_representation(self, instance):
         """Add the file field name and url to the serializer if a file exists"""
@@ -221,7 +268,7 @@ class WebsiteContentDetailSerializer(
 
     class Meta:
         model = WebsiteContent
-        read_only_fields = ["text_id", "type"]
+        read_only_fields = ["text_id", "type", "content_context"]
         fields = read_only_fields + [
             "title",
             "markdown",
--- a/websites/views.py
+++ b/websites/views.py
@@ -330,6 +330,27 @@ def _get_derived_website_content_data(
     return added_data
 
 
+def _get_value_list_from_query_params(query_params, key):
+    """
+    Get a list of values which have keys that start with key[ or key
+    """
+    filter_type_keys = [
+        qs_key
+        for qs_key in query_params.keys()
+        # View should accept "type" as a single query param, or multiple types,
+        # e.g.: "?type[0]=sometype&type[1]=othertype"
+        if qs_key == key or qs_key.startswith(f"{key}[")
+    ]
+    return [query_params[_key] for _key in filter_type_keys]
+
+
+def _parse_bool(value):
+    """
+    Parse a query string value into a bool
+    """
+    return value and value.lower() != "false"
+
+
 class WebsiteContentViewSet(
     NestedViewSetMixin,
     viewsets.ModelViewSet,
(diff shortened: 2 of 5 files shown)


gumaerc:
@noisecapella I'm not sure if I'm doing something wrong here, but I'm getting an error.  I tested this using the `ocw-www` starter.  I created an Image resource and a Testimonial, then linked the image to that testimonial using a RelationField.  When visiting the API URL, I see results but the `content_context` lists are all empty and I see this error in the log:

```
web_1     | [2021-07-23 14:47:01] ERROR 12 [websites.serializers] serializers.py:250 - [a94a029bc151] - Malformed relation field value on WebsiteContent 56479542-70ec-4bea-abfa-395b594dd845 on Website ocw-www
web_1     |   File "./websites/serializers.py", line 242, in get_content_context
web_1     |     content = value["content"]
web_1     | TypeError: 'NoneType' object is not subscriptable
```

websites/views.py
```diff
@@ -342,22 +356,22 @@ class WebsiteContentViewSet(
...
+        text_ids = _get_value_list_from_query_params(
+            self.request.query_params, "text_id"
+        )
+
+        queryset = WebsiteContent.objects.filter(
+            website__name=parent_lookup_website
+        ).select_related("website", "website__starter")
+        if types:
+            queryset = queryset.filter(type__in=types)
+        if search:
```
gumaerc, line 370: Can we add tests for `search` and `text_ids` having a value?

websites/serializers.py
```diff
@@ -209,6 +217,50 @@ def update(self, instance, validated_data):
         update_website_backend(instance.website)
         return instance
 
+    def get_content_context(self, instance):
+        """
+        Create mapping of uuid to a display name for any values in the metadata
+        """
+        if self.context and self.context.get("no_content_context"):
+            # Prevent circular dependencies where two relation fields may refer to one another
+            return None
```
gumaerc, line 226: Can we add a test that creates this condition?

websites/serializers.py
```diff
@@ -209,6 +217,50 @@ def update(self, instance, validated_data):
...
+
+        text_ids = []
+        website_name = None
+        metadata = instance.metadata or {}
+        site_config = SiteConfig(instance.website.starter.config)
+        for field in site_config.iter_fields():
+            if field.field.get("widget") == "relation":
+                try:
```
gumaerc, line 234: Can we add a relation widget to the test fields?  This whole block currently isn't hit.

gumaerc (changes requested):
Looks like we missed some test coverage:

noisecapella:
FYI I just added a migration to put `dbindex=True` on the title field we are using to search

noisecapella:
> @noisecapella I'm not sure if I'm doing something wrong here, but I'm getting an error. I tested this using the `ocw-www` starter. I created an Image resource and a Testimonial, then linked the image to that testimonial using a RelationField. When visiting the API URL, I see results but the `content_context` lists are all empty and I see this error in the log:
> 
> ```
> web_1     | [2021-07-23 14:47:01] ERROR 12 [websites.serializers] serializers.py:250 - [a94a029bc151] - Malformed relation field value on WebsiteContent 56479542-70ec-4bea-abfa-395b594dd845 on Website ocw-www
> web_1     | Traceback (most recent call last):
> web_1     |   File "./websites/serializers.py", line 242, in get_content_context
> web_1     |     content = value["content"]
> web_1     | TypeError: 'NoneType' object is not subscriptable
> ```

Do you have anything in your `metadata` for that instance? The relation field value should be something like `{ "content": ["text-id-1", "text-id-2"], "website": "a-website-here" }`

noisecapella:
OTOH maybe we shouldn't be logging an exception there. If the relation field value is malformed, it's not going to show any value in the frontend so there's no need for `content_context`

gumaerc:
@noisecapella This is the `__dict__` from the `WebsiteContent` object for the testimonial I created:

```
In [4]: WebsiteContent.objects.all().get(text_id="4c399468-b4e4-4e38-a27c-802659653df2").__dict__
Out[4]: 
{'_state': <django.db.models.base.ModelState at 0x7fa877d54550>,
 'id': 3,
 'deleted': None,
 'created_on': datetime.datetime(2021, 7, 23, 14, 40, 15, 632295, tzinfo=<UTC>),
 'updated_on': datetime.datetime(2021, 7, 23, 14, 41, 12, 124279, tzinfo=<UTC>),
 'owner_id': 2,
 'updated_by_id': 2,
 'website_id': UUID('95a06edb-2fd5-4d1c-ae82-cfe2eb6c4c8d'),
 'text_id': '4c399468-b4e4-4e38-a27c-802659653df2',
 'title': 'TESTimonial',
 'type': 'testimonials',
 'parent_id': None,
 'markdown': 'Test body',
 'metadata': {'name': 'Test Imonial',
  'image': {'content': 'b63609aa-0b08-478f-908c-5b7a8c9fc772',
   'website': None},
  'location': 'Here',
  'leadquote': 'testing is fun',
  'occupation': 'Stuff'},
 'is_page_content': True,
 'filename': 'testimonial',
 'dirpath': 'content/testimonials',
 'file': ''}
 ```
 As you can see, the `Website` property of the metadata is `None` for some reason.  Could that be the issue?  How could that happen?

**Comment by noisecapella** [comment]
Hmm, I'm not sure why it would be `None`, and `website` being `None` also shouldn't cause a `TypeError` in any case. Can you also paste your starter config here for testimonials?

**Comment by noisecapella** [comment]
Anyway, for testing sake it should be enough to set the `website` manually to `ocw-www`. Then, assuming the field configuration matches properly you should be able to see the context in the API results

**Comment by gumaerc** [comment]
@noisecapella I'm using the `ocw-www` starter from `ocw-hugo-projects`: https://github.com/mitodl/ocw-hugo-projects/blob/main/ocw-www/ocw-studio.yaml

I tried starting from scratch and creating a new `Website` then uploading an image into the Resource collection.  Then I created a testimonial and associated the image with it, saving the record.  This is what the API view shows me:

```
HTTP 200 OK
Allow: GET, POST, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "count": 2,
    "next": null,
    "previous": null,
    "results": [
        {
            "text_id": "2d227026-47d8-4182-a45d-a77bd26017a1",
            "type": "testimonials",
            "content_context": [],
            "title": "Test Testimonial",
            "markdown": "Test content",
            "metadata": {
                "name": "TESTimonial",
                "image": {
                    "content": "16aae235-ab89-4b45-8b5e-ab678dcedbc8",
                    "website": null
                },
                "location": "Here",
                "leadquote": "Test quote",
                "occupation": "Stuff"
            },
            "file": null
        },
        {
            "text_id": "16aae235-ab89-4b45-8b5e-ab678dcedbc8",
            "type": "resource",
            "content_context": [],
            "title": "Test Image",
            "markdown": null,
            "metadata": {
                "filetype": "Image",
                "metadata": {
                    "credit": "",
                    "caption": "",
                    "image-alt": ""
                },
                "description": ""
            },
            "file": "/media/ocw-www-2/16aae235ab894b458b5eab678dcedbc8_2021-01-05.jpg"
        }
    ]
}
```

**Comment by gumaerc** [comment]
After manually setting the `WebsiteContent` object for the testimonial's metadata.image.website property to `ocw-www-2` (the new `Website` I created), I was able to see the `content_context` property:

```
HTTP 200 OK
Allow: GET, POST, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "count": 2,
    "next": null,
    "previous": null,
    "results": [
        {
            "text_id": "2d227026-47d8-4182-a45d-a77bd26017a1",
            "type": "testimonials",
            "content_context": [
                {
                    "text_id": "16aae235-ab89-4b45-8b5e-ab678dcedbc8",
                    "type": "resource",
                    "content_context": null,
                    "title": "Test Image",
                    "markdown": null,
                    "metadata": {
                        "filetype": "Image",
                        "metadata": {
                            "credit": "",
                            "caption": "",
                            "image-alt": ""
                        },
                        "description": ""
                    },
                    "file": "/media/ocw-www-2/16aae235ab894b458b5eab678dcedbc8_2021-01-05.jpg"
                }
            ],
            "title": "Test Testimonial",
            "markdown": "Test content",
            "metadata": {
                "name": "TESTimonial",
                "image": {
                    "content": "16aae235-ab89-4b45-8b5e-ab678dcedbc8",
                    "website": "ocw-www-2"
                },
                "location": "Here",
                "leadquote": "Test quote",
                "occupation": "Stuff"
            },
            "file": null
        },
        {
            "text_id": "16aae235-ab89-4b45-8b5e-ab678dcedbc8",
            "type": "resource",
            "content_context": [],
            "title": "Test Image",
            "markdown": null,
            "metadata": {
                "filetype": "Image",
                "metadata": {
                    "credit": "",
                    "caption": "",
                    "image-alt": ""
                },
                "description": ""
            },
            "file": "/media/ocw-www-2/16aae235ab894b458b5eab678dcedbc8_2021-01-05.jpg"
        }
    ]
}
```

So I guess the question is why `website` would be None there in the first place.  If you think it's not related to this PR I can file an issue separately and investigate.

noisecapella:
I doubt it's related to the PR since at the moment the relation field value is entirely constructed in the frontend. It's possible #409 may fix it, but it would be good to file an issue to keep track of it and verify whether #409 actually fixes the issue

websites/views.py
noisecapella: I already added tests for these cases: https://github.com/mitodl/ocw-studio/blob/gs/async-search-backend/websites/views_test.py#L473

websites/serializers.py
noisecapella: It looks like this is covered by tests. If I add a `raise Exception()` at this line the `test_website_content_detail_serializer_content_context` test fails in certain conditions

websites/serializers.py
noisecapella: We are testing this already: https://github.com/mitodl/ocw-studio/blob/gs/async-search-backend/websites/serializers_test.py#L170

websites/serializers.py
gumaerc: It seems like for this and the above coverage issues I mentioned, the Codecov data was out of date and it appeared that these cases weren't covered.  Either way, it looks good now.

noisecapella:
@gumaerc @gsidebo FYI I'm going to remove the migration website 0031 I added here since it will soon conflict with #405. Since it's trivial I'll just file a ticket to add it back later

gumaerc (approved):
:+1: This looks good to me now, although I'm interested in what @gsidebo thinks.

gsidebo (changes requested):
I guess my main concern with this is that for every single request to fetch website content detail, we're adding this extra serialized field, an extra query across 2 foreign keys (`instance.website.starter.config`), and all this extra logic to understand. Fetching this content context is only relevant for config items with 1 or more relation fields, though. It seems like a code smell to me that we doing this extra work and adding this extra field in the serialized results even when it's not relevant. What do you think of the component on the frontend component (`SiteContentEditor`) being responsible for appending a querystring param to include the content context when it makes the content detail request? It would run some logic to check its own fields for a `"relation"` widget, and if it finds one, the qs param is added to the request. On the backend, the serializer does not add `content_context` if that qs param is absent. DRF doesn't love the idea of a serializer with conditional fields, but we've done it before by overriding `to_representation`. To me it makes more sense that we think of the frontend component as the client of our API, and it asks for that extra information according to some condition.

gumaerc:
@gsidebo Those are good points.  I think that would be a sensible addition.  If I interpreted this PR correctly, `content_context` is there to provide a fuzzy find type search for relation widgets with the values already associated with that field right?  Would an alternative approach be to just return the `text_id`'s and then leave it up to the client to query the API for the rest of the data?  Would that save on extra queries?

noisecapella:
@gsidebo I can add a query parameter to enable `content_context` though I think it should be on `RelationField` since that's where the data fetch originates

noisecapella:
@gsidebo there should be a query parameter `content_context=true` which you can add to enable the data. If `content_context=false` or the query parameter is absent it will pass a null value instead for `content_context`

noisecapella:
@gsidebo One other thought, should we also add data to `content_context` for menu fields? At the moment we don't, and in the frontend PR we instead pass the `text_id` to the content listing API to fetch that information to display the label for the internal link uuid.

gsidebo (approved):
I'd prefer that we omit the field if the qs param isn't included, but that's just a matter of preference. This looks great! 👍

gsidebo:
@noisecapella Great idea re: the content context and nav menus

Issue #382, opened 2021-07-13
Import instructor list from parsed JSON


noisecapella:
As a maintainer of the instructor list, I'd like to populate the instructor list in ocw-studio so that I don't need to add each instructor manually.

#### Acceptance Criteria:

- [ ] Modify existing OCW import sync code to create or update `WebsiteContent` objects for each instructor found in a course
- [ ] Only add instructors that don't already exist. I think we can assume that two instructors with the same first, middle initial, last name are the same since the instructor list will only be used to show instructor names.
- [ ] If there is no course metadata in the site config, or if there is no `instructor` collection for `ocw-www`, we should log an error and move on
- [ ] `display_name` should be populated from concatening the other fields

Edit: I made a couple changes:
 - if there is no course metadata, we will create it
 - `title` is populated by concatenating the other fields

noisecapella:
Fixed by #409

PR #409, opened 2021-07-23
Create and sync sitemetadata with instructors

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Part of #382 

#### What's this PR do?
Creates and populates a `WebsiteContent` for sitemetadata for each imported course. At the moment only instructors are read into the site metadata. The metadata for each course should have a list of text_ids which match the instructors for that course.

I updated the courses in the test data in order to have up to date versions of `instructors`. I didn't make any changes other than running the courses through ocw-data-parser and ocw-to-hugo.

 - Run `./manage.py override_site_config` or otherwise update your site config with the instructors field which was added in this PR in `ocw-course-site-config.yml`
 - Make sure you have a site called `ocw-www`. If not, create one
 - Run `./manage.py import_ocw_course_sites -b ocw-to-hugo-output-qa --limit 5`. This should take about a minute, or you can remove `--limit 5` to populate based on all the courses, which will take a half hour or so.
 - In your browser, go to one of the sites you imported, for example `1-011-project-evaluation-spring-2011`. Click on "Metadata" in the sidebar. You should see a relationfield with two instructors (see screenshot). It may take 30 seconds or so to load the instructors so you may see a uuid for a number of seconds. This should be addressed in #402 
 - You should be able to add and remove instructors, and save your changes.

#### Screenshots
Screenshot from 2021-07-23 13-46-08


mbertrand:
Needs a rebase

mbertrand (approved):
Backend code works great.  On the front-end, I noticed that if I go from one site's metadata to another, the instructors shown do not update unless I refresh the page.  But there isn't any front end code in this PR so it might be worth opening up an issue unless the RelationField PR fixes it.

noisecapella:
@mbertrand Thanks for catching that. It looks like the metadata problem is still there after a rebase so I'll file an issue for it.

PR #414, opened 2021-07-26
Fix null website error with RelationField

### LINKED ISSUE #412: RelationField saves with website=None when creating a new site metadata
### Steps to Reproduce

- Go to `/new-site/` and create a new site using the [ocw-course starter](https://github.com/mitodl/ocw-hugo-projects/blob/main/ocw-course/ocw-studio.yaml)
- Without refreshing your browser, go to site metadata. Open the network tab in your browser and click save. In the network tab you should see a POST to `/api/websites/.../content/` with the metadata having something like `{"website": null, "content": [ ... ]}`

### Expected Behavior

The `website` should be `ocw-www`

### Actual Behavior

The `website` is saved as null

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #412 

#### What's this PR do?
 - Sets website to the correct value (which is specified in the config) when creating a new form or adding a field without data on the backend
 - Sets website to the correct value when the relation field is updated

 - Create a new website with the ocw-course starter from [here](https://github.com/mitodl/ocw-hugo-projects/blob/main/ocw-course/ocw-studio.yaml)
 - Go to metadata. You should see a relation field for instructors. (You do *not* need to populate instructors, an empty list is fine here)
- Open your network tab. Add a department number to bypass validation, then click "Save"
- In your network tab you should see a POST to create the metadata:

        {
          "type": "sitemetadata",
          "metadata": {
            "department_numbers": [
              "6"
            ],
            "instructors": {
              "website": "course-7",
              "content": []
            }
          },
          "text_id": "sitemetadata"
        }

The `website` value should be the same as the name of your site. If you run through these steps on master the value will be `null` instead

Internal menu links should also still work with no change of behavior. To verify this:
 - Create a new site using the omnibus starter
 - Create a couple resources
 - Go to "Menu" on the side. Click "Add new"
 - In the side drawer create an internal link. You should see both resources in the dropdown. You should be able to update them, refresh the page and see the changes you made persist.

### DIFF
--- a/static/js/components/forms/SiteContentForm.tsx
+++ b/static/js/components/forms/SiteContentForm.tsx
@@ -27,6 +27,7 @@ import {
 } from "../../types/websites"
 import { ContentFormType, SiteFormValues } from "../../types/forms"
 import { getContentSchema } from "./validation"
+import { useWebsite } from "../../context/Website"
 
 interface Props {
   onSubmit: (
@@ -46,12 +47,13 @@ export default function SiteContentForm({
   content,
   formType
 }: Props): JSX.Element {
+  const website = useWebsite()
   const initialValues: SiteFormValues = useMemo(
     () =>
       formType === ContentFormType.Add ?
-        newInitialValues(fields) :
-        contentInitialValues(content as WebsiteContent, fields),
-    [fields, formType, content]
+        newInitialValues(fields, website) :
+        contentInitialValues(content as WebsiteContent, fields, website),
+    [fields, formType, content, website]
   )
   const contentContext = content?.content_context ?? null
 
--- a/static/js/components/forms/SiteContentForm.tsx
+++ b/static/js/components/forms/SiteContentForm.tsx
@@ -81,7 +81,7 @@ export default function SiteContentForm({
       initialValues={initialValues}
       enableReinitialize={true}
     >
-      {({ isSubmitting, status, setFieldValue, values }) => (
+      {({ isSubmitting, status, values }) => (
         <Form className="row">
           {fieldsByColumn.map((columnFields, idx) => (
             <div className={columnClass} key={idx}>
@@ -92,14 +92,12 @@ export default function SiteContentForm({
                     <ObjectField
                       field={field}
                       key={field.name}
-                      setFieldValue={setFieldValue}
                       contentContext={contentContext}
                     />
                   ) : (
                     <SiteContentField
                       field={field}
                       key={field.name}
-                      setFieldValue={setFieldValue}
                       contentContext={contentContext}
                     />
                   )
--- a/static/js/components/widgets/RelationField.tsx
+++ b/static/js/components/widgets/RelationField.tsx
@@ -26,6 +26,7 @@ type BaseProps = {
   website?: string
   valuesToOmit?: Set<string>
   contentContext: WebsiteContent[] | null
+  onChange: (event: any) => void
 }
 
 /* NOTE: Either setFieldValue or onChange should be passed in, not both.
@@ -63,8 +64,7 @@ export default function RelationField(
     value,
     filter,
     valuesToOmit,
-    onChange,
-    setFieldValue
+    onChange
   } = props
 
   const formatOptions = useCallback(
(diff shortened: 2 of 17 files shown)


aliceriot:
confirmed the issue on main and then confirmed that this addresses it, reading code now

static/js/components/forms/SiteContentForm.tsx
```diff
@@ -37,21 +38,23 @@ interface Props {
...
 }
 
 export default function SiteContentForm({
   onSubmit,
   fields,
   configItem,
   content,
-  formType
+  formType,
+  website
```
aliceriot, line 50: if you want you could use the `useWebsite` hook here instead of passing in a prop

static/js/components/widgets/RelationField.tsx
```diff
@@ -155,20 +154,26 @@ export default function RelationField(
...
+      const updatedEvent = {
+        target: {
+          name:  event.target.name.replace(/\.content$/, ""),
+          value: {
+            website: website.name,
+            content: event.target.value
+          }
+        }
       }
+      // @ts-ignore
```
aliceriot, line 173: any way we can avoid the `@ts-ignore` here? I think generally it is a good goal for us to only use them in test files. Maybe you can add a cast on the type of `updatedEvent` or adjust the type signature for the `onChange` prop?

static/js/components/widgets/RelationField.tsx
```diff
@@ -155,20 +154,26 @@ export default function RelationField(
...
-          website: website.name,
-          content
-        })
+      // When we run renameNestedFields we add a '.content' suffix to the name of the field for RelationField
+      // because the data structure looks like this: { content, website }
+      // where 'content' is either a string or a list of strings.
+      // Validation by yup is easier to work with when the field name ends with .content because
+      // formik validates the value of 'content' instead of '{ content, website }'.
+      // But while that's easier for validation, we still need to send the whole object.
+      // So we need to remove that .content suffix and also add back 'website' when onChange is called.
```
aliceriot, line 163: thanks for updating this comment :)

aliceriot (changes requested):
a few comments / questions

static/js/components/forms/SiteContentForm.tsx
noisecapella: It works fine but I think we would need to rewrite many of the tests in `SiteContentForm.test.tsx` because it would require `helper.configureRenderer`. I think some things like whether `isSubmitting` is disabling the button become impossible to test since we would not have access to the inner function anymore

static/js/components/widgets/RelationField.tsx
noisecapella: Yes, it was not in the prop types actually so it should be easy to fix

static/js/components/forms/SiteContentForm.tsx
aliceriot: hmm what if we did a `jest.mock` on `useWebsite` and just made it return a website we made with out website factory?

aliceriot: possibly that would let us still use the shallow renderer

noisecapella: That's an interesting idea, I'll try that

noisecapella: It worked out, I updated the tests for it

aliceriot (changes requested):
ack wrote this comment a bit ago but didn't submit the review!

static/js/components/widgets/RelationField.tsx
```diff
@@ -155,20 +155,25 @@ export default function RelationField(
...
+          value: {
+            website: website.name,
+            content: event.target.value
+          }
+        }
       }
+      onChange(updatedEvent)
     },
-    [setFieldValue, onChange, name, website]
+    [onChange, website]
```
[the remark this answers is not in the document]
aliceriot: can we / should we in this PR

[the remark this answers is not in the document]
noisecapella: Looks like I need to adjust `FileUploadField` and `BooleanField`. I'll try it out and see if it worked out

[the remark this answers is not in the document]
noisecapella: Ok I think that's everything. I checked a site using the omnibus starter and the boolean field still works there, as well as the file fields in the resources

aliceriot (approved):
LGTM! 👍

PR #421, opened 2021-07-27
Add index to WebsiteContent.title and Website.title

### LINKED ISSUE #411: Add db_index=True on WebsiteContent.title
This would be part of #407 but I'm delaying to avoid extra migration conflicts as part of the review

noisecapella:
#### Pre-Flight checklist

- [x] Migrations

Fixes #411 

#### What's this PR do?
Adds an index to `WebsiteContent.title` and `Website.title`. In #407 we are adding a REST API to do searches on this field, so it makes sense to index the column to speed up text queries on it. This is broken out of #407 to avoid blocking other migration-related changes in other PRs. Additionally we started doing searches on `Website.title` in #391 so it makes sense to add an index there too.

I don't think there's a way to test this before #407 merges, but it's a simple enough change that I don't think it's necessary


aliceriot (approved):
LGTM! 👍

PR #428, opened 2021-07-28
Remove text_id parameter, instead use content_context for MenuField

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Related to #402 

#### What's this PR do?
There are a couple optimizations I broke apart from the other work in #402:
 - Previously `content_context` was passed to `RelationField` for most cases but it was `null` for internal links in `MenuField`. We added code in the backend to produce `content_context` for menu fields in #407. This includes that `content_context` value so that the `RelationField` for internal links in `MenuField` can make use of it
 - This means we don't need to make an extra request for content for a given `text_id`. I removed the code related to the `text_id` parameter since it's now unused

Nothing should appear different for the end user. See the manual testing instructions in this PR for testing internal links and other uses of `RelationField`: https://github.com/mitodl/ocw-studio/pull/402

### DIFF
--- a/static/js/components/forms/MenuItemForm.test.tsx
+++ b/static/js/components/forms/MenuItemForm.test.tsx
@@ -4,18 +4,25 @@ import { shallow } from "enzyme"
 import MenuItemForm from "./MenuItemForm"
 import { defaultFormikChildProps } from "../../test_util"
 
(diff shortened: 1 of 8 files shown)


mbertrand (approved):
👍

PR #432, opened 2021-07-28
Use debounced fetch for website collection course search

### LINKED ISSUE #394: Delay fetches in incremental search for course select in collections UI
Related to #391

Currently the course selection UI in the edit collections dialog makes an API request on each keypress. We should figure out how to delay fetches so that they go out every 300 milliseconds or so, for example. https://stackoverflow.com/questions/42217121/how-to-start-search-only-when-user-stops-typing#answer-61629055

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #394 

#### What's this PR do?
Make fetch debounced so that only the latest in a certain period of time is used. A request should only be made every 300 ms or so instead of on every keystroke

Go to `/collections` and edit a collection. Open the network tab. Type some website names in the course search widget. In the network tab you should see requests made only every 300 ms or so. You should also have correct search results when you stop typing.

### DIFF
--- a/static/js/components/forms/WebsiteCollectionItemForm.tsx
+++ b/static/js/components/forms/WebsiteCollectionItemForm.tsx
@@ -7,6 +7,7 @@ import { WCItemCreateFormFields } from "../../types/forms"
 import { Website } from "../../types/websites"
 import { useMutation } from "redux-query-react"
 import { siteApiListingUrl } from "../../lib/urls"
+import { debouncedFetch } from "../../lib/api/util"
 import { WebsiteListingResponse } from "../../query-configs/websites"
 import { WebsiteCollection } from "../../types/website_collections"
 import { createWebsiteCollectionItemMutation } from "../../query-configs/website_collections"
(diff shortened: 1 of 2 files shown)


static/js/components/forms/WebsiteCollectionItemForm.tsx
```diff
@@ -32,7 +33,13 @@ export default function WebsiteCollectionItemForm(props: Props): JSX.Element {
...
       // use-case doesn't exactly jibe with redux-query: we need to issue
       // a request programmatically on user input.
-      const response = await fetch(url)
+      const response = await debouncedFetch("website-collection", 300, url, {
+        credentials: "include"
+      })
+      if (!response) {
+        // this happens if this fetch was ignored in favor of a later fetch
+        return
+      }
```
aliceriot, line 42: would it make sense for us to use an [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) to cancel requests that are no longer relevant?

aliceriot: in addition to debouncing. although it looks like we implemented our own debounce instead of using `_.debounce` or something, so not sure that will play nicely with that

aliceriot (changes requested):
one question!

static/js/components/forms/WebsiteCollectionItemForm.tsx
noisecapella: I don't think `_.debounce` handles promises and I wanted something that could be awaited on. We can modify the debounce code to allow for aborts, though I don't think we have a UI to let the user request an abort

aliceriot: gotcha, that makes sense why we'd need our own in that case

aliceriot (approved):
👍

PR #433, opened 2021-07-28
Handle valuesToOmit when value is a list

### LINKED ISSUE #401: RelationField valuesToOmit code should handle value with an array
The `value === entry.text_id` in this line should be updated to handle `RelationField` when value is a list, which would happen when the `multiple` prop is true 
https://github.com/mitodl/ocw-studio/blob/42f56d3faea049794329164edb017c1fa96c4d80/static/js/components/widgets/RelationField.tsx#L136

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #401 

#### What's this PR do?
Fixes bug where `valuesToOmit` didn't handle `value` when it is a list, which is the case when `multiple` is true.

I don't think there's a way to demonstrate this bug since the only code using `valuesToOmit` is the internal link code and that only uses `value` as a string. I would just make sure the internal link UI still works fine.

### DIFF
--- a/static/js/components/widgets/RelationField.test.tsx
+++ b/static/js/components/widgets/RelationField.test.tsx
@@ -264,8 +264,6 @@ describe("RelationField", () => {
     // @ts-ignore
     wrapper.update()
     // @ts-ignore
-    global.fetch.mockClear()
-    // @ts-ignore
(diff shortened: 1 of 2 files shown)


mbertrand (approved):
👍

PR #435, opened 2021-07-29
Use both website name and content textId for lookup

### LINKED ISSUE #429: Metadata shows up incorrectly when navigating from one site metadata to another
Found in a code review for a separate PR: https://github.com/mitodl/ocw-studio/pull/409#pullrequestreview-717112729
### Steps to Reproduce

 - Find or create two courses based on the ocw-course starter. Ideally they should be right next to each other in the site listing
 - Add some content to the metadata for each of the two sites. It doesn't matter what fields you update, it just needs to be distinct for each site. 
 - Go to `/sites` and find the link for the first of the two sites. Click on that link, then click on "Metadata". The metadata fields should look correct.
 - Click the back button repeatedly until you get back to `/sites`. Click on the link for the second site then click on "Metadata". You should see metadata fields from the first site, instead of the correct fields from the second site.

### Expected Behavior

Each metadata page should have correct information

### Actual Behavior

The second metadata page you visit (without refreshing the browser) has data from the first metadata page

noisecapella:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Fixes #429 

#### What's this PR do?
We were using `textId` as the only lookup value for our frontend content detail data. However we also need to include the website name to make it a unique key. This updates the key to 

 - Find or create two courses based on the ocw-course starter. Ideally they should be right next to each other in the site listing
 - Add some content to the metadata for each of the two sites. It doesn't matter what fields you update, it just needs to be distinct for each site. 
 - Go to `/sites` and find the link for the first of the two sites. Click on that link, then click on "Metadata". The metadata fields should look correct.
 - Click the back button repeatedly until you get back to `/sites`. Click on the link for the second site then click on "Metadata". You should see the correct data for the website, which is different from the metadata for the first site.
 - Additionally, go to a page listing page and make sure things still work in general

### DIFF
--- a/static/js/components/RepeatableContentListing.test.tsx
+++ b/static/js/components/RepeatableContentListing.test.tsx
@@ -13,6 +13,7 @@ import {
   siteApiContentListingUrl
 } from "../lib/urls"
 import {
+  contentDetailKey,
   contentListingKey,
   WebsiteContentListingResponse
 } from "../query-configs/websites"
@@ -67,7 +68,9 @@ describe("RepeatableContentListing", () => {
     ]
     websiteContentDetailsLookup = {}
     for (const item of contentListingItems) {
-      websiteContentDetailsLookup[item.text_id] = item
+      websiteContentDetailsLookup[
+        contentDetailKey({ name: website.name, textId: item.text_id })
+      ] = item
     }
 
     const listingParams = {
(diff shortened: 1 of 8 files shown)


mbertrand (approved):
👍

PR #378, opened 2021-07-13
Add website collection editing functionality

aliceriot:
#### Pre-Flight checklist

- [x] Screenshots and design review for any changes that affect layout or styling
  - [x] Desktop screenshots
  - [x] Mobile width screenshots
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?

closes #357 

#### What's this PR do?

this adds functionality for editing basic attributes of WebsiteCollection records (i.e. their title, description) and for therefore creating new ones.

This does not yet include editing entries in a collection, that will be covered for #376

#### How should this be manually tested?

Visit the collections list page (`/collections`). If you click 'Add' you should get a drawer where you can create a new collection. Then in the list page if you click on the title for a collection you should open an editing drawer where you can change it's title and whatnot.

#### Screenshots (if appropriate)

Screen Shot 2021-07-13 at 10 48 06 AM
Screen Shot 2021-07-13 at 10 48 13 AM
Screen Shot 2021-07-13 at 10 48 21 AM

PR #437, opened 2021-07-29
switch site content over to user ModalState

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

none, just a sort of tech debt / doing things the same way in different places PR

#### What's this PR do?

in #378 I added a module called `modal_state` which exports a way to manage the state of a modal. Basically, there are three options, adding, editing, and closed, with some helper functions to create these states. The type `ModalState` is like a Data type in Haskell or an Enum in Rust (think `Option` or `Result` if you've written some Rust before.

Anyhow - this PR is just some code to manage the modal for editing site content in the same way that we manage the modal for editing website collections.

Exercise the whole site content UI and make sure everything works as it should! Metadata editing and whatnot too.

### DIFF
--- a/static/js/components/RepeatableContentListing.tsx
+++ b/static/js/components/RepeatableContentListing.tsx
@@ -25,9 +25,10 @@ import { getWebsiteContentListingCursor } from "../selectors/websites"
 import {
   ContentListingParams,
   RepeatableConfigItem,
-  WebsiteContentListItem
+  WebsiteContentListItem,
+  WebsiteContentModalState
 } from "../types/websites"
-import { ContentFormType } from "../types/forms"
+import { createModalState } from "../types/modal_state"
 
 export default function RepeatableContentListing(props: {
   configItem: RepeatableConfigItem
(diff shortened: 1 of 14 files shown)


static/js/components/RepeatableContentListing.tsx
```diff
@@ -71,43 +72,42 @@ export default function RepeatableContentListing(props: {
     event: ReactMouseEvent<HTMLLIElement | HTMLAnchorElement, MouseEvent>
   ) => {
     event.preventDefault()
-    setPanelState({
-      textId,
-      formType: textId ? ContentFormType.Edit : ContentFormType.Add
-    })
+
+    setDrawerState(
+      textId ? createModalState("editing", textId) : createModalState("adding")
```
noisecapella, line 77: Can we make enums of `adding`, `editing`, and `closed`?

noisecapella (changes requested):
Looks good, just the one comment

static/js/components/RepeatableContentListing.tsx
aliceriot: I considered doing so but decided not to, this was my reasoning:

1. It makes calls to the `createModalState` function longer because we'd have to reference the enum member (i.e. `createModalState("closed")` vs `createModalState(ModalStateVariant.Closed)`)
2. it adds another import, so that instead of just importing `createModalState` we'd need to import the enum as well
3. it doesn't provide any additional type safety beyond what we get from the function overloading. If you try doing something like `createModalState("dancing")` you'll get a type error saying `"No overload matches this call"`

so that made me think it wasn't actually _so_ helpful to create an enum for these values as opposed to just using these little string constants. what do you think?

aliceriot:
@noisecapella btw just rebased and fixed some merge conflicts with your changes in https://github.com/mitodl/ocw-studio/pull/435, I think it all made sense but would probably be worthwhile to take another look 🙂

noisecapella (approved):
Things look good after the rebase :+1:

PR #446, opened 2021-07-30
Reduce default chunk size for import_ocw_course_sites

### LINKED ISSUE #442: Memory use prevents import_ocw_course_sites from completing on RC
We need to do something to reduce memory use under 512mb per worker. I think reducing the default chunk size to something like 10 would be a good quick way to do this

### Steps to Reproduce

Run `./manage.py import_ocw_course_sites -q ocw-to-hugo-output-qa` on RC. There are many memory warnings and the task times out beyond the expected 30 minute running time.

noisecapella:
#### Pre-Flight checklist

- [x] Testing

Fixes #442 

#### What's this PR do?
Reduces default chunk size for `import_ocw_course_sites` from 100 to 10. On RC this completed in 19 minutes so I think that chunk size should be adequate while reducing memory use

Run `./manage.py import_ocw_course_sites` locally, though I'm not sure there's a need to manually test this since it's a trivial change

### DIFF
--- a/ocw_import/management/commands/import_ocw_course_sites.py
+++ b/ocw_import/management/commands/import_ocw_course_sites.py
@@ -34,9 +34,9 @@ class Command(BaseCommand):
             "-c",
             "--chunks",
             dest="chunks",
-            default=100,
+            default=10,
             type=int,
-            help="Number of courses to process per celery task (default 100)",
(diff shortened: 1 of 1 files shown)


mbertrand (approved):
👍

Issue #430, opened 2021-07-28
import_ocw2hugo_course does not connect parent for certain pages


noisecapella:
After https://github.com/mitodl/ocw-to-hugo/pull/311 was merged, output from ocw-to-hugo no longer includes `menu` in the section frontmatter. This is causing some `WebsiteContent.parent` values to no longer have a value.

It looks like [this code](https://github.com/mitodl/ocw-studio/blob/13ea13a6f1d6a9e92043014200caa692ae54cae1/ocw_import/api.py#L123) in `convert_data_to_content` uses the course uid if no parent exists on the menu, but if there is no menu then it uses `None`. I'm not sure if that behavior makes sense. It seems like the best fix for this is to update ocw-to-hugo to add a `parent` uid so that it can be used by `import_ocw2hugo_course`.

In any case we should remove the separate behavior for `menu` in `convert_data_to_content`, and also update our data in `test_hugo2ocw` to use the latest output of `ocw-to-hugo`

#### Steps to reproduce behavior
I found it when replacing the two courses in `test_hugo2ocw` with the latest output from `ocw-to-hugo` and running `ocw_import/api_test::test_import_ocw2hugo_course`

Issue #410, opened 2021-07-26
OCW import should set is_page_content=True


noisecapella:
### Steps to Reproduce

 - Run an import via `./manage.py import_ocw_course_sites`.
 - In a Django shell run `WebsiteContent.objects.filter(is_page_content=True)`. None of the imported content has this value set

### Expected Behavior

Navigable pages should have `is_page_content` set to true (I think navigable pages means pages and resources). 

### Actual Behavior

No imported content has `is_page_content` set to true. This is needed since it is used to filter valid content for `RelationField` in the frontend.

PR #448, opened 2021-07-30
import menus.yaml files as navmenus and update test data

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/430
Closes https://github.com/mitodl/ocw-studio/issues/410

#### What's this PR do?
A number of changes have been made to the way `ocw-to-hugo` outputs data and the `ocw_import` code here in `ocw-studio` needs to be updated to reflect those changes.  Firstly, the left navigation menu `leftnav` now has its entries stored in a menu specific configuration file stored in each course under `config/_default/menus.yaml` instead of the front matter of each page that had a nav entry.  The primary purpose of this PR is to correct that by removing the code that pulls nav entries from front matter and instead sources them from the `menus.yaml` file.  While working on this, a few other issues were uncovered that were addressed here.  The parent / child relationship of `WebsiteContent` objects was being determined from these front matter based menu entries, and that has been moved over to simply use a `parent_uid` property introduced in https://github.com/mitodl/ocw-to-hugo/pull/320.  Also, `WebsiteContent` objects created by the import are not currently visible in the UI because pages did not have `is_page_content` set to `True`.  This has been corrected so imported pages properly show up in the UI.  Aside from that, the last change was updating the `ocw-to-hugo` test data to the latest version and updating the tests.

Proper manual testing of this PR requires a release of https://github.com/mitodl/ocw-to-hugo/pull/320.  Once this is complete:
 - Set up `ocw-studio` to run locally and make sure S3 access is properly configured
 - Run `docker-compose run web python manage.py import_ocw_course_sites --bucket ocw-to-hugo-output-production --filter 15-401`
 - Browse to the imported site locally at http://localhost:8043/sites/15-401-finance-theory-i-fall-2008
 - Ensure that pages were imported in the Page section and they are visible in the UI
 - Ensure that the menu was imported and items under the Video Lectures and Slides section are indented to indicate that they are children of that section

 One test in particular was problematic, `test_import_ocw2hugo_course_bad_date`.  After updating the test data, there are more `WebsiteContent` objects that get imported and it revealed a problem with the way this test was implemented.  The test was running `mocker.patch` in a way that caused unintentional side effects.  In `api.py`, we were assigning `dateutil.parser` to a variable like:

`from dateutil import parser as dateparser`

...and in the test we were patching it like:

```
    mocker.patch(
        "ocw_import.api.dateparser.parse", side_effect=["2021-01-01", ValueError()]
    )
```

When there were less `WebsiteContent` objects being imported this was fine, but when the test data was updated and more items were added the side effect became apparent.  Even in the old data, the parse function should have only been called once and the array of side effects shouldn't have been necessary.  Calling `mocker.patch` and passing in `ocw_import.api.dateparser.parse` here does not limit the patch to just the locally assigned `dateparser` variable in `api.py` as one might think.  Instead *any* call to `dateutil.parser.parse` is patched.  This means that updates to external dependencies outside of our code even could potentially add extra calls to that that would get patched, resulting in a `StopIteration` error.  Instead of patching the function in this way, I made a small wrapper function in `api.py` for parsing dates and patched that function instead.  Now the patch is only called once (as expected) and the test behaves as it should.


ocw_import/api.py
```diff
@@ -40,6 +40,19 @@
...
 
 
+def parse_date(date_string):
+    """
+    Small utility function for calling dateutil.parser.parse, done so we can
+    mock out date parsing in unit tests
+
+    Args:
+        date_string (str): A date string to parse
+    Yeilds:
```
noisecapella, line 50: It spelled `Yields` but here it should be:

    Returns:
        datetime: A date object

Or alternatively, is there a type annotation we can use for datetime?

noisecapella (changes requested):
Just a tiny docstring change, everything else looks great. Going to test functionality now

noisecapella:
I'm not seeing anything in the menu page for `15-401-finance-theory-i-fall-2008`. I'm seeing this error in my logs:

       File "/src/content_sync/decorators.py", line 97, in wrapper
         return_value = func(*args, **kwargs)
       File "/src/content_sync/tasks.py", line 124, in sync_website_content
         backend.sync_all_content_to_backend()
       File "/src/content_sync/backends/github.py", line 79, in sync_all_content_to_backend
         return self.api.upsert_content_files()
       File "/src/content_sync/apis/github.py", line 246, in upsert_content_files
         self.upsert_content_files_for_user(user_id)
       File "/src/content_sync/decorators.py", line 26, in wrapper
         return func(*args, **kwargs)
       File "/src/content_sync/apis/github.py", line 275, in upsert_content_files_for_user
         data = serialize_content_to_file(
       File "/src/content_sync/serializers.py", line 331, in serialize_content_to_file
         serializer = ContentFileSerializerFactory.for_content(site_config, website_content)
       File "/src/content_sync/serializers.py", line 318, in for_content
         raise ValueError(
     ValueError: Website content cannot be serialized to a file (navmenu).
     
     During handling of the above exception, another exception occurred:
     
       File "/usr/local/lib/python3.8/site-packages/celery/app/trace.py", line 412, in trace_task
         R = retval = fun(*args, **kwargs)
       File "/usr/local/lib/python3.8/site-packages/celery/app/trace.py", line 704, in __protected_call__
         return self.run(*args, **kwargs)
       File "/usr/local/lib/python3.8/site-packages/sentry_sdk/integrations/celery.py", line 171, in _inner
         reraise(*exc_info)
       File "/usr/local/lib/python3.8/site-packages/sentry_sdk/_compat.py", line 57, in reraise
         raise value
       File "/usr/local/lib/python3.8/site-packages/sentry_sdk/integrations/celery.py", line 166, in _inner
         return f(*args, **kwargs)
       File "/usr/local/lib/python3.8/site-packages/celery/app/autoretry.py", line 35, in run
         return task._orig_run(*args, **kwargs)
       File "/src/content_sync/decorators.py", line 102, in wrapper
         lock.release()
       File "/usr/local/lib/python3.8/site-packages/redis/lock.py", line 232, in release
         self.do_release(expected_token)
       File "/usr/local/lib/python3.8/site-packages/redis/lock.py", line 238, in do_release
         raise LockNotOwnedError("Cannot release a lock"
     redis.exceptions.LockNotOwnedError: Cannot release a lock that's no longer owned

noisecapella:
Looks like that bug is addressed in a merged PR, can you rebase on master?

noisecapella:
Can we update `ocw-course-site-config.yml` to match these changes? https://github.com/mitodl/ocw-hugo-projects/pull/44

gumaerc:
@noisecapella I rebased and updated the docstring on the `parse_date` function. As far as the `ocw-course-site-config.yml` file, I wrote up my thoughts on it in my standup card yesterday:

"Something I noticed when working on my recent PR in `ocw-studio` is that imported courses are assigned a website starter matching the slug `course`, which is assumed to have been imported from the `ocw-course-site-config.yml` file. The course starter coming from `ocw-hugo-projects` has a slug of `ocw-course`, and was partly named that to avoid a collision with the one inside the `ocw-studio` repo. I think we should remove the `ocw-course-site-config.yml` file and instead have a provisioning step that pulls in the starters in `ocw-hugo-projects` repo. I tried simply changing `COURSE_STARTER_SLUG` and started getting a bunch of errors during testing that didn't have an apparent cause, so that will take a bit of investigation."

Basically, I think if the course starter config in `ocw-hugo-projects` is going to be what we use to make course sites going forward then it doesn't make sense to maintain two files. We should simply have a step during provisioning of `ocw-studio` that gets the latest version of these configs and imports them based on a configuration value for the slug. This functionality already exists but is only fired off by a webhook from Github. What do you think?

noisecapella:
Can we make an issue for that?

noisecapella (approved):

gumaerc:
@noisecapella https://github.com/mitodl/ocw-studio/issues/457

PR #460, opened 2021-08-03
fix drawer width issue

### LINKED ISSUE #459: site content drawer width isn't determined correctly
### Steps to Reproduce

1. create a new site with the `OCW Course Hugo Starter` config (the one we have checked in to the repo)
2. try to add a new page
3. observe that the modal is squished! it should have `.right wide` but it just has `.right`. If you add the `.wide` class in the DOM inspector you should see if occupy the full width.

### Expected Behavior

When we've got a two-column layout for the form we should have a wide drawer.

### Actual Behavior

It's narrow!

### what's going on?

We look at the output of `splitIntoColumns` in `lib/site_content` to determine how wide we should make the drawer. If this returns an array of length 1 then it's a single column, else it's two columns.

The problem is that we do not call this on the same value everywhere. In particular

- in `RepeatableContentListing.tsx`, where the classname is set on the modal, we call `splitIntoColumns` with a 'raw' version of `website.config.fields`
- then in `SiteContentEditor.tsx` we call `addDefaultFields` with `website.config.fields` as an argument, and we store the return value using `useMemo`
- then in `SiteContentForm` we _again_ modify the value by calling `renameNestedFields` on it.

The result of not pushing the `addDefaultFields` call up the component tree is that when we call `splitIntoColumns` in `RepeatableContentListing` there will not be a 'title' field for the course starter, because that field is added by `addDefaultFields`. So we have a bit of inconsistency here.

iirc `renameNestedFields` needs to be called in `SiteContentForm` and not elsewhere because it's there to basically allow validation logic to work right, but I think we should hoist the `addDefaultFields` call as high as we can (i.e. to `SiteContentListing`) because otherwise we're jsut going to get confused by things like this.

### Screenshot or Screencast

Screen Shot 2021-08-03 at 3 43 47 PM

aliceriot:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #459

#### What's this PR do?

this PR fixes an issue where the drawer width wasn't always being set correctly, because we weren't being so consistent about where in the component hierarchy we were calling the `addDefaultFields` function on the fields in a collection. To fix it, I basically hoisted that call up to the `SiteContentListing` component, refactoring the function a bit in the process, so that we have one consistent value throughout that part of the component tree.

Read the issue for more details!

Check the repro steps on the issue, make sure you can repro this on the main branch. Then check out my branch here and make sure it fixes it.

the problem:

Screen Shot 2021-08-03 at 3 43 47 PM

the solution:

Screen Shot 2021-08-03 at 5 08 33 PM

### DIFF
--- a/static/js/components/SiteContentEditor.test.tsx
+++ b/static/js/components/SiteContentEditor.test.tsx
@@ -8,7 +8,6 @@ import * as formikFuncs from "formik"
 import SiteContentEditor from "./SiteContentEditor"
 import WebsiteContext from "../context/Website"
 
-import { DEFAULT_TITLE_FIELD } from "../lib/site_content"
 import * as siteContentFuncs from "../lib/site_content"
 import { siteApiContentDetailUrl, siteApiContentUrl } from "../lib/urls"
 import IntegrationTestHelper, {
@@ -17,7 +16,6 @@ import IntegrationTestHelper, {
 import {
   makeRepeatableConfigItem,
   makeSingletonConfigItem,
-  makeWebsiteConfigField,
   makeWebsiteContentDetail,
   makeWebsiteDetail
 } from "../util/factories/websites"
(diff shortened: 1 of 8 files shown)


abeglova (approved):
LGTM

PR #475, opened 2021-08-05
add resource embed widget

### LINKED ISSUE #453: add UI for fetching resources, inserting them into CKEditor
We need a for #451 which will fetch insertable items and display them so that the user can pick from them. We can base it on the code here: https://ckeditor.com/docs/ckeditor5/latest/framework/guides/tutorials/using-react-in-a-widget.html

- [x] fetch and display collection items in some sort of picker UI (a drawer? scrollable list? something simple)
- [x] add a CKEditor plugin for inserting a new `SiteContentEmbed` entry into the editor (nodetype added in #450)
- [x] call this insertion code from our React UI. We should be able to have our picker UI live as a child of the `MarkdownEditor` component and reference the editor that way.
- [x] tests!

aliceriot:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #453

#### What's this PR do?

This adds the first part of the resource embed UI: an interface for selecting resources and inserting them into the Markdown editor.

This does not include display a rich preview of those resources in the editor itself.

I updated `ocw-course-site-config.yml`, so you should be able to run the `override_site_config` management command to get some course sites configured to use this feature.

Then you should create some resource records, and then create a page in the same course. You should see a picker for resources, basically a scrollable list of them and a filter. Typing text into the filter should narrow the results shown (it will make a new request to the backend with the filter input text set to the `search` query param). Clicking on an entry in the list should add a node to the Markdown editor with the resource's `.text_id` property in it.

Screen Shot 2021-08-05 at 6 25 59 PM

Screen Shot 2021-08-05 at 6 26 38 PM

Screen Shot 2021-08-05 at 6 27 10 PM

### DIFF
--- a/static/scss/resource-embed-widget.scss
+++ b/static/scss/resource-embed-widget.scss
@@ -1,0 +1,14 @@
+.resource-embed-widget {
+  .resource-list {
+    max-height: 225px;
+    overflow: scroll;
+    border-left: 1px solid $form-control-gray;
+    border-right: 1px solid $form-control-gray;
+    border-bottom: 1px solid $form-control-gray;
+    border-radius: 0 0 0.25rem 0.25rem;
+
+    .resource {
+      cursor: pointer;
+    }
+  }
+}
--- a/localdev/configs/ocw-course-site-config.yml
+++ b/localdev/configs/ocw-course-site-config.yml
@@ -10,6 +10,7 @@ collections:
         name: title
         widget: string
         required: true
+        attach: "resource"
 
       - label: Body
         name: body
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "@dnd-kit/core": "^3.1.1",
     "@dnd-kit/sortable": "^4.0.0",
     "@sentry/browser": "^6.10.0",
+    "@testing-library/react-hooks": "^7.0.1",
     "@types/ckeditor__ckeditor5-core": "^11.0.3",
     "@types/enzyme": "^3.10.9",
     "@types/enzyme-adapter-react-16": "^1.0.6",
--- a/static/js/components/widgets/MarkdownEditor.test.tsx
+++ b/static/js/components/widgets/MarkdownEditor.test.tsx
@@ -50,6 +50,11 @@ describe("MarkdownEditor", () => {
     })
   })
 
+  it("should pass attach down ato a ResourceEmbedField", () => {
+    const wrapper = render({ attach: "resource" })
+    expect(wrapper.find("ResourceEmbedField").prop("attach")).toBe("resource")
+  })
+
   //
   ;[
     ["name", "name"],
--- a/static/js/components/widgets/MarkdownEditor.tsx
+++ b/static/js/components/widgets/MarkdownEditor.tsx
@@ -1,12 +1,14 @@
-import React from "react"
+import React, { useCallback, useRef } from "react"
 import { CKEditor } from "@ckeditor/ckeditor5-react"
-
+import { editor } from "@ckeditor/ckeditor5-core"
 import ClassicEditor from "@ckeditor/ckeditor5-editor-classic/src/classiceditor"
 
 import {
   FullEditorConfig,
   MinimalEditorConfig
 } from "../../lib/ckeditor/CKEditor"
+import ResourceEmbedField from "./ResourceEmbedField"
+import { RESOURCE_EMBED_COMMAND } from "../../lib/ckeditor/plugins/ResourceEmbed"
 
 export interface Props {
   value?: string
@@ -14,6 +16,7 @@ export interface Props {
   onChange?: (event: { target: { value: string; name: string } }) => void
   children?: React.ReactNode
   minimal?: boolean
+  attach?: string
 }
 
 /**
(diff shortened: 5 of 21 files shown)


static/scss/resource-embed-widget.scss
```diff
@@ -0,0 +1,14 @@
+.content-embed-widget {
```
gumaerc, line 1: Some of the naming here feels a bit confusing.  This component is referred to both as a "site content embed widget" and a "resource embed widget."  "Resource embed widget" feels more correct but it also refers to a type of content defined in a config file that could change.  Either way, I think we should refer to it as one name to reduce confusion.

static/js/components/widgets/SiteContentEmbedField.tsx
```diff
@@ -0,0 +1,81 @@
...
+      <input
+        placeholder="filter"
+        type="text"
+        onChange={onChangeFilter}
+        value={filterInput}
+        className="form-control filter-input"
+      />
+      <div className="content-list">
+        {listing.results ?
+          listing.results.map(item => (
```
gumaerc, line 67: I'm not sure if this is outside of the scope of this PR, but this list does not currently paginate.  I tested this both creating my own site and with an imported course.  The filtering function makes it usable because you can type what you're looking for, but it still only lists up to 10 results.  An indication of how many results were returned by your filter might be useful as well, but I think that's definitely out of scope.

static/js/lib/ckeditor/plugins/SiteContentEmbed.ts
```diff
@@ -41,6 +45,35 @@ class SiteContentMarkdownSyntax extends MarkdownSyntaxPlugin {
   }
 }
 
+/**
+ * A CKEditor Command for inserting a new SiteContentEmbed (siteContent)
+ * node into the editor.
+ */
+class InsertSiteContentEmbedCommand extends Command {
```
gumaerc, line 52: I'm not exactly sure where to put this comment, but I think there is an issue with rendering the shortcode in the markdown output.  I created a test site with a resource and embedded it in a page, and then published.  This is the markdown output I got for that page:

```
---
title: Test Page
type: page
uid: 65048845-2e79-4346-93ed-2d06a31df9a5
---
Test content

5b3431f6-c62e-4257-a1e2-644271dfa65d
```
It seems to just be putting the raw UID in there, whereas it should be surrounded in a shortcode.  I think the strategy of using a generic "resource" shortcode is a good idea, although we may want to think about the naming concerns I noted in one of my other comments.  I think maybe we should be a bit more generic about which collection contains the content to embed, since it's named with `name: resource` in the course template and might not always be that way.  Maybe we could do something like `{{< content-embed ${name} ${content} >}}`, that way Hugo knows what `type` of content to look for, and then can fetch the right one by UID.  Since the resource itself contains the rest of the data in its front matter, from there we can fetch the `filetype` and have conditional templates for each type of content defined.

As far as the types of resources, since those are just currently arbitrary strings defined in the config file ("Image, Video, Document, Other"), we should explicitly document which types of resources we support embedding in the `ocw-hugo-themes`' documentation when we get to that part.  I imagine for "Other" we would use some kind of file download fallback tempalte with just a download button, and we could also use that when a `filetype` is defined that doesn't match one supported by the theme.  Either way, we should make sure it's wrapped in some kind of shortcode, otherwise the UID will just appear as text.

static/js/components/widgets/SiteContentEmbedField.tsx
```diff
@@ -0,0 +1,81 @@
...
+            <div
```
gumaerc, line 68: Again, I'm not sure if this is outside of the scope of this PR but it would be nice to see some additional info here.  At least `filetype` would be nice.  For example with transcripts, there might be 3 or 4 different files labeled with the same title but a different `filetype`.

gumaerc (changes requested):
I did some testing on this and overall I like how it was implemented.  It functions well and the filter functionality is nice for finding the content I want, but there are some improvements I think we could do.  Let me know what makes sense to do in this PR and what doesn't.  The two concerns I had that I'd like to see addressed here are the naming issue and the shortcode implementation:

gumaerc:
Also worth noting, my open PR (https://github.com/mitodl/ocw-studio/pull/473) adds significant changes to the structure of `ocw-course-site-config.yml` so if that merges before this you'll want to look out for a rebase.

static/js/components/widgets/SiteContentEmbedField.tsx
aliceriot: for the first pass I want to stick to the filtering alone, if we feel we want to add pagination later I think we can

static/js/components/widgets/SiteContentEmbedField.tsx
aliceriot: maybe I'll do a switch on the filetype to display a different icon, to keep it small and lightweight

static/scss/resource-embed-widget.scss
aliceriot: oh yeah I think I sort of changed my mind and forgot to go through fully. I'll just call it resource embed widget

static/js/lib/ckeditor/plugins/SiteContentEmbed.ts
aliceriot: oh hmm I'm not sure what's going on there - could you check the storage format too?

aliceriot: When I add a resource in the editor and then check the saved content in the API, I see the shortcodes I expect:

Screen Shot 2021-08-06 at 2 55 40 PM

so maybe we have an issue on the pipeline rendering part stripping the shortcodes out somehow?

static/scss/resource-embed-widget.scss
aliceriot: I went through and renamed everything in line with that

static/js/components/widgets/SiteContentEmbedField.tsx
aliceriot: I think the problem we start to run into then is that the `filetype` field is defined in the website start config file, so if we add a display of that field here then we're sort of coupling the UI field here with a specific type of course config. I'm not sure the best way around that - we could for instance add some more config options that define how a metadata field from the 'resource' content items should be grabbed, but that could get kind of complicated too.

Anyhow, not sure what you think about what would be the best way forward there

gumaerc: I think it's certainly outside the scope of this PR but definitely worth a discussion with the group.  I don't think just the title is enough context overall, especially when a lot of resources start piling up.

static/js/lib/ckeditor/plugins/SiteContentEmbed.ts
gumaerc: The resulting `WebsiteContent` object created in the database on my instance did not have the shortcode:

```

Comment (gumaerc) [reply static/js/lib/ckeditor/plugins/SiteContentEmbed.ts]: I re-tested this with the latest changes and tried creating a new Page and everything worked:

```
HTTP 200 OK
Allow: GET, PUT, PATCH, DELETE, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "text_id": "1d7acff5-215d-40f5-94d8-c71c8c35774f",
    "type": "page",
    "content_context": null,
    "title": "Test Page 2",
    "markdown": "Test content 2\n\n{{< resource 5b3431f6-c62e-4257-a1e2-644271dfa65d >}}",
    "metadata": null,
    "file": null
}
```

gumaerc (approved):
:+1:

PR #488, opened 2021-08-12
Don't error on minimal site config

### LINKED ISSUE #434: Error loading page listing for minimalist starter config
### Steps to Reproduce

Create a website using this starter config:

    {
      "collections": [
        {
          "category": "Page",
          "fields": [
            {
              "label": "Title",
              "name": "title",
              "required": true,
              "widget": "string"
            }
          ],
          "label": "Page",
          "name": "page"
        }
      ],
      "root-url-path": "/"
    }

### Expected Behavior

Page listing site should show. Starter configs which are valid according to the schema should work in general for all our functionality.

### Actual Behavior

Page does not show and a stack trace is listed on the config

### Stacktrace

Screenshot from 2021-07-29 09-06-59

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #434 

#### What's this PR do?
If the content page is neither a repeatable content listing nor a singleton, it will render a null rather than error.

I think it would be better to fix this in the validation somehow to enforce one or the other but I couldn't see how we could easily make yamale do that

### DIFF
--- a/static/js/components/SiteContentListing.test.tsx
+++ b/static/js/components/SiteContentListing.test.tsx
@@ -1,3 +1,5 @@
+import casual from "casual-browserify"
+
 const mockUseRouteMatch = jest.fn()
 
 import React from "react"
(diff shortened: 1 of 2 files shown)


noisecapella:
Closing, going to solve this in the validation stage instead

PR #489, opened 2021-08-12
Use yaml.SafeLoader

### LINKED ISSUE #465: Use yaml.SafeLoader on public data to avoid security vulnerabilites
There is a warning message here: https://pyyaml.org/wiki/PyYAMLDocumentation#loading-yaml

I don't think there's a big security risk but it is a trivial change to swap in `SafeLoader` so we may as well do it

noisecapella:
#### Pre-Flight checklist

- [x] Testing

Fixes #465 

#### What's this PR do?
Replaces yaml.Loader with yaml.SafeLoader. yaml.Loader can execute arbitrary code 

Edit and save a starter config in the Django admin


abeglova (approved):
👍

Issue #493, opened 2021-08-16
url of internal menu items wrong


gumaerc:
### Steps to Reproduce

Create a site using a starter with a `navmenu` widget and some other basic content like pages.  Create a few pages, then create menu items linking to them internally.  When done, publish your site.

### Expected Behavior

A `config/_default/menus.yaml` file is rendered and the URLs point to where the content is expected to be on the rendered site

### Actual Behavior

The `url` property on each internal item is the path to the markdown file in the site source, not the expected URL once the site is rendered

PR #364, opened 2021-07-06
Added back-end for configuring navigation menus

gsidebo:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
Closes #323

#### What's this PR do?
Adds the back-end functionality for nav menus (building off of #340). If a `WebsiteContent` record is created for a config item that has a `"menu"` field, the logic to sync that object to the backend (Github) will namespace that menu data under a `"menu"` key (see [Hugo docs](https://gohugo.io/content-management/menus/#add-non-content-entries-to-a-menu)) and will add a `"url"` value for internal links, ensuring that the destination URLs of the menu items are accurate. When re-syncing from the backend, data that was entered for "menu" fields is "flattened" (i.e.: the "menu" namespace is removed) so the data can be edited as normal

#### How should this be manually tested?
**DB TO GITHUB:**
- Add/edit some data for a field with a "menu" widget (see #340) and save it
- Either thru automatic sync or the `sync_website_to_backend` management command, sync the website to Github
- Check the target file in Github

The data for each "menu" field should be namespaced under a "menu" key, and all internal links should have an accurate "url" destination

Example:

```yaml
menu:
  sidemenu:
  - identifier: c5047db5-5d30-481f-878c-4fe79eebeeb1
    name: Page 1
    url: content/page-1.md
    weight: 0
  - identifier: 38223bd4-8eae-4a81-91c2-b36cac529d69
    name: Page 3
    url: content/page-3.md
    weight: 10
```

**GITHUB TO DB:**
- Note the contents of your `WebsiteContent` record, then hard-delete it
- Run the `sync_backend_to_db` command to recreate that record

The `WebsiteContent` record should be recreated with the same data

#### Any background context you want to provide?
The Hugo-specific implementation here is intentional to save time. I made an effort to leave a comment in the places where corners were cut

PR #496, opened 2021-08-16
add the get_destination_url helper function and use it for menu urls

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/493

#### What's this PR do?
In https://github.com/mitodl/ocw-studio/pull/425 and https://github.com/mitodl/ocw-studio/pull/364 we added backend support for generating a Hugo nav menu configuration.  A file path to a piece of content's markdown was used for the `url` property, when in reality what is expected here is a URL; relative for internal, absolute for external.  This PR adds a helper function called `get_destination_url` that generates the expected relative URL of a piece of content in a rendered site and uses it in the Hugo menu serializer.

 - Spin up `ocw-studio` locally on this branch and make sure you are set up to sync to github.mit.edu
 - Make sure you have a website starter that includes a nav menu widget, like https://github.com/mitodl/ocw-hugo-projects/blob/main/ocw-course/ocw-studio.yaml
 - Create a website using this starter
 - Add some pages or other pieces of content that can be linked to internally
 - Add some menu entries that point to this content you've created
 - Publish your site and inspect `config/_default/menus.yaml`, the URL properties should now not contain the `content/` prefix nor the `.md` file extension

### DIFF
--- a/content_sync/utils.py
+++ b/content_sync/utils.py
@@ -46,7 +46,7 @@ def get_destination_url(
     """
     if content.is_page_content:
         filename = "" if content.filename == "_index" else content.filename
-        prefix = "{}/{}".format(content.website.name, site_config.content_dir)
+        prefix = f"{content.website.name}/{site_config.content_dir}"
         dirpath = (
             content.dirpath.replace(prefix, "", 1)
             if content.dirpath.startswith(prefix)
--- a/content_sync/utils.py
+++ b/content_sync/utils.py
@@ -46,8 +46,12 @@ def get_destination_url(
     """
     if content.is_page_content:
         filename = "" if content.filename == "_index" else content.filename
-        dirpath = content.dirpath.replace(content.website.name, "", 1)
-        dirpath = dirpath.replace("/{}".format(site_config.content_dir), "", 1)
+        prefix = "{}/{}".format(content.website.name, site_config.content_dir)
+        dirpath = (
+            content.dirpath.replace(prefix, "", 1)
+            if content.dirpath.startswith(prefix)
+            else content.dirpath
+        )
         return os.path.join(dirpath, filename)
     log.error(
         "Cannot get destination URL because is_page_content is false (content: %s)",
--- a/content_sync/utils.py
+++ b/content_sync/utils.py
@@ -45,9 +45,10 @@ def get_destination_url(
     Returns the URL a given piece of content is expected to be at
     """
     if content.is_page_content:
-        return os.path.join(
-            content.dirpath.replace(site_config.content_dir, "", 1), content.filename
-        )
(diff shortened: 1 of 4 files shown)


noisecapella:
FYI I found this bug when testing this issue, where URLs are not generated on output at all. https://github.com/mitodl/ocw-studio/issues/497. Were you able to reproduce this PR on your laptop?

gumaerc:
> FYI I found this bug when testing this issue, where URLs are not generated on output at all. #497. Were you able to reproduce this PR on your laptop?

I checked this out and yea, I see that after updating the menu items they have no `url` property and only `identifier` with the UUID.  I think you are correct in that internally we should only be storing the UUID and then upon exporting a course to Github, use `get_destination_url` to generate the URL.  Either way, `get_destination_url` uses `content.dirpath` as the base for building the URL, so that shouldn't really matter.  I think maybe `ocw_import.api.import_ocw2hugo_menu` should just conform with the version 4 UUID in that case as you showed in the issue, because it's not currently doing that on import:

```
menu_data["leftnav"][i]["identifier"] = str(
                uuid.UUID(menu_data["leftnav"][i]["identifier"])
            )
```

`url` is imported simply because `ocw-to-hugo` generates the URL and puts it into `menus.yaml`, and since all of the properties of all menu items are imported as metadata it's brought in alongside everything else.

gumaerc:
@noisecapella I added some additional transformation that was necessary, stripping away the website slug as well as `_index` from index pages, and I reformatted for readability.  Based on our discussion, I commented on the other issue with `is_valid_uuid` and will follow up on that tomorrow.

content_sync/utils.py
```diff
@@ -36,3 +36,21 @@ def get_destination_filepath(
...
+
+def get_destination_url(
+    content: WebsiteContent, site_config: SiteConfig
+) -> Optional[str]:
+    """
+    Returns the URL a given piece of content is expected to be at
+    """
+    if content.is_page_content:
+        filename = "" if content.filename == "_index" else content.filename
+        dirpath = content.dirpath.replace(content.website.name, "", 1)
```
noisecapella, line 49: Can we combine this line and the next one, and also check that the search string is a prefix?

    prefix = f'{content.website.name}/{site_config.content_dir}'
    if content.dirpath.startswith(prefix):
        dirpath = content.dirpath.replace(prefix, "", 1)

noisecapella (changes requested):

noisecapella:
Functionality works great after I made the edit to `is_valid_uuid`

gumaerc:
@noisecapella all cleaned up

noisecapella (approved):

PR #498, opened 2021-08-17
Hide hidden fields inside object fields

### LINKED ISSUE #494: Hidden fields inside object fields not handled correctly
### Steps to Reproduce
Create a website with the starter:

    {
      "collections": [
        {
          "category": "Content",
          "fields": [
            {
              "label": "Text field",
              "name": "textfield",
              "widget": "text"
            },
            {
              "fields": [
                {
                  "condition": {
                    "equals": "visible",
                    "field": "textfield"
                  },
                  "default": "hiddenvalue",
                  "label": "Hidden field",
                  "name": "hiddenfield",
                  "widget": "hidden"
                }
              ],
              "label": "Object field",
              "name": "objectfield",
              "widget": "object"
            }
          ],
          "folder": "content",
          "label": "Page",
          "name": "page"
        }
      ],
      "content-dir": "content",
      "root-url-path": "/"
    }

Go to the dialog to create a new page. There are 2 bugs here:
 - The hidden field inside the object field should not be visible, but it is shown as a text field. Hidden fields inside object fields should not be visible.
 - The condition on the hidden field is not being checked to determine what data is sent to the server. If there is a condition on the hidden field and it doesn't match, then no data for that field should be included in the payload sent to the server on `POST` or `PATCH`, which means the field should not appear in the metadata at all.

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #494 

#### What's this PR do?
Fixes handling of fields inside object fields, and hidden fields in particular:
 - Validation for fields within object fields is disabled when that field is not visible due to a condition that doesn't match
 - Fields within object fields are not visible in the UI if they are not visible due to a condition that doesn't match (or if they're hidden fields)
 - Fields within object fields will send default values as part of the payload if they are not visible due to a condition that doesn't match. This matches the current behavior for fields which aren't nested.
 - Hidden fields will now always send the default value. Previously, the default value was sent when creating a new content object, which meant it was stored in the metadata, and subsequent `PATCH`es would send back the same value. That usually works fine except if you change the `default` value in the schema to something else, or if you add a hidden field after the fact.

Create a new site with a starter having this config:

    {
      "collections": [
        {
          "category": "Content",
          "fields": [
            {
              "label": "Text field",
              "name": "textfield",
              "widget": "text"
            },
            {
              "fields": [
                {
                  "condition": {
                    "equals": "visible",
                    "field": "textfield"
                  },
                  "default": "hiddenvalue",
                  "label": "Hidden field",
                  "name": "hiddenfield",
                  "widget": "hidden"
                }
              ],
              "label": "Object field",
              "name": "objectfield",
              "widget": "object"
            }
          ],
          "folder": "content",
          "label": "Page",
          "name": "page"
        }
      ],
      "content-dir": "content",
      "root-url-path": "/"
    }

- Open your network tab in your browser to record `POST`s. Create two pages: one page should have "visible" in the "Text field" field, and the other should have "invisible" in that field. You should see `"hiddenfield": "hiddenvalue"` in the `POST` request for the page with "visible", and the other should have `"hiddenfield": ""`
- Edit both pages, swapping the values in the "Text field" field so that the page that used to have `visible` now has `invisible`, and vice versa. You should see two `PATCH`es in the network tab. The response for the page with "visible" should now have `"hiddenfield": "hiddenvalue"`and the other one should have `"hiddenfield": ""`

### DIFF
--- a/static/js/components/widgets/ObjectField.tsx
+++ b/static/js/components/widgets/ObjectField.tsx
@@ -46,7 +46,7 @@ export default function ObjectField(props: Props): JSX.Element {
       {collapsed ? null : (
         <div className="object-sub-fields">
           {field.fields
-            ?.filter(innerField => fieldIsVisible(innerField, values))
+            .filter(innerField => fieldIsVisible(innerField, values))
             .map((innerField: ConfigField) => (
               <SiteContentField
                 field={innerField}
--- a/static/js/components/widgets/ObjectField.tsx
+++ b/static/js/components/widgets/ObjectField.tsx
@@ -1,16 +1,19 @@
 import React, { useState, useCallback } from "react"
 
 import SiteContentField from "../forms/SiteContentField"
+import { fieldIsVisible } from "../../lib/site_content"
 
 import {
   ConfigField,
   ObjectConfigField,
   WebsiteContent
 } from "../../types/websites"
+import { SiteFormValues } from "../../types/forms"
 
 interface Props {
   field: ObjectConfigField
   contentContext: WebsiteContent[] | null
+  values: SiteFormValues
 }
 
 /**
@@ -18,7 +21,7 @@ interface Props {
  * to be edited.
  **/
 export default function ObjectField(props: Props): JSX.Element {
-  const { field, contentContext } = props
+  const { field, contentContext, values } = props
 
   const [collapsed, setCollapsed] = useState(field.collapsed ?? false)
   const toggleCollapse = useCallback(
@@ -42,13 +45,15 @@ export default function ObjectField(props: Props): JSX.Element {
       </div>
       {collapsed ? null : (
         <div className="object-sub-fields">
-          {field.fields?.map((field: ConfigField) => (
-            <SiteContentField
-              field={field}
-              key={field.name}
-              contentContext={contentContext}
-            />
-          ))}
+          {field.fields
+            ?.filter(innerField => fieldIsVisible(innerField, values))
+            .map((innerField: ConfigField) => (
+              <SiteContentField
+                field={innerField}
+                key={innerField.name}
+                contentContext={contentContext}
+              />
+            ))}
         </div>
       )}
       {field.help && <span className="help-text">{field.help}</span>}
--- a/static/js/lib/site_content.ts
+++ b/static/js/lib/site_content.ts
@@ -131,7 +131,8 @@ export const isSingletonCollectionItem = (
 ): configItem is SingletonConfigItem => "file" in configItem
 
 /**
- * Creates a value for a payload to be sent to the server, for a given field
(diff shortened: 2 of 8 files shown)


static/js/components/widgets/ObjectField.tsx
```diff
@@ -42,13 +45,15 @@ export default function ObjectField(props: Props): JSX.Element {
...
         <div className="object-sub-fields">
-          {field.fields?.map((field: ConfigField) => (
-            <SiteContentField
-              field={field}
-              key={field.name}
-              contentContext={contentContext}
-            />
-          ))}
+          {field.fields
+            ?.filter(innerField => fieldIsVisible(innerField, values))
```
alicewriteswrongs, line 49: I think we can get rid of the `?` here because as long as we're dealing with an object field is should always have fields defined

static/js/lib/site_content.ts
```diff
@@ -126,6 +126,39 @@ export const isSingletonCollectionItem = (
   configItem: BaseConfigItem
 ): configItem is SingletonConfigItem => "file" in configItem
 
+const contentFormValueForField = (
```
alicewriteswrongs, line 129: would you mind adding a comment here describing what this function does and why it's necessary?

alicewriteswrongs (changes requested):
tried it out and I think it works correctly, just had two small questions when looking through

noisecapella:
I added the comment above the function, does it read clearly to you?

static/js/lib/site_content.ts
```diff
@@ -130,6 +130,43 @@ export const isSingletonCollectionItem = (
   configItem: BaseConfigItem
 ): configItem is SingletonConfigItem => "file" in configItem
 
+/**
+ * Creates a value for a payload to be sent to the server, for a given field. This is used by to
+ * contentFormValuesForPayload to figure out what should go in the payload for a field and nested fields.
+ */
```
alicewriteswrongs, line 136:
```suggestion
 * Returns a value for a payload to be sent to the server for a field, given a ConfigField
 * and the current state of the form (`values`).
 *
 * We need to look at the values for the entire form in order to handle
 * the case of Object fields, where for the 'nested' entries in the 
 * Object we need to look at the value for their 'parent' field in order
 * to set them correctly.
 */
```

I was thinking something like this, provided you think it makes sense. It's just been a little confusing to me in this file to have a series of functions that do fairly closely related things, so I want to make sure we name things well and keep track of what everything is supposed to be doing somehow.

alicewriteswrongs (changes requested):
one suggestion

static/js/lib/site_content.ts
noisecapella: Sounds good

alicewriteswrongs (approved):
lgtm!

Issue #512, opened 2021-08-20
imported courses have the wrong dirpath for content when exported


gumaerc:
### Steps to Reproduce

 - Make sure you have Github integration set up to both pull in `ocw-to-hugo` data and export published content to Github under your own test organization
 - Import any number of courses using `import_ocw_course_sites`
 - Browse to the site's UI in `ocw-studio` and click the publish button
 - Browse to the repository in Github where your content has been published

### Expected Behavior

All content is under the `content` folder at the root of the repo

### Actual Behavior

The `content` folder is inside a folder named with the `course_id`

Issue #260, opened 2021-05-10
Remove course id from filepaths in imported OCW course sites


gsidebo:
When importing OCW course sites, we prepend `WebsiteContent.course_filepath` values with the full course id.

Example: `1-040-project-management-spring-2004/content/sections/syllabus.md`

The prepended course id is unnecessary for this project, and is almost definitely a relic from the way our pre-Studio OCW course site builds work. The example above should instead be `content/sections/syllabus.md`

If PR #259 is merged before this is complete, this issue should also include removing the prepended course id from `WebsiteContent.dirpath` values.

**Acceptance criteria:**
- Data migration added to remove course id prefixes from `content_filepath` values
- (If #259 is merged) Data migration added to remove course id prefixes from `dirpath` values
- Course site import logic updated to avoid prepending course id to `content_filepath`/`dirpath` values

gumaerc:
Closed by https://github.com/mitodl/ocw-studio/pull/513

PR #513, opened 2021-08-20
remove course_id prefix on dirpath of imported course content

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/512

#### What's this PR do?
When importing of `ocw-to-hugo` content was added, the `dirpath` property of all imported content was set based on the S3 key from the bucket it was pulled from.  In this bucket, each course is stored in a folder matching the course ID.  The dirpath on the imported content should have this prefix removed, as when the course is subsequently exported to Github by `ocw-studio` it will put the `content` folder in the wrong place.  This PR corrects that by removing the prefix when `dirpath` is set.

 - Make sure you have Github integration set up to both pull in `ocw-to-hugo` data and export published content to Github under your own test organization
 - Import any number of courses using `import_ocw_course_sites`
 - Browse to the site's UI in `ocw-studio` and click the publish button
 - Browse to the repository in Github where your content has been published
 - Ensure that the `content` folder is at the root of the repository

### DIFF
--- a/ocw_import/api.py
+++ b/ocw_import/api.py
@@ -85,13 +85,17 @@ def import_ocw2hugo_content(bucket, prefix, website):  # pylint:disable=too-many
         prefix (str): S3 prefix for filtering by course
         website (Website): Website to import content for
     """
+    site_prefix = f"{prefix}{website.name}"
     for resp in bucket.meta.client.get_paginator("list_objects").paginate(
-        Bucket=bucket.name, Prefix=f"{prefix}{website.name}/content"
+        Bucket=bucket.name, Prefix=f"{site_prefix}/content"
     ):
         for obj in resp["Contents"]:
             s3_key = obj["Key"]
             s3_content = get_s3_object_and_read(bucket.Object(s3_key)).decode()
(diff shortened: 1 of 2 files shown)


mbertrand:
This will also close #260

noisecapella (approved):

noisecapella:
@gumaerc do we need to adjust anything here or elsewhere for import? I'm not sure if there's any possibility of importing what we exported to github https://github.com/mitodl/ocw-studio/blob/master/ocw_import/api.py#L89

gumaerc:
@noisecapella Good point.  I think if we want to pull published content Github back into `ocw-studio`, we would add another function.  Something like `import_github_content` that would clone the repo and then iterate the `content` directory, calling `convert_data_to_content` on it all.  The function you noted is specifically `import_ocw2hugo_content` so I would only really expect it to be compatible with the way `ocw-to-hugo` outputs multiple courses, each to their own directory named with the course ID.

PR #516, opened 2021-08-23
Make starter required for OCW_IMPORT_STARTER_SLUG

### LINKED ISSUE #514: import_ocw_course_sites sets starter to None
### Steps to Reproduce

On my local instance I ran `./manage.py import_ocw_course_sites -b ocw-to-hugo-output-qa --filter res-8-001-applied-geometric-algebra-spring-2009`, and it set the value for the corresponding website's `starter` field to `None`. I set the starter to the ocw-course starter manually, then reran the command, and it set it to `None` again, so you should be able to reproduce this easily I think.

noisecapella:
#### Pre-Flight checklist

- [x] Testing

Fixes #514 

#### What's this PR do?
 - Remove a constant which was also being used as the course starter slug. Instead testing code will use the setting
 - If the starter does not exist on import, an exception will be thrown

We also need to make sure our production environments have environment variables which match the starter slugs

Run `./manage.py import_ocw_course_sites -b ocw-to-hugo-output-qa --filter res-8-001-applied-geometric-algebra-spring-2009`. The website should have the starter populated correctly, or else there should be an error in the import or when Django starts and finds a missing environment variable.


localdev/management/commands/override_site_config.py
```diff
@@ -6,51 +6,65 @@
...
+    def _get_pairs(self):
+        """
+        Get yaml paths and starter slugs.
+        Since we are referencing settings we can't populate this until
+        Django is initialized, so it has to go inside a function
+        """
+        return (
+            (
+                "localdev/configs/ocw-course-site-config.yml",
+                settings.COURSE_STARTER_SLUG,
```
mbertrand, line 27: Should this be `settings.OCW_IMPORT_STARTER_SLUG` ?

mbertrand (changes requested):

mbertrand (approved):
👍

Issue #510, opened 2021-08-20
imported instructors should be published with ocw-www


gumaerc:
### Steps to Reproduce

Make sure you have an `ocw-www` site (created using the `ocw-www` starter) and then import any number of courses using `import_ocw_course_sites`, then publish `ocw-www`

### Expected Behavior

Instructors are published with `ocw-www` as headless content under `content/instructors`

### Actual Behavior

They are not published

PR #518, opened 2021-08-24
publish instructors

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/510

#### What's this PR do?
In https://github.com/mitodl/ocw-studio/pull/409, we added the creation of instructors in the "root website" (`ocw-www`) when importing courses from `ocw-to-hugo`.  Currently, these instructors are not published with the rest of `ocw-www`'s content.  In order for them to be accessible in the published site, we need to change a few things.  This PR sets `is_page_content` to true for imported instructors, and sets the proper dirpath as well as a filename equal to the `text_id`.

 - Spin up `ocw-studio` on this branch and ensure that you have both the `ocw-www` and `ocw-course` starers from `ocw-hugo-projects` in your local instance.  Also, ensure that your local instance is set up to sync to Github and pull content from S3.  To test this, you're going to want to have your own test Github organization.  You'll then need to create a personal access token within Github and put that in your `.env` file before spinning up `ocw-studio`.  The values should look something like:
 ```
GIT_API_URL=https://api.github.com
GIT_ORGANIZATION=<github org here>
GIT_TOKEN=<personal access token here>
```
 - Create a site using the `ocw-www` starter.  You don't need to create any content, the site just needs to exist.
 - Run `docker-compose run web ./manage.py import_ocw_course_sites -b ocw-to-hugo-output-qa --filter 4-123`
 - Browse to the Instructors section of your `ocw-www` site.  You should see Prof. Meejin Yoon and Prof. Muhamet Yildiz.
 - Click the Publish button
 - Browse to your Github org that you created and go to the `ocw-www` repo under there.  Browse to the instructors folder and verify that there are two markdown files there.  The filenames should match the `text_id` of the instructors imported into `ocw-www` and the front matter should include the instructor information.  It should look like this: https://github.com/gumaerc-testorg/ocw-www/tree/main/content/instructors

### DIFF
--- a/ocw_import/api.py
+++ b/ocw_import/api.py
@@ -192,6 +192,7 @@ def import_ocw2hugo_sitemetadata(
     instructor_contents = []
     for instructor in course_data["instructors"]:
         uid = instructor["uid"]
+        text_id = str(uuid.UUID(uid))
         first_name = instructor.get("first_name", "")
         last_name = instructor.get("last_name", "")
         middle_initial = instructor.get("middle_initial", "")
@@ -201,7 +202,7 @@ def import_ocw2hugo_sitemetadata(
 
         instructor_content, _ = WebsiteContent.objects.update_or_create(
             website=website_root,
-            text_id=str(uuid.UUID(uid)),
+            text_id=text_id,
             defaults={
                 "metadata": {
                     "headless": True,
@@ -212,6 +213,9 @@ def import_ocw2hugo_sitemetadata(
                 },
                 "title": f"{salutation_plus_space}{first_name} {middle_initial_plus_space}{last_name}",
                 "type": CONTENT_TYPE_INSTRUCTOR,
+                "dirpath": "content/instructors",
+                "filename": text_id,
+                "is_page_content": True,
             },
         )
         instructor_contents.append(instructor_content)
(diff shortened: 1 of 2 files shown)


mbertrand (commented):
Works great. One concern though - If a course is published with new instructors, but ocw-www hasn't been published lately, will this break the course's instructors list?  Should ocw-www automatically be previewed/published every time another site is published, assuming it has 1+ instructors?

gumaerc:
> Works great. One concern though - If a course is published with new instructors, but ocw-www hasn't been published lately, will this break the course's instructors list? Should ocw-www automatically be previewed/published every time another site is published, assuming it has 1+ instructors?

@mbertrand That's a good point.  I think in `ocw-hugo-themes` when pulling these in regardless, there should be error handling for that case.  I think your proposal makes sense as well, but maybe only if the instructors list on a course has changed?  Maybe it should be more generic, something like "If I'm publishing a course that has relation widgets tied to other websites, publish those sites before this one."

gumaerc:
@mbertrand After discussing this with Ferdi, it seems that the actual expected behavior here might be that the relation widget's select field doesn't even display unpublished content as a selectable option.  The expected outcome of this might change, but I logged an issue for it here: https://github.com/mitodl/ocw-studio/issues/520

PR #525, opened 2021-08-26
Filter out unpublished content from REST API

### LINKED ISSUE #520: relation widget allows selection of unpublished content across websites
### Steps to Reproduce

 - Define a relation widget in a starter that references content in another website
 - Create some content in the related website's field and save it but don't publish it
 - Create a site with the starter using the relation widget, and click the select field to bring up the available options

### Expected Behavior

Unpublished content is not selectable

### Actual Behavior

Unpublished content is selectable

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #520 

#### What's this PR do?
Adds a filter to the website content REST API so that unpublished content is not visible

TBD

View the website content list view REST (for example `http://localhost:8043/api/websites/res-8-001-applied-geometric-algebra-spring-2009/content/`) for a website you imported but never published. It should have no content. A website which is published should have some content in the REST API.

### DIFF
--- a/static/js/components/widgets/RelationField.tsx
+++ b/static/js/components/widgets/RelationField.tsx
@@ -71,6 +71,7 @@ export default function RelationField(props: Props): JSX.Element {
   // if website: websitename param is set, then we want to use the context
   // website, else we want to use the cursor to fetch the specified website
   const websiteName = props.website ? props.website : contextWebsite.name
+  const differentSite = !props.website || (props.website !== contextWebsite.name)
 
   const handleChange = useCallback(
     (event: any) => {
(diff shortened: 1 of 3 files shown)


noisecapella:
It looks like `WebsiteContent` which is created is automatically synced to `main` but not to the preview or release branches. So there isn't going to be an easy way to produce a case to show that the relation field correctly filters out unpublished content, without manually unpublishing it via the Django admin. 

This also brings up another question. If it is published to main but not to preview or release, then it effectively still does not exist in the git repository for those two branches. I don't think we keep track of whether we published to a particular branch, just whether we published at all. I'm not sure how to proceed there

PR #573, opened 2021-09-10
WIP resource type

noisecapella:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

(Required)

#### What's this PR do?
(Required)

(Required)

(Optional)

(Optional)

(Optional)

(Optional)


websites/api.py
```diff
@@ -204,3 +206,10 @@ def mail_website_admins_on_publish(website: Website, version: str, success: bool
                     "version": version,
                 },
             )
+
+
+def detect_mime_type(uploaded_file: UploadedFile) -> str:
+    """Detect mime type of an uploaded file"""
+    magic = Magic(mime=True)
+    chunk = next(uploaded_file.chunks(chunk_size=2048))
+    return magic.from_buffer(chunk)
```
mbertrand, line 215: An uploaded JSON file got assigned 'text/plain'  instead of 'application/json', not sure if that's worth worrying about.

mbertrand (changes requested):
Needs a rebase.

There are still references to "filetype" in the code (mostly tests though), including:
`FILE_FIELDTYPE` in `main.settings`
`WebsiteContentViewSet.get_queryset` in `websites.views`
`unassigned_youtube_ids` in `websites.api`
`update_youtube_metadata` in `videos.youtube`

These should be updated too.

websites/api.py
noisecapella: Looks like it's a known bug that's not easily fixed https://github.com/ahupp/python-magic/issues/193

noisecapella:
@mbertrand Thanks, that was an oversight on my part. Should be all renamed now

websites/serializers.py
```diff
@@ -222,6 +222,12 @@ def update(self, instance, validated_data):
             update_youtube_thumbnail(
                 instance.website.uuid, validated_data.get("metadata"), overwrite=True
             )
+        if "file" in validated_data:
+            if "metadata" not in validated_data:
+                validated_data["metadata"] = {}
+            validated_data["metadata"]["file_type"] = detect_mime_type(
```
mbertrand, line 228: Just realized this will insert `file_type` into the metadata even if that field is not part of the starter config.  Doesn't seem like this will necessarily break anything though.  Do you think its worth checking first to see if this field is in the config definition, and/or if the website is an ocw course (starter.slug == settings.OCW_IMPORT_STARTER_SLUG)?

Ditto for the `WebsiteContentCreateSerializer`

mbertrand (commented):
Looks great overall, just one question

websites/serializers.py
noisecapella: I don't know. Any thoughts @Ferdi @gumaerc @pdpinch?

pdpinch: mime type seems like a generally useful thing, when uploading a file. I don't think we need to make it conditional on presence in the starter.

mbertrand (approved):
👍

Issue #536, opened 2021-08-31
add support for linking to resources to the Markdown editor


alicewriteswrongs:
As a course author, I'd like to both embed 'resource' content and also link to it!

Currently we support embedding resources into Markdown via a 'resource' node (`"resourceEmbed"`) which is written out to markdown as the `resource` shortcode. Currently this is a block-level node, meaning it can only be rendered at the top level (not, for instance, within a paragraph).

In addition to support a block-level embed (think an embedded video or something like that) we'd also like to support linking to resource content. These links should be inline-able within text content too, to support linking to resources freely within any part of the course content (see the 'Grading' section of this syllabus for an example: https://ocw.mit.edu/courses/chemistry/5-310-laboratory-chemistry-fall-2019/syllabus/).

To support this we should add a 'resource-link' node type to CKEditor, and serialize it to and from a `resource_link` shortcode. This will let us enforce different restrictions in the markdown editor for where these two nodes can be inserted, and it will also let us render them differently with Hugo.

#### Questions

- Should we rename the existing shortcode from `resource` to `resource-embed` ?

#### Acceptance Criteria:

- [ ] add resource_link node type to our CKEditor config
- [ ] add React component to display resource_link nodes in the CKEditor content
- [ ] make modifications to Hugo templates to support rendering these links.

alicewriteswrongs:
unassigning myself, I left some comments on #615 on how to go forward on that, and I think #616 is pretty straightforward to do (it should be fairly similar to toggling italic or bold on a span of text).

PR #589, opened 2021-09-20
update CKEditor config to use our resource-link fork

### LINKED ISSUE #607: add Link resource and Embed resource buttons to the editor toolbar
on the toolbar on the Markdown editor we'd like to have three separate buttons relating to links:

- the standard CKEditor 'link' button for making normal HTML links to external sites
- a 'Link resource' button for adding links to resources, which opens the resource picker to select the resource
- an 'Embed resource' button for adding a resource embed, which also opens the resource picker to select the resource

### LINKED ISSUE #612: allow inline resource links with user-definable text
we want to allow content editors to write resource links into the course content in a familiar way - they should behave like normal HTML links, with text inside of them and so on.

alicewriteswrongs:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #607
closes #612
part of #536

#### What's this PR do?

This PR changes how we support linking to resources. Instead of a block-level element (which is sort of like a paragraph in terms of how it fits in to the document) we now support inline linking. Basically, this means that the resource link is actually an attribute on a span of text, instead of a block element, and so the text inside of it can be edited and so on.

The user experience is not _the greatest_ at present, but here's how it works:

- the user highlights some text in the editor
- the user clicks 'Link resource'
- the user uses the resource picker dialog to pick a resource, and clicks the 'Link resource' confirmation button
- a resource link is added to the editor state, serializing to the `{{< resource_link >}}` shortcode

To get this all working I had to make a few other changes. Some things of note:

- we now have a fork of ckeditor here: https://github.com/mitodl/ckeditor. this is being used just to fork the `Link` plugin, but since ckeditor is developed as a monorepo we need to fork the whole thing. The changes I've made to the plugin to support our weird use-case are here: https://github.com/mitodl/ckeditor5/pull/1 Once this PR is approved I will merge that PR as well and publish a non-beta version of our custom package to npm.
- I changed how the resource picker works a little bit. Instead of having one 'Add resource' button we now have 'Embed resource' and 'Link resource' buttons. Together these make it a little less ambiguous for the user (at least, that is the hope!)

follow the steps I outlined above. Keep in mind that the UX is still basic now. We do not yet support editing which resource a resource link points to once it is created, for instance. For now the user should just delete and recreate to do that. But generally play around with it and make sure it at least does what it's supposed to.

Screen Shot 2021-09-28 at 12 38 23 PM


noisecapella:
Functionality looks good other than the editing issue you mentioned in the description

static/js/lib/ckeditor/plugins/ResourcePicker.ts
```diff
@@ -18,17 +24,33 @@ export default class ResourcePicker extends Plugin {
...
+    })
+
+    editor.ui.componentFactory.add(ADD_RESOURCE_EMBED, (locale: any) => {
+      const view = new ButtonView(locale)
+
+      view.set({
+        label:    "Embed resource",
+        withText: true
+      })
+      // TODO: icon? how to right-justify?
```
noisecapella, line 50: Can we remove this comment and the same above?

noisecapella (approved):

noisecapella:
Looks great, just noticed a stray comment

PR #597, opened 2021-09-23
Add pragma: allowlist secret to ignore false positives for secret detection

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

None

#### What's this PR do?
Adds comment to tell the git precommit hook to ignore these two example values which look like secrets

N/A, this change is entirely inside a docstring so there should be no changes in functionality


abeglova (approved):
LGTM

PR #591, opened 2021-09-22
Sync all files in Google Drive

mbertrand:
#### Pre-Flight checklist
- [x] Migrations
  - [x] Migration is backwards-compatible with current production code
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested
- [x] Settings
  - [x] New settings are documented and present in `app.json`
  - [x] New settings have reasonable development defaults, if applicable
  - [x] Opened issue for DevOps regarding necessary configuration changes to deployed environments
 

#### What are the relevant tickets?
- Implements backend for #578 

#### What's this PR do?
- New celery tasks to import non-video files on Google Drive and auto-create WebsiteContent resources for them, including a scheduled task to import all recent non-video files, that runs every hour by default.
- A new view that will trigger a task to import the non-video files for a specific website
- Refactor some tasks that were being called from other tasks without `raise self.replace` and/or grouping/chaining.

#### How should this be manually tested?
- Set `DRIVE_SERVICE_ACCOUNT_CREDS` and `DRIVE_SHARED_ID` with credentials from RC
- Set `DRIVE_VIDEO_UPLOADS_PARENT_FOLDER_ID=1H4HCvbmY7v5YZFeqSlbCI1TFC5MXTMY4` in your .env file.
- Create a new site with starter whose slug equals the value of your `OCW_IMPORT_STARTER_SLUG` value
- Check the google drive folder to make sure there is a new folder with name == `website.short_id` and 2 subfolders, "video_final" and "files_final"
- Repeat the instructions for importing videos (#484), but place the drive videos under the "video_final" folder - other than that it should work as before.
- Place one or more files in the "files_final" folder (images, pdfs, etc)
- Go to `http://localhost:8043/api/websites/<your_site_name>/content/gdrive_sync` and click the `POST` button.
- Very shortly you should have new resources for all the files in that folder.  The resource type ("Document", "Image", etc) for each should be a reasonable guess based on the S3 content type or file extension.  The title should equal the original filename.
- Save some changes to the new resources (for instance, change the titles).
- Add another 1 or 2 files in the drive folder.
- Go to `http://localhost:8043/api/websites/<your_site_name>/content/gdrive_sync` and click the `POST` button again.
- Check that there are again new resources for these files, the old resources still exist with the changes you made to the titles still intact, and there are no dupes. 

#### Any background context you want to provide?
- A frontend button to trigger the drive sync will be added in a future PR.
- Videos still do not have resources automatically created.  This may change in a future PR.

PR #611, opened 2021-09-28
Tweak s3 path for gdrive-imported nonvideo files

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Related to #591 

#### What's this PR do?
Changes the S3 key path of google drive nonvideo files to be the same as manually uploaded files so that they will be copied over by the publishing pipelines.

Follow steps for #604, check that the s3 path is `<website.starter.config.root-url-path>/<site-name>/<filename>`

### DIFF
--- a/gdrive_sync/api.py
+++ b/gdrive_sync/api.py
@@ -167,8 +167,10 @@ def streaming_download(drive_file: DriveFile) -> requests.Response:
     return response
 
 
-def stream_to_s3(drive_file: DriveFile, prefix=settings.DRIVE_S3_UPLOAD_PREFIX):
+def stream_to_s3(drive_file: DriveFile, prefix: str = None):
     """ Stream a Google Drive file to S3 """
+    if prefix is None:
+        prefix = settings.DRIVE_S3_UPLOAD_PREFIX
     service = get_drive_service()
     permission = (
         service.permissions()
--- a/gdrive_sync/api.py
+++ b/gdrive_sync/api.py
@@ -167,7 +167,7 @@ def streaming_download(drive_file: DriveFile) -> requests.Response:
     return response
 
 
-def stream_to_s3(drive_file: DriveFile):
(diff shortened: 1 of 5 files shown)


gdrive_sync/api.py
```diff
@@ -167,7 +167,7 @@ def get_confirm_token(response: requests.Response) -> str:
     return response
 
 
-def stream_to_s3(drive_file: DriveFile):
+def stream_to_s3(drive_file: DriveFile, prefix=settings.DRIVE_S3_UPLOAD_PREFIX):
```
noisecapella, line 170: I think `settings.DRIVE_S3_UPLOAD_PREFIX` is evaluated during file import rather than during the function call here. Can we change this to `prefix=None` and then have a `if prefix is None: prefix = settings.DRIVE_S3_UPLOAD_PREFIX` at the beginning of the function?

noisecapella (changes requested):

gdrive_sync/api.py
mbertrand: Thanks, changed it

gdrive_sync/tasks.py
```diff
@@ -32,11 +32,13 @@
 
 
 @app.task(bind=True)
-def stream_drive_file_to_s3(self, drive_file_id: str):
+def stream_drive_file_to_s3(
+    self, drive_file_id: str, prefix=settings.DRIVE_S3_UPLOAD_PREFIX
```
noisecapella, line 36: Can we make a similar change here?

noisecapella (changes requested):

noisecapella:
Got functionality working, just the one comment

gdrive_sync/tasks.py
mbertrand: Changed

noisecapella (approved):

PR #600, opened 2021-09-23
Add slug field to set a different value for filenames of content

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
Fixes #594 
Fixes #499 

#### What's this PR do?
Adds a `slug` field for the schema which describes a field to use to set the filename when the content is exported

#### How should this be manually tested?
 - Update your local ocw-www starter in Django admin. Find the collection for instructors, then add `"slug": "text_id"` to that collection. Click save. It should save successfully if you added this in the right place, otherwise there will be a validation error.
 - In the ocw-studio UI, go to the ocw-www site and add a new instructor
 - View the `WebsiteContent` for the new instructor. The filename field should have a uuid, rather than a slugified title.

PR #625, opened 2021-09-30
Fix bug so slug field is used for filename

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Bug fix for #600 

#### What's this PR do?
Add `text_id` to be passed through to the website context. We are generating `text_id` in `_get_derived_website_content_data` which is used by the serializer context. However it was being filtered out of `WebsiteContentCreateSerializer`, which meant the default uuid generated by the model was used instead for `text_id`. This changes that to use the uuid passed in so that `text_id` and `filename` match up.

Update the ocw-www schema to the latest on ocw-hugo-projects if you haven't already. Then go to site ocw-www and create a new instructor. In the django admin or shell look at that instructor. Verify that the `filename` matches `text_id` on that `WebsiteContent`

### DIFF
--- a/websites/serializers.py
+++ b/websites/serializers.py
@@ -315,7 +315,7 @@ class WebsiteContentCreateSerializer(
         user = self.user_from_request()
         added_context_data = {
             field: self.context[field]
-            for field in {"is_page_content", "filename", "dirpath"}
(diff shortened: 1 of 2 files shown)


gumaerc (approved):
:+1: LGTM

PR #631, opened 2021-10-01
[proposal] change how URL matching works in IntegrationTestHelper

alicewriteswrongs:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

no ticket

#### What's this PR do?

This is a proposal for changing a bit how we manage mocking out requests for our frontend tests. I recently ran into a few pain points with our current setup that I think this will address. What does it change?

The main thing is making an unmocked request into an error that will fail a test. Previously, an unmatched request (i.e. a call made by one of our components which was not matched by the `handleRequestStub` on `IntegrationTestHelper`) would do one of two things:

1. if the `defaultResponse` of `{ body: {}, status: 200 }` didn't cause any application logic errors it would simply allow the test to pass
2. if the `defaultResponse` did cause an error you'd be confronted with a sometimes-opaque error message, and figuring out exactly which URL wasn't mocked out, what wasn't quite right about it, etc could require a good deal of spelunking (I'd sometimes find myself adding things like `throw(url)` to query configs and the like).

My proposed solution here is to basically make two changes:

1. add a Jest plugin that causes `console.error` and `console.warn` calls to fail tests
2. add a `console.error` call in our request mocking code on `IntegrationTestHelper` which will write an error message to the console whenever `handleRequestStub` is called with a URL and HTTP verb combination which has not been previously explicitly mocked using `withArgs`.

This means that now we have some guarantees that in our test suite we are mocking 100% of API calls made by the application. It also means that if you are writing a test and you have an unmocked URL you will get a straightforward and unambiguous error message that causes the test to fail early. The error message prints out the unexpected URL to the console, which makes it easy to figure out where you need to add a mock.

I think this is a better situation overall. I think we want to always either explicitly mock API responses or, if the requests are being made by a child component, we can also mock out that child component to achieve sort of the same thing.

Interested in some feedback about the change - is this too disruptive? Should we print a warning message without failing the test instead? Personally I'm partial to require this to be dealt with and thereby disallowing tests which don't mock out all their API dependencies correctly, but open to other ideas about how to go about this.

Testing only change, so tests should be passing, changes should make sense, etc.


noisecapella:
This all makes sense to me. I think erroring in more cases is a good idea since it helps us diagnose problems faster.

noisecapella (approved):

alicewriteswrongs:
I think so too!

Issue #561, opened 2021-09-08
Copy all file resources from the legacy ocw bucket to the ocw studio bucket


mbertrand:
All pdfs, images, non-youtube videos, and other files in the OCW legacy S3 bucket need to be copied over the OCW Studio bucket, under appropriate site-specific paths, so that they are included in preview/publish tasks.

#### Designs and Mockups

-  Add an additional option on the existing import command to copy all legacy ocw files (videos, pdf's, images, docs, etc) from the legacy ocw bucket to the OCW Studio bucket, using the correct S3 object prefix in the destination.
- Modify the ocw import code so that the `WebsiteContent.file` field is populated correctly, pointing to the appropriate OCW Studio S3 url.

pdpinch:
@mbertrand can we close this now?

PR #635, opened 2021-10-04
Save file location to WebsiteContent.file for imported OCW courses

mbertrand:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

#561 

#### What's this PR do?
- Saves the `file_location` value of imported OCW course resources to the `WebsiteContent.file` field, with some modifications based on the starter config (default is to replace `coursemedia` with `courses`).

- Run `manage.py import_ocw_course_sites -b ocw-to-hugo-output-qa. --limit 10`
- Check that the `WebsiteContent` resources have populated `file` values equal to the `file_location` in the metadata, but with `coursemedia` replaced by `courses`.  

Current assumption is that the OCW S3 files will be copied by devops from the legacy bucket to the ocw-studio RC/prod buckets, so no code to do so has been added in this PR.

### DIFF
--- a/ocw_import/api.py
+++ b/ocw_import/api.py
@@ -135,8 +135,19 @@ def convert_data_to_content(filepath, data, website):  # pylint:disable=too-many
             parent, _ = WebsiteContent.objects.get_or_create(
                 website=website, text_id=str(uuid.UUID(parent_uid))
             )
+        # Assumes that s3 objects will be independently synced to the ocw-studio bucket via devops
+        file_location = content_json.get("file_location", None)
+        if file_location:
+            ocw_prefix = (
+                website.starter.config.get("root-url-path", "courses")
+                if website.starter
+                else "courses"
+            )
+            file_location = file_location.replace("coursemedia", ocw_prefix)
+
         base_defaults = {
             "is_page_content": True,
+            "file": file_location,
             "metadata": content_json,
             "markdown": (s3_content_parts[1] if len(s3_content_parts) == 2 else None),
             "parent": parent,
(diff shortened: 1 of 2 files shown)


ocw_import/api.py
```diff
@@ -135,8 +135,19 @@ def convert_data_to_content(filepath, data, website):  # pylint:disable=too-many
...
             )
+        # Assumes that s3 objects will be independently synced to the ocw-studio bucket via devops
+        file_location = content_json.get("file_location", None)
+        if file_location:
+            ocw_prefix = (
+                website.starter.config.get("root-url-path", "courses")
+                if website.starter
+                else "courses"
+            )
+            file_location = file_location.replace("coursemedia", ocw_prefix)
```
gumaerc, line 146: Since we only want to replace a prefix here, can we check that `file_location` starts with `coursemedia`, and then only replace the one instance?  It's unlikely that `coursemedia` would be elsewhere in the URL, but if it is this code would replace it with `ocw_prefix`

gumaerc (changes requested):
Looks good, just one thing I would change:

ocw_import/api_test.py
```diff
@@ -88,6 +88,9 @@ def test_import_ocw2hugo_course_content(mocker, settings):
     assert lecture_pdf.metadata.get("file_type") == "application/pdf"
     assert lecture_pdf.filename == "file-20"
     assert lecture_pdf.dirpath == "content/resources"
+    assert lecture_pdf.file == lecture_pdf.metadata.get("file_location").replace(
```
gumaerc, line 91: Can we update this test to use the `re.sub` function as well?

gumaerc (approved):
:+1: LGTM

Issue #569, opened 2021-09-09
Tables


pdpinch:
As a course author, I'd like to be able to create, edit and publish tables in my course

#### Acceptance Criteria:

- [x] Turn on the CKEditor tables widget & configure it to work with markdown
  - [x] does it handle escaping `|` characters? 
  - [x] how/does it handle colspan?
  - [x] how/does it handle line breaks?
- [x] export markdown for tables
- [x] make any necessary theme updates 
   - [x] desktop
   - [x] mobile 
- [x] allow course author to edit tables they created in studio
- [ ] allow course author to edit tables that were imported via ocw-to-hugo 
- [x] #663
- [x] #672
- [x] #673

pdpinch:
@alicewriteswrongs can you update this issue? Do you think we could close it yet?

alicewriteswrongs:
the only thing I'm not sure about is editing imported tables (i.e. tables produced by `ocw-to-hugo` - @gumaerc do you know if `ocw-to-hugo` has been updated / if that's supported yet?

pdpinch:
Editing imported tables on RC is working fine for me. I'm closing this epic.

PR #645, opened 2021-10-05
enable tables in CKEditor

### LINKED ISSUE #445: add support for tables to WYSIWYG editor
As a course author, I'd like to be able to add tables to pages.

See https://ckeditor.com/docs/ckeditor5/latest/features/table.html for information about table support in ckeditor. It looks quite robust.

There is a known issue with markdown rendering of nested tables, but I don't consider that a required feature.

alicewriteswrongs:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

closes #445
part of #569

#### What's this PR do?

This adds the CKEditor table plugin to our CKEditor configuration, giving us a table button and whatnot.

The markdown serialization is supported OOB because CKEditor's markdown stuff which we're already using already supports it.

Test out adding a table to a page. You should be able to do things like add text or numbers inside, maybe add some lines, and so on.

Screen Shot 2021-10-05 at 4 17 16 PM
Screen Shot 2021-10-05 at 4 17 21 PM


alicewriteswrongs:
A new thing to test is that

- the tables should _always_ have a header row when you insert them
- tables should serialize to markdown correctly, and not be saved as HTML. You can check by just looking in the django admin.

gumaerc:
So I took a look at this and overall I think it's fine, but there are some things we should consider.  Inside table cells, formatting options like bold, italic and underscore as well as links work fine.  However, anything that causes a line break like bulleted / numbered lists, block quotes or simply manually inserting a line break by hitting enter will break the table because they insert line breaks into the cell.  In `ocw-to-hugo`, we handle this by replacing line breaks with a line break shortcode (`{{< br >}}`) so maybe we could add something to our ckeditor configuration that would do that upon saving the markdown?

Another concern are pipe characters.  Since pipes are what define the table structure in the resulting markdown, if you insert any pipe characters into the table in ckeditor it breaks the table on the other end.  In `ocw-to-hugo`, we translate these to the HTML character entity representing a pipe, `&#124;`.  We have other concerns in `ocw-to-hugo` but anything besides what I've already stated here is mostly a concern with converting legacy data.  Either way, there's no any of this would be a hard stop in a Hugo build.  The worst thing that would happen would be an improperly rendered table.

PR #719, opened 2021-10-19
add 'sortable' option to the Relation field

### LINKED ISSUE #687: Relation widget:  Add "ordered" option
We need support collections ( in studio config) like video gallery , or  image gallery . Currently we can do something like this

```
collections:
  - category: Content
    folder: content/videolectures
    label: Video Gallery
    name: videogallery
    fields:
      - label: Title
        name: title
        widget: string
        required: true

      - label: Body
        name: body
        widget: markdown

      - label: Videos
         name: gallery_videos
         widget: relation
         collection: resource
         display_field: title
         multiple: true
         filter:
            field: "resourcetype"
            filter_type: "equals"
            value: "Vidoe"
```

This would be ok, it's not ideal. It would be hard to arrange the order of videos in the gallery.  

The suggestion here is to add a new attribute `sortable` to the relation widget. If `true`,  The relation widget UI would looks something like :
videogallery

alicewriteswrongs:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #687 

#### What's this PR do?

This adds a new option to the 'relation' field which is `sortable`. If `sortable === true` then instead of using a normal react-select multiselect input element, we show a more complicated UI with two elements: 1) a single-select which adds like an 'add' input for selecting new pieces of content, and 2) a 'reordering' area which uses a drag-and-drop UI to allow the user to freely reorder the elements already selected.

I added a 'Video Gallery' content item to the omnibus config, so you can use that to test out how it works. You'll need to create some resources first and then, once you have those, you should be able to create a video gallery and try out the reorderable select there.

#### screens

Screen Shot 2021-10-19 at 2 08 16 PM
Screen Shot 2021-10-19 at 2 08 25 PM
Screen Shot 2021-10-19 at 2 08 30 PM
Screen Shot 2021-10-19 at 2 08 36 PM
Screen Shot 2021-10-19 at 2 08 44 PM

### DIFF
--- a/localdev/configs/omnibus-site-config.yml
+++ b/localdev/configs/omnibus-site-config.yml
@@ -40,6 +40,23 @@ collections:
           - { label: "State", name: "state", widget: "string", required: false }
           - { label: "Zip Code", name: "zip", widget: "string", required: false }
 
+  - category: Content
+    folder: content/video_galleries
+    label: "Video Gallery"
+    name: video_gallery
+    fields:
+      - label: Videos
+        name: videos
+        widget: relation
+        multiple: true
+        collection: resource
+        display_field: title
+        sortable: true
+        filter:
+          field: "resourcetype"
+          filter_type: "equals"
+          value: "Video"
+
   - name: resource
     label: Resources
     label_singular: Resource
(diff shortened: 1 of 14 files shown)


noisecapella:
Are we allowing duplicates for the sortable relation field? I don't think we do for the non-sortable or multiselect ones
Screenshot from 2021-10-20 10-15-08

noisecapella:
Functionality and code look good, just the one question

PR #726, opened 2021-10-21
Use name instead of short_id for resource S3 keys

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

N/A

#### What's this PR do?
Uses `website.name` instead of `website.short_id` for file resources imported from google drive.  Adds a management command for fixing resources that have the wrong s3 keys.

Run `manage.py move_misplaced_s3_keys`.  When complete, make sure the `WebsiteContent.file` fields for existing non-video resources have `Website.name` instead of `Website.short_id` (for websites where they are different), that they are publicly viewable, and that the keys are also updated for the resources in github.

Add new files to Google Drive for a site with differing `name` vs `short_id` values and sync them.  Make sure their `file` field values include the name and not the short_id.

The management command doesn't modify any existing video file s3 keys because they are not directly linked to in Studio (youtube only).

### DIFF
--- a/gdrive_sync/management/commands/move_misplaced_s3_keys.py
+++ b/gdrive_sync/management/commands/move_misplaced_s3_keys.py
@@ -34,6 +34,8 @@ class Command(BaseCommand):
                         f'{drive_file.s3_prefix}/{site["name"]}',
                         1,
                     )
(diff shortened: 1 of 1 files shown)


mbertrand:
@abeglova Ready for another look

PR #757, opened 2021-10-29
Handle level import, add term and year to metadata

noisecapella:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Related to https://github.com/mitodl/ocw-hugo-projects/issues/89

#### What's this PR do?
Adjusts import to handle updated format for level, and also include data for term and year. The import code should remain backwards compatible for a short while so imports don't break while ocw-to-hugo runs.

 - Run `./manage.py import_ocw_course_sites -b ocw-to-hugo-output-qa --limit 1`. This should take about 5 minutes
 - In the Django shell run `WebsiteContent.objects.get(website__name="1-00-introduction-to-computers-and-engineering-problem-solving-spring-2012", type="sitemetadata").metadata`. You should see a "year" field with `None` in the metadata.

When we switch to the newer ocw-to-hugo output it should also work without changes, but we can't test that yet.


ocw_import/api.py
```diff
@@ -268,6 +289,9 @@ def import_ocw2hugo_sitemetadata(
         "content": [content.text_id for content in instructor_contents],
     }
 
+    metadata["term"] = course_data["term"]
+    metadata["year"] = course_data.get("year")
```
mbertrand, line 293: Minor thing, wondering if we should use `course_data.get()` consistently in this function, or is `course_data[]` used when it is guaranteed to be present?

mbertrand (approved):
Looks good, just 1 minor comment, take or leave.

ocw_import/api.py
noisecapella: Right, `term` exists in both the legacy and updated format so I used `[]` for that

PR #763, opened 2021-11-01
Fix poll_build_status_until_complete to use celery countdown instead of sleep

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

None

#### What's this PR do?
Fixes `poll_build_status_until_complete` to use celery to call a delayed task rather than `sleep` to delay within the task. `sleep` caused the celery worker to do nothing for 5 seconds, which caused problems with other tasks in queue

Do a publish on a site. You should see a task being executed every 5 seconds or so in the celery logs. At some point you should see a success or failure in the UI, and the task should stop being called in the celery logs

### DIFF
--- a/app.json
+++ b/app.json
@@ -207,6 +207,10 @@
       "description": "Max retry attempts to get an S3 object",
       "required": false
     },
+    "MAX_WEBSITE_POLL_SECONDS": {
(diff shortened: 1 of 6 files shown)


abeglova (approved):
LGTM

PR #771, opened 2021-11-03
add 404 page, show it when the user gets a 404 from the site API

### LINKED ISSUE #764: no error message when requesting a site that doesn't exist
### Steps to Reproduce

1. Request a site that doesn't exist, like https://ocw-studio-rc.odl.mit.edu/sites/foo-bar 

### Expected Behavior

- user gets an error message, ideally with a 404 header 

### Actual Behavior

- no visible error message
- sites API returns a 404

### Screenshot or Screencast

image

alicewriteswrongs:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #764

#### What's this PR do?

This adds a 404 page and renders it in the case that the user tries to visit a nonexistent website.

Try going to a site normally - everything should be, well, normal. Then try going to like `/sites/foobasdfasdfasdf` or something (unless of course you happen to already have that site!). You should see a 404 page like in the screenshot.

I also added a 'wildcard match' (see [here](https://reactrouter.com/web/example/no-match)) at the bottom of our `<Switch>` in `App.tsx`. I'm not sure this will ever match anything because we only render certain routes as react routes on the backend, but it's a little security I guess.

Screen Shot 2021-11-03 at 11 32 30 AM

### DIFF
--- a/static/js/components/NotFound.tsx
+++ b/static/js/components/NotFound.tsx
@@ -1,0 +1,19 @@
+import React from "react"
+
+interface Props {
+  children?: React.ReactNode
+}
+
+/**
+ * Our 404 component.
+ */
+export default function NotFound(props: Props) {
+  const { children } = props
+
+  return (
+    <div className="w-50 m-auto pt-5">
+      <h1>That's a 404!</h1>
+      {children}
+    </div>
+  )
+}
--- a/static/js/pages/App.test.tsx
+++ b/static/js/pages/App.test.tsx
@@ -13,7 +13,8 @@ describe("App", () => {
   let helper: IntegrationTestHelper,
     render: TestRenderer,
     website: Website,
-    siteDetailApiUrl: string
+    siteDetailApiUrl: string,
+    siteDetailUrl: string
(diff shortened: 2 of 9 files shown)


mbertrand (commented):
Is the new 404 page supposed to show up for every 404 response or just the ones returned by the API?  I got a "Not Found" response but not the new 404 page for these URL's, but they aren't API requests:

http://localhost:8043/nonsense
http://localhost:8043/sitees
http://localhost:8043/privacy-policy/oops

Screen Shot 2021-11-03 at 2 58 50 PM

alicewriteswrongs:
@mbertrand yeah so those 404 pages are basically coming from the backend before the JS app has an opportunity to do anything. This PR is basically designed to cover cases where we boot the frontend app (i.e. the user navigates to `/sites/foobar`) but the API returns a 404 there. It's a little funky to have two different types of 404 pages but I think it's ok :)

mbertrand (approved):
👍

PR #772, opened 2021-11-03
Tweak WebsiteContent permissions

### LINKED ISSUE #738: ocw-www permissions issue
### Steps to Reproduce

visit a course site as a non-global admin user and try to edit the metadata for the course.

### Expected Behavior

One should be able to edit the metadata!

### Actual Behavior

It seems like in some cases (not sure of the exact incidence of this) instead there is an issue where the `GET` request to the website content API for ocw-www to fetch instructors fails with a 403 because the user doesn't have permissions for that API.

Maybe we want to make sure the read-only access to such things is working properly? Or maybe we need to extend/modify some of our permission code to correctly account for this case? not sure!

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #738 

#### What's this PR do?
Makes the API request to list `WebsiteContent` objects  viewable to any logged in user, not just collaborators on that associated `Website`.

- Make a new normal user (not superuser) and add it as an admin collaborator to a `Website` (but not to ocw-www)
- Log in as that user, and go to the metadata page for that site.
- You should be able to add new instructors from a populated dropdown list.  There should be no 403 API responses.

### DIFF
--- a/websites/permissions.py
+++ b/websites/permissions.py
@@ -267,7 +267,9 @@ class HasWebsiteContentPermission(BasePermission):
             Website, name=view.kwargs.get("parent_lookup_website", None)
         )
         if request.method in SAFE_METHODS:
-            return check_perm(request.user, constants.PERMISSION_VIEW, website)
+            # Give permission to any logged in user, to facilitate content relations between different sites.
+            # In particular, to allow instructor relations from ocw-www in site metadata.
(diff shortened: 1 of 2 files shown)


alicewriteswrongs (approved):
lgtm!

PR #773, opened 2021-11-03
update yarn configuration

alicewriteswrongs:

none

#### What's this PR do?

This upgrades us to Yarn 3, which is setup and operates fairly differently than before. Read about the differences [here](https://yarnpkg.com/getting-started/migration).

The biggest difference I actually opted out of (moving to the plug n' play dependency resolving scheme for node) because it didn't work out of the box and I didn't want to fiddle with it.

CI should be passing. If you pull down the branch the webpack build should work and whatnot.


noisecapella (approved):
Works great :+1:

Issue #495, opened 2021-08-16
menu items with 0 weight not ordering properly


gumaerc:
### Steps to Reproduce

Create a site using a starter with a `navmenu` widget and some other basic content like pages. Create a few pages, then create menu items linking to them internally. When done, publish your site.

### Expected Behavior

No menu items should have a weight of zero

### Actual Behavior

The lowest order menu items have a weight of zero

### Additional Info
As seen in [this](https://discourse.gohugo.io/t/menu-weight-and-other-oddity/5653) post, Hugo will order menu items with a weight of zero by name.  If we want to have items at the top of the list, we should use a weight of 1 instead.

PR #775, opened 2021-11-04
make sure menu items never have zero weight

gumaerc:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/495

#### What's this PR do?
In https://github.com/mitodl/ocw-studio/pull/364, we added the back end support for managing menus.  When this was done, it was set up so the weighting of these items starts at zero.  Hugo does not support weights of zero, so in this PR the code has been altered so the weight starts at 10 and increments from there.  A migration was also added to adjust existing menu items up by 10 for sites created through `ocw-studio`.

 - Spin up a local instance of `ocw-studio` and ensure that Github integration is enabled and pointing at a test organization you've created as detailed here: https://github.com/mitodl/ocw-studio#enabling-github-integration
 - Create a site at http://localhost:8043/sites using a starter that has a navmenu collection, such as [`ocw-course`](https://github.com/mitodl/ocw-hugo-projects/blob/main/ocw-course/ocw-studio.yaml)
 - Go to the Pages section and create some pages in your site
 - Go to the Menu section and add some internal and external link menu items, nesting some of them
 - Save and publish the site to your test Github organization
 - View the published output and browse to `config/_default/menus.yaml` and ensure that no menu items have weights of zero


websites/migrations/0040_no_zero_menu_weights.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 3.1.13 on 2021-11-04 18:10
+
+from django.db import migrations
+from websites.models import WebsiteContent
+from websites.constants import WEBSITE_SOURCE_STUDIO
+
+
+def add_ten_to_weights(apps, schema_editor):
+    navmenus = WebsiteContent.objects.filter(
```
noisecapella, line 9: We need to use `apps.get_model` to get the `WebsiteContent` class because it refers to the historical version of the model. That way we can change the model in the future but the version of `WebsiteContent` in the migration here still has the old columns. https://docs.djangoproject.com/en/3.2/topics/migrations/#data-migrations

noisecapella (changes requested):

websites/migrations/0040_no_zero_menu_weights.py
```diff
@@ -0,0 +1,24 @@
...
+        navmenu.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("websites", "0039_gdrive_sync_progress"),
+    ]
+
+    operations = [migrations.RunPython(add_ten_to_weights)]
```
noisecapella, line 24: We should also add a function to allow backwards migrations here

noisecapella (changes requested):

gumaerc:
@noisecapella should be set for another look

noisecapella (approved):

Issue #786, opened 2021-11-08
migration 0040_no_zero_menu_weights fails on navmenu with no leftnav


gumaerc:
### Steps to Reproduce

 - Create a Website with the omnibus config
 - Create a menu in the site and save
 - Run migrations

### Expected Behavior

The migrations complete without error

### Actual Behavior

There is an error

### Stacktrace

```
Applying websites.0040_no_zero_menu_weights...Traceback (most recent call last):
......
         File "/tmp/build_e812639a/websites/migrations/0040_no_zero_menu_weights.py", line 9, in migrate_weights_forward
           increment_decement_weights(apps, True)
         File "/tmp/build_e812639a/websites/migrations/0040_no_zero_menu_weights.py", line 23, in increment_decement_weights
           for menuitem in navmenu.metadata["leftnav"]:
       KeyError: 'leftnav'
```

### Related Issues

(Optional)

### Screenshot or Screencast

(Optional)

PR #787, opened 2021-11-08
check if leftnav exists in 0040_no_zero_menu_weights

gumaerc:
#### Pre-Flight checklist

- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/786

#### What's this PR do?
In https://github.com/mitodl/ocw-studio/pull/775, we changed the weights of menu items to start at 10 instead of 0 and included a migration to increment all existing weights by 10 to compensate.  This migration does not check for the existence of the `leftnav` key before accessing it and therefore fails on databases with other types of menus.  This PR adds an `isnull` filter to the `WebsiteContent` objects the migration is applied to, making sure it only applies to `navmenu` objects with a `metadata.leftnav` key.

 - Spin up `ocw-studio` locally and make sure you have the [`ocw-course`](https://github.com/mitodl/ocw-hugo-projects/blob/main/ocw-course/ocw-studio.yaml) starter
 - Create a site using the `ocw-course` starter
 - Add a few pages to your course
 - Go to the Menu section and add nav menu entries for your pages and save everything
 - Go to Django admin and find the `navmenu` `WebsiteContent` object for your site and manually edit the metadata, replacing "leftnav" with something else like "rightnav"
 - In a terminal, run `docker-compose run web ./manage.py migrate --fake websites 0039_gdrive_sync_progress` to fake back to the previous migration
 - Run `docker-compose run web ./manage.py migrate websites 0040_no_zero_menu_weights` and make sure you see no errors

### DIFF
--- a/websites/migrations/0040_no_zero_menu_weights.py
+++ b/websites/migrations/0040_no_zero_menu_weights.py
@@ -17,7 +17,9 @@ def increment_decement_weights(apps, forward):
     WebsiteContent = apps.get_model("websites", "WebsiteContent")
     weight_change = 10 if forward else -10
     navmenus = WebsiteContent.objects.filter(
-        type="navmenu", website__source=WEBSITE_SOURCE_STUDIO
+        type="navmenu",
+        website__source=WEBSITE_SOURCE_STUDIO,
+        metadata__leftnav__isnull=False,
     )
     for navmenu in navmenus:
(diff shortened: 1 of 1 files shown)


mbertrand (approved):
👍

PR #813, opened 2021-11-16
add global option to RelationField component

### LINKED ISSUE #780: add 'global' option to RelationWidget to allow implementing cross-site resource collections
As a course author, I'd like to author collections of resources like some of the resource lists which currently exist in production OCW.

One example is the 'Online textbooks' collection which is an index of all the resources on OCW which could be classified as textbooks. Importantly, this is a _cross-site_ collection, meaning a given resource's eligibility for inclusion does not depend on what site it is saved within.

#### Acceptance Criteria:

- [x] add a `cross_site` option to the RelationWidget which will add an input for picking a course to look within
- [x] once a course is picked the widget will allow searching and adding resources within that course
- [x] this should play well with the `sortable` option which we're already using for video galleries on the `RelationField`
- [x] once an item is added to the list it should be displayed with the course it is located within
- [x] in addition to adding support for this option in ocw-studio we should make a change to the course config to allow course authors to create these things in the `ocw-www` course

#### Out of Scope

- Not going to do

alicewriteswrongs:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #780

#### What's this PR do?

this adds a new option to the relation widget which is called `cross_site`. This option basically enables the same 'linking' behavior we have now, but with the added option of picking which website you want to 'look within' in order to find items to link to. This functionality can be used to, for instance, implement resource lists which group together diverse resources from across a swath of websites.

I added a new ocw yaml file based on the one in `ocw-hugo-projects` (by which I mean I copy-pasted). This has a cross-site relation field implementing a resource list set up in it already, so if you set up a new site which uses that configuration you should be able to test it out.

In the UI test out that:

- clicking in to fields you always have some options in the dropdowns
- you see options in the second field after picking a website in the first
- when you add things to the list you can then smoothly reorder them, delete them, etc
- you should be able to save and then re-open things

generally the UI should be unsurprising I think. Hopefully it is!

https://user-images.githubusercontent.com/6207644/142086507-832f1641-e480-4235-b400-5f11cded6771.mov


noisecapella:
When I refresh the browser and reopen I see the uuid:
Screenshot from 2021-11-17 14-55-48

alicewriteswrongs:
yeah, I think that may be somewhat unavoidable for the time being - if we want to avoid that we have to basically crawl the list and issue a request for each separate item / website combo, since each item can potentially be from a different website, in order to get the info from the backend that we need to render the title...but it isn't a great experience.

Maybe we could add some sort of bulk site content API that takes an array of UUIDs and returns the objects?

noisecapella:
I found a different error:
Screenshot from 2021-11-17 16-37-25

I added this field to the "Pages" part of the omnibus starter:

        {
          "collection": "resource",
          "display_field": "title",
          "label": "PDFs (singular)",
          "multiple": false,
          "name": "pdfs_singular",
          "widget": "relation",
          "cross_site": true
        },

Then I went to Pages and clicked "Add Page". I went to that field and typed in "nuts" to get "Nuts and Bolts of Business Plans", then clicked on "3play caption file" which was the first item in the dropdown.

noisecapella:
@alicewriteswrongs regarding the uuid issue, I think we should be including it in the `content_context` part of the API response, which should include any `WebsiteContent` which is referred to in the field values. I'll investigate and see what's going on there

noisecapella:
Ah, it looks like you just need to update the `get_content_context` function to handle the tuple including the website. I think this is the area you would want to modify: https://github.com/mitodl/ocw-studio/blob/master/websites/serializers.py#L360

alicewriteswrongs:
cool, I'll fix that so that we get the context

also - I have been sort of assuming that this will only be used for relation fields with `multiple = true` - I think if we did that for now it would be easiest

alicewriteswrongs:
@noisecapella ok I just pushed two fixes, one to fix the context generation on the backend and then on the frontend a change to properly initialize the `contentMap` object from the `contentContext` value passed in. afaict without that change we would also miss titles for things in non-cross-site relation widgets because we weren't putting that `contentContext` value into the `contentMap` `Map` (which is where we pull out `WebsiteContent` objects to get the title in the drag-and-drop zone)

static/js/components/widgets/RelationField.tsx
```diff
@@ -304,16 +383,43 @@ export default function RelationField(props: Props): JSX.Element {
...
+      focusedContent,
+      setFocusedContent,
+      onChangeShim,
+      focusedWebsite,
+      value,
+      crossSite
+    ]
   )
 
+  console.log(contentMap)
```
noisecapella, line 407: Stray console.log here

noisecapella:
I'm still seeing this bug https://github.com/mitodl/ocw-studio/pull/813#issuecomment-972089288, but the uuid default loading issue looks solved

alicewriteswrongs:
@noisecapella I think that should be fixed for now - since our use case doesn't require it, I'm going to leave supporting non-multiple select w/ cross_site off for the future

noisecapella (approved):

PR #822, opened 2021-11-19
remove some unnecessary mocks of `global.fetch`

alicewriteswrongs:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

none

#### What's this PR do?

In #813 I added a mock of `global.fetch` in our 'global' `beforeEach` function, making a few places where we do the same elsewhere in tests redundant, so I went and removed them. Just doing a little housekeeping I guess :)

This is a testing change only, so it should be good enough to just confirm that the build is passing.


abeglova (approved):
LGTM

PR #820, opened 2021-11-17
Use webhooks and periodic task for updating publishing statuses

mbertrand:
#### Pre-Flight checklist
- [x] Migrations
  - [x] Migration is backwards-compatible with current production code
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested
- [x] Settings
  - [x] New settings are documented and present in `app.json`
  - [x] New settings have reasonable development defaults, if applicable

#### What are the relevant tickets?
Closes #817 

#### What's this PR do?
- Adds webhooks back to the pipeline definitions.  The requests send back various build statuses: "started" at the beginning,  "succeeded" if it completes, "failed" if anything breaks.
- Adds a periodic task to query concourse for websites with pending builds whose statuses have not been updated in at least X (default 10) minutes.  If the status has not changed after a certain cutoff number of minutes (default 30) then the build is aborted.
- Tweaks the pipeline definition to skip requiring a git private key for public repos if `settings.CONCOURSE_IS_PRIVATE_REPO=True` (default is `False`).

#### How should this be manually tested?
- Use same AWS, CONCOURSE, GIT settings as RC
- Download ngrok and run `./ngrok http 8043`.  Copy the ngrok url to `OCW_STUDIO_BASE_URL=<ngrok_url>` in your .env file
- Restart docker instances
- Run `manage.py import_ocw_course_sites --filter res-7-001-pre-7-01-getting-up-to-speed-in-biology-summer-2019`
- Run `manage.py sync_backend_to_db --website res-7-001-pre-7-01-getting-up-to-speed-in-biology-summer-2019`
- Run `manage.py backpopulate_pipelines --filter  res-7-001-pre-7-01-getting-up-to-speed-in-biology-summer-2019`
- Publish a new draft version of the site from the front end.  Check that the status indicator updates as the build starts on concourse, and eventually changes to failed or completed.  You should also get an email at the end.
- When done testing, run this on heroku RC to get the pipeline back to what it should be: `manage.py backpopulate_pipelines --filter  res-7-001-pre-7-01-getting-up-to-speed-in-biology-summer-2019`

PR #832, opened 2021-11-23
Use task decorator to prevent multiple instances of recurring scheduled tasks

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #820

#### What's this PR do?
Modifies the `single_website_task` decorator for more generic use, and adds it to various scheduled tasks that should only have 1 instance running at a time.

- Add these lines to the beginning of `videos.tasks.update_youtube_statuses` and restart docker containers:
```python
    log.error("Running update_youtube_statuses")
    sleep(10)
```    
- Run the following in a shell:
```python
from videos.tasks import update_youtube_statuses
[update_youtube_statuses.delay() for i in range(5)]
```
- Check the console log, after entering the above command there should be only 1 log output of  "Running update_youtube_statuses"

- Modify the decorator and function kwargs for `videos.tasks.update_youtube_statuses` and restart containers:
```python
@app.task(autoretry_for=(BlockingIOError,), retry_backoff=True, max_retries=10)
@single_task(timeout=settings.YT_STATUS_UPDATE_FREQUENCY, raise_block=True)
```
- Run the same loop as above in a new shell. 
- Check the console log, after entering the above shell command there should eventually be 5 log outputs of  "Running update_youtube_statuses" (and lots of Retry log statements).

### DIFF
--- a/content_sync/decorators.py
+++ b/content_sync/decorators.py
@@ -2,7 +2,7 @@
 import functools
 import logging
 from time import sleep
-from typing import Callable
+from typing import Callable, Optional
 
 from django.conf import settings
 from github.GithubException import RateLimitExceededException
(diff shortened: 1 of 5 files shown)


gumaerc (approved):
:+1:

PR #833, opened 2021-11-23
put title as default text in a resource link instead of the uuid

### LINKED ISSUE #767: resource link with no link text should use title instead of uuid
As a course author, I'd like to quickly add links to resources to a page. 

Most of the time, I would like the link text to be exactly the same as the resource title. If I add a link with no link text, the wysiwyg editor should use the resource title -- the same title that's used in the resource pop-up.

alicewriteswrongs:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes  #767

#### What's this PR do?

This sets the resource title to be the default text inside of a resource link, rather than the UUID that we're doing currently.

This PR should work as advertised right now. Create a resource with a title, and then add a page to your site (this is all in the `ocw-course` starter). Then you should be able to use the markdown editor to add a link to your nicely-titled resource, and the title of that resource should be the text inside of the link, rather than the uuid.

Before this can merge we need to merge and release https://github.com/mitodl/ckeditor5/pull/8 and update this PR to reference the newly-published package.

Screen Shot 2021-11-23 at 4 16 41 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "@ckeditor/ckeditor5-widget": "^29.2.0",
     "@dnd-kit/core": "^3.1.1",
     "@dnd-kit/sortable": "^4.0.0",
-    "@mitodl/ckeditor5-resource-link": "29.2.0",
+    "@mitodl/ckeditor5-resource-link": "^29.2.1",
     "@sentry/browser": "^6.10.0",
     "@testing-library/react-hooks": "^7.0.1",
     "@types/ckeditor__ckeditor5-core": "^11.0.3",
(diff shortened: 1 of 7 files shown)


mbertrand (approved):
LGTM! There was one strange problem where after 1 save [everything reverted to text](https://github.com/mbtestorg1a/mb-local-1-2/commit/8e8c83e769d9136095dbc4d107c9776d6ed17132#diff-b89bea5837d07e46cfab5a1b9d4ab1cd795e433aa9a5c0dd6eed3eeae36a72d0), but I haven't been able to make that happen again, and I don't see anything in the PR that would obviously cause it, so it was probably an unrelated fluke.

alicewriteswrongs:
hmm that's odd, I wonder if there's a problem with the regex we use to recognize shortcodes...lmk if you run into that again

PR #836, opened 2021-11-29
move from casual-browserify to the normal package

alicewriteswrongs:

none

#### What's this PR do?

Remove [casual-browserify](https://github.com/klowner/casual-browserify) in favor of [casual](https://github.com/boo1ean/casual), since the 'browserify' package is looking more and more unmaintained.

Tests should pass.


mbertrand (approved):
👍

PR #852, opened 2021-12-07
Concourse in a docker container

mbertrand:
Optionally starts a concourse docker container for some partial testing of pipeline functionality.

#### What's this PR do?
- Adds definitions for a concourse container to `docker-compose.yml` and updates the README.

Follow the new README instructions.  Try creating new sites and/or running the `manage.py backpopulate_pipelines` command for existing sites.  Try publishing a site.


docker-compose.yml
```diff
@@ -87,3 +87,39 @@ services:
...
+    privileged: true
+    depends_on: [concourse-db]
+    ports: ["8080:8080"]
+    stdin_open: true
+    tty: true
+    environment:
+      CONCOURSE_POSTGRES_HOST: concourse-db
+      CONCOURSE_POSTGRES_USER: postgres
+      CONCOURSE_POSTGRES_DATABASE: concourse
+      CONCOURSE_EXTERNAL_URL: http://host.docker.internal:8080
```
gumaerc, line 112: Since `host.docker.internal` isn't a thing in Linux, I had another issue related to this setting.  I was confused because I thought that the `CONCOURSE_URL` setting in the `.env` would control the URL `backpopulate_pipelines` would hit Concourse on, but this doesn't seem to be the case.  Once I changed this to `http://concourse:8080` and recreated my containers, `backpopulate_pipelines` worked correctly.  Instead of having `host.docker.internal` in my laptop's `hosts` file, I added an entry for `127.0.0.1  concourse` there and the web UI was available in my browser.  Also if `CONCOURSE_URL` in `.env` is supposed to control where commands like `backpopulate_pipelines` point to then you should probably look into that because it doesn't seem to be the case.

gumaerc (changes requested):
I was able to get this to work, but there are some Linux specific issues that need to be addressed:

docker-compose.yml
mbertrand: Thanks, made that change.  Both `CONCOURSE_URL` and `CONCOURSE_EXTERNAL_URL` are required: the former determines the url that studio uses to connect, and the latter determines the redirect url for login/authentication.

gumaerc (approved):
:+1:

PR #855, opened 2021-12-07
Prevent users from creating content with forbidden filenames

### LINKED ISSUE #829: Don't allow index or _index for a filename
As a user, I'd like to be able to upload files or create pages without worrying how it could affect the hugo build service. We should either reject or slugify filenames which could potentially cause problems when the content is created or imported

#### Acceptance Criteria:

- [ ] Files published should not be named `index.md` or `_index.md`

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #829 

#### What's this PR do?
- Renames any pages or resources created via the UI that are equal to a list of "forbidden" words (currently "index" and "_index"), but does not apply this to imported OCW sites.

- Import this course:  `manage.py import_ocw_course_sites --bucket ocw-to-hugo-output-qa --filter 1-011-project-evaluation-spring-2011`
- Verify that the `Projects` page for the course has a `filename` of `_index`
- Add a page to a new or existing site and give it a title of `index` or `_index`
- Check that the `filename` for the page is `index-page` (or `_index-page`)
- Add another page with the same title.  The `filename` should end with `-page2`
- Add a file to the site's `final_files` drive folder with a name like `index.jpg` or `index.pdf` and sync the drive.
- Verify that the `WebsiteContent.filename` for the drive file is `index-resource`

### DIFF
--- a/websites/utils.py
+++ b/websites/utils.py
@@ -40,3 +40,10 @@ def set_dict_field(obj: Dict, field_path: str, value: Any):
 def get_dict_query_field(dict_field_name: str, sub_field: str):
     """Generate django query key for searching a nested json feild"""
     return dict_field_name + "__" + sub_field.replace(".", "__")
+
+
+def get_valid_base_filename(filename: str, content_type: str) -> str:
+    """Avoid forbidden filenames that could confuse hugo"""
+    if filename in constants.CONTENT_FILENAMES_FORBIDDEN:
+        return f"{filename}-{content_type}"
+    return filename
--- a/gdrive_sync/api.py
+++ b/gdrive_sync/api.py
@@ -31,6 +31,7 @@ from videos.constants import VideoStatus
 from videos.models import Video
 from websites.api import get_valid_new_filename
 from websites.constants import (
+    CONTENT_TYPE_RESOURCE,
     RESOURCE_TYPE_DOCUMENT,
     RESOURCE_TYPE_IMAGE,
     RESOURCE_TYPE_OTHER,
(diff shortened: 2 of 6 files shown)


websites/utils.py
```diff
@@ -40,3 +40,13 @@ def set_dict_field(obj: Dict, field_path: str, value: Any):
 def get_dict_query_field(dict_field_name: str, sub_field: str):
     """Generate django query key for searching a nested json feild"""
     return dict_field_name + "__" + sub_field.replace(".", "__")
+
+
+def get_valid_base_filename(filename: str, content_type: str) -> str:
+    """Avoid forbidden filenames that could confuse hugo"""
+    if (
+        content_type in (constants.CONTENT_TYPE_PAGE, constants.CONTENT_TYPE_RESOURCE)
+        and filename in constants.CONTENT_FILENAMES_FORBIDDEN
```
alicewriteswrongs, line 49: is it possible we want to do this for other content types too?

alicewriteswrongs (changes requested):
tested things out and it seems to work correctly, just had one question

websites/utils.py
mbertrand: @gumaerc ?

gumaerc: It should cover anything that goes into the `content` folder.  Right now, "Pages" and "Resources" cover all bases but in the future if we add additional types of content this will have to cover that.  In addition, in the future if we add explicit support for nested content (`folders with `_index.md` and child content) or [Page Bundles](https://gohugo.io/content-management/page-bundles/) (folders with `index.md` and static content alongside it) then we will want to intentionally publish these files.  However, there's no scenario where user-generated content should place these files unintentionally.

alicewriteswrongs: We also have other types of content in the ocw-www course where I imagine we'd also want this protection

alicewriteswrongs (approved):
LGTM! 👍

PR #861, opened 2021-12-09
[proposal] try using swc for building assets

alicewriteswrongs:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

none

#### What's this PR do?

This is a dev tooling change which I want to propose and see what everyone things. It basically swaps out babel and the typescript compiler in our build / test toolchain for [swc](https://swc.rs/), which is a series of tools built in Rust for compiling and bundling web things. Because the tools are written in Rust and compiled to native code they are generally much faster than tools written in JS (like babel, the typescript compiler, etc) so our builds and tests run quite a bit quickly.

In my testing I observed the tests running in a little under half the time, and CI builds seem to go from ~4 minutes to ~2.5. Nice!

Test a few things:

- the build should be passing!
- you should be able to run tests locally
- the local dev build (in the `watch` container) should work without issue

My thought is that it would be good to get some feedback on this. I like the speed increase, but I'm not 100% sure we want to adopt it. What do y'all think?


mbertrand:
I tried it out, it does run tests faster, seems to be regularly updated and popular on github so unlikely to disappear or go unmaintained anytime soon.  Seems nice overall.

Maybe one cause for concern is if we someday need a plugin that's available for babel but not swc, not sure how likely that is.

I read [this recent article about swc](https://datastation.multiprocess.io/blog/2021-11-13-benchmarking-esbuild-swc-typescript-babel.html) which was overall positive but mentioned that its error reporting isn't very good:
>SWC's error messages need some love; add an extra comma in .swcrc and it blows up without any indiciation where the problem is.

I tested this out by adding an extra comma, and got a detailed error message about it.  So error reporting already seems to have improved since this article was written.

alicewriteswrongs:
I think we're probably not likely to need babel plugins in the future, as we've mostly moved away from them already in newer projects and we tend to just write standard TypeScript nowadays (and I think we should do that on all new projects going forward).

mbertrand (approved):
👍

PR #865, opened 2021-12-10
remove an unused JS dependency

alicewriteswrongs:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

none

#### What's this PR do?

This removes a dependency on `redux-actions`, which we're not using in this project.

it would be weird if this broke anything...


mbertrand (approved):
👍

PR #869, opened 2021-12-13
split reordering UI off from RelationWidget into SortableSelect component

alicewriteswrongs:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

none

#### What's this PR do?

I'm doing this in preparation for work we want to refactor course collections. We want to add a new field type so that we can have course collections (i.e. content author defined collections of courses, like 'Python-based programming courses', 'Math courses with video lectures,' or something like that) be defined as content collections, like pages, resources, and the like, which live in a site like `ocw-www`.

For this we could add more functionality to our existing `RelationField` component, but I'm a little reluctant to do that because it's already quite a large, complicated component. Instead I think it makes sense to add a new `WebsiteCollectionField` field which will implement this functionality.

In order to do _that_, I want to pull out the portion of the current `RelationField` implementation which is concerned with reordering a list, so that we can reuse that UI for the `WebsiteCollectionField`. This PR basically does that, pulling that code out into a new `SortableSelect` component which we'll be able to depend on for both things.

Try out some fields which use the `RelationField` with the `cross-site` and `sortable` option. It should all work as before with no differences at all.

### DIFF
--- a/static/js/components/widgets/RelationField.test.tsx
+++ b/static/js/components/widgets/RelationField.test.tsx
@@ -1,7 +1,7 @@
 import React from "react"
 import { act } from "react-dom/test-utils"
 import R from "ramda"
-import sinon, { SinonStub } from "sinon"
+import { SinonStub } from "sinon"
 
 import RelationField from "./RelationField"
 import { debouncedFetch } from "../../lib/api/util"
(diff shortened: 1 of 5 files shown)


mbertrand (approved):
Looks great, everything still seems to work!  👍

PR #881, opened 2021-12-21
Use static version file as trigger for sites other than ocw-www

### LINKED ISSUE #880: Out-of-sync static js/css content on mass publish events triggered by a theme change
When a new ocw-hugo-themes tgz file is created, pipelines for all sites are triggered.  But only the `ocw-www` pipeline extracts the contents from that file into the static directory.  Any sites whose pipelines complete prior to that one will have broken js/css until the ocw-www pipeline runs.

A possible solution is to trigger only the ocw-www pipeline when a new themes tarball is created.  A new version/hash file can be included in the tarball and extracted to the `/static` folder on S3 by the ocw-www pipeline.  When that file changes, trigger all the other pipelines.

mbertrand:
Depends on https://github.com/mitodl/ol-infrastructure/pull/516

#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #880 

#### What's this PR do?
Modifies the concourse pipeline so that only the "ocw-www" site is triggered by a a new themes tarball, and other sites should be triggered by a new version file in the static folder uploaded to S3 by the ocw-www pipeline.

This is a bit tricky.  I've already updated the `ocw-www` pipeline and one site pipeline (`mb-local-1-2`) to use the modified pipeline definition.

You can trigger a new build of the release-candidate themes pipeline, which should include a new `version.<dt-num>.txt` file in the generated tarball.  That should still trigger the ocw-www build, but it will also trigger all the unmodified sites too.  Once the ocw-www build finishes, a `mb-local-1-2` build should be triggered within a minute or two.

If you want to avoid triggering all. sites, you can try this:
- Manually start another run of the ocw-www pipeline from the UI
- Soon after it completes, start a shell in that build:
    `fly -t rc intercept -j draft/site:ocw-www/build-ocw-site  -b <latest build num> -s copy-s3-buckets  --team ocw`
- Run the following to create a new version file and copy to the static s3 folder:
   ```
   dt=$(date +%s)
   echo $dt >  ocw-hugo-themes/theme/base-theme/dist/static/version.$dt.txt
   aws s3 cp ocw-hugo-themes/theme/base-theme/dist/static/version.$dt.txt  s3://ocw-content-draft-qa/static/version.$dt.txt
   ```
- Keep an eye on the draft pipeline for `mb-local-1-2`, it should start a new build shortly.

All pipelines should be updated on RC and production once this change makes it to each environment.
```manage.py backpopulate_pipelines```

### DIFF
--- a/content_sync/pipelines/concourse.py
+++ b/content_sync/pipelines/concourse.py
@@ -152,7 +152,14 @@ class ConcourseGithubPipeline(BaseSyncPipeline):
 
         site_config = SiteConfig(self.website.starter.config)
         site_url = f"{site_config.root_url_path}/{self.website.name}".strip("/")
-        base_url = "" if self.website.name == settings.ROOT_WEBSITE_NAME else site_url
+        if self.website.name == settings.ROOT_WEBSITE_NAME:
+            base_url = ""
+            themes_trigger = "true"
+            js_trigger = "false"
+        else:
+            base_url = site_url
+            themes_trigger = "false"
+            js_trigger = "true"
         purge_header = (
             ""
             if settings.CONCOURSE_HARD_PURGE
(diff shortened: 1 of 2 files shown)


gumaerc (approved):
:+1: I was able to test this and it works well.  I got an error in the build of your test site, but I assume that that has to do with the pipeline being created from your local instance of `ocw-studio`.The code and strategy looks good, so feel free to merge but I did have one thought about the variable naming, take it or leave it.  The variable names are `themes-trigger` and `js-trigger`, but really what they represent is "trigger on publishing the zip" and "trigger on publishing the contents of the zip" respectively.  Maybe something like `theme-created-trigger` and `theme-deployed-trigger`?

PR #888, opened 2022-01-04
add support for 'website collection' field

alicewriteswrongs:
#### Pre-Flight checklist

- [x] Screenshots and design review for any changes that affect layout or styling
  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?

closes #886 

#### What's this PR do?

This PR adds a new type of site content field called a `WebsiteCollection`. This is designed to be a reorderable list of other websites in the application, which can be used to support something along the lines of the course collection feature that exists on current OCW prod. Under the hood it uses the `SortableSelect` component I introduced in #869.

#### How should this be manually tested?

I modified the checked-in ocw-www configuration, so if you have a site configured to use that starter you should be able to just `override_site_configs` and then see the relevant changes pop up. Basically, there will now be a `Course Collections` resource, and editing them should work like in the video below. You should be able to search for sites in the select box, add them, remove them, etc.

#### Screenshots (if appropriate)

this shows some basic use of the field:

https://user-images.githubusercontent.com/6207644/148134197-2c6bbb6f-3830-4a9d-a0bd-9f92b0b2bc8d.mov

PR #890, opened 2022-01-05
remove legacy website collections implementation

### LINKED ISSUE #889: remove old website collections implementation
As part of #885 (re-implementing course/website collections as a piece of course content) we should circle back and remove the "legacy" code that I implemented for them before.

Right now there are `WebsiteCollection` and `WebsiteCollectionItem` models and associated APIs and views and whatnot. On the frontend there is query config code, UI components, forms, etc. We should just remove all of that.

alicewriteswrongs:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

closes #889 

#### What's this PR do?

Before the new 'Website Collection' field type was added in #888 we had a separate implementation of website collections which didn't save them as course content but instead using separate `WebsiteCollection` and `WebsiteCollectionItem` models. Since we're not going to continue with that approach we might as well delete the code!

All other stuff on the app should work just as it did before.


gumaerc (approved):
:+1:

PR #895, opened 2022-01-10
switch from storing website UUID to the website name property

alicewriteswrongs:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

none

#### What's this PR do?

I thought for course collections that storing a website's UUID would be what we need to then be able to render websites in a list in ocw-www, but I goofed and missed that actually the `name` property is what's going to end up being used in the URL to the `data.json` file for each course at render-time, so that's actually the information we need.

Anyway - this PR just makes a small adjustment so that the website `name` is saved in the course collection array rather than the `uuid`.

Course collections should look as normal. If you look at the JSON for one you should see website names rather than UUIDs. The UX should be identical.


abeglova (approved):
👍

Issue #896, opened 2022-01-11
theme asset extraction step in course pipeline fails on retry


gumaerc:
### Steps to Reproduce

This is tough to reproduce reliably, Concourse has to be in an overloaded state where it will retry the step in `site-pipeline.yml` that extracts the theme assets:

```
          run:
            dir: course-markdown
            path: sh
            args:
            - -exc
            - |
              mkdir ../ocw-hugo-themes/theme
              tar -xvzf ../ocw-hugo-themes/ocw-hugo-themes-*.tgz -C ../ocw-hugo-themes/theme
              hugo --config ../ocw-hugo-projects/((config-slug))/config.yaml --baseUrl /((base-url)) --themesDir ../ocw-hugo-themes/theme
```

### Expected Behavior

The theme assets are extracted, overwriting anything that already exists

### Actual Behavior

After the first attempt, the `ocw-hugo-themes/theme` directory has already been created and running `mkdir` on it causes an error.

### Stacktrace

`mkdir: cannot create directory '../ocw-hugo-themes/theme': File exists`

PR #900, opened 2022-01-12
add -p to mkdir command before theme asset extraction

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/896

#### What's this PR do?
This PR adds the `-p` argument to a `mkdir` command in the site pipeline definition that's run before extracting the theme assets tarball.  It solves an issue where between retries the directory already exists so an error is thrown without the `-p` argument.

I can't really think of a great way to manually test this change, as the original issue only happens when Concourse is overloaded and specifically retries the `build-course-task` step.  I think it will just have to be deployed to RC and then have a mass publish triggered, watching the latest builds to see if the error in the original issue comes up again.

### DIFF
--- a/content_sync/pipelines/definitions/concourse/site-pipeline.yml
+++ b/content_sync/pipelines/definitions/concourse/site-pipeline.yml
@@ -108,7 +108,7 @@ jobs:
             args:
             - -exc
             - |
-              mkdir ../ocw-hugo-themes/theme
+              mkdir -p ../ocw-hugo-themes/theme
               tar -xvzf ../ocw-hugo-themes/ocw-hugo-themes-*.tgz -C ../ocw-hugo-themes/theme
               hugo --config ../ocw-hugo-projects/((config-slug))/config.yaml --baseUrl /((base-url)) --themesDir ../ocw-hugo-themes/theme
         on_failure:


alicewriteswrongs (approved):
looks straightforward enough, and I can't imagine a scenario in which adding `-p` causes new issues so even without testing I think this looks good

content_sync/pipelines/definitions/concourse/site-pipeline.yml
```diff
@@ -108,7 +108,7 @@ jobs:
             args:
             - -exc
             - |
-              mkdir ../ocw-hugo-themes/theme
+              mkdir -p ../ocw-hugo-themes/theme
               tar -xvzf ../ocw-hugo-themes/ocw-hugo-themes-*.tgz -C ../ocw-hugo-themes/theme
```
alicewriteswrongs, line 112: I'm wondering though, does this re-running create the possibility of some 'leakyness,' where some files could be in `../ocw-hugo-themes/theme` from an older theme? like for instance if a file is changed the unpacking the `.tgz` archive will overwrite it, but if a file were delete (and not present in the newer theme) then we wouldn't overwrite it with anything

maybe this isn't relevant at all given how the builds are triggered (i.e. there's no possibility of a retry pulling in a newer version of the theme than the original 'try') but I don't think I fully grok how the concourse stuff gets triggered and whatnot.

so anyhow - maybe not anything to worry about, but just something I thought I'd raise

gumaerc: You're right to be concerned, but this situation only happens during "retries."  The version of the theme that a given pipeline run is pulling in is set at the beginning of the build.  In the PR I noted above, Matt added `attempts: 3` to each step which means if a step in the pipeline times out or fails for any reason, it will retry it with the same inputs.  This is to cover cases of timeouts looking for workers or API outages, etc.  When a new build is started, all fresh docker containers are built.

alicewriteswrongs: ok sounds good!

PR #904, opened 2022-01-18
add 'published' filter to website listing API

### LINKED ISSUE #887: filter website API calls based on published status
As part of #885 we want to ensure that websites which are added to a website collection are published. To do so we'll need to add support for filtering the API results to only include publishes sites, via a query param or something. We should, at the same time, add an option to the `useWebsiteSelectOptions` hook to control that param and then set it accordingly in the `WebsiteCollectionField`.

alicewriteswrongs:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #887

#### What's this PR do?

This PR adds a query param (`published`) to the website API which accepts a boolean value and will filter accordingly on whether websites have already been published or not.

It also makes a change to the `WebsiteCollection` widget to only show published websites in it, ensuring that course content authors can't accidentally create a course collection which references unpublished courses.

You'll need to have both unpublished and published websites locally. Then confirm that if you're making a course collection in `ocw-www` you can only see the published ones as options.

### DIFF
--- a/websites/views.py
+++ b/websites/views.py
@@ -90,6 +90,7 @@ class WebsiteViewSet(
         website_type = self.request.query_params.get("type", None)
         search = self.request.query_params.get("search", None)
         resourcetype = self.request.query_params.get("resourcetype", None)
+        published = self.request.query_params.get("published", None)
 
         user = self.request.user
         if self.request.user.is_anonymous:
@@ -116,6 +117,11 @@ class WebsiteViewSet(
 
         if website_type is not None:
             queryset = queryset.filter(starter__slug=website_type)
+
+        if published is not None:
+            published = _parse_bool(published)
+            queryset = queryset.filter(publish_date__isnull=not published)
+
         return queryset.select_related("starter").order_by(ordering)
 
     def get_serializer_class(self):
--- a/static/js/components/widgets/WebsiteCollectionField.test.tsx
+++ b/static/js/components/widgets/WebsiteCollectionField.test.tsx
@@ -42,6 +42,11 @@ describe("WebsiteCollectionField", () => {
     helper.cleanup()
   })
 
+  it("should pass published=true to the useWebsiteSelectOptions", async () => {
+    await render()
+    expect(useWebsiteSelectOptions).toBeCalledWith("name", true, true)
+  })
+
   it("should pass things down to SortableSelect", async () => {
     const value = websites.map(website => ({
       id:    website.name,
--- a/static/js/components/widgets/WebsiteCollectionField.tsx
+++ b/static/js/components/widgets/WebsiteCollectionField.tsx
@@ -35,7 +35,7 @@ export default function WebsiteCollectionField(props: Props): JSX.Element {
     [onChange, websiteMap, name]
   )
 
-  const { options, loadOptions } = useWebsiteSelectOptions("name")
+  const { options, loadOptions } = useWebsiteSelectOptions("name", true, true)
 
   useEffect(() => {
     setWebsiteMap(cur => {
--- a/static/js/hooks/websites.test.ts
+++ b/static/js/hooks/websites.test.ts
@@ -52,6 +52,28 @@ describe("website hooks", () => {
       expect(debouncedFetch).toBeCalledTimes(0)
     })
 
+    //
+    ;[true, false].forEach(published => {
+      it(`should set published=${String(
+        published
+      )} if you pass the option`, async () => {
+        const { waitForNextUpdate } = renderHook(() =>
+          useWebsiteSelectOptions("uuid", true, published)
+        )
+        await act(async () => {
+          await waitForNextUpdate()
+        })
+        expect(global.fetch).toBeCalledWith(
+          siteApiListingUrl
+            .query({ offset: 0 })
+            .param({ search: "" })
+            .param({ published })
+            .toString(),
+          { credentials: "include" }
+        )
+      })
+    })
+
     it("should let you issue a debounced request with a search param if you pass a callback", async () => {
       const { result, waitForNextUpdate } = renderHook(useWebsiteSelectOptions)
       const cb = jest.fn()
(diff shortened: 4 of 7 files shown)


websites/views.py
```diff
@@ -116,6 +117,11 @@ def get_queryset(self):
 
         if website_type is not None:
             queryset = queryset.filter(starter__slug=website_type)
+
+        if published is not None:
+            published = _parse_bool(published)
+            queryset = queryset.filter(publish_date__isnull=not published)
```
gumaerc, line 123: So I have some concerns here about using `publish_date` as the mechanism for this functionality.  Currently when courses are imported, `publish_date` is brought in from the imported course.  This means that the value will reflect when the site was "published" according to the legacy Plone system.  We should either change this so that we import that as something like `legacy_publish_date` and ensure `publish_date` only refers to when it was published by `ocw-studio`, or use another property to check when the site was last published to Github.  I think in order to be referenced in a relation widget, we should also be checking this for both draft and live, so maybe we should `latest_build_id_draft` and `latest_build_id_live` and only display the result if both of those are not null?  What do you think, @mbertrand ?

gumaerc (changes requested):
I have a concern about how this is implemented:

websites/views.py
alicewriteswrongs: sure, that sounds straightforward enough

gumaerc: Thinking about this some more, it might not be necessary to check `latest_build_id_draft` because `STATIC_API_BASE_URL` should be set to `https://ocwnext-rc.odl.mit.edu/` in RC and `https://ocwnext.odl.mit.edu` in Production, which would mean that both draft and live in RC for example would hit `https://ocwnext-rc.odl.mit.edu` with `getJSON`.

alicewriteswrongs: I was mislead, btw, by our website factory, which if you ask for a published or unpublished website only varies the `publish_date` property, so I'll change that a bit to reflect this

mbertrand: Are there any imported sites that have a `publish_date` but might not be actually published via concourse by the time this goes live?   If so, then it would make sense to add a separate `legacy_publish_date` for imported sites and upsert the `publish_date` field only when published via concourse.  If not, then I think it should be okay as is, though the additional field would provide clarification and might be worth doing anyway.  The `id` fields might not be a good indicator because those specific builds might have failed?

gumaerc: @mbertrand That's a good point that I hadn't actually thought about, the builds could indeed have failed.  However, at least you know that if the site has been published at some point that the expected JSON file will exist.  When it comes to using that data in `ocw-hugo-themes`, it's up to the template itself to have logic to sanity check the data.  The main issue is when the file doesn't exist at all because Hugo throws a hard error if it gets anything other than a 200 when using `getJSON`.  I think the most ideal route here is to change `import_ocw_course_sites` to store the legacy publish date into a separate field and keep `publish_date` reserved for "when was the site last published live by `ocw-studio."  However, we would need to make that change likely in a separate PR first, merge it before this one and make sure all courses are re-imported before this is merged.  Said PR would also need to include a migration to make sure `publish_date` is null where it's supposed to be for imported courses that haven't been published yet.

alicewriteswrongs: am I understanding correctly that based on the data we currently store in the database there is not a foolproof way to determine whether a course has already been published through concourse?

I think we could determine the following things, correct me if I'm wrong:

1. If there is no `latest_build_id_live` (or `..._draft`) then the site has _not_ been published through concourse
2. if there is a value for `latest_build_id_{draft|live}`, then:
    1. if `{live|draft}_publish_status` is `PUBLISH_STATUS_SUCCEEDED`, then the course was published! 🎉 
    2. if `{live|draft}_publish_status` is _not_ `PUBLISH_STATUS_SUCCEEDED`, then we know that the _most recent_ build failed, aborted, is pending, etc. But just looking around a little bit it seems like we don't have a way of knowing whether a _previous_ build has ever succeeded - another way to say that is that I don't see anywhere where we're storing the build history. If we were then I think we could easily just check to see if there has ever been a successful build of the site.

I'm wondering in particular about the 2-ii case - is it true that we can't determine the status of previous attempts to publish this course?

if that's correct, then I think the suggestion to add `legacy_publish_date` makes sense, so that then we can have a property (`publish_date`) which serves as an unambiguous record to indicate that _yes, this has been published_.

About the legacy data - is there anything we need it for? If not, would it also be possible to just drop that info on the floor? the publish date on the plone CMS may not be important, I don't know.

In any case, my assumption reading the code and not knowing about the whole publish pipeline was that I'd be able to rely on a property like `publish_date` to tell me when something had been published through our system. If that value doesn't represent that, then I think we should refactor it so that it does store that.

gumaerc: @pdpinch What do you think about this?  Should we be importing `publish_date` from the Plone data at all?  Should we preserve it as `legacy_publish_date`?

gumaerc: I spoke with @pdpinch about this offline and here's the plan:

1. Alter `import_ocw_course_sites` to store `publishdate` from the `ocw-to-hugo` data as a new field called `first_published_to_production` (this is what it's called in the original source data) instead of in `publish_date`
2. On first publish of a course in `ocw-studio`, set both `first_published_to_production` and `publish_date`, then on subsequent publishes only set `publish_date`

gumaerc: I made an issue about this here: https://github.com/mitodl/ocw-studio/issues/916

Once it's addressed, we should be able to simply merge this as-is.  Until then, I think we can consider this blocked.

alicewriteswrongs:
@gumaerc now that #916 is addressed I think this should be ready for review again

gumaerc (approved):
:+1: LGTM

The changes work well.  I had to use postman to fake a "succeeded" publish locally, but once I made that webhook call to mark a course as published I was able to add it to a collection and publish `ocw-www`, seeing it there.

PR #909, opened 2022-01-19
Throttle batch management commands that invoke the git API

### LINKED ISSUE #905: Git API rate limit needs to be checked wherever/whenever appropriate
### Steps to Reproduce

Run `backpopulate_pipelines` (with `--create_backend` option) or `mass_publish` command on all sites on production

### Expected Behavior

The command eventually completes.

### Actual Behavior

The command eventually fails because production uses a non-enterprise github organization with limited hourly API calls, and that rate limit is quickly exceeded.

### Potential solution

- Add a setting specifying if there is a git api rate limit (`GITHUB_CHECK_RATE_LIMIT`) that should be checked (`False` on RC, `True` on prod).
- If `True`, check that limit, and throttle calls if necessary, wherever appropriate, such as during `mass_publish` command execution. 
- Use very large batch sizes (equal to site count?) when running `mass_publish` or `backpopulate_pipelines --create_backend` if rate limits are in place.

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested
- [x] Settings

Closes #905

#### What's this PR do?
Adds a `throttle_git_backend_calls` function that is used in various website batch tasks:
- sync_unsynced_websites
- publish_website_batch
- upsert_website_pipeline_batch

This function will check the remaining git api calls available (if `settings.GITHUB_RATE_LIMIT_CHECK=True`), and if it is below a certain threshold (defined by `settings.GITHUB_RATE_LIMIT_CUTOFF`) then sleep until the rate limit is reset. 

Set the following in your .env file:
```
GITHUB_RATE_LIMIT_CHECK=True
OCW_STUDIO_LOG_LEVEL=DEBUG
```

You can start running the `mass_publish` command (or `sync_unsynced_websites.delay(create_backends=True)` in a shell ) and look at the logs - search for `Remaining github calls`.  Eventually, if the remaining requests drops below 100, execution should pause until the limit is reset.  

If running `mass_publish`, it might be safer to use a larger-than-default chunk size (for instance, 2000 instead of 500).

Related: https://github.com/mitodl/salt-ops/pull/1493

### DIFF
--- a/app.json
+++ b/app.json
@@ -147,6 +147,14 @@
       "description": "Google analytics tracking ID",
       "required": false
     },
+    "GITHUB_RATE_LIMIT_CHECK": {
+      "description": "True if the github domain has API rate limits",
+      "required": false
+    },
+    "GITHUB_RATE_LIMIT_CUTOFF": {
+      "description": "Number of remaining Github API calls that triggers throttling",
+      "required": false
+    },
     "GITHUB_WEBHOOK_BRANCH": {
       "description": "Github branch to filter webhook requests against",
       "required": false
--- a/content_sync/api.py
+++ b/content_sync/api.py
@@ -1,13 +1,17 @@
 """ Syncing API """
 import logging
+from datetime import datetime
+from time import sleep
 from typing import List, Optional
 
+import pytz
 from django.conf import settings
 from django.utils.module_loading import import_string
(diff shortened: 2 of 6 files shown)


gumaerc (approved):
:+1:

Issue #902, opened 2022-01-13
Linking to pages within a site


pdpinch:
In PR we added the link button to the WYSIWYG editor so users can link from markdown fields to resources 

#### Designs and Mockups

<img width="672" alt="image" src="https://user-images.githubusercontent.com/430126/149414364-79f18959-2aa3-462e-89af-673079f5ff7f.png">

#### Acceptance Criteria:

- [ ] Add a tab "Pages" to the pop-up for linking resources
- [ ] Populate the tab with a list of pages on the site 
- [ ] Allow the user to filter the list of pages
- [ ] When a page is chosen and the user clicks the link button, write a resource short code into the editor
- [ ] Since pages aren't technically resources, we should update the UI to be more generic, changing "Link Resource" to "Add LInk" 

#### Questions

- Should the short code be a resource short code, or something else?

#### Out of Scope

- Add a switch for selecting pages to link to from anywhere in OCW Studio, not just the current site

ChristopherChudzicki:
![](https://github.trello.services/images/mini-trello-icon.png) [Chris Chudzicki](https://trello.com/c/lOAcmyzC/8664-chris-chudzicki)

ChristopherChudzicki:
Closed by #915

PR #915, opened 2022-01-20
Link to page within course

ChristopherChudzicki:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

#### What's this PR do?
This PR enables authors to create links to pages in their course through the markdown editor..
- "Link resource" has been renamed to "Add link". This dialog increases a fourth "Pages" tab.
- "Embed resource" is unchanged (it still has only 3 tabs).

In a course with multiple pages:
1. Edit a page, then "Add link". Select the pages tab. Check filter works. Select a page. Click "Add Link".
2. Check link works in published course. 

#### Where should reviewer start
I suggest starting with `static/js/components/widgets/ResourcePickerDialog.tsx`, and the associated github comments.

#### Out of scope
Pages tab does not have pagination: It shows at most 10 pages. This is consistent with the other tabs (documents, images, videos).

#### Screenshots 
link_to_page

### DIFF
--- a/static/js/components/widgets/ResourcePickerDialog.tsx
+++ b/static/js/components/widgets/ResourcePickerDialog.tsx
@@ -5,42 +5,82 @@ import Dialog from "../Dialog"
 import {
   RESOURCE_TYPE_DOCUMENT,
   RESOURCE_TYPE_IMAGE,
-  RESOURCE_TYPE_VIDEO
+  RESOURCE_TYPE_VIDEO,
+  CONTENT_TYPE_RESOURCE,
+  CONTENT_TYPE_PAGE
 } from "../../constants"
 import ResourcePickerListing from "./ResourcePickerListing"
 import { useDebouncedState } from "../../hooks/state"
 import {
   CKEResourceNodeType,
-  ResourceDialogState,
-  RESOURCE_EMBED
+  ResourceDialogMode,
+  RESOURCE_EMBED,
+  RESOURCE_LINK
 } from "../../lib/ckeditor/plugins/constants"
 import { WebsiteContent } from "../../types/websites"
 
 interface Props {
-  state: ResourceDialogState
+  mode: ResourceDialogMode
+  isOpen: boolean
   closeDialog: () => void
   insertEmbed: (id: string, title: string, variant: CKEResourceNodeType) => void
-  attach: string
 }
 
-const RESOURCE_PICKER_TABS = [
+interface ResourceTabSettings {
+  title: string
+  id: string
+  contentType: "resource" | "page"
+  resourcetype: string | null
+  embeddable: boolean
+}
+const RESOURCE_PICKER_TABS: ResourceTabSettings[] = [
   {
     title:        "Documents",
-    resourcetype: RESOURCE_TYPE_DOCUMENT
+    id:           RESOURCE_TYPE_DOCUMENT,
+    contentType:  CONTENT_TYPE_RESOURCE,
+    resourcetype: RESOURCE_TYPE_DOCUMENT,
+    embeddable:   true
   },
   {
     title:        "Videos",
-    resourcetype: RESOURCE_TYPE_VIDEO
+    id:           RESOURCE_TYPE_VIDEO,
+    contentType:  CONTENT_TYPE_RESOURCE,
+    resourcetype: RESOURCE_TYPE_VIDEO,
+    embeddable:   true
   },
   {
     title:        "Images",
-    resourcetype: RESOURCE_TYPE_IMAGE
+    id:           RESOURCE_TYPE_IMAGE,
+    contentType:  CONTENT_TYPE_RESOURCE,
+    resourcetype: RESOURCE_TYPE_IMAGE,
+    embeddable:   true
+  },
+  {
+    title:        "Pages",
+    id:           CONTENT_TYPE_PAGE,
+    contentType:  CONTENT_TYPE_PAGE,
+    resourcetype: null,
+    embeddable:   false
   }
 ]
 
-export default function ResourcePickerDialog(props: Props): JSX.Element {
-  const { state, closeDialog, insertEmbed, attach } = props
+const modeText = {
+  [RESOURCE_EMBED]: {
+    headerText: "Resources",
+    acceptText: "Embed resource"
+  },
+  [RESOURCE_LINK]: {
+    headerText: "Resources & Pages",
+    acceptText: "Add link"
+  }
+}
 
+export default function ResourcePickerDialog(props: Props): JSX.Element {
+  const { mode, isOpen, closeDialog, insertEmbed } = props
+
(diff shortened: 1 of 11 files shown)


static/js/components/widgets/ResourcePickerDialog.tsx
```diff
@@ -5,42 +5,82 @@ import Dialog from "../Dialog"
...
+  ResourceDialogMode,
+  RESOURCE_EMBED,
+  RESOURCE_LINK
 } from "../../lib/ckeditor/plugins/constants"
 import { WebsiteContent } from "../../types/websites"
 
 interface Props {
-  state: ResourceDialogState
+  mode: ResourceDialogMode
+  isOpen: boolean
```
ChristopherChudzicki, line 24: Previously the state variable had values roughly `["link", "embed", "closed"]. Using a single prop for both "content/behavior" state and "open/closed" state meant we could not close the modal without changing its content. See gif below (which is from rc). This was a pretty minor visual bug but became more noticeable with the addition of "Pages" to the Link dialog variant (since now the Link/Embed variants are more different).

modal_contents_change

static/js/components/widgets/ResourcePickerDialog.tsx
```diff
@@ -5,42 +5,82 @@ import Dialog from "../Dialog"
...
 
-const RESOURCE_PICKER_TABS = [
+interface ResourceTabSettings {
+  title: string
+  id: string
+  contentType: "resource" | "page"
+  resourcetype: string | null
+  canEmbed: boolean
+}
+const RESOURCE_PICKER_TABS: ResourceTabSettings[] = [
```
ChristopherChudzicki, line 36: Basically all the functionality for this feature was already present, we just needed a way for the tab to differentiate whether it should make API calls with `type=resource` or `type=page`. This array controls those settings for the tabs.

ChristopherChudzicki:
@alicewriteswrongs or anyone: There was a note in the issue (#902):

> Should the short code be a resource short code, or something else?

This PR is using the resource short codes. I don't have a great sense of the consequences of using resource short codes "or something else". But the resource short codes were working automatically.

ChristopherChudzicki:
[Chris Chudzicki](https://trello.com/c/r0dRJClA/8665-chris-chudzicki)

ChristopherChudzicki:
[Chris Chudzicki](https://trello.com/c/r0dRJClA/8665-chris-chudzicki)

alicewriteswrongs:
@ChristopherChudzicki so yeah, for this we want to just use the `resource_link` shortcode, it is now slightly mis-named (should be more of a `content_link` or something like that) but I think it's ok to not worry too much about that 🙂

static/js/components/widgets/ResourcePickerDialog.tsx
alicewriteswrongs: one other option for resolving the funny UI flash could be to hide the button in the case that the user has already clicked 'cancel'...which, come to think of it, we might want to do anyway.

Thinking about this made me wonder if it was possible to still click the 'link resource' button _after_ clicking 'cancel' and while you have to have pretty good timing I was able to do so. So we might want to hide the 'submit' button after cancel is hit.

what do you think?

static/js/components/widgets/ResourcePickerListing.test.tsx
```diff
@@ -156,6 +174,28 @@ describe("ResourcePickerListing", () => {
...
+  it("should fetch and display pages", async () => {
+    const { wrapper } = await render({
+      contentType:  "page",
+      resourcetype: null
+    })
+    expect(
+      wrapper
+        .find(".resource-picker-listing .resource-item")
+        .map(el => el.find("h4").text())
+    ).toEqual(contentListingItems[1].map(item => item.title))
```
alicewriteswrongs, line 186: here it might be good to add an assertion that the API request we expected went through, because I think looking at the API mocks up top we're returning `contentListingItems[1]` for both document and page requests. An alternative I guess would be to add a new, separate return value just for the pages, either way would disambiguate it I think

static/js/components/widgets/ResourcePickerDialog.tsx
```diff
@@ -5,42 +5,82 @@ import Dialog from "../Dialog"
...
+    headerText: "Resources & Pages",
+    acceptText: "Add link"
+  }
+}
+
 export default function ResourcePickerDialog(props: Props): JSX.Element {
-  const { state, closeDialog, insertEmbed, attach } = props
+  const { mode, isOpen, closeDialog, insertEmbed } = props
 
+  const pickerTabs = RESOURCE_PICKER_TABS.filter(
```
alicewriteswrongs, line 81: small naming comment - maybe `activeTabs`, `currentTabs`, something to that effect? up to you, I just always dither over naming things

static/js/components/widgets/ResourcePickerDialog.tsx
```diff
@@ -5,42 +5,82 @@ import Dialog from "../Dialog"
...
+  canEmbed: boolean
+}
+const RESOURCE_PICKER_TABS: ResourceTabSettings[] = [
   {
     title:        "Documents",
-    resourcetype: RESOURCE_TYPE_DOCUMENT
+    id:           RESOURCE_TYPE_DOCUMENT,
+    contentType:  CONTENT_TYPE_RESOURCE,
+    resourcetype: RESOURCE_TYPE_DOCUMENT,
+    canEmbed:     true
```
alicewriteswrongs, line 42: possibly `embeddable`?

alicewriteswrongs (changes requested):
tested it out, and functionality looks great! I can link pages to my heart's content.

just had a few comments / questions on the code, nothing too serious

static/js/components/widgets/ResourcePickerDialog.tsx
ChristopherChudzicki: BTW: Your question about whether the button should disappear when clicking close made me wonder how other sites handle this, so I spent a few minutes looking. I checked a few sites, of which only Linkedin and Trello had modals thats. Both Linkedin and Trello make their modals disappear immediately (no transition). Now I will be on the lookout for sites with transitiony modals, and how they handle this. 🤷 

But... getting rid of the transition definitely stuff like this easier to avoid, so that could be another possibility.

alicewriteswrongs: yeah, I think I'd prefer no transition but we might be a bit stuck with it

it might be a good idea to clear the selection when the tab changes, but I don't think we need to do so in this PR

and yeah, that makes sense, I think we can just leave the state as is

alicewriteswrongs (approved):
LGTM! Congrats on your first PR! 🎉

ChristopherChudzicki:
[Chris Chudzicki](https://trello.com/c/W8h7BkWL/8756-chris-chudzicki)

Issue #455, opened 2021-08-03
Upgrade local version of postgres to 12.x


noisecapella:
As a developer, I'd like to have a version of the database which matches the version being used on production (12.5) so I can be confident that there aren't any differences between the versions

Probably best to wait until after MVP since this could be a little disruptive to people's local environments

#### Acceptance Criteria:

- [ ] Set up docker-compose and other parts of our development application stack to run postgres 12.x without issues

#### Notes
I did this already because I thought there was an issue in our migrations being different on local and production, though it turned out to be was unrelated to the postgres version. Here's what I did to upgrade:

1. Add these lines to `Dockerfile`:

    ```sh
    # Add repository to get postgres-client-12
    RUN sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt buster-pgdg main" > /etc/apt/sources.list.d/pgdg.list'
    RUN wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -
    ```
2. Replace `postgres-client` with `postgres-client-12` in `apt.txt`
3. Update `docker-compose.yml` to add a password to the database string: `DATABASE_URL: postgres://postgres:postgres@db:5432/postgres`
4. And update the `db` section too:
    ```yaml
    image: postgres:12.5
    environment:
      POSTGRES_PASSWORD: postgres
    ```

PR #923, opened 2022-01-21
update docker postgres to 12.5

ChristopherChudzicki:
## Manual migration steps
If you upgrade the postgres container from 11 to 12 without migrating the database files, you'll get an error along the lines of
```
ocw-studio-db-1      | 2022-01-26 21:52:44.993 UTC [1] FATAL:  database files are incompatible with server
ocw-studio-db-1      | 2022-01-26 21:52:44.993 UTC [1] DETAIL:  The data directory was initialized by PostgreSQL version 11, which is not compatible with this version 12.8 (Debian 12.8-1.pgdg110+1).
```
Destroying the pg-11 database seemed like easiest upgrade path. So to keep your local data intact, we'll dump the data, then upgrade, then restore the data.

⚠️  Alert! Do not just dump these all into a shell script. If you do, step 6 will run before the database server starts in step 5, and then the data import will fail.

```sh
# 1. Using your old 11.6 database:
# (If you've already upgraded, just change the docker-compose file back to 11.6 for this step.)
docker-compose exec -T db pg_dumpall -h db -U postgres > dump.sql
# 2. stop all containers
docker-compose stop
# 3 destroy the old database container
docker-compose rm -sv db
# 4. build+start the new database container
docker-compose up db -d
# 5. restore the database from backup
docker-compose exec -T -e PGPASSWORD=postgres db psql -h db -U postgres < dump.sql
# 6. start the remaining containers
docker-compose up
# 7 verify postgres 12 is in use:
docker-compose exec -e PGPASSWORD=postgres db psql -h db -U postgres
```

#### Pre-Flight checklist

- [x] Testing

#### What's this PR do?
Update local postgres to version 12.8 in `db` container only. Does not update concourse-db postgres.

Follow migration steps at top of PR, then try using ocw-studio and observe that it works as expected.

### Notes
- Again, this does not update concourse-db postgres
- The original issue (https://github.com/mitodl/ocw-studio/issues/455) described 4 steps to upgrade. As far as I can tell, only steps 1 and 2 are currently unnecessary
- I am fairly new to docker, so hopefully the above is a reasonable way to do this...

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ x-environment:
   NODE_ENV: 'production'
   DEV_ENV: 'True'  # necessary to have nginx connect to web container
   SECRET_KEY: local_unsafe_key
-  DATABASE_URL: postgres://postgres@db:5432/postgres
+  DATABASE_URL: postgres://postgres:postgres@db:5432/postgres
   OCW_STUDIO_SECURE_SSL_REDIRECT: 'False'
   OCW_STUDIO_DB_DISABLE_SSL: 'True'
   CELERY_TASK_ALWAYS_EAGER: 'False'
@@ -19,9 +19,11 @@ x-environment:
 
 services:
   db:
-    image: postgres:11.6
+    image: postgres:12.8
     ports:
       - "5432"
+    environment:
+      POSTGRES_PASSWORD: postgres
 
   redis:
     image: redis:6.0.10


ChristopherChudzicki:
[Chris Chudzicki](https://trello.com/c/W8h7BkWL/8756-chris-chudzicki)

gumaerc:
So I tried this on my local database which had a full compliment of imported legacy courses.  The database import failed with a number of errors, too many to post in Github comments but the most common seemed to be `invalid command \N`.  It seems as if the imported content of the legacy courses is causing issues with the way the data is dumped / restored.  Stringified line breaks seem to be causing an issue and I'm not sure why or how we would resolve it.

pdpinch:
@mitodl/devops do we need to upgrade the postgres attached to ocw-studio in production too? Is this the way we would do it?

gumaerc:
@pdpinch The RC and prod instances of `ocw-studio` use RDS databases.  This PR simply upgrades the version of Postgres that the local Docker based setup uses and doesn't affect any of our Heroku setups.

gumaerc (approved):
:+1: LGTM

I gave this another shot from scratch.  It seems that the errors I encountered had to do with my database not being totally stopped when I went to do the restore.  I attempted to put all of the steps into a bash script and simply execute it, which resulted in the errors.  As long as you run each step one by one, this process works.  My database was successfully upgraded to 12.5 and the restore succeeded.

PR #924, opened 2022-01-21
update ocw import + concourse build time handling of `publish_date` field

### LINKED ISSUE #916: import_ocw_course_sites improperly sets publish_date
In `ocw-to-hugo`, the `publishdate` value in the course metadata is sourced from the `first_published_to_production` value from the original OCW source data.  This value is currently being set on `publish_date` on the `Website` object created during runs of `import_ocw_course_sites`.  This isn't desired, as `publish_date` is supposed to reflect when the given site was last published live by Concourse.

Instead of setting `publish_date` on import, we should:

 - Add a new nullable field on `Website` objects called `first_published_to_production`
 - During import, populate `first_published_to_production` with the `publishdate` value from the `ocw-to-hugo` course metadata
 - On first publish of courses in `ocw-studio`, if `first_published_to_production` is null, set it to the current date
 - Set `publish_date` on `Website` objects every time a site is published live

alicewriteswrongs:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

closes #916

#### What's this PR do?

This PR adjusts the role played by the `publish_date` field in a few ways.

1. Introduce a new field on the `Website` model called `first_published_to_production`. The migration to introduce this field also sets `publish_date` to `None` for all existing `Website` records
2. during the OCW import, instead of setting `Website.publish_date` to the value for `publishdate` on the ocw2hugo data, stick that value into the `first_published_to_production` field instead
3. when doing a Concourse build, update the `Website` record's `publish_date` and whatnot a bit more carefully. In particular, for both `draft` and `live`:
    1. only set `publish_date` to `datetime.now()` when the build finished successfully
    2. if a `Website` record's `.first_publish_to_production` field is `None`, then set that to `datetime.now()` as well

Together this should ensure that we have an unambiguous ability to tell, given the data we store, whether or not a course has been successfully published to production through our Concourse pipeline. More specifically, if a `Website` record has a non-null value for `publish_date` then it has been published through our system. We also, by storing the OCW legacy `publishdate` value preserve information about when those courses have been updated and whatnot.

I think it will be necessary to test out the migration, run a course import, and then test out publishing a course, to make sure that everything works as described above.


gumaerc (approved):
:+1: 

I was able to run the migration locally and it successfully wiped out the publish date on all my sites.  I wasn't able to manually verify that publishing a site successfully updates the publish date, as you can do the part where it publishes the content to Github locally but triggering a local Concourse pipeline and have it complete to the point where it will call back to your local instance of studio and report that is currently not possible.  Either way though, this looks good.  We should re-import / re-publish all courses after this merges.

Issue #917, opened 2022-01-21
issue with resource link regex


alicewriteswrongs:
### Steps to Reproduce

1. create two resource links on a single line (i.e. within a single paragraph) in the Markdown editor
2. save the page
3. re-open the content editor, and note that the first resource link has gobbled up the second (see screens below)

### Expected Behavior

links shouldn't cannibalize each other

### Screenshot or Screencast

1. create a paragraph with two links:
<img width="838" alt="Screen Shot 2022-01-21 at 9 06 46 AM" src="https://user-images.githubusercontent.com/6207644/150540966-fb33a09e-354a-4011-990a-3f5d4294b884.png">

2. it saves as the markdown we expect:
<img width="889" alt="Screen Shot 2022-01-21 at 9 09 12 AM" src="https://user-images.githubusercontent.com/6207644/150541018-643ddccb-b7fb-4538-8e42-783d4cc36762.png">

```md
a thing {{< resource_link 9f27dc7b-2d7e-f67f-fac9-9260d4ae99dc \"is linked\" >}} and then so is {{< resource_link c46db57c-9699-3827-ec0c-6590087f8ace \"another thing\" >}} ok great

1. sdf
2. asdf
3. asdf
```

3. re-open the editor, and this will happen :/

<img width="832" alt="Screen Shot 2022-01-21 at 9 10 29 AM" src="https://user-images.githubusercontent.com/6207644/150541183-14857664-d378-4b0e-94d8-4bd23b08a802.png">

pdpinch:
possibly the same issue as #903

ChristopherChudzicki:
![](https://github.trello.services/images/mini-trello-icon.png) [Chris Chudzicki](https://trello.com/c/W8h7BkWL/8756-chris-chudzicki)

Issue #903, opened 2022-01-13
The last link in a table gets corrupted after save and load


pdpinch:
WIP bug description

### Steps to Reproduce

1. Create a site 
2. Create a resource, like a PDF
3. Create a page
4. Create a table on the page with three rows
5. In the first row, add a link to the PDF Resource. Repeat for each row
6. Save the page
7. Open the page

### Expected Behavior

- When the page is opened, it should look the same as it did when it was saved, e.g. 

### Actual Behavior

- When the page is opened, the code for the last link is visible, instead of displaying as a link. 

### Stacktrace

(Optional)

### Related Issues

(Optional)

### Screenshot or Screencast

(Optional)

ChristopherChudzicki:
Fixed by #927

PR #927, opened 2022-01-24
fix resource_link regex, make non-greedy

ChristopherChudzicki:
Before the regex was greedy so it would not stop at the end of the first resource_link. Two resource links on the same lin would be gobbled up as one.

The new regex is still pretty simple but will be fooled by resource_link titles that include literal `" >}}` values. This is a pretty edgy edge case. We could get around this in the future by escpaing quotations in the link titles and using a different regex (or parsing some other way). But right now the unescaped quotation characters are not causing any problems other than this, so lets keep this fix simple.

#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Fixes https://github.com/mitodl/ocw-studio/issues/917 and https://github.com/mitodl/ocw-studio/issues/903.

#### What's this PR do?
Fixes the regex that was turning resource_link shortcodes into HTM.

- #917 add multiple resource links on the same line; save the page and re-open editor. Verify links look the same.
- #903 add multiple resource links within a table; save the page and re-open editor. Verify table / links look the same.

**Before / After**
Screen Shot 2022-01-24 at 3 05 13 PM Screen Shot 2022-01-24 at 3 04 54 PM

### DIFF
--- a/static/js/lib/ckeditor/plugins/ResourceLinkMarkdownSyntax.test.ts
+++ b/static/js/lib/ckeditor/plugins/ResourceLinkMarkdownSyntax.test.ts
@@ -1,7 +1,7 @@
 jest.mock("@ckeditor/ckeditor5-utils/src/version")
 
 import ResourceLink from "@mitodl/ckeditor5-resource-link/src/link"
-import Markdown from "./Markdown"
+import Markdown, { MarkdownDataProcessor } from "./Markdown"
 import { createTestEditor, markdownTest } from "./test_util"
 import { turndownService } from "../turndown"
 
@@ -44,6 +44,25 @@ describe("ResourceLink plugin", () => {
       editor,
       '{{< resource_link asdfasdfasdfasdf "text here" >}}',
       '<p><a class="resource-link" data-uuid="asdfasdfasdfasdf">text here</a></p>'
+    )
+  })
+
+  it("should serialize multiple links to and from markdown", async () => {
+    const editor = await getEditor("")
+    markdownTest(
+      editor,
+      'dogs {{< resource_link uuid1 "woof" >}} cats {{< resource_link uuid2 "meow" >}}, cool',
+      '<p>dogs <a class="resource-link" data-uuid="uuid1">woof</a> cats <a class="resource-link" data-uuid="uuid2">meow</a>, cool</p>'
+    )
+  })
+
+  it("[BUG] does not behave well if link title ends in backslash", async () => {
+    const editor = await getEditor("")
+    const { md2html } = (editor.data
+      .processor as unknown) as MarkdownDataProcessor
+    expect(md2html('{{< resource_link uuid123 "bad \\" >}}')).toBe(
+      // This is wrong. Should not end in &lt;/a&gt;
+      '<p><a class="resource-link" data-uuid="uuid123">bad &lt;/a&gt;</p>'
     )
   })
 })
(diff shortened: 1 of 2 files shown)


static/js/lib/ckeditor/plugins/ResourceLinkMarkdownSyntax.test.ts
```diff
@@ -46,4 +46,23 @@ describe("ResourceLink plugin", () => {
...
+  it("should serialize multiple links to and from markdown", async () => {
+    const editor = await getEditor("")
+    markdownTest(
+      editor,
+      'dogs {{< resource_link uuid1 "woof" >}} cats {{< resource_link uuid2 "meow" >}}, cool',
+      '<p>dogs <a class="resource-link" data-uuid="uuid1">woof</a> cats <a class="resource-link" data-uuid="uuid2">meow</a>, cool</p>'
+    )
+  })
+
+  it("[BUG] does not behave well if link title ends in backslash", async () => {
```
ChristopherChudzicki, line 59: This test was just to document the behavior https://github.com/mitodl/ocw-studio/issues/928 🤷

static/js/lib/ckeditor/plugins/ResourceLinkMarkdownSyntax.ts
```diff
@@ -27,6 +25,17 @@ import {
...
+
+export const RESOURCE_LINK_SHORTCODE_REGEX = /{{< resource_link (\S+) "(.*?)" >}}/g
+/**
+ * (\S+) to match and capture the UUID
+ * "(.*?)" to match and capture the label text
+ *
+ * Limitations:
+ *   - gets fooled by label texts that include literal `" >}}` values. For
+ *     example, < resource_link uuid123 "silly " >}} link" >}}.
+ */
```
alicewriteswrongs, line 37: I think you might have accidentally moved the link and it's jsdoc comment between the `ResourceLinkMarkdownSyntax` class declaration and it's jsdoc comment here

alicewriteswrongs (changes requested):
tested it out and it works great, just noticed one formatting thing

static/js/lib/ckeditor/plugins/ResourceLinkMarkdownSyntax.ts
ChristopherChudzicki: Ahhh silly me. I've moved the regex mostly back up (but put it below the imports)

alicewriteswrongs: one more thing, and sorry I should have been more clear! but it would be good to put the block comment that you wrote for it above the variable declaration too, like this

```ts
/**
 * Here I can describe
 * What something does or doesn't
 * do through a haiku
 **/
const REGEX = /etc/
```

then various code tools will associate that comment with the `REGEX` variable as a documentation comment - for instance, in my editor if I ask about the type of a variable it will pop up a little window with type information and also print the associated comment, like this:

Screen Shot 2022-01-24 at 4 39 34 PM

ChristopherChudzicki: Neat—for some reason I never realized JSDoc comments worked for constants like they do for functions and classes... but why wouldn't they, I suppose.

alicewriteswrongs: yeah, I don't know if they'll work like that for all documentation tools, but they seem to be well supported w/ typescript ecosystem tools at any rate! super handy to have them pop up when something is confusing me later (usually something I wrote originally haha)

alicewriteswrongs (approved):
lgtm!

ChristopherChudzicki:
[Chris Chudzicki](https://trello.com/c/FhFHwFmV/8738-chris-chudzicki)

Issue #936, opened 2022-01-26
concourse webhook calls stalling site pipelines


gumaerc:
### Steps to Reproduce

Trigger basically any build

### Expected Behavior

The `ocw-studio-webhook` steps fire off quickly

### Actual Behavior

The webhook steps stall out and retry, sometimes hanging a build for up to 1.5 hours (for the two webhook calls)

PR #937, opened 2022-01-26
move pipeline api callbacks to jgriff/http-resource

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/936

#### What's this PR do?
Currently, we are using [`aequitas/http-api-resource`](https://github.com/aequitas/concourse-http-api-resource) to form HTTP requests back to the `ocw-studio` API and set the status of a build.  This resource has been problematic, causing massive timeouts.  It uses the Python `requests` library which has to be installed from pip each time the Docker container is re-created, and also doesn't allow you to specify a no-op in the `get` step.  This PR moves over to using [`jgriff/http-resource`](https://github.com/jgriff/http-resource) for the API calls instead, which simply uses `curl` behind the scenes and also allows for a no-op on the `get` step.

 - Spin up your local `ocw-studio` instance on this PR branch using all of the `CONCOURSE_` env variables from the RC instance of `ocw-studio` in Heroku as well as `OCW_STUDIO_BASE_URL=https://ocw-studio-rc.odl.mit.edu`
 - Create a site in [`ocw-studio-rc`](https://ocw-studio-rc.odl.mit.edu/sites) and leave it empty
 - Create a site in your local Concourse instance with the exact same name / short id
 - In a shell, run `docker-compose run --rm web ./manage.py backpopulate_pipelines --filter cpg-test-course-3`, replacing `cpg-test-course-3` with the short id of the course you created
 - Browse to https://cicd-qa.odl.mit.edu and log in, then find your site under the `live` or `draft` folder
 - Trigger a new build and watch the `ocw-studio-webhook` steps carefully, making sure that the pipeline doesn't hang at that step for more than a minute or so.  It should be very quick and definitely not time out
 - Keep in mind that while the webhook requests should succeed, the course build itself will fail because there is no content for it yet.  Browse to your site in [`ocw-studio-rc`](https://ocw-studio-rc.odl.mit.edu/sites) and ensure that you see "failed" in the upper right corner with a red circle.

### DIFF
--- a/content_sync/pipelines/definitions/concourse/site-pipeline.yml
+++ b/content_sync/pipelines/definitions/concourse/site-pipeline.yml
@@ -35,7 +35,6 @@ resources:
         headers:
             Content-Type: "application/json"
             Authorization: "Bearer ((api-token))"
-        text: "{json}"
         out_only: true
 jobs:
   - name: build-ocw-site
@@ -46,7 +45,7 @@ jobs:
           timeout: 5m
           attempts: 3
           params:
-            json: |
+            text: |
               {
                 "version": "((version))",
                 "status": "started"
@@ -58,7 +57,7 @@ jobs:
           try:
             put: ocw-studio-webhook
             params:
-                json: |
+                text: |
                   {
                     "version": "((version))",
                     "status": "errored"
@@ -71,7 +70,7 @@ jobs:
           try:
             put: ocw-studio-webhook
             params:
-                json: |
+                text: |
                   {
                     "version": "((version))",
                     "status": "errored"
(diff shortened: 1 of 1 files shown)


blarghmatey:
Looks good to me :+1:

content_sync/pipelines/definitions/concourse/site-pipeline.yml
```diff
@@ -47,15 +45,23 @@ jobs:
           timeout: 5m
```
mbertrand, line 45: Mike suggested dropping the timeout value to 1m for all the webhook steps

mbertrand (approved):
Just one suggestion, otherwise looks good

PR #941, opened 2022-01-27
reconcile ckedidtor, showdown multiline list items

ChristopherChudzicki:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

#### What's this PR do?
Adds a markdown "extension" that addresses an incompatibility between
 - the HTML that showdown produces for lists
 - the HTML that CKEditor expects for lists

When a markdown list item contains multiple lines, showdown puts `<p>` tags around all list items (and multipple `<p>` tags for the multi-line list item). See https://github.com/showdownjs/showdown/wiki/Showdown's-Markdown-syntax#known-differences-and-gotchas.

CKEditor, on the other hand, forbids paragraphs inside list items. Both are "block" elements in its schema; see https://ckeditor.com/docs/ckeditor5/latest/framework/guides/deep-dive/schema.html#defining-additional-semantics and block elements cannot contain other block elements. (There is discussion of having a list variant that can contain block elements; see https://github.com/ckeditor/ckeditor5/issues/2973 for more.)

So here we add a Showdown rule to replace list item paragraph tags with linebreaks, the same as CKEditor produces.

Create a page with list items where the list item contains line breaks (shift+enter). Save the page and re-open. It should look the same as it did before you saved it.

_Except it won't look the same if you insert multiple consecutive line breaks. They'll get collapsed into one. That's a separate issue, and not related to list items. And it seems like a less serious issue than #935_

list_items_with_new_lines

### DIFF
--- a/static/js/lib/ckeditor/plugins/MarkdownListSyntax.ts
+++ b/static/js/lib/ckeditor/plugins/MarkdownListSyntax.ts
@@ -1,0 +1,47 @@
+import { ShowdownExtension } from "showdown"
+import { TurndownRule } from "../../../types/ckeditor_markdown"
+import MarkdownSyntaxPlugin from "./MarkdownSyntaxPlugin"
+
+/**
+ * This "extension" is addresses an incompatibility between
+ *  - the HTML that showdown produces for lists
+ *  - the HTML that CKEditor expects for lists
+ *
+ * When a markdown list item contains multiple lines, showdown puts <p> tags
+ * around all list items (and multiple <p> tags for the multi-line list item).
+ * See https://github.com/showdownjs/showdown/wiki/Showdown's-Markdown-syntax#known-differences-and-gotchas.
+ *
+ * CKEditor, on the other hand, forbids paragraphs inside list items. Both are
+ * "block" elements in its schema; see https://ckeditor.com/docs/ckeditor5/latest/framework/guides/deep-dive/schema.html#defining-additional-semantics
+ * and block elements cannot contain other block elements. (There is discussion
+ * of having a list variant that can contain block elements; see
+ * https://github.com/ckeditor/ckeditor5/issues/2973 for more.)
+ *
+ * So here we add a Showdown rule to replace list item paragraph tags with
+ * linebreaks, the same as CKEditor produces.
+ *
+ */
+export default class MarkdownListSyntax extends MarkdownSyntaxPlugin {
+  get showdownExtension(): () => ShowdownExtension[] {
+    return () => [
+      {
+        type:   "output",
+        filter: htmlString => {
+          const container = document.createElement("div")
+          container.innerHTML = htmlString
+
+          container.querySelectorAll("li > p").forEach(node => {
+            const suffix = node.nextSibling === null ? "" : " <br><br>"
+            node.outerHTML = `${node.innerHTML}${suffix}`
+          })
+
+          return container.innerHTML
+        }
+      }
+    ]
+  }
+
+  get turndownRules(): TurndownRule[] {
+    return []
+  }
+}
--- a/static/js/lib/ckeditor/plugins/MarkdownSyntaxPlugin.ts
+++ b/static/js/lib/ckeditor/plugins/MarkdownSyntaxPlugin.ts
@@ -29,12 +29,12 @@ export default abstract class MarkdownSyntaxPlugin extends MarkdownConfigPlugin 
         ...currentConfig.showdownExtensions,
         this.showdownExtension
       ],
-      turndownRules: [...currentConfig.turndownRules, this.turndownRule]
+      turndownRules: [...currentConfig.turndownRules, ...this.turndownRules]
     }
     this.setMarkdownConfig(newConfig)
   }
 
   abstract get showdownExtension(): () => Showdown.ShowdownExtension[]
 
-  abstract get turndownRule(): TurndownRule
+  abstract get turndownRules(): TurndownRule[]
 }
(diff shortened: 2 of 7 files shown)


static/js/lib/ckeditor/plugins/MarkdownListSyntax.ts
```diff
@@ -0,0 +1,52 @@
...
+ * "block" elements in its schema; see https://ckeditor.com/docs/ckeditor5/latest/framework/guides/deep-dive/schema.html#defining-additional-semantics
+ * and block elements cannot contain other block elements. (There is discussion
+ * of having a list variant that can contain block elements; see
+ * https://github.com/ckeditor/ckeditor5/issues/2973 for more.)
+ *
+ * So here we add a Showdown rule to replace list item paragraph tags with
+ * linebreaks, the same as CKEditor produces.
+ *
+ */
+export default class MarkdownListSyntax extends MarkdownSyntaxPlugin {
```
ChristopherChudzicki, line 24: @alicewriteswrongs You mentioned a few possibilities in slack
> but anyway - we can make an intervention at a few points, [1] we could try to fix how Showdown handles this, or [2] we could try to edit and fix up it’s html output before it’s passed to ckeditor, or [3] we could try to see if there’s a way to make ckeditor tolerate <p> tags inside of list items

Since this is an "output" showdown rule, it basically seems like [2] to me. Using an output rule seemed more convenient than a "lang" rule since having the html tree is convenient for locating `li > p`. 

I did not explore (3) _too_ much beyond reading about the schema, and block elements not containing block elements and reading through some of https://github.com/ckeditor/ckeditor5/issues/2973.

static/js/lib/ckeditor/plugins/MarkdownListSyntax.ts
```diff
@@ -0,0 +1,52 @@
...
+import { TurndownRule } from "../../../types/ckeditor_markdown"
+import MarkdownSyntaxPlugin from "./MarkdownSyntaxPlugin"
+
+/**
+ * This "extension" is addresses an incompatibility between
+ *  - the HTML that showdown produces for lists
+ *  - the HTML that CKEditor expects for lists
+ *
+ * When a markdown list item contains multiple lines, showdown puts <p> tags
+ * around all list items (and multipple <p> tags for the multi-line list item).
```
alicewriteswrongs, line 11: small spelling thing, `s/multipple/multiple/`

alicewriteswrongs (changes requested):
looks good, just had one question about a change we might want to make to `MarkdownSyntaxPlugin` while we're in here

static/js/lib/ckeditor/plugins/MarkdownListSyntax.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  get turndownRule(): TurndownRule {
+    return {
+      name: "MarkdownListSyntax",
+      rule: {
+        filter: () => false
+      }
+    }
+  }
```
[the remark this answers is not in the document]
ChristopherChudzicki: True. How about changing the property to `turndownRules: TurndownRule[]`. That way:
- it's still required, so you have to actively opt-out rather than just forgetting to make the rule.
- but the array can be empty, so the boilerplate is minimal
- more symmetry between `turndownRules` and `showdownExtension`, since the latter is already a list of showdown rules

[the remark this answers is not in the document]
ChristopherChudzicki: The only downside that I can see is a bigger diff / cluttering the git history due to formatting the existing rules as an array. Still seems like a reasonable solution, though, so I went with it.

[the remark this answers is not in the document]
alicewriteswrongs: yeah that sounds like a good approach to me! arrays are nice for this kind of thing

static/js/lib/ckeditor/plugins/MarkdownSyntaxPlugin.ts
```diff
@@ -29,12 +29,12 @@ export default abstract class MarkdownSyntaxPlugin extends MarkdownConfigPlugin
...
-      turndownRules: [...currentConfig.turndownRules, this.turndownRule]
+      turndownRules: [...currentConfig.turndownRules, ...this.turndownRules]
     }
     this.setMarkdownConfig(newConfig)
   }
 
   abstract get showdownExtension(): () => Showdown.ShowdownExtension[]
 
-  abstract get turndownRule(): TurndownRule
+  abstract get turndownRules(): TurndownRule[]
```
alicewriteswrongs, line 39: maybe it's getting too fiddly, but I tested this out locally and it would work for allowing _no boilerplate at all_ on subclasses of `MarkdownSyntaxPlugin`

```suggestion
  get turndownRules(): TurndownRule[] {
    return []
  }
```

then basically subclasses that want to set a turndown rule would override this method, and others wouldn't. We could make the same change to 'de-abstract' the showdown method if we wanted as well

anyway - consider that a fully optional suggestion 😄

alicewriteswrongs (approved):
lgtm! one fully-optional small-reduction-in-boilerplate suggestion

static/js/lib/ckeditor/plugins/MarkdownSyntaxPlugin.ts
ChristopherChudzicki: I'm going to leave as-is for now because I like the idea that you have to opt-in to "no, I really don't need Turndown rules for this extension". Maybe that's silly—we would probably catch that sort of error in the unit tests 🤷

PR #942, opened 2022-01-27
fix issue with website name not being saved in resource collection

### LINKED ISSUE #940: relation widget not storing website name when using crossSite option
### Steps to Reproduce

1. use the `ocw-www` configuration and try to create a resource collection
2. check out the saved markdown, and 😱 because the website name isn't being saved! here's an example of what happens: https://github.mit.edu/ocw-content-rc/ocw-www/blob/main/content/resource_collections/open-textbooks.md

### Expected Behavior

We need to save both the content UUID and the website name so that we have all the info we need to make a request for the content later.

alicewriteswrongs:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #940

#### What's this PR do?

This fixes an issue where the website name wasn't properly being stored in resource collections (aka `cross_site` Relation  fields). We need the website name so that we'll be able to construct a URL to get info about the resource when rendering in ocw-www.

Create a resource collection and inspect the saved markdown. The value for the field should be an array of tuples like

```
[
  [resource-uuid, website-name]
]
```

the issue this fixes was causing the website name to just be `null`.

### DIFF
--- a/static/js/components/widgets/RelationField.test.tsx
+++ b/static/js/components/widgets/RelationField.test.tsx
@@ -27,6 +27,7 @@ import {
 import { ReactWrapper } from "enzyme"
 import { FormError } from "../forms/FormError"
 import { formatOptions, useWebsiteSelectOptions } from "../../hooks/websites"
+import SortableSelect from "./SortableSelect"
 
 jest.mock("../../lib/api/util", () => ({
   ...jest.requireActual("../../lib/api/util"),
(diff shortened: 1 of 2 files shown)


static/js/components/widgets/RelationField.test.tsx
```diff
@@ -249,6 +249,27 @@ describe("RelationField", () => {
           label: item.title
         }))
       )
+
+      await act(async () => {
+        // @ts-ignore
+        wrapper.find("SortableSelect").prop("onChange")(
+          // @ts-ignore
+          [contentListingItems[0].text_id]
+        )
```
ChristopherChudzicki, line 258: **Requested Change:** If you pass the `SortableSelect` function to `find` instead of the string then ts will know that (a) the prop is required and (b) the argument is `string[]` not ChangeEvent. Then we can get rid of both the `@ts-ignore`.

ChristopherChudzicki (changes requested):
Change looks good and it's working locally for me! I made one tiny request in the test that should help with ts stuff.

static/js/components/widgets/RelationField.test.tsx
alicewriteswrongs: ah nice catch! just made that change

ChristopherChudzicki (approved):

Issue #908, opened 2022-01-19
pointer cursor on `/sites` doesn't match up to link placement


alicewriteswrongs:
### Steps to Reproduce

1. go to `/sites`
2. hover over a site card, and note that the cursor changes to a pointer
3. notice as well that you can't navigate into that site unless you click specifically on the title

### Expected Behavior

the pointer should match up with triggering some behavior, so we should _either_ 1) make the whole card an `<a>` tag which links to the site page or 2) make only the title portion (which is an `<a>` tag) have `cursor: pointer`

### Actual Behavior

see video below

### Screenshot or Screencast

https://user-images.githubusercontent.com/6207644/150142744-d274787b-6ce6-4257-a913-870fc22dc84e.mov

PR #944, opened 2022-01-28
Course link clickability

ChristopherChudzicki:
#### Pre-Flight checklist

  - [x] Desktop screenshots _see comment_
- [x] Testing
  - [x] Code is tested

#### What's this PR do?
This PR makes `<StudioListItems />`'s cursor reflect its clickability better. Previously, all StudioListItems had `cursor:pointer` on hover. Now they only get that css effect if the onClick prop is passed.

In #908 two ideas were presented:
> the pointer should match up with triggering some behavior, so we should either 1) make the whole card an <a> tag which links to the site page or 2) make only the title portion (which is an <a> tag) have cursor: pointer

I went with option (2) but also made the site link area a bit bigger so it's easier to click. 

StudioListItems pointer status should now match the affect that clicking has. So for example:
- collaborator listings should no longer have cursor:pointer
- ...but pages, resources still should
- the site listing should only show pointer for the clickable part

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "autoprefixer": "9",
     "bootstrap": "4",
     "casual": "^1.6.2",
+    "classnames": "^2.3.1",
     "core-js": "^3.16.0",
     "css-loader": "^4.2.1",
     "enzyme": "^3.11.0",
--- a/static/js/components/Card.tsx
+++ b/static/js/components/Card.tsx
@@ -9,7 +9,7 @@ export default function Card(props: Props): JSX.Element {
 
   return (
     <div className="studio-card">
-      <div className="card-contents p-4">{children}</div>
+      <div className="p-4">{children}</div>
     </div>
   )
 }
--- a/static/js/components/StudioList.tsx
+++ b/static/js/components/StudioList.tsx
@@ -1,6 +1,7 @@
 import React, { useCallback, useState } from "react"
 import { Link } from "react-router-dom"
 import OutsideClickHandler from "react-outside-click-handler"
+import classNames from "classnames"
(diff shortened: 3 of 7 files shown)


static/js/components/StudioList.tsx
```diff
@@ -63,10 +64,16 @@ export function StudioListItem(props: ListItemProps): JSX.Element {
   )
 
   return (
-    <li className="py-2" onClick={onClick}>
```
ChristopherChudzicki, line 66: Switched from padding to margin so that the space between items would not be clickable:

**Before / After**

between_items_beforebetween_items_now

The margin is twice as big since these adjacent margins are collapsed.

package.json
```diff
@@ -68,6 +68,7 @@
     "autoprefixer": "9",
     "bootstrap": "4",
     "casual": "^1.6.2",
+    "classnames": "^2.3.1",
```
ChristopherChudzicki, line 71: Reactstrap was already requiring this package, so I added it for our use.

static/js/components/Card.tsx
```diff
@@ -9,7 +9,7 @@ export default function Card(props: Props): JSX.Element {
 
   return (
     <div className="studio-card">
-      <div className="card-contents p-4">{children}</div>
```
ChristopherChudzicki, line 12: As far as I can tell, this class is unused

alicewriteswrongs: ah yup, I think we copied over this component from another project and then refactored a bit and that class slipped through the net

alicewriteswrongs (approved):
looks good!

Issue #938, opened 2022-01-27
add theme assets pipeline definition


gumaerc:
The theme assets pipeline definition should be moved from [`ol-infrastructure`](https://github.com/mitodl/ol-infrastructure/blob/main/pipelines/ocw/pipeline-theme-assets.yml) to here in `ocw-studio`.  It should:

 - [ ] Be able to be upserted via an optional argument on `backpopulate_pipelines`, something like `--upsert-theme-assets-pipeline`
 - [ ] Instead of creating a tarball with the built theme assets, it should simply run `npm run build:webpack` and then publish the contents of `base-theme/dist` to the proper bucket, templated by environment
 - [ ] The file generated during the webpack build at `base-theme/data/webpack.json` should be uploaded to `ol-eng-artifacts` templated by the branch
 - [ ] In addition, the `site-pipeline.yml` definition should be updated to source `ocw-hugo-themes` from git and the code that pulls in the tarball should be removed
 - [ ] The `webpack.json` file should be added as a resource in the `site-pipeline.yml` definition and it should be injected into the site build before Hugo is run

PR #891, opened 2022-01-05
"waterfall" triggering scheme for ocw-hugo-themes changes

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Changes have been manually tested

#### What are the relevant tickets?
Closes https://github.com/mitodl/ocw-studio/issues/880

#### What's this PR do?
Modifies the concourse pipeline so that only the "ocw-www" site is triggered by a a new themes tarball, and other sites should be triggered by a new version file in the static folder uploaded to S3 by the ocw-www pipeline.

#### How should this be manually tested?
This is a bit tricky.  I've already updated the `ocw-www` pipeline and one site pipeline (`cpg-test-course-3`) to use the modified pipeline definition.

You can trigger a new build of the release-candidate themes pipeline, which should include a new `version.<dt-num>.txt` file in the generated tarball.  That should still trigger the ocw-www build, but it will also trigger all the unmodified sites too.  Once the ocw-www build finishes, a `cpg-test-course-3` build should be triggered within a minute or two.

If you want to avoid triggering all. sites, you can try this:
- Manually start another run of the ocw-www pipeline from the UI
- Soon after it completes, start a shell in that build:
    `fly -t rc intercept -j draft/site:ocw-www/build-ocw-site  -b <latest build num> -s copy-s3-buckets  --team ocw`
- Run the following to create a new version file and copy to the static s3 folder:
   ```
   dt=$(date +%s)
   echo $dt >  ocw-hugo-themes/theme/base-theme/dist/static/version.$dt.txt
   aws s3 cp ocw-hugo-themes/theme/base-theme/dist/static/version.$dt.txt  s3://ocw-content-draft-qa/static/version.$dt.txt
   ```
- Keep an eye on the draft pipeline for `cpg-test-course-3`, it should start a new build shortly.
- Do this for both draft and live pipelines in QA

PR #945, opened 2022-01-29
theme assets pipeline

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
- [ ] Settings

Closes https://github.com/mitodl/ocw-studio/issues/938

#### What's this PR do?
In https://github.com/mitodl/ocw-studio/pull/891, we set up a "waterfall" like triggering strategy where publishes would happen in this sequence:

 - Create RC in `ocw-hugo-themes`
 - `ocw-theme-assets`build triggers from merge to `release-candidate`
 - `ocw-www` build triggers from artifacts being deployed to `ol-eng-artifacts`
 - `ocw-www` build creates a static version file upon completion
 - static version file triggers all other course builds

In theory this is a good approach, but in practice we've had a number of issues:

 - The compounding number of static version files in the bucket is making the trigger for it take a longer amount of time with each run
 - Through some analysis, the unzipping of the theme assets tar file is causing a compounding issue with CPU usage

This PR adjusts our whole approach for deploying new theme assets from start to finish:

 - First of all, the pipeline definition for the theme assets build has been moved from `ol-infrastructure` to here in `ocw-studio`.
 - The pipeline is also now templated and deployed using a new management command, `upsert_theme_assets_pipeline`.  It only creates a single pipeline based on the configured `GITHUB_WEBHOOK_BRANCH`, so that everything done on RC uses the `release-candidate` branch, and everything in production uses the `release` branch.
 - The theme assets pipeline has been updated to not zip up the artifacts.  Instead, they are deployed directly to the publish buckets for draft and live, then the `webpack.json` file is uploaded to `ol-eng-artifacts`.
 - The site pipeline definition was updated so that upload of the `webpack.json` file is what triggers the course build, and `ocw-hugo-themes` is pulled in directly from git for the templates
 
 - Set the following in your `.env` file:
```
CONCOURSE_URL=http://concourse:8080
CONCOURSE_PASSWORD=test
CONCOURSE_USERNAME=test
CONCOURSE_TEAM=main
```
 - Start up `ocw-studio` with the local Concourse profile using `docker-compose --profile concourse up`
 - Run `docker-compose run --rm web ./manage.py upsert_theme_asssets_pipeline`
 - Go to http://localhost:8080 in a browser and log into your local Concourse instance with the info above
 - View the theme assets pipeline and make sure it was properly created
 - Create a test course in the `ocw-studio` UI if you haven't already
 - Run `docker-compose run --rm web ./manage.py backpopulate_pipelines --filter <your site id here>`
 - Ensure that your site pipeline was created successfully and that the triggers look correct

If you actually try running any of the pipelines you will likely encounter errors as your local instance doesn't have AWS access to the buckets required.  Further testing will need to be done once this PR is merged to RC.


content_sync/pipelines/concourse.py
```diff
@@ -249,3 +259,98 @@ def get_build_status(self, build_id: int):
...
+        return f"/api/v1/teams/{settings.CONCOURSE_TEAM}/pipelines/ocw-theme-assets/config?vars={self.instance_vars}"
+
+    def _make_job_url(self, job_name: str):
+        """Make URL for fetching job info"""
+        return f"/api/v1/teams/{settings.CONCOURSE_TEAM}/pipelines/ocw-theme-assets/jobs/{job_name}?vars={self.instance_vars}"
+
+    def _make_pipeline_unpause_url(self):
+        """Make URL for unpausing a pipeline"""
+        return f"/api/v1/teams/{settings.CONCOURSE_TEAM}/pipelines/ocw-theme-assets/unpause?vars={self.instance_vars}"
+
```
mbertrand, line 309: May want to consider moving these to the base pipeline to avoid nearly identical code, see this PR:  https://github.com/mitodl/ocw-studio/pull/950/files#diff-979f137beba738446b3fc46c89be033a4f04e2e285ce7e68d3028b0e1b83c11fR104-R138

In this case, `ocw-theme-assets` would be a value for the `version` parameter (which I will probably change to `pipeline_name` in my PR).
For example:
```
    def _make_pipeline_config_url(self, pipeline_name: str):
        """Make URL for fetching pipeline info"""
        return f"/api/v1/teams/{settings.CONCOURSE_TEAM}/pipelines/{pipeline_name}/config?vars={self.instance_vars}"
```

and call it for `ThemeAssetsPipeline` in `upsert_theme_assets_pipeline`. like this:
```
url_path = self._make_pipeline_config_url(self.pipeline_name)
```

where `self.pipeline_name` could be set in the `init()` function to `"ocw-theme-assets"`

content_sync/pipelines/concourse.py
```diff
@@ -249,3 +259,98 @@ def get_build_status(self, build_id: int):
...
         """Abort a build"""
         return self.api.abort_build(build_id)
+
+
+class ThemeAssetsPipeline(BaseThemeAssetsPipeline):
+    """
+    Concourse-CI pipeline for publishing theme assets
+    """
+
+    MANDATORY_SETTINGS = [
```
mbertrand, line 269: I moved these in my PR so they could be used for all concourse pipelines: https://github.com/mitodl/ocw-studio/pull/950/files#diff-979f137beba738446b3fc46c89be033a4f04e2e285ce7e68d3028b0e1b83c11fR23

content_sync/pipelines/concourse.py
```diff
@@ -249,3 +259,98 @@ def get_build_status(self, build_id: int):
...
+        """Unpause the pipeline"""
+        self.api.put(self._make_pipeline_unpause_url())
+
+    def get_build_status(self, build_id: int):
+        """Retrieve the status of the build"""
+        return self.api.get_build(build_id)["status"]
+
+    def abort_build(self, build_id: int):
+        """Abort a build"""
+        return self.api.abort_build(build_id)
```
mbertrand, line 356: Same as comment above (put in the base pipeline to avoid dupe or nearly identical code)

content_sync/pipelines/concourse.py
```diff
@@ -249,3 +259,98 @@ def get_build_status(self, build_id: int):
...
+    """
+    Concourse-CI pipeline for publishing theme assets
+    """
+
+    MANDATORY_SETTINGS = [
+        "CONCOURSE_URL",
+        "CONCOURSE_USERNAME",
+        "CONCOURSE_PASSWORD",
+        "GITHUB_WEBHOOK_BRANCH",
+        "SEARCH_API_URL",
```
mbertrand, line 274: Also "THEME_ASSETS_PIPELINE" (currently, though that may change if one of the prior suggestions is implemented).

mbertrand (changes requested):

content_sync/pipelines/definitions/concourse/site-pipeline.yml
```diff
@@ -130,9 +127,8 @@ jobs:
             args:
             - -exc
             - |
-              mkdir -p ../ocw-hugo-themes/theme
-              tar -xvzf ../ocw-hugo-themes/ocw-hugo-themes-*.tgz -C ../ocw-hugo-themes/theme
-              hugo --config ../ocw-hugo-projects/((config-slug))/config.yaml --baseUrl /((base-url)) --themesDir ../ocw-hugo-themes/theme
+              cp ../webpack-json/webpack.json data
+              hugo --config ../ocw-hugo-projects/((config-slug))/config.yaml --baseUrl /((base-url)) --themesDir ../ocw-hugo-themes/
```
mbertrand, line 131: What does `cp ../webpack-json/webpack.json data` do?  Isn't `webpack.json` already in the `ocw-hugo-themes` directory?

app.json
```diff
@@ -107,6 +107,10 @@
       "description": "The pipeline to preview/publish websites with",
       "required": false
     },
+    "THEME_ASSETS_PIPELINE": {
```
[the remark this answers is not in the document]
gumaerc: I'll rename it to `CONTENT_SYNC_THEME_PIPELINE` for now and leave any other refactoring to you

content_sync/pipelines/concourse.py
gumaerc: The thing is, the theme pipeline doesn't require the Github settings.  I don't know if you saw, but I added a utility function called `check_mandatory_settings` to utils.py to abstract the actual check, then each pipeline can define what settings are mandatory.  I think in `concourse.py` at least, it makes sense for each pipeline to include the `CONCOURSE_` prefixed settings, then define what they will use beyond that?

mbertrand: ok, makes sense

content_sync/pipelines/definitions/concourse/site-pipeline.yml
gumaerc: No, it won't be.  `webpack.json` is generated at build time by an `AssetsPlugin` defined in the Webpack configuration and it's deposited in the theme's `base-theme/data` folder.  The `ocw-hugo-themes` folder in this case will also now just be the raw Github repo, cloned directly from github without any built artifacts.  The `webpack.json` file is uploaded to `ol-eng-artifacts` by the theme assets pipeline upon a completed build, and it has been added to the site pipeline definition as a versioned file.  The idea is that when a new version is detected, it will trigger site builds.  Although, once we merge your mass publish pipeline we'll want to remove `trigger: true` from it.  This line simply copies it from the resource import into the site's `data` file at build time so Hugo knows what URLs to write in.

mbertrand (approved):
Looks good, need a salt-ops PR to add `CONTENT_SYNC_THEME_PIPELINE` and `SEARCH_API_URL` setting values, otherwise 👍

gumaerc:
@mbertrand Yep, I'll merge it after https://github.com/mitodl/salt-ops/pull/1496 is reviewed and merged.

PR #951, opened 2022-02-01
No duplicate entries in ResourceCollection, CourseCollection

ChristopherChudzicki:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

#### What's this PR do?
Disables already-selected values in Relation and WebsiteCollection fields.  

Check that allowed in
- resource collections
- video galleries
- website collections

Any other Relation/WebsiteCollection fields you can think of. (_Aside: Duplicate entries were already filtered out in RelationFields with multi-true, since that seems to be the default behavior of react-select in the multi scenario._)

Screen Shot 2022-02-01 at 2 08 42 PM

### DIFF
--- a/static/js/components/widgets/SelectField.tsx
+++ b/static/js/components/widgets/SelectField.tsx
@@ -1,7 +1,7 @@
 import React, { useCallback, ChangeEvent } from "react"
 import Select from "react-select"
 import AsyncSelect from "react-select/async"
-import { is, isNil } from "ramda"
+import { isNil } from "ramda"
 
 export interface Option {
   label: string
@@ -10,13 +10,14 @@ export interface Option {
 
 interface Props {
   name: string
-  value: any
+  value: null | undefined | string | string[]
   onChange: (event: ChangeEvent<HTMLSelectElement>) => void
   multiple?: boolean
   options: Array<string | Option>
   loadOptions?: (s: string, cb: (options: Option[]) => void) => void
   placeholder?: string
   defaultOptions?: Option[]
+  isOptionDisabled?: (option: Option) => boolean
 }
 
 export default function SelectField(props: Props): JSX.Element {
@@ -27,11 +28,12 @@ export default function SelectField(props: Props): JSX.Element {
     options,
     loadOptions,
     defaultOptions,
-    placeholder
+    placeholder,
+    isOptionDisabled
   } = props
   const multiple = props.multiple ?? false
-  const selectOptions = options.map((option: any) =>
-    is(String, option) ? { label: option, value: option } : option
+  const selectOptions = options.map(option =>
+    typeof option === "string" ? { label: option, value: option } : option
   )
 
   const changeHandler = useCallback(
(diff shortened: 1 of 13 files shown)


static/js/lib/api/util.test.ts
```diff
@@ -22,16 +22,32 @@ describe("api utility functions", () => {
...
+    global.mockFetch.mockResolvedValue("meow")
+
+    const results = await Promise.all([
+      debouncedFetch("key", 30, "url1", { credentials: "include" }),
+      debouncedFetch("key", 30, "url2", { credentials: "omit" }),
+      debouncedFetch("key", 30, "url3", { credentials: "same-origin" }),
+      debouncedFetch("key", 30, "url3", { credentials: "omit" })
+    ])
+
+    expect(results).toEqual([null, null, null, "meow"])
```
ChristopherChudzicki, line 51: Before this PR, line 51 would return `[null, null, "meow", null]`. Ie, before the real fetch response was being returned by oldest invocation whose first arg matches the most recent invocation's first arg.

The change to `debouncedFetch` is not directly related to the rest of this PR. But (a) it seems like this was the intended behavior, and (b) caused me some confusion on Monday while exploring potential refactor of some react-select usage.

static/js/components/widgets/SortableSelect.tsx
```diff
@@ -101,6 +110,7 @@ export default function SortableSelect(props: Props) {
           options={options}
           loadOptions={loadOptions}
           defaultOptions={defaultOptions}
+          {...(isOptionDisabled ? { isOptionDisabled } : {})}
```
alicewriteswrongs, line 113: could also just do `isOptionDisabled={isOptionDisabled}` since the types are the same (`isOptionDisabled?: (option: Option) => boolean`)

alicewriteswrongs: sometimes I wish there was a JSX syntax like the object literal property shorthand (`let foo = "bar"; let obj = { foo }`)

static/js/components/widgets/SelectField.tsx
```diff
@@ -82,7 +87,8 @@ export default function SelectField(props: Props): JSX.Element {
         border:    0,
         boxShadow: "none"
       })
-    }
+    },
+    ...(isOptionDisabled ? { isOptionDisabled } : {})
```
alicewriteswrongs, line 91: here too, the props match so why not just put the value into the object, whatever it might be? 

```suggestion
    },
    isOptionDisabled
```

alicewriteswrongs (changes requested):
works great! just had a few code comments here and there

static/js/components/widgets/SortableSelect.tsx
ChristopherChudzicki: 👍 I was trying to distinguish between `props = { isOptionDisabled: undefined }` and `props = { }`. But that's probably overly paranoid and nice to keep the code simpler.

alicewriteswrongs: My thought is that typescript gives us a lot of assurance that if a prop is `undefined` (vs not set) we still won't end up trying to call it, so I think it's fine to just pass the value down

alicewriteswrongs (approved):
lgtm! 👍

PR #952, opened 2022-02-01
small rename of two functions for clarity

alicewriteswrongs:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

no ticket

#### What's this PR do?

Just noticed while reviewing #951 that we had two functions named `formatOptions` hanging around in the frontend codebase, so I thought I'd go ahead and rename them to be more specific about what they do.

test should pass, select fields and whatnot should work.


ChristopherChudzicki (approved):
🎉 _I checked selects still work. They do!_

PR #957, opened 2022-02-02
fix up eslint camelcase rule disables

alicewriteswrongs:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

none

#### What's this PR do?

Just a little neatness thing. For `types/websites.ts` we're declaring the values we're getting from the backend, so we're stuck with camelcase stuff there. I went ahead and disabled the `camelcase` rule for the whole file, so we don't have to have a whole bunch of per-line disables.

Then I also made a few small changes to the `RelationField` widget to improve the situation a bit.

tests should pass, etc.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,6 +20,12 @@ module.exports = {
       "error",
       { argsIgnorePattern: "_" }
     ],
-    "mocha/no-top-level-hooks": "off"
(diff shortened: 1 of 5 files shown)


ChristopherChudzicki (approved):
🐫

PR #962, opened 2022-02-03
scroll to form errors on submission

ChristopherChudzicki:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

#### What's this PR do?
Adds logic to scroll to form errors upon submission. Should be no other changes.

This should affect all forms using `<SiteContentForm />` that have field validation. The one I am aware of is metadata, but if there are others, feel free to check those, too!

<details open>
<summary>Collapse/expand distracting gif</summary>

form_scroll_on_errors

</details>

### DIFF
--- a/static/js/components/forms/SiteContentForm.tsx
+++ b/static/js/components/forms/SiteContentForm.tsx
@@ -17,6 +17,7 @@ import {
   newInitialValues,
   renameNestedFields
 } from "../../lib/site_content"
+import { scrollToElement } from "../../util/dom"
 
 import {
   ConfigField,
@@ -93,8 +94,17 @@ export default function SiteContentForm({
       initialValues={initialValues}
       enableReinitialize={true}
     >
-      {({ isSubmitting, status, values, handleChange }) => (
-        <Form>
+      {({ isSubmitting, status, values, handleChange, handleSubmit }) => (
+        <Form
+          onSubmit={async event => {
+            handleSubmit(event)
+            const { target } = event // get target before the await; https://reactjs.org/docs/legacy-event-pooling.html
+            const errors = await validate(values)
+            if (Object.keys(errors).length > 0) {
+              scrollToElement(target as HTMLElement, ".form-error")
+            }
+          }}
+        >
           <div>
             {renamedFields
               .filter(field => fieldIsVisible(field, values))
--- a/static/js/test_setup.ts
+++ b/static/js/test_setup.ts
@@ -44,6 +44,7 @@ declare global {
 // cleanup after each test run
 // eslint-disable-next-line mocha/no-top-level-hooks
 afterEach(function() {
+  jest.resetAllMocks()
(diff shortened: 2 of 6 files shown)


static/js/components/forms/SiteContentForm.tsx
```diff
@@ -76,6 +77,42 @@ export default function SiteContentForm({
...
+    }
+    const { matches: prefersReducedMotion } = window.matchMedia(
+      "(prefers-reduced-motion: reduce)"
+    )
+    errorElement.scrollIntoView({
+      behavior: prefersReducedMotion ? "auto" : "smooth",
+      block:    "center"
+    })
+    errorElement.focus({ preventScroll: true })
+  }
```
alicewriteswrongs, line 114: I think we could simplify this function a bit - instead of querying for the input element, I think we can get the same functionality by just looking for the first instance on the page of `.form-error` - the class that we put on our form error messages. If we did that I think we'd also remove any runtime dependencies for this function, so it could go into a library file and get some isolated tests

alicewriteswrongs (changes requested):
works great! I had one though about the implementation

static/js/components/forms/SiteContentForm.tsx
ChristopherChudzicki: yes that sounds much better 👍

ChristopherChudzicki: That is a ... much cleaner approach 😄

static/js/test_setup.ts
```diff
@@ -44,6 +44,7 @@ declare global {
 // cleanup after each test run
 // eslint-disable-next-line mocha/no-top-level-hooks
 afterEach(function() {
+  jest.resetAllMocks()
```
ChristopherChudzicki, line 47: This change will apply to all our jest mocks... But the tests still pass, so it seems like something we should do.

static/js/util/dom.test.ts
```diff
@@ -0,0 +1,103 @@
...
+describe("scrollToError", () => {
+  const stringToHtml = (string: string): HTMLElement => {
+    const template = document.createElement("template")
+    template.innerHTML = string.trim()
+    const node = template.content.firstChild
+    if (node instanceof HTMLElement) return node
+    throw new Error("Expected an HTMLElement")
+  }
+  const getTestPage = () => {
+    const page = stringToHtml(`
```
alicewriteswrongs, line 13: this is totally fine as is, but you can also do JSX and use `ReactDOMServer.renderToString` for stuff like this

static/js/test_setup.ts
alicewriteswrongs: makes sense I think, I can't really think of a case where we'd want mocks sticking around between tests

alicewriteswrongs (approved):
looks good to me!

PR #965, opened 2022-02-03
refactor the content form dirty logic to use Formik's state

alicewriteswrongs:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

none

#### What's this PR do?

Our existing implementation of setting a dirty variable was basically duplicating Formik's logic. I refactored the `SiteContentForm` to add a `useEffect` hook between `Formik` and our form fields which listens to Formik's `dirty` property and echos that out via a callback.

I also refactored the associated test file to be flat (no nesting) and to not use any mutable shared variables. I think it's much cleaner and easier to figure out where things are coming from, and I think I'd like to write frontend tests like this going forward.

The 'are you sure?' confirmation behavior should work 100% as it did before (i.e. if you go into a site content, type something, and then try to leave without saving it should make you confirm that you want to throw that input out).

### DIFF
--- a/static/js/components/forms/SiteContentForm.test.tsx
+++ b/static/js/components/forms/SiteContentForm.test.tsx
@@ -1,12 +1,14 @@
 import { times } from "ramda"
 import React from "react"
-import { Formik, Form } from "formik"
-import sinon, { SinonSandbox, SinonStub } from "sinon"
-import { shallow } from "enzyme"
+import { Formik } from "formik"
+import { mount } from "enzyme"
+import { act } from "react-dom/test-utils"
 
-import SiteContentForm from "./SiteContentForm"
-
-import { defaultFormikChildProps, shouldIf } from "../../test_util"
+import SiteContentForm, {
+  FormFields,
+  InnerFormProps,
+  FormProps
+} from "./SiteContentForm"
 import {
   makeEditableConfigItem,
   makeWebsiteConfigField,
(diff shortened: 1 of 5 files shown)


static/js/components/forms/SiteContentForm.test.tsx
```diff
@@ -1,239 +1,236 @@
...
+]
+
+type ESBoolMatrix = [WebsiteContentModalState, boolean][]
+
+const EDITOR_STATE_BOOLEAN_MATRIX: ESBoolMatrix = EDITOR_STATES.map(
+  editorState => [
+    [editorState, true],
+    [editorState, false]
+  ]
+).flat() as any // :/ unfortunate but necessary
```
ChristopherChudzicki, line 110: You can use `as ESBoolMatrix` here instead of `as any` if you want... `as ESBoolMatrix` still does some checking, e.g. `5 as ESBoolMatrix` will error.

ChristopherChudzicki (commented):
The code makes sense and looks good, I have not tested it much locally but might tomorrow (Will coordinate with @mbertrand ... @mbertrand I have another PR working on this component yesterday/today, so I was curious to see what Alice did with it 😄)

@alicewriteswrongs Did separating `SiteContentForm` and `FormFields` make an aspect of testing easier, or just general goodness around keeping their scopes separate? _(I did see it means we don't need the `renderProp` enzyme method anymore.)_

static/js/components/forms/SiteContentForm.test.tsx
alicewriteswrongs: mmm will change that, I thought I tried it but I must not have

static/js/components/forms/SiteContentForm.test.tsx
```diff
@@ -1,239 +1,236 @@
...
+  const { form } = setup({
+    configItem
+  })
+  const objectField = form.find(ObjectField)
+  expect(objectField.exists()).toBeTruthy()
+  const renamedField = renameNestedFields([configField])[0]
+  expect(objectField.prop("field")).toEqual(renamedField)
+})
+
+test.each`
```
ChristopherChudzicki, line 187: I did know know about `test.each`. Cool....!

alicewriteswrongs:
The main reason to separate it was that 1) by doing so I was able to have a stable, statically defined function instead of one which is created inline, which allows me to use hooks and whatnot, and 2) then I can render the `FormFields` function separately so it's easier to manually set a single prop on it for testing (before we had that render prop hacky thing)

static/js/components/forms/SiteContentForm.test.tsx
alicewriteswrongs: yeah we've been doing it manually, living in ignorance haha

alicewriteswrongs: the only downside is that you can using a template for the test title, and it's based on a format string instead, but that's not too bad I think

static/js/components/forms/SiteContentForm.test.tsx
```diff
@@ -1,239 +1,236 @@
...
-    onSubmitStub = sandbox.stub()
-    setDirtyStub = sandbox.stub()
-  })
+function setupData() {
+  const content = makeWebsiteContentDetail()
+  content.content_context = times(() => makeWebsiteContentDetail(), 3)
+  const configItem = makeEditableConfigItem(content.type)
+  const website = makeWebsiteDetail()
+  // @ts-ignore
+  useWebsite.mockReturnValue(website)
```
ChristopherChudzicki, line 51: If you want:

```ts
import * as Website from "../../context/Website"

const { useWebsite: mockUseWebsite } = Website  as jest.Mocked<typeof Website>

mockUseWebsite.mockReturnValue(website)
```

That's the best way I've currently found to deal with mock typings. 

I came across [`jest.mocked`](https://jestjs.io/docs/jest-object#jestmockedtitem-t-deep--false) the other day which also looks useful...but as far as I can tell it doesn't exist on `@types/jest` yet. Which is kinda funny, since it's all about typing.

ChristopherChudzicki (approved):
Tested it out locally—everything looks good to me ! 👍

ChristopherChudzicki:
> ...which allows me to use hooks and whatnot

Ah, I forgot about that. That makes sense.

static/js/components/forms/SiteContentForm.test.tsx
alicewriteswrongs: ooo nice!

Issue #990, opened 2022-02-10
search on `/sites` needs to match on more fields


alicewriteswrongs:
### Steps to Reproduce

Search for a course number

### Expected Behavior

the associated course should pop up

### Actual Behavior

it most certainly does not!

### what's going on

Right now the search field on `/sites` _only_ matches on the `title` field on the website model. This means that for a lot of courses things that course authors might think they can search for (like the course number, or a fragment of the `name` field) actually will not match at all.

We should improve this by searching on `title`, `name`, and `short_id`.

pdpinch:
Also, course number, for example `8.701`

alicewriteswrongs:
yes, course number is found in the `short_id` field

PR #992, opened 2022-02-10
improvements to site search [frontend]

### LINKED ISSUE #963: /sites Course search loses focus + cursor position as typing
### Steps to Reproduce
1. Use the search bar in upper right to search for a site https://ocw-studio-rc.odl.mit.edu/sites
2. Type in a phrase, but type kinda slowly

### Expected Behavior
Search input should maintain focus + cursor position so user can continue typing

### Actual Behavior
Search input loses focus, so if you want to keep typing you have to click on it again.

### Screenshot or Screencast

<details open>
<summary> Open/close gif </summary>

course_search_reloads_studio

</details>

alicewriteswrongs:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #963 

this is sort of a twin PR to #993 but just addresses the frontend part.

#### What's this PR do?

This fixes a few issues on the site search (the little box on `/sites`). In particular, previously a full page refresh (well, actually a react-router reboot of the whole route) was triggered when the search changed. now the search results are updated, the search box retains focus, and the UI around the results doesn't go away.

Verify that the site search has a few properties:

- if you type in a new search string (like `ocw-www`) after a short debounce (600ms) there should be a brief loading indicator and then results relevant to your query should appear. the rest of the UI should now refresh.
- if you load the page with a search query param (`?q=foobar`) the value of the search param should show up in the search box, and the results shown should be relevant to that search (e.g. at `/sites/?q=ocw-www` you should see ocw-www)
- pagination should work as normal. in particular:
    - without a search param it should just work as you'd expect
    - and _with_ a search param it should just work as you'd expect :)
    - if you have paginated into the result set and you change the search param, then it should take you to the _first page_ of result for that search parameter, rather than retaining the `offset` parameter when the search input changes.

This PR does not address #990 (changing the search to match on more fields).

This showcases some of the changes in behavior (compare with the video on the linked issue):

https://user-images.githubusercontent.com/6207644/153459788-fdd8efcd-9f27-4421-a5c8-0bbf7f6536ed.mov

### DIFF
--- a/static/js/pages/SitesDashboard.test.tsx
+++ b/static/js/pages/SitesDashboard.test.tsx
@@ -8,7 +8,7 @@ import {
   sitesBaseUrl
 } from "../lib/urls"
 import { WebsiteListingResponse } from "../query-configs/websites"
-import { isIf, wait } from "../test_util"
+import { wait } from "../test_util"
 import {
   makeWebsiteListing,
   makeWebsiteDetail
@@ -66,7 +66,7 @@ describe("SitesDashboard", () => {
     helper.cleanup()
   })
 
-  it("lists a page", async () => {
+  test("lists a page", async () => {
     const { wrapper } = await render()
     let idx = 0
     for (const website of websites) {
@@ -80,7 +80,7 @@ describe("SitesDashboard", () => {
     }
   })
 
-  it("lets the user filter the sites", async () => {
+  test("lets the user filter the sites", async () => {
     const { wrapper } = await render()
     const filterInput = wrapper.find(".site-search-input")
     const event = {
@@ -90,12 +90,10 @@ describe("SitesDashboard", () => {
     } as React.ChangeEvent<HTMLInputElement>
     filterInput.simulate("change", event)
     await wait(800)
-    expect(helper.browserHistory.location.search).toBe(
-      "offset=0&q=my-search-string"
-    )
+    expect(helper.browserHistory.location.search).toBe("?q=my-search-string")
   })
 
-  it("should issue a request based on the 'q' param", async () => {
+  test("should issue a request based on the 'q' param", async () => {
     helper.browserHistory.replace({
       pathname: "/",
       search:   "q=searchfilter"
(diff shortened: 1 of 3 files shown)


static/js/pages/SitesDashboard.test.tsx
```diff
@@ -120,74 +118,73 @@ describe("SitesDashboard", () => {
...
-  it("sets the page title", async () => {
+  test("sets the page title", async () => {
     const { wrapper } = await render()
     expect(wrapper.find("DocumentTitle").prop("title")).toBe(
       "OCW Studio | Sites"
     )
   })
 
-  it("has an add link to the new site page", async () => {
+  test("has an add link to the new site page", async () => {
```
ChristopherChudzicki, line 128: Do you prefer `test`?

static/js/pages/SitesDashboard.tsx
```diff
@@ -98,29 +114,39 @@ export default function SitesDashboard(): JSX.Element {
...
-                <div className="text-success">Published</div>
-              ) : (
-                <div className="text-dark">Draft</div>
-              )}
-            </StudioListItem>
-          ))}
-        </StudioList>
+        {isPending || !listing ? (
+          <div className="site-page container">Loading...</div>
+        ) : (
```
ChristopherChudzicki, line 119: **Opinion / Minor Suggestion:** I think we not show the "Loading ...". These results should load fairly quickly, and if we don't show the loading, the previous results stay visible until the new results arrive, right? The very fast "loading" flash looks strange to me (opinion based!)

static/js/pages/SitesDashboard.tsx
```diff
@@ -29,51 +29,67 @@ export function siteDescription(site: Website): string | null {
...
+   */
   useDebouncedEffect(
     () => {
-      const currentSeach = searchString ?? ""
-      if (searchInput !== currentSeach) {
+      const currentSearch = listingParams.search ?? ""
+      if (searchInput !== currentSearch) {
         const newParams = new URLSearchParams()
-        newParams.set("offset", String(offset))
         newParams.set("q", searchInput)
```
ChristopherChudzicki, line 70: I noticed that if you delete the search text, the url ends in `sites/?offset=0&q=`.

I can't imagine that the empty string `q` would ever matter? But we could do `if (searchInput)` here if avoiding it is desirable.

static/js/pages/SitesDashboard.test.tsx
```diff
@@ -120,74 +118,73 @@ describe("SitesDashboard", () => {
...
-        const nextWrapper = wrapper.find(".pagination Link.next")
-        expect(nextWrapper.exists()).toBe(hasNextLink)
-        if (hasNextLink) {
-          expect(nextWrapper.prop("to")).toBe(
-            sitesBaseUrl
-              .query({ offset: startingOffset + WEBSITES_PAGE_SIZE })
-              .toString()
-          )
-        }
+  test.each([
```
ChristopherChudzicki, line 133: 🎉

ChristopherChudzicki (approved):
Looks good and works great! Left some minor comments.

static/js/pages/SitesDashboard.tsx
alicewriteswrongs: mmm that's definitely a little bit cleaner that way, I'll make that change

static/js/pages/SitesDashboard.tsx
alicewriteswrongs: I think you're right, the old results should stick around. and since I also changed it to show the surrounding initially it's less important to have a loading state I think

ChristopherChudzicki (approved):

PR #993, opened 2022-02-10
improvements to site search [backend]

### LINKED ISSUE #990: search on `/sites` needs to match on more fields

alicewriteswrongs:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #990 

this is sort of a twin PR to #992 but addresses the backend portion

#### What's this PR do?

This PR changes how we find results when the `search` parameter is passed to the `WebsiteListing` API. Previously we were doing an `icontains` filter on only the `title` field, which resulted in surprising behavior -- for instance, searching for course numbers would not return the relevant courses.

This changes things around a bit so that instead of a more basic filter we instead leverage Django's support for Postgres' built-in full text search. I set it up to search on the `title`, `name`, and `short_id` fields, which should hopefully give us what we need to provide a search that works the way course authors expect.

Check out the course search, and confirm the following:

- if you search for a course number you should get the relevant course back, hopefully as the first result.
- ditto with a fragment of the course's `title` (or the whole title) and
- ditto with the course's `name` 

It hopefully won't return anything unexpected or weird in the result set!

#### screencast

this just shows searching for a course number:

https://user-images.githubusercontent.com/6207644/153464473-116efa1f-3ce6-4951-b908-2355dc5f4928.mov

### DIFF
--- a/websites/views.py
+++ b/websites/views.py
@@ -5,6 +5,7 @@ import os
 
 from django.conf import settings
 from django.contrib.auth.models import Group
+from django.contrib.postgres.search import SearchQuery, SearchVector
 from django.db.models import Case, CharField, OuterRef, Q, Value, When
 from django.utils.functional import cached_property
 from django.utils.text import slugify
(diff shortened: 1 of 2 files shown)


ChristopherChudzicki (approved):
👍 matching on short ids and names works for me!

One thing I noticed is that this does change the search behavior to requeire matching on full words. For example, search "perspect" on master and courses with "perspective" in the title come back.

alicewriteswrongs:
oh that's a good point - I'll look at the options for the Django <-> Postgress full-text search and see if we can get some better matching. There are a fair few options that can set, so we can probably do better!

alicewriteswrongs:
just looked into it a bit and it seems there's not a great way to support fuzzy matches with the postgres full-text search apparently...so I think it probably makes sense to just merge this as is for now

PR #1002, opened 2022-02-14
add management command markdown_cleanup_baseurl

ChristopherChudzicki:
Converts baseurl links to resource_link shortcodes. Current version cleans up 90% of the baseurl links.

#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#### What's this PR do?
This PR adds a management command to clean up baseurl links in our markdown and convert them to resource_link shortcodes.

This management command converts about 90% of our current baseurl usage. The remaining 10% falls in two categories:

1. **Links we can convert soon:** baseurl links we can convert soon, but just aren't handled by this command yet. See https://github.com/mitodl/ocw-studio/pull/1002/files#r806218276 for more info.

2. **Links whose conversion to resource_link is blocked:** This includes things like links whose titles are images.

The management command can be invoked via
```
docker-compose run --rm web python manage.py markdown_cleanup_baseurl -o out.csv
```
This will:

1. Run the find-and-replace cleanup on all website content markdown, writing the replacements made as a csv `out.csv`
2. but NOT commit the changes

To commit the changes, add `--commit true`.

To test this locally, I would recommend running this on some courses and inspecting `out.csv`, then running in commit-mode and checking that markdown has changed as you would expect.  _If you want to run the commit multiple times, or restore your original data afterwards, dump your database beforehand and recreate the container a la #923_

For example, you can check the Calendar page in "Uncertainty in Engineering" on rc vs local studio. On rc the calendar links do not show up correctly in CKEditor, but after this management command, they will.

#### How I tested this
My primary concern when testing this was ensuring that it didn't screw up our data. And I found some issues with earlier versions, e.g., an earlier version was capturing way too much text as link title. I tested this pretty thoroughly and am now confident that the converter is correctly converting the stuff it changes.

What I did to test this was
- import all courses from `ocw-to-hugo-output-qa`
- run some queries on our website tables to investigate baseurl usage
- generate the csv of replacements
- imported csv into a psotgres db and did some analysis

The numbers below are based on 2642 courses imported on 2022-02-14:
```
92607 instances of 'baseurl' across 10533 markdown files in 2415 courses
91227 instances of 'baseurl' picked up by markdown_cleanup_baseurl's regex
83966 instances of 'baseurl' converted by markdown_cleanup_baseurl
```
Of the 83966 replacements made, all but 646 involved links whose
    - title was shorter than 80 characters, AND
    - title matched regex `[0-9\.a-zA-z\s]` (alphanumeric + period + space)

The 646 longer or more complicated titles I scanned over briefly. They all look reasonable.

#### Other Notes:
- took about 90 seconds locally on the 2600 courses 
- [CSV output from my testing](https://docs.google.com/spreadsheets/d/1VYMMiVkpY7jTrC6UlmgTFgHOcGCyNjqEoEeWCIaHtk8/edit?usp=sharing)

### DIFF
--- a/websites/management/commands/markdown_cleanup_baseurl.py
+++ b/websites/management/commands/markdown_cleanup_baseurl.py
@@ -128,11 +128,8 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         self.validate_options(options)
-        self.do_handle(**options)
-
-    @staticmethod
-    def do_handle(commit=False, out=None):
-        """Replace baseurl with resource_link"""
+        commit = options['commit']
+        out = options['out']
 
         with ExitStack() as stack:
             wc_list = WebsiteContent.all_objects.all().only(
--- a/websites/management/commands/markdown_cleanup_baseurl.py
+++ b/websites/management/commands/markdown_cleanup_baseurl.py
@@ -133,7 +133,7 @@ class Command(BaseCommand):
     @staticmethod
     def do_handle(commit=False, out=None):
         """Replace baseurl with resource_link"""
-        print(WebsiteContent.all_objects)
+
         with ExitStack() as stack:
             wc_list = WebsiteContent.all_objects.all().only(
                 "dirpath", "filename", "markdown", "website_id"
--- a/websites/management/commands/markdown_cleanup_baseurl.py
+++ b/websites/management/commands/markdown_cleanup_baseurl.py
@@ -144,17 +144,17 @@ class Command(BaseCommand):
 
             content_lookup = ContentLookup(wc_list)
             replacer = BaseurlReplacer(content_lookup)
-            clear = WebsiteContentMarkdownCleaner(
+            cleaner = WebsiteContentMarkdownCleaner(
                 BaseurlReplacer.baseurl_regex, replacer
             )
 
             wc: WebsiteContent
             for wc in progress_bar(wc_list):
-                clear.update_website_content_markdown(wc)
+                cleaner.update_website_content_markdown(wc)
 
             if commit:
-                wc_list.bulk_update(clear.updated_website_contents, ["markdown"])
+                wc_list.bulk_update(cleaner.updated_website_contents, ["markdown"])
 
         if out is not None:
             outpath = os.path.normpath(os.path.join(os.getcwd(), out))
-            clear.write_matches_to_csv(outpath)
+            cleaner.write_matches_to_csv(outpath)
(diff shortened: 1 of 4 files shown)


websites/management/commands/markdown_cleanup_baseurl.py
```diff
@@ -0,0 +1,131 @@
...
+        Example:
+        =======
+        content_lookup.get_content_by_url('some-uuid', '/resources/graphs/cos')
+        """
+
+        content_relative_dirpath, filename = os.path.split(content_relative_url)
+
+        dirpath = f"content{content_relative_dirpath}"
+
+        return self.website_contents[(website_id, dirpath, filename)]
```
ChristopherChudzicki, line 44: This expects that the after-baseurl portion of url `{{< baseurl >}}/path/to/the/filename` is of form `dirpath/filename`. This is true most of the time, but not always, and is something that should be improved.

For example, sometimes:
- `path/to/the/thing` is just `dirname`, and there's no `filename`. The corresponding websitecontent entry has filename `_index`
- sometime the url filename contains periods `path/to/u.s.-census-info` but the filename in our database has dashes periods replaced with dashes `u-s--census-info`
- sometimes the urlfiname has different casing, e.g., `path/to/Syllabus` where the syllabus file has filename `syllabus` with a lowercase s.

These should all be straightforward improvements that should get us from 90% conversion to high 90s.

websites/management/commands/util.py
```diff
@@ -0,0 +1,147 @@
...
+    print_progress_bar(0)
+    # Update Progress Bar
+    for i, item in enumerate(iterable):
+        yield item
+        print_progress_bar(i + 1)
+    # Print New Line on Complete
+    print()
+
+
+class WebsiteContentMarkdownCleaner:
```
ChristopherChudzicki, line 55: Making this class for one management command was perhaps a premature abstraction, but I am hoping we could reuse this for https://github.com/mitodl/ocw-to-hugo/issues/461 and other similar issues.

gumaerc (approved):
:+1: LGTM

Before we merge this though, can you create Github issues for the edge cases not caught by this PR, such as document fragments, etc.?  That way we can keep track of those and come back and address them in the future.  I also had one suggestion about the progress bar but you can take it or leave it:

ChristopherChudzicki:
> It didn't seem to save my comment about the progress bar, but essentially what I was thinking is that we might want to use something like https://github.com/tqdm/tqdm instead of rolling our own progress bar as it's kind of re-inventing the wheel, but it's up to you.

Definitely agree. I will include that in PR for #1012 .

ChristopherChudzicki:
> Before we merge this though, can you create Github issues for the edge cases not caught by this PR, such as document fragments, etc.?

https://github.com/mitodl/ocw-studio/issues/1015

PR #1013, opened 2022-02-16
support non-destructive editing of content w/ legacy shortcodes

alicewriteswrongs:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

(Required)

#### What's this PR do?

This implements support for nondestructively editing markdown files which have 'legacy' shortcodes in them (these are shortcodes inserted into converted content by ocw-to-hugo to support a series of things which aren't supported in plain markdown).

You'll need to get some shortcode-filled markdown into one of your pages and then test out that it displays like in the little video below.

- Arguments to shortcodes should be retained between saves
- shortcode opening / closing state should be saved

etc

This is a little snippet that you might want to try:

```
Soundness and Validity
----------------------
  
{{< quiz_multiple_choice questionId="Q1_div" >}}{{< quiz_choices >}}{{< quiz_choice isCorrect="false" >}}&nbsp; sound, satisfiable &nbsp;{{< /quiz_choice >}}
{{< quiz_choice isCorrect="false" >}}&nbsp; valid, satisfiable &nbsp;{{< /quiz_choice >}}
{{< quiz_choice isCorrect="true" >}}&nbsp; sound, valid &nbsp;{{< /quiz_choice >}}
{{< quiz_choice isCorrect="false" >}}&nbsp; valid, true &nbsp;{{< /quiz_choice >}}{{< /quiz_choices >}}
{{< quiz_solution / >}}{{< /quiz_multiple_choice >}}
```

it's an example of a quiz, which uses a whole bunch of shortcodes, opening and closing ones, with arguments, etc.

https://user-images.githubusercontent.com/6207644/154374070-73180ec2-fd79-4cff-858b-d756c3a452b4.mov

### DIFF
--- a/static/js/lib/ckeditor/plugins/LegacyShortcodes.ts
+++ b/static/js/lib/ckeditor/plugins/LegacyShortcodes.ts
@@ -1,0 +1,194 @@
+import { ShowdownExtension } from "showdown"
+import Plugin from "@ckeditor/ckeditor5-core/src/plugin"
+import { editor } from "@ckeditor/ckeditor5-core"
+
+import MarkdownSyntaxPlugin from "./MarkdownSyntaxPlugin"
+import { TurndownRule } from "../../../types/ckeditor_markdown"
+import { LEGACY_SHORTCODES } from "./constants"
+
+const shortcodeClass = (shortcode: string) => `legacy-shortcode-${shortcode}`
+
+const DATA_ISCLOSING = "data-isclosing"
+const DATA_ARGUMENTS = "data-arguments"
+
+class LegacyShortcodeSyntax extends MarkdownSyntaxPlugin {
+  static get pluginName(): string {
+    return "LegacyShortcodeSyntax"
+  }
+
+  get showdownExtension() {
+    return function legacyShortcodeExtension(): ShowdownExtension[] {
+      return LEGACY_SHORTCODES.map(
+        (shortcode: string): ShowdownExtension => {
+          const shortcodeRegex = new RegExp(`{{< /?${shortcode} (.*?)>}}`, "g")
+          const closingShortcodeRegex = new RegExp(
+            `{{< /${shortcode} (.*?)>}}`,
+            "g"
+          )
+
+          return {
+            type:    "lang",
+            regex:   shortcodeRegex,
+            replace: (stringMatch: string, shortcodeArgs: string | null) => {
+              const isClosing = JSON.stringify(
+                closingShortcodeRegex.test(stringMatch)
+              )
+
+              const tag = `<span ${DATA_ISCLOSING}="${isClosing}" ${
+                shortcodeArgs ?
+                  `${DATA_ARGUMENTS}="${encodeURIComponent(shortcodeArgs)}"` :
+                  ""
+              } class="${shortcodeClass(shortcode)}"></span>`
+
+              return tag
+            }
+          }
+        }
+      )
+    }
+  }
+
+  get turndownRules(): TurndownRule[] {
+    return LEGACY_SHORTCODES.map(shortcode => ({
+      name: `LegacyShortcodeSyntax-${shortcode}`,
+      rule: {
+        filter: node => {
+          return (
+            node.nodeName === "SPAN" &&
+            node.className === shortcodeClass(shortcode)
+          )
+        },
+        replacement: (_content: string, node: any): string => {
+          const isClosingTag = JSON.parse(node.getAttribute(DATA_ISCLOSING))
+          const rawShortcodeArgs = node.getAttribute(DATA_ARGUMENTS)
+
+          return `{{< ${isClosingTag ? "/" : ""}${shortcode} ${
+            rawShortcodeArgs !== undefined && rawShortcodeArgs !== null ?
+              decodeURIComponent(rawShortcodeArgs) :
+              ""
+          }>}}`
+        }
+      }
+    }))
+  }
+}
+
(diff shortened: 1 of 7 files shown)


static/js/lib/ckeditor/plugins/LegacyShortcodes.ts
```diff
@@ -0,0 +1,194 @@
...
+
+          return el
+        }
+      })
+    })
+  }
+}
+
+export default class LegacyShortcodes extends Plugin {
+  static get requires(): Plugin[] {
```
ChristopherChudzicki, line 186: **Comment:** *I'm not suggesting we should necessarily worry about this now... just writing it down*

I think there's something weird here... I think the `Plugin[]` type annotation is referring to `globalThis.Plugin`. https://developer.mozilla.org/en-US/docs/Web/API/Plugin , not to CKEditor plugins.

static/js/lib/ckeditor/plugins/LegacyShortcodes.ts
```diff
@@ -0,0 +1,194 @@
...
+            type:    "lang",
+            regex:   shortcodeRegex,
+            replace: (stringMatch: string, shortcodeArgs: string | null) => {
+              const isClosing = JSON.stringify(
+                closingShortcodeRegex.test(stringMatch)
+              )
+
+              const tag = `<span ${DATA_ISCLOSING}="${isClosing}" ${
+                shortcodeArgs ?
+                  `${DATA_ARGUMENTS}="${escape(shortcodeArgs)}"` :
```
ChristopherChudzicki, line 39: **Question/Suggestion**: I was unfamiliar with `escape` and VSCode highlighted it, so I wlooked it up on MDN. MDN says not to use it, but doesn't really explain why. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape Would `encodeURIComponent` work instead?

ChristopherChudzicki (changes requested):
👍 The looks good to me and is working well in local trials.

One request: I think we should add baseurl to the legacy list. No baseurls have been converted yet (hopefully will today) but even after that conversion some will remain for a while.

static/js/lib/ckeditor/plugins/constants.ts
```diff
@@ -64,3 +64,25 @@ export const TABLE_ALLOWED_ATTRS: string[] = ["colspan", "rowspan"]
  * with a double quote and captures anything in between the quotes.
  */
 export const ATTRIBUTE_REGEX = /(\S+)=["']?((?:.(?!["']?\s+(?:\S+)=|\s*\/?[>"']))+.)["']?/g
+
+export const LEGACY_SHORTCODES = [
```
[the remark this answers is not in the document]
alicewriteswrongs: ok cool, I was too optimistic 😃

static/js/lib/ckeditor/plugins/LegacyShortcodes.ts
alicewriteswrongs: oooo

alicewriteswrongs: that's why it's throwing errors...I'll take a minute to see if I can fix that

static/js/lib/ckeditor/plugins/constants.ts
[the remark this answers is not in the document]
ChristopherChudzicki: > also am I right in thinking that there's no harm in having it in the list even after the issue has been address via your PR?

yes, i agree.

static/js/lib/ckeditor/plugins/LegacyShortcodes.ts
alicewriteswrongs: weird...so I changed `import Plugin from ...` to `import { default as CKEditorPlugin } from ...` and then I get an error about being unable to use that type in a static method...I think I might just leave this off for now in the interest of time

but it would be good to fix!

static/js/lib/ckeditor/plugins/LegacyShortcodes.ts
alicewriteswrongs: I think we can use encoreURIComponent, I'll swap that in and make sure the tests pass real quick. The reason to do it is just that if it's not escaped then the text w/in the arguments to a shortcode can cause issues correctly finding the boundaries of the HTML tags later on and things get all messed up. When I was hacking on it I actually used `btoa` and `atob` to be 100% sure I wasn't going to get any characters I didn't expect in the arguments string, haha

static/js/lib/ckeditor/plugins/constants.ts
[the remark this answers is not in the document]
gumaerc: If it's not there then the rule won't apply right?  There shouldn't be any harm to simply including it in this list, as even if we have a need for it in the future for some reason it would be manually inserted by us and won't be touched by editors.

gumaerc (approved):
I'm approving this, as it tested out well for me locally.  I think the things @ChristopherChudzicki noted should be checked out still though.  The `escape` thing isn't critical as it's not going away anytime soon, but might be better to use `encodeUriComponent`.

static/js/lib/ckeditor/plugins/LegacyShortcodes.ts
alicewriteswrongs: yep, I already made the change since all the tests pass with it

PR #1016, opened 2022-02-16
convert resource_file to resource shortcodes

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#### What's this PR do?
This refactors `markdown_cleanup_baseurl` management command to be `markdown_cleanup` which should now be invoked with one of multiple rules. Right now, one of:
```
# Either "baseurl" or "resource_file_to_resource"
# plus the old
docker-compose run --rm web python manage.py markdown_cleanup baseurl -o out.csv
docker-compose run --rm web python manage.py markdown_cleanup resource_file_to_resource -out.csv
```
or with `--commit` to commit the result.

There are no changes to the `baseurl` command besides refactoring. I have tested it locally with a baseurl-intact database.

Run the commands above and verify resource_file links are converted as expected.

The "Instructor Insights" page of `8-591j-systems-biology-fall-2014` is a good one to check in studio. See screenshots below.

**Before**
Screen Shot 2022-02-16 at 4 10 02 PM

**After**
Screen Shot 2022-02-16 at 4 09 22 PM


websites/management/commands/markdown_cleanup.py
```diff
@@ -0,0 +1,97 @@
...
+    WebsiteContentMarkdownCleaner,
+)
+from websites.management.commands.markdown_cleaning.resource_file_rule import (
+    ResourceFileReplacementRule,
+)
+from websites.management.commands.util import progress_bar
+from websites.models import WebsiteContent
+
+
+class Command(BaseCommand):
```
ChristopherChudzicki, line 22: This command is mostly the same as the old `markdown_cleanup_baseurl.py`, except it has a new positional argument "alias" that selects a rule from the `Rules` class property.

websites/management/commands/markdown_cleaning/baseurl_rule.py
```diff
@@ -67,19 +61,28 @@ def get_content_by_url(self, website_id, content_relative_url: str):
...
-    baseurl_regex = r"\[(?P<title>[^\[\]\n]*?)\]\({{< baseurl >}}(?P<url>.*?)\)"
+    regex = r"\[(?P<title>[^\[\]\n]*?)\]\({{< baseurl >}}(?P<url>.*?)\)"
+
+    alias = "baseurl"
 
-    def __init__(self, content_lookup: ContentLookup):
-        self.content_lookup = content_lookup
+    def __init__(self):
+        website_contents = get_all_website_content()
+        self.content_lookup = ContentLookup(website_contents)
```
ChristopherChudzicki, line 85: Previously the django command was passing `website_contents` to this replacer function. 

Now:
1. The django command gets all `website_content` so that it can iterate over them, making replacements
2. separately, this replacer function also gets all `website_contents` so that it can create a lookup to find content by url.

So there's some redundant fetching from the database... but it seemed simpler to make the rule responsible for getting its own data if it needs it.

gumaerc (approved):
:+1: LGTM

In my local DB I had 3349 matches in my CSV, all of which looked like reasonable replacements

PR #1023, opened 2022-02-17
filter out unpublished content in the website content listing API

### LINKED ISSUE #520: relation widget allows selection of unpublished content across websites

alicewriteswrongs:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #520

#### What's this PR do?

This adds a query param to the website content listing API called `published` which filters the content returns based on whether it has been published or not.

Note that we do not specifically store data about whether a specific piece of content has been published. Instead, I came up with a heuristic that I think should cover us in most cases.

Basically, we defined published content to be content which was created _before_ the website to which it is attached was most recently published. I think this ensures that, even if the _current state of the content on studio_ hasn't been published, the content must have been published in the past _in some form_.

I believe this will be safe to use to determine whether to create a reference to the content in a relation widget (remember that content linked to via a relation widget is resolved at hugo-build-time via the `data.json` file for that content --- in order for that `data.json` file to be present that particular piece of content must have been previously published).

Anyhow, this PR implements that sorting on the website content listing API and then also sets the relation widget to set `published=true` so that it will only allow the user to select published content.

Please read through the code and make sure my query logic makes sense!

Also, test that you cannot get a reference to unpublished course content (unpublished according to the information in Studio and the logic described above) to appear in the course collection relation widget on ocw-www.

### DIFF
--- a/websites/views.py
+++ b/websites/views.py
@@ -6,7 +6,7 @@ import os
 from django.conf import settings
 from django.contrib.auth.models import Group
 from django.contrib.postgres.search import SearchQuery, SearchVector
-from django.db.models import Case, CharField, OuterRef, Q, Value, When
+from django.db.models import Case, CharField, F, OuterRef, Q, Value, When
 from django.utils.functional import cached_property
 from django.utils.text import slugify
 from guardian.shortcuts import get_groups_with_perms, get_objects_for_user
@@ -441,6 +441,7 @@ class WebsiteContentViewSet(
         search = self.request.query_params.get("search")
         resourcetype = self.request.query_params.get("resourcetype")
         types = _get_value_list_from_query_params(self.request.query_params, "type")
+        published = self.request.query_params.get("published", None)
 
         queryset = WebsiteContent.objects.filter(
             website__name=parent_lookup_website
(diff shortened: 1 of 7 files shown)


websites/views.py
```diff
@@ -460,7 +461,24 @@ def get_queryset(self):
...
+                    website__publish_date__gt=F("created_on"),
+                )
+            else:
+                # unpublished content is any content where the website has never
+                # been published
+                queryset = queryset.filter(
+                    website__publish_date__isnull=True
+                    # or any content where the website was most recently published
+                    # before it was created
+                ) | queryset.filter(website__publish_date__lt=F("created_on"))
```
mbertrand, line 481: I think this might give identical results, take or leave:
```python
queryset = queryset.filter(
 Q(website__publish_date__isnull=True) | Q(website__publish_date__lt=F("created_on"))
 )

mbertrand (approved):
Works great, just 1 minor suggestion

PR #1024, opened 2022-02-17
add markdown cleanup rule for legacy data fix

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#### What's this PR do?
This PR fixes extra escapes caused by the legacy-shortcode editing issue.

I suggest:

Run
```
docker-compose exec web python manage.py markdown_cleanup legacy_shortcode_datafix_1_of_2 -o one.csv
docker-compose exec web python manage.py markdown_cleanup legacy_shortcode_datafix_2_of_2 -o two.csv
```
against a database with some corrupted shortcodes. For example, 
```
\[!\[Small groups of students talking at tables in a classroom. In the foreground, an instructor stands near a pair of students talking to each other.\]({{\< resource\_file 68aab975-18a3-39e4-91a6-bd54c8f35356 >}})\]({{\<
```

Or if you want to test it against production backup:

```
docker-compose stop
docker-compose rm -sv db
docker-compose up db -d
docker-compose exec -T -e PGPASSWORD=postgres db psql -h db -U postgres < prod_dump.sql
```

### DIFF
--- a/websites/management/commands/markdown_cleaning/legacy_shortcodes_data_fix.py
+++ b/websites/management/commands/markdown_cleaning/legacy_shortcodes_data_fix.py
@@ -1,0 +1,52 @@
+"""
+These rules are intended to help fix data after the legacy shortocde incident.
+Prior to https://github.com/mitodl/ocw-studio/pull/1013, "legacy" shortcodes
+like {{< baseurl >}} would become corrupted in studio during editing. The editor:
+    - displayed the shortcodes as raw text in studio edit
+    - then escaped special characters like "[" and "<" when saving (only if edits were made)
+
+This rule selectively unescapes content.
+"""
+import re
+
+from websites.management.commands.markdown_cleaning.cleanup_rule import (
+    MarkdownCleanupRule,
+)
+
+
+LINK_WITHOUT_IMAGE = r"\\\[[^\[\]]*\\\]\({{\\<"
+LINK_WITH_IMAGE = r"\\\[!\\\[[^\[\]]*\][^\[\]]*\\\]\({{\\<"
+
+
+class LegacyShortcodeFixOne(MarkdownCleanupRule):
+    """
+    Use this BEFORE LegacyShortcodeTwo because this matches against "{{\\<" for
+    increased selectivity.
+    """
+
+    regex = f"{LINK_WITH_IMAGE}|{LINK_WITHOUT_IMAGE}"
+
+    alias = "legacy_shortcode_datafix_1_of_2"
+
+    def __call__(self, match: re.Match, _website_content):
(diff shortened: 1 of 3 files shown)


websites/management/commands/markdown_cleaning/legacy_shortcodes_data_fix.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+
+    regex = f"{LINK_WITH_IMAGE}|{LINK_WITHOUT_IMAGE}"
+
+    alias = "legacy_shortcode_datafix_1_of_2"
+
+    def __call__(self, match: re.Match, _website_content):
+        original_text = match[0]
+        fixed = (
+            original_text.replace("\\[", "[").replace("\\]", "]").replace("\\_", "_")
```
gumaerc, line 34: I think we might be missing a replace rule here for escaped angle braces in the shortcode opening delimiters.  In my testing with the prod database, I saw lines like this:

`"\[A Proposal\]({{\<","[A Proposal]({{\<","bfa33eaf-740c-6b15-4330-2ba484afaf5f","a099bbd6-c091-427d-8ef2-c398b6566ed6"`

The replacement doesn't remove the backslash from before the angle brace in the opening delimiter of the shortcode.

gumaerc (changes requested):
Overall this looks good, but I noticed one replacement that might have been overlooked:

websites/management/commands/markdown_cleaning/legacy_shortcodes_data_fix.py
ChristopherChudzicki: This was intentional... thought there very well might be a better way to organize this logic.

In general, I think there are two things we need to fix. Consider the texts
```
blah \[Some Tile\]{{\< resource\_file uuid >}} blah
blah blah {{\< sup 2 >}} blah
```
The two things are:
1. stuff before the shortcode, namely link/image title
2. the shortcode and its contents

The `\<` in the first line will be caught by the second rule.

If I make the first rule fix the angle bracket, the the first rule would also need to go to the end of the shortocde so it can fix the `\_`. Which we can do, but it seemed simple to keep the "fix before stuff" and "fix shortcode stuff" separate.

Though, it's definitely a little awkward that rules expect to be run in a particular order, but that’s why I gave them the 1_of_2 and 2_of_2 aliases.

gumaerc: Gotcha, thanks for the explanation.  This makes sense now.

gumaerc (approved):
:+1: LGTM

PR #1036, opened 2022-02-18
convert baseurl links w/ fragments

ChristopherChudzicki:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

#### What's this PR do?
Add markdown_cleanup rule to convert baseurl links with fragments `[title]({{< baseurl >}}/pages/whatever/file#some-heading)` to resource_links `< resource_link uuid "title" "some-heading" >`.

Includes changes to our CKeditor so that these resource_links **display** properly. But DOES NOT support saving these links. Currently saving markdown like `< resource_link uuid "title" "some-heading" >` results in the `some-heading` portion being lost. (@alicewriteswrongs is working on the necessary change to our CKEditor ResourceLink plugin.)

Run
```
# To generate csv of replacements only
docker-compose exec web python manage.py markdown_cleanup baseurl -o meow2.csv
# to commit changes to database
docker-compose exec web python manage.py markdown_cleanup baseurl --commit
```
and view the changes in Studio editor.

Some pages that would be affected:
```
b7cc5225-96b8-bb67-252f-d5048c2b43f4
    course: The Biology of Aging: Age-Related Diseases and Interventions
    page: Readings
df5e63ff-8c3f-cf7e-32cd-20fae4ed96c7
    course: Design of Electromechanical Robotic Systems
    page: Calendar
db02fb8f-2784-8eb9-d53c-72fe22be0540
    Immune Cell Migration: On the Move in Response to Pathogens and Cancer Immunotherapy
    Readings
```

### DIFF
--- a/static/js/lib/ckeditor/plugins/ResourceLinkMarkdownSyntax.test.ts
+++ b/static/js/lib/ckeditor/plugins/ResourceLinkMarkdownSyntax.test.ts
@@ -6,7 +6,9 @@ import { createTestEditor, markdownTest } from "./test_util"
 import { turndownService } from "../turndown"
 
 import { RESOURCE_LINK } from "@mitodl/ckeditor5-resource-link/src/constants"
-import ResourceLinkMarkdownSyntax from "./ResourceLinkMarkdownSyntax"
+import ResourceLinkMarkdownSyntax, {
+  encodeShortcodeArgs as encode
+} from "./ResourceLinkMarkdownSyntax"
 import Paragraph from "@ckeditor/ckeditor5-paragraph/src/paragraph"
 
 const getEditor = createTestEditor([
@@ -45,8 +47,9 @@ describe("ResourceLink plugin", () => {
     markdownTest(
       editor,
       '{{< resource_link 1234-5678 "link text" "#some-header-id" >}}',
-      `<p><a class="resource-link" data-uuid="${encodeURIComponent(
-        JSON.stringify(["1234-5678", "#some-header-id"])
+      `<p><a class="resource-link" data-uuid="${encode(
+        "1234-5678",
+        "#some-header-id"
       )}">link text</a></p>`
     )
   })
@@ -56,7 +59,9 @@ describe("ResourceLink plugin", () => {
     markdownTest(
       editor,
       '{{< resource_link asdfasdfasdfasdf "text here" >}}',
-      '<p><a class="resource-link" data-uuid="asdfasdfasdfasdf">text here</a></p>'
+      `<p><a class="resource-link" data-uuid="${encode(
+        "asdfasdfasdfasdf"
+      )}">text here</a></p>`
     )
   })
 
@@ -65,7 +70,11 @@ describe("ResourceLink plugin", () => {
     markdownTest(
       editor,
       'dogs {{< resource_link uuid1 "woof" >}} cats {{< resource_link uuid2 "meow" >}}, cool',
-      '<p>dogs <a class="resource-link" data-uuid="uuid1">woof</a> cats <a class="resource-link" data-uuid="uuid2">meow</a>, cool</p>'
+      `<p>dogs <a class="resource-link" data-uuid="${encode(
+        "uuid1"
+      )}">woof</a> cats <a class="resource-link" data-uuid="${encode(
+        "uuid2"
+      )}">meow</a>, cool</p>`
     )
   })
 
@@ -75,7 +84,9 @@ describe("ResourceLink plugin", () => {
       .processor as unknown) as MarkdownDataProcessor
     expect(md2html('{{< resource_link uuid123 "bad \\" >}}')).toBe(
       // This is wrong. Should not end in &lt;/a&gt;
-      '<p><a class="resource-link" data-uuid="uuid123">bad &lt;/a&gt;</p>'
+      `<p><a class="resource-link" data-uuid="${encode(
+        "uuid123"
+      )}">bad &lt;/a&gt;</p>`
     )
   })
 })
--- a/static/js/lib/ckeditor/plugins/ResourceLinkMarkdownSyntax.test.ts
+++ b/static/js/lib/ckeditor/plugins/ResourceLinkMarkdownSyntax.test.ts
@@ -44,9 +44,9 @@ describe("ResourceLink plugin", () => {
 
     markdownTest(
       editor,
-      '{{< resource_link 1234-5678 "link text" "some-header-id" >}}',
+      '{{< resource_link 1234-5678 "link text" "#some-header-id" >}}',
       `<p><a class="resource-link" data-uuid="${encodeURIComponent(
-        "1234-5678#some-header-id"
+        JSON.stringify(["1234-5678", "#some-header-id"])
       )}">link text</a></p>`
     )
   })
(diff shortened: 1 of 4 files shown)


websites/management/commands/markdown_cleaning/baseurl_rule_test.py
```diff
@@ -43,13 +43,13 @@ def get_markdown_cleaner(website_contents):
...
-            R"This link includes a fragment [text title]({{< baseurl >}}/resources/path/to/file1#some-fragment) cool",
+        (
             R"This link includes a fragment [text title]({{< baseurl >}}/resources/path/to/file1#some-fragment) cool",
+            R'This link includes a fragment {{< resource_link content-uuid-1 "text title" "some-fragment" >}} cool',
         ),
-        (  # should not touch fragments with / before #
-            R"This link includes a fragment with slash first [text title]({{< baseurl >}}/resources/path/to/file1/#some-fragment) cool",
+        (
             R"This link includes a fragment with slash first [text title]({{< baseurl >}}/resources/path/to/file1/#some-fragment) cool",
+            R'This link includes a fragment with slash first {{< resource_link content-uuid-1 "text title" "some-fragment" >}} cool',
```
gumaerc, line 52: I was looking at the updated `resource_link.html` shortcode in `ocw-hugo-themes`:

```
{{- $uuid := index .Params 0 -}}
{{- $title := index .Params 1 -}}
{{- $anchor_id := index .Params 2 | default ""}}
{{- range where $.Site.Pages "Params.uid" $uuid -}}
    <a href="{{- .Permalink -}}{{- $anchor_id -}}">{{ $title }}</a>
{{- end -}}

```

We're fetching `fragment_id` there from the second argument, but it looks like the fragment itself here doesn't include the `#` character.  Are we expecting that character to be in the template, or passed in?  If we're not going to put it in the theme, this new argument could be used even to just put in any query string on the URL.  Either way, it doesn't currently look like the `#` will be preserved, or am I missing something?

gumaerc (changes requested):
Question about how the URL is going to be reconstructed in the theme:

websites/management/commands/markdown_cleaning/baseurl_rule_test.py
ChristopherChudzicki: Ahhhhh good catch. I was testing this in studio but hadn't followed a link yet.

And yes, I agree—leaving the second arg flexible for `#` or `?` is good.

ChristopherChudzicki:
@gumaerc Thanks for catching this. I've made two updates to the PR:
- change slightly how `ResourceLinkMarkdownSyntax` was handling the non-uuid fragment/hash portion. Before we were relying on splitting at `'#'`. Now we aren't, which makes it a little easier to include the `'#'` and would allow `?` querystrings in future if needed
- changed the management command to keep the `'#'`

I've now tested this with ocw-hugo-thmes locally using https://github.mit.edu/ocw-content-rc/16.90-spring-2014 ... resource _links link-to-sections are working for me in ocw-hugo-themes now. Should have tested that before.

static/js/lib/ckeditor/plugins/ResourceLinkMarkdownSyntax.test.ts
```diff
@@ -75,7 +84,9 @@ describe("ResourceLink plugin", () => {
       .processor as unknown) as MarkdownDataProcessor
     expect(md2html('{{< resource_link uuid123 "bad \\" >}}')).toBe(
       // This is wrong. Should not end in &lt;/a&gt;
-      '<p><a class="resource-link" data-uuid="uuid123">bad &lt;/a&gt;</p>'
+      `<p><a class="resource-link" data-uuid="${encode(
```
ChristopherChudzicki, line 87: Using `encodeURIComponent(JSON.stringify(args))` for multiple args seems simple, but it means that `'uuid1'` shows up as `'%5B%22uuid1%22%5D'` instead of just `'uuid1'. We could probably avoid this if we really want, but handling 1 args the same as 2 args seems clean, if aesthetically unappealing in tests.

gumaerc: I'm a little confused by this, what you're saying is that because it's encoding both `class` and `data-uuid` that it will include the quotations for both as part of the string.?

gumaerc (approved):
:+1: LGTM

This looks good now, although we should consider renaming `$anchor_id` in `resource_link.html` in `ocw-hugo-themes` to `$queryString` instead.  I just have one question about one of your comments for my own clarification:

static/js/lib/ckeditor/plugins/ResourceLinkMarkdownSyntax.test.ts
ChristopherChudzicki: _I'm going to write this a slightly longer reply because I also want to share it with Alice, whom I have not talked to about this change yet but should probably make aware_

I'm not sure what `class` you refer to, but btw: `data-uuid` is an outdated name. Previously it was a uuid. Now it's a string that encodes a uuid and possibly more information (like the anchor). Changing the name of that attribute would break our CKEditor ResourceLink plugin (which is in a different repo, https://github.com/mitodl/ckeditor5). We probably should change the attribute name from `data-uuid` to something else, like `data-link` (which ironically might have been the original name in the Link plugin, I'm not sure). 

If that sounds weird...

My understanding from talking with @alicewriteswrongs is: In CKEditor, links are modeled as an attribute on text and text in CKEditor can only have one attribute. For our resource link stuff, the value of that attribute is determined by the `data-uuid` attribute on the anchor tag.

```
# to/from CKEditor:
Markdown --> HTML --> CKEditor internal representation
         <--      <--
1. convert shortcode to text + metadata (metadata = uuid + anchor)
2. store metadata on anchor data-uuid attribute
3. CKEditor uses the data-uuid attribute as the single attribute for the
   link text in its internal representation
```
The important thing is: In CKEditor's model of a link, it can store the text and one other value. (No more than one other value). For normal links, that "one value" is the url. For our resource_links, that one value was previously the uuid.

But now we need to store two pieces of information in that one value: the uuid and the fragment id.

The way ResourceLink plugin is set up, the "one value" is the same as "data-uuid" attribute on the `<a></a>` tag. 

In Alice's PR (#1037) the two pieces of information were encoded as: `uuid#fragment`, i.e.,  `"piece1#piece2" with `'#'` as the separator. That was really simple but (1) makes including. the '#' piece in the shortcode value slightly harder and also precludes us from using this for search params `?foo=1&bar=2` if we ever needed to. 

So in this PR, I switched the encoding to just use JSON.stringify/parse to encode/decode the multiple pieces (uuid+anchor) to/from a single value. Except the output of JSON.stringify be an HTML attribute value since it has quotations, so I had to use `encodeURIComponent/decode...`. 

This works well for passing 2 (or more) args as a single value to/from CKEditor.

All I meant in my original comment was: **It also works for passing 1 argument, but it's kinda ugly** because the single argument also has the quotations and square brackets from JSON.stringify:
```
const uuid = 'uuid1'
const asJson = JSON.stringify(['uuid1']) // stringify an array for consistency with the two-argument case
// out: the string '["uuid1"]'
const encoded = encodeURIComponent(asJson)
// out: the string '%5B%22uuid1%22%5D'
```

gumaerc: @ChristopherChudzicki thanks for the explanation, I now have a greater understanding of how this works and what you meant by a single argument being "ugly."  I don't think it matters as long as the data comes out as expected on the other side.

Issue #1020, opened 2022-02-17
add filtering / searching, sorting to the WebsiteContent listing UI page


alicewriteswrongs:
As a course author, I'd like to navigate to something specific. Right now once you have more than 20 or so resources that becomes very difficult! You have to just page through until you find it.

#### Designs and Mockups

UI can be very similar to the filter on `/sites`

#### Acceptance Criteria:

- [ ] we should add some predictable sort order and some controls (MRU, LRU, Created ascending/descending, alphabetical, others?)
- [ ] we should add a filter input (like we have on `/sites`) which searches the website content items and makes it easy to navigate to them. Should we support some type of full text support? Title matching might be good enough to start.

alicewriteswrongs:
A good deal of this is already done, the only thing remaining would be the sort order if we wanted to add that.

PR #1046, opened 2022-02-22
add content filtering to the website content listing page

alicewriteswrongs:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

for  #1020

#### What's this PR do?

this PR implements a basic sort UI on the content listing pages. I pulled out the logic I wrote for the same type of sort on the `/sites` page into a hook and then added it to the content page, so it works the same way.

go to a content page like `/sites/$SITENAME/type/page` and try it out!

Screen Shot 2022-02-22 at 11 50 09 AM

### DIFF
--- a/static/js/components/RepeatableContentListing.test.tsx
+++ b/static/js/components/RepeatableContentListing.test.tsx
@@ -1,8 +1,8 @@
 import React from "react"
 import { act } from "react-dom/test-utils"
-import useInterval from "@use-it/interval"
+import { default as useIt } from "@use-it/interval"
 import sinon from "sinon"
-import { useLocation } from "react-router-dom"
+import * as rrDOM from "react-router-dom"
 
 import RepeatableContentListing from "./RepeatableContentListing"
 import {
@@ -12,7 +12,7 @@ import {
 import WebsiteContext from "../context/Website"
 
 import useConfirmation from "../hooks/confirmation"
-import { isIf, shouldIf } from "../test_util"
+import { twoBooleanTestMatrix } from "../test_util"
 import {
   siteApiContentDetailUrl,
   siteContentListingUrl,
@@ -41,6 +41,10 @@ import {
 } from "../types/websites"
 import { createModalState } from "../types/modal_state"
 import configureStore from "../store/configureStore"
+import { singular } from "pluralize"
+
+const { useLocation } = rrDOM as jest.Mocked<typeof rrDOM>
+const useInterval = useIt as jest.Mocked<typeof useIt>
 
 // ckeditor is not working properly in tests, but we don't need to test it here so just mock it away
 function mocko() {
@@ -100,12 +104,9 @@ describe("RepeatableContentListing", () => {
       setConfirmationModalVisible,
       conditionalClose
     })
-    // @ts-ignore
     useLocation.mockClear()
     // @ts-ignore
-    useLocation.mockReturnValue({
-      pathname: "/path/to/pages"
-    })
+    useLocation.mockReturnValue({ pathname: "/path/to/pages" })
 
     const listingParams = {
       name:   website.name,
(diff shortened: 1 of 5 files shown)


static/js/components/RepeatableContentListing.test.tsx
```diff
@@ -254,6 +255,31 @@ describe("RepeatableContentListing", () => {
...
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+      preventDefault() {},
+      target: { value: "my-search-string" }
+    } as React.ChangeEvent<HTMLInputElement>
+    filterInput.simulate("change", event)
+    await wait(800)
+    wrapper.update()
+    expect(wrapper.find(".site-search-input").prop("value")).toBe(
+      "my-search-string"
+    )
```
ChristopherChudzicki, line 280: **Question / Requested Change**: What's the goal of this test? I think it might not be doing what is expected. 

Doesn't the input update immediately, but the history / url bar has debounced updating? I tried this test out and the assertion passes without the `wait(800)`.

Is the goal to test the debouncing history thing? If yes, two suggestions:
- I think `jest.spyOn(helper.browserHistory, 'push')` should let us spy on the url bar... but maybe theres a better way. Unsure.
- Instead of using `wait`, you could use `jest.useFakeTimers` and `jest.advanceFakeTimers` ... :timelord:

    Although, I imagine `wait` isn't a significant contributor to our test runtime 🤷 

Anyway, requested change is: remove the wait(800) if the test is the way it is is supposed to be, or make it fail without the wait(800).

static/js/components/RepeatableContentListing.test.tsx
```diff
@@ -254,6 +255,31 @@ describe("RepeatableContentListing", () => {
...
+      expect(li.text()).toContain(item.title)
+    })
+  })
+
+  test("should let the user filter via text input", async () => {
+    const { wrapper } = await render()
+    const filterInput = wrapper.find(".site-search-input")
+    const event = {
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+      preventDefault() {},
```
ChristopherChudzicki, line 272: **Random question:** I noticed `useTextInputState` in your hook, so I went to go look at it... do you know why we call `preventDefault` in `useTextInputState`? When is preventing default on a text input important?

ChristopherChudzicki (changes requested):
Nice 🪝 ... So cool that it can be reused from the dashboard to the listings.

Ferdi:
why just "Pages" ?

static/js/components/RepeatableContentListing.test.tsx
alicewriteswrongs: I had a heck of a time with this test! I'll give `spyOn` a try, but it was the strangest thing - although I had an almost identical test to the corresponding one for the `SitesDashboard`, for some reason the calls to `history.push` in the hook weren't changing any kind of history anywhere that I could pull out and assert, so I got a little frustrated and then just asserted that the new search string was present as the `value` prop on the input. I'll give it another go today :)

alicewriteswrongs:
@Ferdi it's not just present on pages, I just neglected to screenshot them :)

Here's what it looks like elsewhere:

Screen Shot 2022-02-23 at 9 14 05 AM

Screen Shot 2022-02-23 at 9 25 48 AM

static/js/components/RepeatableContentListing.test.tsx
alicewriteswrongs: ok `spyOn` worked, but still if I `console.log(helper.browserHistory.location)` the `.search` prop is `""`... 🤷‍♀️

ChristopherChudzicki (approved):

PR #1047, opened 2022-02-22
Update ContentSyncState checksums when bulk updating WebsiteContent

### LINKED ISSUE #1040: Multiple git commits for one small change
### Steps to Reproduce

1. I made a small change to the ocw-www site on RC, resulting in this commit in the main branch https://github.mit.edu/ocw-content-rc/ocw-www/commit/dfffdf08a61f89362b2251a63fe8f860052ed1ff
2. Publishing to the staging server, resulted in 8 additional, empty commits to the main branch, starting with https://github.mit.edu/ocw-content-rc/ocw-www/commit/94911d020189310d713f7e3b09045dec7e8df243 and ending with https://github.mit.edu/ocw-content-rc/ocw-www/commit/4233cbdd8a4333dcf49e82927671f0b7ec75115f
3. There was one additional, merge commit on the preview branch: https://github.mit.edu/ocw-content-rc/ocw-www/commit/9d53b62a88bd73470c5a7e0762aeac684be16033

### Expected Behavior

I think only one commit is necessary, to the main branch, followed by one merge commit to the preview branch

### Actual Behavior

9 commits? And why are the authors @mbertrand and @gumaerc ?

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Fixes #1040 

#### What's this PR do?
Bulk updates `ContentSyncState.current_checksum` values, while running the `markdown_cleanup` management command.

In a shell, pick a site, save a page's markdown to a corrupted value, and set checksums for `WebsiteContent` objects with markdown to the wrong value:
```
from content_sync.models import *
from websites.models import *
content = WebsiteContent.objects.filter(type="page").first()
content.markdown = '\[!\[Small groups of students talking at tables in a classroom. In the foreground, an instructor stands near a pair of students talking to each other.\]({{\< resource\
    ...: _file 68aab975-18a3-39e4-91a6-bd54c8f35356 >}})\]({{\<'
content.save()
sync_state = content.content_sync_state
sync_state.current_checksum = "wrong"
sync_state.save()
ContentSyncState.objects.filter(current_checksum="wrong").count()
```

Run `docker-compose exec web python manage.py markdown_cleanup legacy_shortcode_datafix_1_of_2 -o one.csv -c`

Check that the checksum values are no longer "wrong".
```
ContentSyncState.objects.filter(current_checksum="wrong").count()
```

### DIFF
--- a/websites/management/commands/markdown_cleaning/baseurl_rule_test.py
+++ b/websites/management/commands/markdown_cleaning/baseurl_rule_test.py
@@ -3,6 +3,7 @@ from unittest.mock import Mock, patch
 
 import pytest
 
+from content_sync.factories import ContentSyncStateFactory
 from websites.factories import WebsiteContentFactory, WebsiteFactory
 from websites.management.commands.markdown_cleaning.baseurl_rule import (
     CONTENT_FILENAME_MAX_LEN,
@@ -78,6 +79,7 @@ def test_baseurl_replacer_specific_title_replacements(markdown, expected_markdow
     target_content = WebsiteContentFactory.build(
         markdown=markdown, website_id=website_uuid
     )
+    ContentSyncStateFactory.build(content=target_content)
 
     linkable = WebsiteContentFactory.build(
         website_id=website_uuid,
@@ -119,6 +121,7 @@ def test_baseurl_replacer_handle_specific_url_replacements(
     target_content = WebsiteContentFactory.build(
         markdown=markdown, website_id=website_uuid
     )
+    target_sync_state = ContentSyncStateFactory.build(content=target_content)
 
     linkable = WebsiteContentFactory.build(
         website_id=website_uuid,
@@ -131,6 +134,7 @@ def test_baseurl_replacer_handle_specific_url_replacements(
     cleaner.update_website_content_markdown(target_content)
 
     assert target_content.markdown == expected_markdown
+    assert target_sync_state.current_checksum == target_content.calculate_checksum()
 
 
 def test_baseurl_replacer_handles_index_files():
@@ -141,6 +145,7 @@ def test_baseurl_replacer_handles_index_files():
     target_content = WebsiteContentFactory.build(
         markdown=markdown, website_id=website_uuid
     )
+    target_sync_state = ContentSyncStateFactory.build(content=target_content)
 
     linkable = WebsiteContentFactory.build(
         website_id=website_uuid,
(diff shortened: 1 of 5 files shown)


ChristopherChudzicki (commented):
I tried this out locally and it is behaving as described: Running

```
docker-compose exec web python manage.py markdown_cleanup baseurl --commit # or any rule besides baseurl
```
on `master` results in `current_checksum` and `synced_checksum` being different. On this branch, the two checksums are the same.

**Question:** But I noticed that the markdown was not syncing with github's master branch. That makes sense to me if the checksums are different (out of sync) but if the checksums are the same, shouldn't we be committing the new markdown to github?

I very well could be misunderstanding something... I will try and spend some time understanding the db -> github connection today.

mbertrand:
@ChristopherChudzicki the changes should be synced to the git main branch when:
- [the user saves edits from the UI ](https://github.com/mitodl/ocw-studio/blob/master/websites/serializers.py#L170)
- [the user publishes to draft or live from the UI](https://github.com/mitodl/ocw-studio/blob/master/content_sync/api.py#L117-L121)
- the `mass-publish` management command is run (which should hit the same publish code as above)

The files chosen for syncing should be ones where [their associated `ContentSyncState.synced_checksum` != `current_checksum`](https://github.com/mitodl/ocw-studio/blob/master/content_sync/apis/github.py#L319-L323)

mbertrand:
There is also a [`content_sync.tasks.sync_unsynced_websites` task](https://github.com/mitodl/ocw-studio/blob/master/content_sync/tasks.py#L47-L51) which can be run to sync to git without publishing.

mbertrand:
Maybe that `sync_unsynced_websites` task should be run as the last step of this command, actually.  What do you think?  Unless the `mass-publish` task is typically run after this command anyway, in which case it would be duplicating effort.

mbertrand:
Sure, just added it in, with an option to skip it, but will run by default as long as `commit=True`

ChristopherChudzicki:
@mbertrand Sooooo I am pretty sure I totally misunderstood this yesterday, but I think I get it now.
- Prior to this PR `markdown_cleanup` was editing the markdown entries, but since this was done through bulk update, the `current_checksum` was not being updated. So the database recorded things as "in sync" even though they were not really in sync. Consequently, a subsequent sync would miss changes we had made.
- With this PR:
    - whenever commiting, we update the current checksum, so that a subsequent sync will update github appropriately 
    - and by default (i.e., without --skip-sync) we db --> github

Does that sound correct? Thanks for your patience with my misunderstandings :)

ChristopherChudzicki (approved):

mbertrand:
Yup, that sounds right!

PR #1048, opened 2022-02-22
Link to Collections

ChristopherChudzicki:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

#### What's this PR do?
Adds support for linking to course collections and resource collections via the CKEditor "Add Link" tool.

Try adding resource and course collections to a page. Click "Save" and verify that the markdown is saved as expected, i.e., that shortcodes of the form `{{< resource_link some-collection-uuid "Some Title" >}} are added to the markdown with the correct UUID. 

#### Screenshots
Screen Shot 2022-02-22 at 2 31 35 PM

### DIFF
--- a/static/js/components/widgets/ResourcePickerDialog.tsx
+++ b/static/js/components/widgets/ResourcePickerDialog.tsx
@@ -1,5 +1,5 @@
 import React, { SyntheticEvent, useCallback, useState } from "react"
-import { Nav, NavItem, NavLink, TabContent, TabPane } from "reactstrap"
+import { Nav, NavItem, NavLink, TabContent, TabPane, Dropdown, DropdownItem, DropdownToggle, DropdownMenu } from "reactstrap"
 
 import Dialog from "../Dialog"
 import {
@@ -7,7 +7,9 @@ import {
   RESOURCE_TYPE_IMAGE,
   RESOURCE_TYPE_VIDEO,
   CONTENT_TYPE_RESOURCE,
-  CONTENT_TYPE_PAGE
+  CONTENT_TYPE_PAGE,
+  CONTENT_TYPE_COURSE_COLLECTION,
+  CONTENT_TYPE_RESOURCE_COLLECTION
 } from "../../constants"
 import ResourcePickerListing from "./ResourcePickerListing"
 import { useDebouncedState } from "../../hooks/state"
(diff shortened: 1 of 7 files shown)


static/js/components/widgets/ResourcePickerDialog.tsx
```diff
@@ -29,59 +41,99 @@ interface Props {
...
+    sourceWebsiteName: "ocw-www"
+  },
+  {
+    title:             "Resource Collections",
+    id:                CONTENT_TYPE_RESOURCE_COLLECTION,
+    contentType:       CONTENT_TYPE_RESOURCE_COLLECTION,
+    resourcetype:      null,
+    embeddable:        false,
+    singleColumn:      true,
+    sourceWebsiteName: "ocw-www"
```
ChristopherChudzicki, line 106: Hard-coding `ocw-www` here doesn't seem like the best thing to do. Do we have anything in place that says "course X belongs to "ocw-www" vs "ocw-xyz"?

One option might be to put something in course starter config as a "collection source", perhaps. I think the Instructors page does something similar as a source of instructors.

Ferdi:
Nice UX choice with the "More" tab.

static/js/components/widgets/ResourcePickerDialog.tsx
alicewriteswrongs: mmm yeah, we do have something similar with the instructors, which is implemented as a 'relation' widget. The relation field takes a 'collection' prop which gives the name of the content type that it will reference, and then it also has an optional 'website' prop - if 'website' is not set, it will look in the same collection where the field is declared, and 'website' _is_ set it will look within that website. So the instructors field looks like this:

```yml
          - label: Instructors
            name: instructors
            widget: relation
            collection: instructor
            display_field: title
            multiple: true
            website: ocw-www
```

alicewriteswrongs: I think it's not a bad idea to add a field to the markdown widget called `collections-source` or something along those lines, just so we can avoid hard-coding this

static/js/constants.ts
```diff
@@ -155,6 +155,8 @@ export const exampleSiteConfigFields: ConfigField[] = uniq(
 
 export const CONTENT_TYPE_RESOURCE = "resource"
 export const CONTENT_TYPE_PAGE = "page"
+export const CONTENT_TYPE_COURSE_COLLECTION = "course_collections"
+export const CONTENT_TYPE_RESOURCE_COLLECTION = "resource_collections"
```
alicewriteswrongs, line 159: should we declare these as an enum or a union type here in `constants.ts`? might be nice to have something like a `LinkableResource` type which we can use elsewhere

static/js/components/widgets/ResourcePickerDialog.tsx
alicewriteswrongs: I think it's fine to have this hardcoded now, and maybe we can file a tech debt issue to fix that later on (since it will require a change to the config schema and corresponding change to ocw-hugo-projects)

static/js/components/widgets/ResourcePickerDialog.tsx
```diff
@@ -29,59 +41,99 @@ interface Props {
...
 
-  const currentTabs = RESOURCE_PICKER_TABS.filter(
+  const tabsMain = RESOURCE_PICKER_TABS_MAIN.filter(
     tab => mode !== RESOURCE_EMBED || tab.embeddable
   )
-  const [activeTab, setActiveTab] = useState(RESOURCE_TYPE_IMAGE)
+  const tabsMore = RESOURCE_PICKER_TABS_MORE.filter(
+    tab => mode !== RESOURCE_EMBED || tab.embeddable
+  )
+  const tabs = [...tabsMain, ...tabsMore]
```
alicewriteswrongs, line 134: is it possible that this would be a little clearer?

```suggestion
  const tabs = mode === RESOURCE_EMBED
    ? [...tabsMain, ...tabsMore].filter(tab => tab.embeddable)
    : [...tabsMain, ...tabsMore]
```

idk, just took me a minute to follow what's going on so thought maybe we could simplify

static/js/components/widgets/ResourcePickerDialog.tsx
```diff
@@ -29,59 +41,99 @@ interface Props {
...
 
 export default function ResourcePickerDialog(props: Props): JSX.Element {
   const { mode, isOpen, closeDialog, insertEmbed } = props
 
-  const currentTabs = RESOURCE_PICKER_TABS.filter(
+  const tabsMain = RESOURCE_PICKER_TABS_MAIN.filter(
     tab => mode !== RESOURCE_EMBED || tab.embeddable
   )
-  const [activeTab, setActiveTab] = useState(RESOURCE_TYPE_IMAGE)
+  const tabsMore = RESOURCE_PICKER_TABS_MORE.filter(
```
alicewriteswrongs, line 131: what would you think of naming these two things `fullTabs` and `dropdownTabs` or something like that?

alicewriteswrongs (changes requested):
all looks pretty good! functionality works great. Just had a few small naming / organization comments

static/js/components/widgets/ResourcePickerDialog.tsx
ChristopherChudzicki: Cool, I will do that next. https://github.com/mitodl/ocw-studio/issues/1052

static/js/components/widgets/ResourcePickerDialog.tsx
```diff
@@ -28,60 +32,101 @@ interface Props {
...
   },
   [RESOURCE_LINK]: {
-    headerText: "Resources & Pages",
+    title:      "Link to Content",
     acceptText: "Add link"
   }
 }
 
+const isTabEnabled = (mode: ResourceDialogMode) => (tab: ResourceTabSettings) =>
+  mode !== RESOURCE_EMBED || tab.embeddable
```
alicewriteswrongs, line 123: love a good curry!

alicewriteswrongs (approved):
looks great!

Issue #1038, opened 2022-02-18
Link to course collections


ChristopherChudzicki:
As a course author, I should be able to add links to course collections in pages.

#### Designs and Mockups
Horizontal space is a bit limited in our tabs. This might be a good approach for course collections + resource collections, but focus should be on course collections for this feature.

<img width="550" alt="Screen Shot 2022-02-18 at 5 16 12 PM" src="https://user-images.githubusercontent.com/9010790/154774440-a172d704-006c-4de5-b9ba-27b35ae1bbef.png">

#### Acceptance Criteria:

In CKEditor on a course page:
- [ ] Clicking "Add link" should open link-picking dialog with a "Collections"
- [ ] Clicking "Collections" tab should populate dialog with filterable list of **course collections**
- [ ] it should be clear that the list is course collections for now, and not resource collections 

#### Out of Scope

- Adding resource collections

Ferdi:
I'd keep it simple for now - just a "Collection" tab that shows all collections.   We might want to redesign this a bit anyway , the tabs is not scaling.

PR #1055, opened 2022-02-24
Cc/collections limited

ChristopherChudzicki:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

#### What are the relevant tickets? #1038 
( a followup to #1048 ) uses https://github.com/mitodl/ocw-hugo-projects/pull/128

#### What's this PR do?
This PR changes the ResourcePickerDialog to show tabs based on the site configuration. With <PROJECT PR GOES HERE> 
- `ocw-course` sites are configured to link resources & pages and embed resources
- `ocw-www` sites are configured to link collections

**Why only link collections in ocw-www?** Because linking to collections in `ocw-www` sites is the immediate need and ... right now there are some mismatches between `ocw-course` and `ocw-www` configurations which would make linking to other things (pages, resources) a bit annoying. For example, in courses, pages have `type=page`, but in www they have `type=pages`. And in courses, resource types are determined by `resourcename` but in `www` they are determined by `filename`.

First, get the new ocw-www and ocw-course site configs:
- for courses: https://github.com/mitodl/ocw-hugo-projects/blob/cc/collections-limited/ocw-course/ocw-studio.yaml
- for www: https://github.com/mitodl/ocw-hugo-projects/blob/cc/collections-limited/ocw-www/ocw-studio.yaml

Then check that:
- within a course, you can link to pages and resources
- within a course, you can embed resources
- within ocw-www sites, you can link to collections
- within ocw-www sites, there is no "embed" option

And maybe check that the course collection links actually render in hugo output, which was an issue with #1048 

Screen Shot 2022-02-24 at 11 24 23 AM
Screen Shot 2022-02-24 at 11 27 13 AM


static/js/components/widgets/ResourcePickerDialog.test.tsx
```diff
@@ -57,121 +58,95 @@ describe("ResourcePickerDialog", () => {
...
 
-    expect(insertEmbedStub.args[0]).toStrictEqual([
-      resource.text_id,
-      resource.title,
-      RESOURCE_EMBED
-    ])
-  })
+      wrapper.update()
 
-  it("should pass basic props to ResourcePickerListing", async () => {
```
ChristopherChudzicki, line 165: I don't think this test added anything not covered in `"passes the correct props to ResourcePickerListing when main tab $index is clicked"`, so I just removed this test.

static/js/components/widgets/ResourcePickerDialog.tsx
```diff
@@ -189,30 +175,6 @@ export default function ResourcePickerDialog(props: Props): JSX.Element {
                 </NavLink>
               </NavItem>
             ))}
-            {dropdownTabs.length > 0 && (
-              <Dropdown
```
ChristopherChudzicki, line 193: 😭

static/js/components/widgets/MarkdownEditor.tsx
```diff
@@ -102,15 +101,13 @@ export default function MarkdownEditor(props: Props): JSX.Element {
...
         toolbar: {
           ...FullEditorConfig.toolbar,
-          items: FullEditorConfig.toolbar.items.filter(item =>
-            hasAttach ?
-              true :
-              item !== ADD_RESOURCE_LINK && item !== ADD_RESOURCE_EMBED
-          )
+          items: FullEditorConfig.toolbar.items
+            .filter(item => link.length !== 0 || item !== ADD_RESOURCE_LINK)
+            .filter(item => embed.length !== 0 || item !== ADD_RESOURCE_EMBED)
```
alicewriteswrongs, line 106: just me - but I find these conditionals confusing - could we say it more plainly instead, something like:

```suggestion
            .filter(item => {
              if (item === ADD_RESOURCE_LINK) {
                return link.length > 0
              }
              if (item === ADD_RESOURCE_EMBED) {
                return embed.length > 0
              }
              return true
            })
```

more of a nitpick really, I just always get tripped up trying to think through boolean conditions and so prefer to write them out as explicitly as possible 🙂

static/js/components/widgets/ResourcePickerDialog.tsx
alicewriteswrongs: well, it will always be there in `git blame` for us 😄

static/js/lib/site_content.ts
```diff
@@ -97,7 +97,8 @@ export function widgetExtraProps(field: ConfigField): Record<string, any> {
   case WidgetVariant.Markdown:
     return {
       minimal: field.minimal ?? false,
-      attach:  field.attach ?? undefined
+      link:    field.link ?? [],
+      embed:   field.embed ?? []
```
alicewriteswrongs, line 101: if we're always going to return `[]` here then I think we could probably type `link` and `embed` on the Markdown field props like `link: string[]` rather than like `link?: string[]`

alicewriteswrongs (changes requested):
works great! just had a few small comments

static/js/components/widgets/MarkdownEditor.tsx
ChristopherChudzicki: Yes that's much cleaner. I dunno why my brain was having a hard time putting this in one filter =/

alicewriteswrongs (approved):
👍

Issue #1066, opened 2022-02-25
resource_link shortcodes don't preserve line breaks


gumaerc:
### Steps to Reproduce

 - Create a site on RC using `ocw-course`
 - Add a resource to your site in Google Drive and sync
 - Add a page to your site and create multiple resource links (doesn't matter if it's to the same resource or different ones) with line breaks in between them
 - Publish your site and inspect the output

### Expected Behavior

The links have line breaks in between them

### Actual Behavior

The links are all rendered inline

PR #1067, opened 2022-02-25
change resource_link delimiters to % instead of < >

gumaerc:
#### Pre-Flight checklist

- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1066

#### What's this PR do?
The Hugo documentation doesn't really explain this very well, but there are two types of delimiters you can use when calling shortcodes.  The standard method of `{{< shortcode >}}`, using the `< >` syntax, expects that the shortcode will return HTML, and therefore the shortcode template should be HTML.  When you use `{{% shortcode %}}`, Hugo expects that the shortcode will return markdown.  So, in your shortcode template definition you write markdown instead of HTML to be rendered.  This PR changes `resource_link` to be called with `%` delimiters so it is interpreted as markdown.  This way, when Hugo sees actual markdown links with line breaks on either side, it will properly preserve that structure when rendering HTML.

 - Spin up a local instance of `ocw-studio`, making sure you follow the directions in the readme and are set up to publish courses to a test Github Organization.  It's okay to use RC's Google Drive credentials, and you will need them to test this PR
 - Create a test site and verify that a repo has been created in your test org
 - Add a resource to your site in Google Drive and sync
 - Add a page to your site and create multiple resource links (doesn't matter if it's to the same resource or different ones) with line breaks in between them
 - Publish your site and inspect the output, you should see your resource links with `%` delimiters
 - Clone `ocw-hugo-themes` on the branch from [this PR](https://github.com/mitodl/ocw-hugo-themes/pull/485)
 - Clone your test site's repo from your test org and configure `ocw-hugo-themes` to point at it by setting `COURSE_CONTENT_PATH=/path/to/test-org` and `OCW_TEST_COURSE=your-test-course`, replacing with your values
 - Spin up the site with `npm run start:course`
 - Visit http://localhost:3000 and browse to your test page
 - Verify that all of the links are displayed on their own line

Finally, check the new `markdown_cleanup` strategy to verify results on your local database:
 - Make sure you have at least one page saved with `resource_link` shortcodes
 - Run `docker-compose run --rm web ./manage.py markdown_cleanup resource_link_delimiter_swap -o resource_link_delimiters.csv`
 - Inspect the output and make sure your `resource_link` shortcodes are now properly delimited with `%` instead of `< >`

### DIFF
--- a/static/js/lib/ckeditor/plugins/ResourceLinkMarkdownSyntax.ts
+++ b/static/js/lib/ckeditor/plugins/ResourceLinkMarkdownSyntax.ts
@@ -22,10 +22,10 @@ const decodeShortcodeArgs = (encoded: string) =>
  * (?: "(.*?)")? to match the optional anchor ID param
  *
  * Limitations:
- *   - gets fooled by label texts that include literal `" >}}` values. For
- *     example, < resource_link uuid123 "silly " >}} link" >}}.
+ *   - gets fooled by label texts that include literal `" %}}` values. For
+ *     example, < resource_link uuid123 "silly " %}} link" %}}.
  */
-export const RESOURCE_LINK_SHORTCODE_REGEX = /{{< resource_link (\S+) "(.*?)"(?: "(.*?)")? >}}/g
+export const RESOURCE_LINK_SHORTCODE_REGEX = /{{% resource_link (\S+) "(.*?)"(?: "(.*?)")? %}}/g
 
 /**
  * Class for defining Markdown conversion rules for Resource links
@@ -33,7 +33,7 @@ export const RESOURCE_LINK_SHORTCODE_REGEX = /{{< resource_link (\S+) "(.*?)"(?:
  * These are stored in Markdown like this:
  *
  * ```md
- * {{< resource_link AUUIDUNLIKEANYOTHER "Here's a link to my resource" >}}
+ * {{% resource_link AUUIDUNLIKEANYOTHER "Here's a link to my resource" %}}
  * ```
  *
  * The first argument is the uuid of the resource to which we're linking, and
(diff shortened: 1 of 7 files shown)


websites/management/commands/markdown_cleaning/resource_link_delimiters.py
```diff
@@ -0,0 +1,26 @@
...
+from websites.management.commands.markdown_cleaning.cleanup_rule import (
+    MarkdownCleanupRule,
+)
+from websites.models import WebsiteContent
+
+
+class ResourceLinkDelimitersReplacementRule(MarkdownCleanupRule):
+    """Replacement rule for use with WebsiteContentMarkdownCleaner."""
+
+    regex = r"{{<\sresource_link\s(?P<resource_uuid>[a-z0-9\-]*)\s(?P<link_text>\".*?\")\s>}}"
```
ChristopherChudzicki, line 19: **Requested Change:** I don't think this will work for the resource_links that have fragment ids. Also, I'm not 100% sure that all resource link titles have double quotes. I suspect they do? But I'm not sure.

**Suggestion:** We don't really need to parse the arguments into separate uuid/title/fragment for this replacement, so I think something like
```py
regex = r"{{<\sresource_link(?P<the_args>.*?)\s>}}"
```
should work. Plus if we ever needed to do this again in the future, I think the markdown_cleanup rule would be entirely reusable, just with `resource_link` swapped out for a different string.

static/js/lib/ckeditor/plugins/ResourceLinkMarkdownSyntax.ts
```diff
@@ -22,18 +22,18 @@ const decodeShortcodeArgs = (encoded: string) =>
  * (?: "(.*?)")? to match the optional anchor ID param
  *
  * Limitations:
- *   - gets fooled by label texts that include literal `" >}}` values. For
- *     example, < resource_link uuid123 "silly " >}} link" >}}.
+ *   - gets fooled by label texts that include literal `" %}}` values. For
+ *     example, < resource_link uuid123 "silly " %}} link" %}}.
```
ChristopherChudzicki, line 26:
```suggestion
 *     example, % resource_link uuid123 "silly " %}} link" %}}.
```

ChristopherChudzicki (changes requested):
Looks good to me except for the regex issue I mentioned.

gumaerc:
@ChristopherChudzicki Good call on the argument capture.  I made some updates and it's ready for another look.

ChristopherChudzicki (approved):
👍
- I can add new resource_links in studio and they save appropriately
- The csv output shows all replacements are just angle brackets to %.
- My local has 91137 resource_links, same as rows in csv

PR #1081, opened 2022-03-03
fix spacing issue for single-line table cell content

### LINKED ISSUE #1082: issue with single-line content + markdown formatting in tables
We're seeing an issue with single-line content in table cells (i.e. content that has no newline).

Basically, the issue is that if you have markdown formatting on that content it seems to be getting ignored by Hugo unless there is white space around it, so something like this:

```md
{{< tdopen >}}**great stuff!**{{< tdclose >}}
```

_should_ render to

```html
<td><strong>great stuff!</strong></td>
```

but instead it's rendering to 

```html
<td>**great stuff!**</td>
```

alicewriteswrongs:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #1082

#### What's this PR do?

This fixes an issue we're seeing with single-line content in table cells. Basically, the issue is that if you have markdown formatting on that content it seems to be getting ignored by Hugo unless there is white space around it, so something like this:

```md
{{< tdopen >}}**great stuff!**{{< tdclose >}}
```

_should_ render to

```html
<td><strong>great stuff!</strong></td>
```

but instead it's rendering to 

```html
<td>**great stuff!**</td>
```

we can intervene in the html -> markdown conversion process and insert newlines, like so:

```md
{{< tdopen >}}
**great stuff!**
{{< tdclose >}}
```

which fixes the issue afaict

really try to exercise and break this - I think I fixed it in a way that shouldn't cause new problems, but this stuff is finicky.

Make sure at a minimum that you can have a single-line table cell, put markdown formatting on it, and have that show up in the rendered course.

### DIFF
--- a/static/js/lib/ckeditor/plugins/TableMarkdownSyntax.test.tsx
+++ b/static/js/lib/ckeditor/plugins/TableMarkdownSyntax.test.tsx
@@ -155,4 +155,23 @@ describe("table shortcodes", () => {
     </table>`
     expect(htmlBeautify(md2html(md))).toBe(htmlBeautify(html))
   })
+
+  it.each([
+    [
+      "<table><thead><tr><th>my content</th></tr></thead></table>",
+      "{{< tableopen >}}{{< theadopen >}}{{< tropen >}}{{< thopen >}}\nmy content\n{{< thclose >}}{{< trclose >}}{{< theadclose >}}{{< tableclose >}}"
+    ],
+    [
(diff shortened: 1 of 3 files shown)


gumaerc (approved):
:+1: LGTM

This seems like a sensible approach and I wasn't able to produce any side effects in my testing.  One unrelated issue I discovered is that the underline functionality in CKEditor currently doesn't do anything.  I made an issue for that here: https://github.com/mitodl/ocw-studio/issues/1084

image

image

PR #1094, opened 2022-03-08
merge new metadata with old metadata

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#### What's this PR do?
Updates `WebsiteContentDetailSerializer` so that `update` merges new metadata with old metadata, rather than overwriting the old metadata.

A few notes:
- similar to https://datatracker.ietf.org/doc/html/rfc7396
- but does not implement the "null means delete" thing
- not consistent with how we are handling other json fields (e.g., `Website.metadata`

Might be nice to have a more uniform solution to this

_But use your local not RC:_

1. Open an instructor insights page in admin panel, e.g., https://ocw-studio-rc.odl.mit.edu/admin/websites/websitecontent/515029/change/?_changelist_filters=q%3Dc211db95-89db-de17-36b3-806670e359ff
2. Add `"layout": "instructor_insights"` to the metadata if it's not present.
3. Open the corresponding page in studio, https://ocw-studio-rc.odl.mit.edu/sites/18-330-introduction-to-numerical-analysis-spring-2012/type/page/edit/c211db95-89db-de17-36b3-806670e359ff/
4. Save a simple edit

Check that the metadata still exists in admin panel after the studio save.

### DIFF
--- a/websites/serializers.py
+++ b/websites/serializers.py
@@ -376,6 +376,11 @@ class WebsiteContentDetailSerializer(
             validated_data["metadata"]["file_type"] = detect_mime_type(
                 validated_data["file"]
             )
+        if "metadata" in validated_data:
+            validated_data["metadata"] = {
+                **instance.metadata,
+                **validated_data["metadata"],
+            }
         instance = super().update(
             instance, {"updated_by": self.user_from_request(), **validated_data}
         )
--- a/websites/serializers_test.py
+++ b/websites/serializers_test.py
@@ -391,12 +391,18 @@ def test_website_content_detail_serializer_save(mocker):
     mock_create_website_pipeline = mocker.patch(
         "websites.serializers.create_website_publishing_pipeline"
     )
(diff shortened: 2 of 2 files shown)


websites/serializers.py
```diff
@@ -376,6 +376,11 @@ def update(self, instance, validated_data):
             validated_data["metadata"]["file_type"] = detect_mime_type(
                 validated_data["file"]
             )
+        if "metadata" in validated_data:
+            validated_data["metadata"] = {
+                **instance.metadata,
+                **validated_data["metadata"],
+            }
         instance = super().update(
```
ChristopherChudzicki, line 384: Random question: This is the `update` method, which seems to be tied to HTTP PATCH requests. Is it also used for PUT? Django noob.

alicewriteswrongs: I'm not sure, a quick scan of the docs didn't turn anything saying one way or the other. I would guess so?

alicewriteswrongs (approved):
LGTM!

Issue #1073, opened 2022-03-01
can't update instructor salutation


pdpinch:
### Steps to Reproduce

1. Update an instructor missing a salutation, e.g. https://ocw-studio-rc.odl.mit.edu/sites/ocw-www/type/instructor/?q=mattuck
2. Set the salutation to "Prof"
3. Save the instructor
4. Publish the site to stage. 
5. Publish the site to production. 
6. Confirm the update is in github, at https://github.mit.edu/ocw-content-rc/ocw-www/blob/release/content/instructors/583cc582-1f48-236c-af5a-b2f4750fef40.md
7. Confirm the instructor is published on the web site, at https://ocwnext-rc.odl.mit.edu/instructors/583cc582-1f48-236c-af5a-b2f4750fef40/index.json
8. In a different course, such as https://ocw-studio-rc.odl.mit.edu/sites/peters-test-2021-09-29-1/type/metadata/ and open the metadata section
9. Add the instructor you edited. 

### Expected Behavior

Salutation should be part of the instructor's title when added. 

### Actual Behavior

Still no salutation. 

### Screenshot or Screencast

<img width="722" alt="image" src="https://user-images.githubusercontent.com/430126/156102600-d55e7193-26ce-4da2-bdba-aed982b6ba30.png">

MAbdurrehman12:
@alicewriteswrongs I agree with what you stated and I think we can go with this approach
> One approach is to basically change .title to a computed property by programmatically setting in on instructor objects when they're saved

but with hiding the `title` field from course author, so course author will only see the metadata while saving/updating an instructor, and in the backend we can programmatically set the title to `Salutation + first_name + middle ...` so this title field becomes a ready-to-go field. As I also noticed that title is being used as an instructor's full name in a few other places as well.

What do you think?

alicewriteswrongs:
Yeah I think that makes sense to me, especially if we make it so the user can't edit the field, since if they edit it and then it gets overwritten that could be confusing.

pdpinch:
Turns out this is user error. We can improve that data model and/or the UX, but it’s not a release blocker. I’m putting it on hold.

PR #1105, opened 2022-03-10
fix: salutation appended in title

MAbdurrehman12:
#### Pre-Flight checklist

- [ ] Migrations
- [x] Testing
  - [x] Code is tested
- [ ] Settings

Closes https://github.com/mitodl/ocw-studio/issues/1073

#### What's this PR do?
- Appends salutation in title only when type is `instructor` and `salutation` is present in metadata
    - A util to append salutation with title is created in `websites/utils.py`
    - this util is called in `save` method of `WebsiteContent` model in `websites/models.py`
    - A test for this util has also been created. 

- Go to ocw-studio
- Add a new instructor
- Set a title without salutation and add salutation in metadata
- Verify that, on saving, this salutation is appended with title
- Save this instructor again and verify that salutation is not appended again if title already has a salutation.
- Repeat the above steps for 
     - `ocw-studio/admin` and `ocw-studio/sites` 
     - Creating a new instructor and saving a new instructor

image

image

image

#### Note:
`Title` is a different field from metadata which is set differently so I'm not sure whether it's a good solution to only take salutation from metadata and append it in the title. 
In case of instructor, if title is instructor's full name, then I think, title should not be asked upon saving/updating instuctor, title should be auto-populated when instructor is saved/updated. (title = salutation + first_name + middle_name + last_name).
But since in the issue ticket, the expected behavior is: `Salutation should be part of the instructor's title when added.` so I have updated the `save` method, but I don't think this is a good approach as there might be some use-cases where this approach will cause few issues, maybe or maybe not.

### DIFF
--- a/websites/models.py
+++ b/websites/models.py
@@ -31,7 +31,7 @@ from websites.constants import (
     CONTENT_FILEPATH_UNIQUE_CONSTRAINT,
 )
 from websites.site_config_api import SiteConfig
-from websites.utils import permissions_group_name_for_role
+from websites.utils import permissions_group_name_for_role, set_title_salutation
 
 
 def validate_yaml(value):
@@ -279,6 +279,7 @@ class WebsiteContent(TimestampedModel, SafeDeleteModel):
 
     def save(self, **kwargs):  # pylint: disable=arguments-differ
         """Update dirty flags on save"""
+        set_title_salutation(self)
         super().save(**kwargs)
         website = self.website
         website.has_unpublished_live = True
--- a/websites/utils.py
+++ b/websites/utils.py
@@ -47,3 +47,14 @@ def get_valid_base_filename(filename: str, content_type: str) -> str:
     if filename in constants.CONTENT_FILENAMES_FORBIDDEN:
         return f"{filename}-{content_type}"
     return filename
+
+
+def set_title_salutation(website_content):
+    """Set salutation in title if salutation found in metadata and not found in title"""
+    if (
+        website_content.type == constants.CONTENT_TYPE_INSTRUCTOR
+        and website_content.metadata.get("salutation", None)
+    ):
+        salutation = website_content.metadata["salutation"].strip()
+        if not website_content.title.startswith(salutation):
+            website_content.title = f"{salutation} {website_content.title}"
--- a/websites/utils_test.py
+++ b/websites/utils_test.py
@@ -1,12 +1,14 @@
 """Website utils tests"""
+from curses import meta
 import pytest
 
 from websites import constants
-from websites.factories import WebsiteFactory
+from websites.factories import WebsiteFactory, WebsiteContentFactory
 from websites.utils import (
     get_dict_query_field,
     permissions_group_name_for_role,
     set_dict_field,
+    set_title_salutation,
 )
 
 
@@ -79,3 +81,35 @@ def test_set_dict_field():
             "parameter_2": "value2",
         },
     }
+
+
+def test_set_title_salutation():
+    """
+    Verifies that set_title_salutation works as expected:
+    Append salutation in title only when needed
+    """
+    data_arr = [
+        {
+            "title": "John Cena",
+            "metadata": {"salutation": "Dr."},
+            "title_with_salutation": "Dr. John Cena",
+        },
+        {
+            "title": "John Cena",
+            "metadata": {"salutation": ""},
+            "title_with_salutation": "John Cena",
+        },
+        {
+            "title": "Dr. John Cena",
+            "metadata": {"salutation": "Dr."},
+            "title_with_salutation": "Dr. John Cena",
+        },
+    ]
+    for data in data_arr:
+        website_content = WebsiteContentFactory.build(
+            title=data["title"],
+            metadata=data["metadata"],
+            type=constants.CONTENT_TYPE_INSTRUCTOR,
+        )
+        set_title_salutation(website_content)
+        assert website_content.title == data["title_with_salutation"]
--- a/websites/utils_test.py
+++ b/websites/utils_test.py
@@ -1,5 +1,4 @@
 """Website utils tests"""
-from curses import meta
 import pytest
 
 from websites import constants
--- a/websites/utils_test.py
+++ b/websites/utils_test.py
@@ -2,7 +2,7 @@
 import pytest
 
 from websites import constants
-from websites.factories import WebsiteFactory, WebsiteContentFactory
+from websites.factories import WebsiteContentFactory, WebsiteFactory
 from websites.utils import (
     get_dict_query_field,
     permissions_group_name_for_role,


websites/models.py
```diff
@@ -279,6 +279,7 @@ def get_config_file_field(self) -> Dict:
 
     def save(self, **kwargs):  # pylint: disable=arguments-differ
         """Update dirty flags on save"""
+        set_title_salutation(self)
```
MAbdurrehman12, line 282: `Title` is a different field from metadata which is set differently so I'm not sure whether it's a good solution to only take salutation from metadata and append it in the title. 
In case of instructor, if title is instructor's full name, then I think, title should not be asked upon saving/updating instuctor, title should be auto-populated when instructor is saved/updated. (title = salutation + first_name + middle_name + last_name).
But since in the issue ticket, the expected behavior is: `Salutation should be part of the instructor's title when added.` so I have updated the `save` method, but I don't think this is a good approach as there might be some use-cases where this approach will cause few issues, maybe or maybe not.

alicewriteswrongs: what do you think are cases where this will cause an issue?

I think it seems our assumption is that, for instructors, the `.title` field will be a sort of 'ready to go' field we can use for correctly displaying the name + title of a given instructor. If that's the intent then I think it makes sense that it's basically a computed field like `title + " " + first_name + " " + last_name`.

One suggestion to make it a little more clear what's going on would be to move the conditional you have in `set_title_salutation`: 

```python
    if (
        website_content.type == constants.CONTENT_TYPE_INSTRUCTOR
        and website_content.metadata.get("salutation", None)
    ):
```

websites/utils.py
```diff
@@ -47,3 +47,14 @@ def get_valid_base_filename(filename: str, content_type: str) -> str:
...
+
+def set_title_salutation(website_content):
+    """Set salutation in title if salutation found in metadata and not found in title"""
+    if (
+        website_content.type == constants.CONTENT_TYPE_INSTRUCTOR
+        and website_content.metadata.get("salutation", None)
+    ):
+        salutation = website_content.metadata["salutation"].strip()
+        if not website_content.title.startswith(salutation):
+            website_content.title = f"{salutation} {website_content.title}"
```
alicewriteswrongs, line 60: I think this isn't quite the right approach. Because we're only checking
`website_content.title.startswith(salutation)` there can be some strange (and
probably puzzling to the user) behavior when the `title` or the `salutation` is
edited. For instance, if I do the following:

1. create a new instructor with first name `'First'`, title `'First'`, and salutation `'prof'`
2. realize I didn't capitalize `'prof'`, change salution to `'Prof'` and save
3. the title for my instructor is now `'Prof prof First'`

Screen Shot 2022-03-10 at 1 25 31 PM

I might also not realize that the title field is what's going to be rendered,
and think that changing the `first name` field will update what's displayed in
the rendered course - that might be something to sort out separately though.

I think the way to go about this has to be one of two things:

1. either we have to go all-in on the `title` field being computed _purely_ from `salutation` and the name fields or
2. title has to be manually editable and we discard the idea of a separate salutation.

I added a note on the issue that I think `title` is a bit of an odd naming choice here, we're using it as sort of an expanded, full form of address for the instructor rather than a title, which to me means more like 'professor', 'dr', etc.

anyhow, wondering if @pdpinch  or possible @dseaton have thoughts on this.

alicewriteswrongs (changes requested):
some comments on this

websites/utils.py
MAbdurrehman12: Thank you for looking into this @alicewriteswrongs.
As I mentioned [here](https://github.com/mitodl/ocw-studio/issues/1073#issuecomment-1064408827), if, in the case of an instructor, the title field is hidden from the user on adding/updating an instructor and programmatically computed, then wouldn't it resolve the issue which you stated? As now, it can be rendered anywhere with accurate full name of an instructor, including salutation.

Another benefit is that the course author will not have to re-write the same thing in the title field

websites/models.py
MAbdurrehman12: The exact same issues which you just pointed out below :P haha.
Another use-case is that if a course author removes the salutation from metadata then it will be still there in the title field.
Another use-case is that if a course author changes the salutation. e.g: Initially, salutation was `Dr.`, now it's changed to `Prof.` so title will become `Prof. Dr. ....`.

Hence due to these use-cases, I'm also not in the favour of this solution. 
I guess I took the `expected behavior` in the issue ticket a bit too seriously. haha

websites/utils.py
alicewriteswrongs: yeah, my thought is that we should probably just go to setting the title field to be `salutation + " " + first_name + " " + last_name` - wondering what @pdpinch thinks about this too

MAbdurrehman12: Yup, making the title either hidden or uneditable while adding/updating an instructor and setting it programmatically, as you stated, will do the job!

pdpinch:
Oh... I didn't realize that studio uses the instructor's title field for the drop-down UI. 

Does the course home page in ocw-hugo-themes use the instructor title as well? I'm actually more concerned with seeing the correct "Prof." there. I just assumed that ocw-studio and ocw-hugo-themes would use the same fields.

MAbdurrehman12:
> Does the course home page in ocw-hugo-themes use the instructor title as well? 

I guess yes and I think we should get this title thing sorted (as mentioned above) for instructors to avoid such issue or confusion for the course author.

In another issue, related to displaying instructor's full name in search results, title was also used as full_name which was updated to include salutation, first, middle and last name etc in this [PR](https://github.com/mitodl/open-discussions/pull/3529/)

pdpinch:
I'm a little uncomfortable special-casing the behavior for this object, but I guess it's OK.

MAbdurrehman12:
> I'm a little uncomfortable special-casing the behavior for this object, but I guess it's OK.

Is your concern related to the purpose of the `title` field in instructor object or it's related to special-casing a generic model/class (i.e: `website content`)? 
I'm not sure whether we have any other option in this use case. Do you have any suggestions?

MAbdurrehman12:
@alicewriteswrongs 
Off-topic: Shouldn't this `type` field be a dropdown?
image

pdpinch:
Am I correct in understanding that the only field we use, out of the whole instructor content type is `title`?

MAbdurrehman12:
> Am I correct in understanding that the only field we use -- in ocw-studio, and in ocw-hugo-themes -- out of the whole instructor content type is title?

I'm not very sure but I think yes. The only place where I could find first_name, last_name etc being used in studio is in this function [import_ocw2hugo_sitemetadata](https://github.com/mitodl/ocw-studio/blob/6fd1310a180adaf95afb5735dfd97c11c0113bf4/ocw_import/api.py#L408) and that's also to compute title

> I'm wondering if we should just remove the other fields, to avoid user confusion.

This would be optimal, if the condition, you stated above, is fulfilled.

I guess @alicewriteswrongs or @carey might have better insight on this.

alicewriteswrongs:
We could definitely remove the other fields and just have title. It is probably fairly surprising right now that you can change someone's name and not see it get reflected in the rendered site. If we did this, then we'd just have to communicate to the course authors that the title field needs to contain the full name + title for the instructors.

MAbdurrehman12:
> We could definitely remove the other fields and just have title

Thanks for confirming. I'll go on and update this PR to have only `title` field for instructors.

pdpinch:
Thanks for you all for helping my understanding of this. We've decided that studio users should focus on the title and ignore the other fields.

PR #1107, opened 2022-03-10
Cc/metadata relative

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/ocw-to-hugo/issues/481

#### What's this PR do?
This PR adds a markdown_cleanup rule to turn relative links in metadata into root-relative links. (We can't use shortcodes in metadata because they are rendered via Hugo's RenderString, which does not support shortcodes.)

Additionally, this PR refactors
```
websites/management/commands/markdown_cleaning/cleaner.py
websites/management/commands/markdown_cleaning/cleanup_rule.py
```
to be less regex-specific in case we want to do non-regex based replacements in the future. 

Also adds a `validate_urls` rule the goal of which is to help be a bit more proactive about finding broken stuff.

Run
```
docker-compose run --rm web python manage.py markdown_cleanup metadata_relative_urls -o metadata_urls.csv
```
to generate a CSV containing the changes that this command will make. Inspect some rows in the CSV and check that they look reasonable, and maybe check that the new root-relative links work.


MAbdurrehman12 (approved):
Looks good 👍

Issue #1104, opened 2022-03-09
published content filter in relation widget applying to same-site content


gumaerc:
### Steps to Reproduce

 - Create a page without publishing
 - Create a menu item and search for your page to link as an internal item

### Expected Behavior

The page shows up to select

### Actual Behavior

It does not

alicewriteswrongs:
ah! in the relation widget we're always applying the `published` filter, but we should be conditionally doing so only when `cross-site` or `global` is set on the field

PR #1109, opened 2022-03-10
only query with published = true if cross_site is also true

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1104

#### What's this PR do?
In https://github.com/mitodl/ocw-studio/pull/904, we added a filter to the relation widget that filters results in the frontend, filtering out unpublished content.  This was intended to prevent referencing content that hasn't been published yet, but is unnecessary if you're referencing content in the same site.  This is because if you publish a site, that content that you're referencing will be published as part of that.  Thus, this PR changes the filter so that it only applies if the relation widget is cross-site.

 - Spin up `ocw-studio` locally
 - Create a course and a few pages or use an existing course with some pages in it
 - Publish your course
 - Add a new page, but don't publish it
 - Go to the Menu section
 - Click add new and verify that you see both your published and unpublished pages to choose from

### DIFF
--- a/static/js/components/widgets/RelationField.test.tsx
+++ b/static/js/components/widgets/RelationField.test.tsx
@@ -209,6 +209,21 @@ describe("RelationField", () => {
       ).toEqual(formatWebsiteOptions(websites, "name"))
     })
 
+    it("should not filter on published if cross_site is false", async () => {
+      await render({ cross_site: false, value: [] })
+      expect(global.fetch).toHaveBeenCalledWith(
+        siteApiContentListingUrl
+          .query({
+            detailed_list:   true,
+            content_context: true,
+            type:            "page",
+            published:       false
+          })
+          .toString(),
+        { credentials: "include" }
+      )
+    })
+
     it("should let the user pick a website and then content within that website", async () => {
       const { wrapper } = await render({ cross_site: true, value: [] })
       await act(async () => {
--- a/static/js/components/widgets/RelationField.test.tsx
+++ b/static/js/components/widgets/RelationField.test.tsx
@@ -185,7 +185,6 @@ describe("RelationField", () => {
               detailed_list:   true,
               content_context: true,
               type:            "page",
-              published:       true,
               ...(withResourcetypeFilter ? { resourcetype: "Image" } : {})
             })
             .toString()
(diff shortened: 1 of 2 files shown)


static/js/components/widgets/RelationField.tsx
```diff
@@ -177,7 +177,7 @@ export default function RelationField(props: Props): JSX.Element {
       const url = siteApiContentListingUrl
         .query({
           detailed_list:   true,
-          published:       true,
+          published:       crossSite ? true : false,
```
alicewriteswrongs, line 180: I think this may not be what we want to do here - the way this parameter is implemented is that you pass `published: boolean` and it returns only items which match that publication status - in other words, passing `published: false` will return only unpublished content. So if you want to not filter to only published content, the thing to do would be to omit the parameter.

So I think instead what you probably want to do is

```suggestion
          ...(crossSite ? { published: true } : {})
```

alicewriteswrongs (commented):
one comment on it

gumaerc:
@alicewriteswrongs Thanks for your help on this this morning.  I fixed up the code and tests and this is now ready for review.

static/js/components/widgets/RelationField.test.tsx
```diff
@@ -209,6 +208,25 @@ describe("RelationField", () => {
       ).toEqual(formatWebsiteOptions(websites, "name"))
     })
 
+    //
+    ;[true, false].forEach(isCrossSite => {
+      it("should not filter on published if cross_site is not set", async () => {
```
alicewriteswrongs, line 213: if you want, there's a neat facility built-in to Jest that you can use for parameterized tests like this, it looks something like this:

```suggestion
    it.each([true, false])("should not filter on published if cross_site is not set", async (isCrossSite) => {
```

static/js/components/widgets/RelationField.tsx
```diff
@@ -174,10 +174,11 @@ export default function RelationField(props: Props): JSX.Element {
     async (search: string | null, debounce: boolean) => {
       const params = collection ? { type: collection } : { page_content: true }
       const name = crossSite && focusedWebsite ? focusedWebsite : websiteName
+      console.log(`crossSite: ${crossSite}`)
```
alicewriteswrongs, line 177: stray `console.log` here 🙂

alicewriteswrongs (changes requested):
works great! a few code comments

pdpinch:
What's the status on this?

pdpinch:
Does this need to be merged?

gumaerc:
@alicewriteswrongs I made the change you suggested and removed that stray log statement

alicewriteswrongs (approved):
lgtm!

PR #1137, opened 2022-03-17
Remove extra parameter from < resource > shortcodes

ChristopherChudzicki:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

#### What's this PR do?
When converting some images from  markdown `alt text` to resource shortcodes, I erroneously included the alt text as an argument to the short code: `< resource uuid "alt text here" >`. This was *wrong:* resource shortcode only takes 1 uuid argument, the alt text etc is read from metadata.

This PR adds a rule to remove extra shortcode arguments.

This PR also...
- introduces the use of [pyparsing](https://pyparsing-docs.readthedocs.io/en/latest/pyparsing.html) as in markdown_cleanup rules
- removes a few old rules we seem unlikely to use again
- switches from batch updates to one-by-one updates

    This is less efficient but... it's not like we are updating *that* many records. And now we don't have to manually track sync_state changes. 

### How should this be manually tested?

1. Ensure you have some bad resource shortcodes... If you do not have recent rc/prod data, run `docker-compose run --rm web python manage.py markdown_cleanup rootrelative_urls -o rootrelative_urls.csv --commit` **on master branch**. This will make some new resource shortcodes that have too many arguments. (😱). Add `--skip-sync` if you want to skip the sync to github.
2. Checkout this branch.
3. Run `docker-compose run --rm web python manage.py markdown_cleanup remove_extra_resource_args -o remove_extra_args.csv --csv-only-changes` to generate a CSV of changes that will be made. This takes about 5 minutes.

5. Verify nothing has changed. (`dac7dbe9-bb2e-33c6-6027-a326888523c0` is a good page to inspect)
6. `docker-compose run --rm web python manage.py markdown_cleanup remove_extra_resource_args -o remove_extra_args.csv --commit`
7. Verify changes have been made (`dac7dbe9-bb2e-33c6-6027-a326888523c0` again)

You could run `manage markdown_cleanup shortcode_logging -o shortcodes.csv`, too. But it takes a while and does not make any changes, only logs shortcode usage.

Why Pyparsing? Because:
- it's very flexible
- its `parse_action` and `transform_string` methods provide a mechanism similar to regex substitution, so making substitutions on part of the markdown is easy(ish) 
- and I've used it before so I knew it would work. E.g., EdX's formula parser is written with PyParsing (https://github.com/openedx/openedx-calc/blob/master/calc/calc.py)

Downsides... 
- significantly slower than regexp
- no special utilities for helping with markdown

We are not too tied to pyparsing. Most of `markdown_cleanup` is independent of it. https://readthedocs.org/projects/marko/ is maybe worth looking into more, too.


websites/management/commands/markdown_cleaning/parsing_utils_test.py
```diff
@@ -0,0 +1,81 @@
...
+        ),
+        (
+            # Pyparsing hates whitespace.
+            R"The quick {{< sup    2 >}} brown fox",
+            R'The quick {{< sup "2" >}} brown fox',
+        ),
+        (
+            # Example: Convert shortcode delimiter
+            R'Take {{< resource_link uuid "that {{< sup 123 >}}" >}}, regex.',
+            R'Take {{% resource_link "uuid" "that {{< sup 123 >}}" %}}, regex.',
```
ChristopherChudzicki, line 25: This was hard to handle with the regex approach. (Note that this is valid Hugo markdown. Just, {{< sup 123 >}} renders as plaintext, not as a shrotcode. When we did delimiter replacement with regexes, this replacement was wrong and caused the build to fail.

This still won't look good, but it won't cause things to blow up.

websites/management/commands/markdown_cleaning/shortcode_grammar.py
```diff
@@ -0,0 +1,84 @@
...
+            closing_delimiter = ">}}"
+
+        if self.closer:
+            opening_delimiter += "/"
+
+        return opening_delimiter, closing_delimiter
+
+    def to_hugo(self):
+        """
+        Encases all shortcode arguments in double quotes, because Hugo allows it and that's simplest.
```
ChristopherChudzicki, line 32: Three options here:
1. Encase args in double quotes always
2. Encase args in double quotes only if necessary (if the arg contains spaces or quotes)
3. Do (2), but provide a mechanism to allow encasing other tags, too.

Originally I did (2), but that was problematic because our frontend code wants `{{< resource_link uuid "blah" >}}` even though "blah" does not really need to be encased in quotes.

(3) sounds overly complicated.

So that leaves (1), which is the simplest anyway. But our frontend code also does NOT want quotes around uuid args in resource and resource_link. But that is easy to change, so I made a PR to tolerate double quotes around the uuids. #1136

ChristopherChudzicki:
Force push had no changed, just a rebase to incorporate #1136

PR #1154, opened 2022-03-22
Cc/link parser and resolveuid

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#### What's this PR do?
This PR adds two new `markdown_cleanup` rules:
- `resolveuid`: converts markdown links/images with `resolveuid` to shortcodes or rootrelative links, depending on whether they are cross-site or not.
- `link_logging`: Logs all markdown links within a course to a CSV 

Most of the new logic is in `link_parser.py` which defines a Pyparsing-based markdown link parser. See comments that.

1. Run `docker-compose run --rm web python manage.py markdown_cleanup resolveuid -o resolveuid.csv` to generate a list of changes that will be made by the resolve uid rule.

    -  The resulting CSV will include some rows that have not changed, but it is not too big, only ~450 rows
    - the command takes about 3-5 minutes to run.
2. Run the above with `--commit` and check that the appropriate changes were made if you like. 

It might also be fun to run `docker-compose run --rm web python manage.py markdown_cleanup link_logging -o link_logging.csv` to log all markdown links. This takes about 45 minutes. 

Suggest starting with `link_parser.py`.


websites/management/commands/markdown_cleaning/parsing_utils.py
```diff
@@ -49,3 +53,149 @@ def transform_string(self, string: str):
...
+    Otherwise throws an error.
+    """
+    if text.startswith("'") and text.endswith("'"):
+        return unescape_single_quoted_string(text)
+    else:
+        return unescape_double_quoted_string(text)
+
+
+@dataclass
+class ShortcodeTag:
```
ChristopherChudzicki, line 134: This class is not new, just moved files. The two classmethods are new, though.

websites/management/commands/markdown_cleaning/link_parser.py
```diff
@@ -0,0 +1,148 @@
...
+    text_links: list = field(default_factory=tuple)
+
+    def to_markdown(self):
+        """Generate markdown representation of this link/image."""
+        prefix = "!" if self.is_image else ""
+        title_suffix = " " + json.dumps(self.title) if self.title else ""
+        return f"{prefix}[{self.text}]({self.destination}{title_suffix})"
+
+
+class LinkParser(WrappedParser):
```
ChristopherChudzicki, line 47: This is where the main new logic is. A brief look at https://spec.commonmark.org/0.30/#links might be worthwhile, too. This parser does not handle all the CommonMark edge cases, but should be good enough for us.

#### Aside: Why not use a python libraring build for Markdown parsing?
Most of the pyparsing stuff was introduced in #1137 but adding it here for some context...

A specific Markdown library might be useful. The two reasons I shied away from that for now are (A) shortcodes and (B) my impression is that with a Markdown parsing library, we would need to:
1. for a single page, parse the **whole** markdown into HTML
2. make changes on the HTML
3. convert back to markdown

That's a global transformation: Ie, we'd be changing converting and unconverting all of the markdown to make localized edits. In principal it should be fine... the conversion process *should* be lossless. But I suspect there are a lot of quirks involved... `-` vs `*` for lists, `_` vs `*` for italics, inclusion (or not) of extra whitespace... All of those things *shouldn't* matter, but... they might. And I'd rather not tempt fate.

In contrast, Pyparsing lets us make localized transformations of the markdown text. We can change a link in the middle and be confident that the pragraphs above and below did not change **at all**.

websites/management/commands/markdown_cleaning/resolveuid_rule.py
```diff
@@ -0,0 +1,95 @@
...
+            )
+            return new_link.to_markdown(), notes
+
+    def should_parse(self, text: str):
+        """Should the text be parsed?
+
+        If the text does not contain '](', then it definitely does not have
+        markdown links.
+        """
+        return "resolveuid" in text
```
gumaerc, line 95: First of all, I think the docstring here should describe that it's filtering on `resolveuid` in this case.  Secondly, why is it that I see non `resolveuid` links in the output labeled with the text set above "not a resolveuid link?"  Shouldn't `should_parse` be filtering those out?  I'm a little confused as to why anything other than `resolveuid` links are matched here.

websites/management/commands/markdown_cleaning/utils.py
```diff
@@ -15,6 +15,17 @@
 CONTENT_DIRPATH_MAX_LEN = filepath_migration.CONTENT_DIRPATH_MAX_LEN
 
 
+def is_valid_uuid(text: str):
+    """
+    Return True if text is valid uuid, esle False
```
gumaerc, line 20: typo: else

gumaerc (changes requested):
I spent a lot of time going over this and making sure I understand how `pyparsing` works and is being used here.  Overall this works very well at resolving `resolveuid` links and all the ones I manually went through ended up pointing at the right content.  I noted a few things below, and also noticed something interesting in the content itself.  I'm not sure if we can address this using markdown, but I noticed that there were some links like this:

`"[Functions of Two Variables Mathlet](./resolveuid/e1585460781926f359c7b54ce97984d3 ""Open in a new window."")","[Functions of Two Variables Mathlet](./resolveuid/e1585460781926f359c7b54ce97984d3 ""Open in a new window."")","False","markdown","18-02sc-multivariable-calculus-fall-2010","18.02sc-fall-2010","47187485-2c83-ac55-1b87-dcff9cb3f6cc","/courses/18-02sc-multivariable-calculus-fall-2010/pages/2.-partial-derivatives/part-a-functions-of-two-variables-tangent-approximation-and-optimization/session-30-second-derivative-test","False","","","UUID('e1585460-7819-26f3-59c7-b54ce97984d3')"`

You can see that some of these links have "Open in a new window." appended to the end of the URL for some reason.  There's no way to do a `target="_blank"` in markdown and it's out of scope for this PR, but in the future we might want to consider writing a shortcode for `target="_blank"`.  My main question here has to do with the fact that non `resolveuid` links show up in the results at all and I'm wondering why that is:

ChristopherChudzicki:
> "[Functions of Two Variables Mathlet](./resolveuid/e1585460781926f359c7b54ce97984d3 ...

Hey, you found the java applet we were talking about this morning! You likely noticed this, but there were a handful of resolveuid links where the uuid could not be resolved in OCW Next. Presumably this content was not migrated. In this particular case... the applet doesn't work anyway even [on the legacy site](https://ocw.mit.edu/courses/mathematics/18-02sc-multivariable-calculus-fall-2010/2.-partial-derivatives/part-a-functions-of-two-variables-tangent-approximation-and-optimization/problem-set-4/MIT18_02SC_ps4_applet.html). _Well, maybe it would for you on your retro 90s gaming computer 😄 _

> ...but in the future we might want to consider writing a shortcode for target="_blank"
We could certainly do that. I wonder if we could come up with rules like

- all links external to OCW open in a new tab (or maybe external to the current course site?) 
- links to root resources (not pages) open in a new tab (maybe??? unsure)

if we had rules like that, we could handle it with a Hugo renderHook.

Do ~normal people~ non-developers use ctrl+click / cmd+click to open links in new tabs? If so this functionality may be less compelling than it was in the early days of ocw.

> I'm a little confused as to why anything other than resolveuid links are matched here.

Does the updated docstring (for the abstract class) clear things up?

I agree this is slightly confusing.

I added a CLI option `--csv-only-changes` to only include changed matches in the CSV, which might be easier for people reviewing... _I did not want to do this by default. The unchanged matches can be very useful._

TY again for review.

gumaerc (approved):
:+1: LGTM

PR #1115, opened 2022-03-11
set parent_id when overwriting metadata.parent_uid

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
This is a followup to https://github.com/mitodl/ocw-to-hugo/issues/482, toward #1041 

#### What's this PR do?
When using `overwrite_ocw_course_content` to import `metadata.parent_uid`, also set `parent_id`.

#### How should this be manually tested?

1. Have `6-004-computation-structures-spring-2017` locally
2. Inspect WebsiteContent object `c675efdb-a53d-c4ec-1d14-46b3d44fda1e` in the admin panel
3. Observe that it has `metadata.parent_title` but no `metadata.parent_uid` or `parent` 
4. Run `docker-compose run --rm web python manage.py overwrite_ocw_course_content --bucket ocw-to-hugo-output-qa  --content-field="metadata.parent_uid" --filter 6-004-computation-structures-spring-2017`
5. Inspect the WebsiteContent object again and observe that it now has *both* `metadata.parent_uid` and `parent`. 

If you do not have `6-004-computation-structures-spring-2017` locally, do the above with any object that has metadata `parent_title` but not `parent_uid`. There are lots.

PR #1165, opened 2022-03-25
Cc/rootrel with parentage

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#### What's this PR do?
This PR:
- converts `markdown_cleanup rootrel_urls` to use the link parser introduced in #1154 
- changes `markdown_cleanup` management command to skip courses that have *never* been published.
- fix legacy links that link to files with duplicate legacy filenames (the main point of this PR)

1. First, populate `parent_uid`: `docker-compose run --rm web python manage.py overwrite_ocw_course_content --bucket ocw-to-hugo-output-qa --content-field="metadata.parent_uid"` which is needed for this command. See https://github.com/mitodl/ocw-studio/pull/1165/files#diff-c9f4154e325d48a617e2879b73236a5462ed6baa7aa2d2d6503b9106526b8e54R239

    *Note: This populates `metadata.parent_uid` AND sets the `parent_id` FK on WebsiteContent objects, see #1115*
3. Run `docker-compose run --rm web python manage.py markdown_cleanup rootrelative_urls -o rootrel.csv --csv-only-changes` to generate a list of changes that will be made when the command is run in `--comit` mode. Inspect list to makes ure it's reason / links work.

Optionally, run the command in `--commit mode` or without `--csv-only-changes`. If you run the command without `--csv-only-changes`, then the generated CSV will include ALL links on pages that contain ANY rootrelative links. This means that the CSV includes more than just "changes that will be made". In particular, it includes:
1. changes to be made, i.e., rows where `original_text != replacement`. 
2. rows that have not changed, i.e.,`original_text = replacement`. These rows might not have changed for a variety of reasons:
    - the corresponding root-relative links were not broke
    - the corresponding root-relative link was broken but could not be fixed _There are not too many of these_
    - the corresponding link was not a rootrelative link, but was on the same page as a rootrelative link

### DIFF
--- a/websites/management/commands/markdown_cleaning/parsing_utils.py
+++ b/websites/management/commands/markdown_cleaning/parsing_utils.py
@@ -72,12 +72,12 @@ class WrappedParser:
         return self.grammar.scanString(string)
 
 
-
 def escape_double_quotes(s: str):
     """Encase `s` in double quotes and escape double quotes within `s`."""
     return s.replace('"', '\\"')
 
-def unescape_string_quoted_with(text: str, single_quotes = False):
+
+def unescape_string_quoted_with(text: str, single_quotes=False):
     """
     Given a string encased in quotes of type `quote_char` and in which all
     interior instances of `quote_char` are escaped, strip the encasing instances
@@ -85,10 +85,12 @@ def unescape_string_quoted_with(text: str, single_quotes = False):
     """
     q = "'" if single_quotes else '"'
 
-    if f'\\\\{q}' in text:
-        raise NotImplementedError("Unescaping quoted strings in which backslashes precede quotes is not implemented.")
+    if f"\\\\{q}" in text:
+        raise NotImplementedError(
+            "Unescaping quoted strings in which backslashes precede quotes is not implemented."
+        )
 
-    all_escaped = text[1:-1].count(q) == text[1:-1].count(f'\\{q}')
+    all_escaped = text[1:-1].count(q) == text[1:-1].count(f"\\{q}")
     if text.startswith(q) and text.endswith(q) and all_escaped:
         return text[1:-1].replace(f"\\{q}", q)
 
--- a/websites/management/commands/markdown_cleaning/parsing_utils.py
+++ b/websites/management/commands/markdown_cleaning/parsing_utils.py
@@ -1,7 +1,7 @@
 import json
 from dataclasses import dataclass
-from uuid import UUID
 from typing import Union
+from uuid import UUID
 
 from pyparsing import ParserElement, ParseResults, originalTextFor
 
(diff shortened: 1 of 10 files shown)


websites/management/commands/markdown_cleaning/rootrelative_urls.py
```diff
@@ -157,45 +163,66 @@ def fuzzy_find_linked_content(self, url: str):
...
+                        uuid=linked_content.text_id, text=text, fragment=fragment
+                    )
+                return shortcode.to_hugo(), notes
+            except ValueError:
+                # This happens with within-site links to the homepage.
+                # The text_id of the resource is "sitemetadata", which is not
+                # a uuid. The resource link shortcode probably works in this
+                # case, but there are only 3 of these, so let's not worry about
+                # it. Plus, keeping resource_links as true UUIDs will be nice
+                # if we implement cross-site resource_links down the road.
```
ChristopherChudzicki, line 204: @gumaerc Curious if you know whether there's a reason that course homepages have `"sitemetadata"` as their text_id rather than some uuid like everything else.

gumaerc: So this is because the file behind this, `course.json`, is generated during a run of `ocw-to-hugo`.  There's no equivalent file in Plone, hence no UUID.  If we generated a new UUID, it would be different during every run of `ocw-to-hugo`.  When we were regularly running `import_ocw_course_sites`, we needed a way to make sure that changes to the metadata file were updated instead of creating a new `WebsiteContent` object every time.

websites/management/commands/markdown_cleaning/parsing_utils.py
```diff
@@ -72,73 +72,43 @@ def scan_string(self, string: str):
...
-        return json.loads(double_quoted)
-    elif text.startswith('"') and text.endswith('"'):
-        return json.loads(text[1:-1])
+def escape_double_quotes(s: str):
+    """Encase `s` in double quotes and escape double quotes within `s`."""
+    return s.replace('"', '\\"')
 
 
-def unescape_single_quoted_string(text: str):
+def unescape_string_quoted_with(text: str, single_quotes=False):
```
ChristopherChudzicki, line 80: @gumaerc This change (see https://github.com/mitodl/ocw-studio/pull/1165/commits/d062254cf2b77b3d38604bd4348e505819ab7f3b) is a new commit after the rebase.

Using json.loads/dumps for escaping/unescaping stuff was not a great idea. I found this was doing unwanted things, like converting `"«"` to `"\\u00ab"`. BTW: My goal with the unescaping was that
- `Shortcode(name='whatever', args=['This has "quotes" in it']).to_hugo()` should handle the escaping for you...
- So the internal shortcode representation should be unescaped, so the parser needs to unescape args when creating the shortcodes. 

The extra escaping from json dumps/loads should not have caused any problems on prod yet. The two places this was used were
- link titles `[text](destination "title")` which tend not to have quotes in the anyway / aren't really supported in ocw-next yet
- the shortcode parser has only been used so far for removing extra arguments from `< resource >`

Issue #1015, opened 2022-02-16
Data migration: Finish baseurl conversion


ChristopherChudzicki:
Finish the `baseurl -> resource_link` conversion started in #1002 

Known remaining edge cases:
1. [x] links with document fragments, e.g., `[calendar]({{< baseurl >}}/pages/syllabus#Calendar)` 
2. [x] links where the link text has square brackets or newlines (not double newlines—those aren't links according to Markdown). 
3. [ ] links where the link casing is different from dirpath / filename casing, e.g., `[Syllabus]({{< baseurl >}}/path/to/Syllabus)` where the syllabus filename is `syllabus` not `Syllabus`.
4. [ ]  Links where the title is an image, e.g.
    `[![Boundary conditions animation.](/courses/civil-and-environmental-engineering/1-061-transport-processes-in-the-environment-fall-2008/lecture-notes/movie.jpg)]({{< baseurl >}}/resources/bounds3)`

    ⚠️ Investigate the usage images in titles. Do we actually want to migrate/support this in future? 

Related: 
- #969 
- https://github.com/mitodl/ocw-hugo-themes/issues/523

ChristopherChudzicki:
Even after #1013, the square bracket links BEFORE baseurl shortcodes are being escaped. See for example https://github.mit.edu/ocw-content-rc/18.04-fall-1999/commit/97032ead4f7bf3810c758522b30451dfc7538f00#diff-9674b1ed65559a66e97cbc6ada224051L24-R24

Consequences:
- this will break the baseurl link if they were working

BUT: Since we are working on replacing baseurl with resource_link anyway...one option is to not worry about this and just alter to replacement regex to catch both the escaped version (`\[link title\]({{ baseurl}/...)`) and unescaped version (`[link title]({{ baseurl}/...)`)

ChristopherChudzicki:
Closing this. We had 92k originally and about 400 remain. Will add those to the list of may-need-manual-fixes which I am compiling.

PR #1174, opened 2022-03-28
Convert more baseurl links to resource_links 

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

This completes the second task in https://github.com/mitodl/ocw-studio/issues/1015

#### What's this PR do?
This PR does more baseurl --> resource_link conversion using the new LinkParser. In particular, we now convert about 950 more links that have square brackets in their link text or (non-consecutive) newlines. The earlier regex approach could not handle this.

1. **First, unescape some stuff:** Run `docker-compose run --rm web python manage.py markdown_cleanup link_unescape -o link_unescape.csv --commit` to unescape some link square brakcets that ought not to be escaped. Inspect the CSV to ensure replacements look reasonable.
2. **Now convert baseurls** Run `docker-compose run --rm web python manage.py markdown_cleanup baseurl -o baseurl.csv --csv-only-changes to generate a list of changes that will be made when the command is run in --comit mode.
3. Inspect list to makes sure it's reasonable / links work.
4. You may want to specifically check that the new resource links do NOT include newlines. Markdown allows non-consecutive newlines in the link text but does render them. Hugo shortcodes do NOT allow newlines in the shortcode arguments, as far as I can tell.


mbertrand (approved):
`markdown_cleanup baseurl ` took a really long time to run but the results look good to me.  Just noticed one minor docstring typo,

PR #599, opened 2021-09-23
update ocw_import to support ocw-to-hugo 1.30.3

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested
  - [ ] Changes have been manually tested

#### What are the relevant tickets?
Closes https://github.com/mitodl/ocw-studio/issues/598

#### What's this PR do?
In https://github.com/mitodl/ocw-to-hugo/pull/350 and some subsequent PR's, we began altering `course.-json` and created `course_legacy.json` to preserve the state of `course.json` before modifications started being made to conform to the structure of the `ocw-course` starter config at https://github.com/mitodl/ocw-to-hugo/pull/350.  This PR adjusts the `ocw_import` code to use this file instead and updates the import process to support `ocw-to-hugo` 1.30.3` which includes changes to topics as well as changes to the structure of content files from PR's like https://github.com/mitodl/ocw-to-hugo/pull/337 that change the location of resource markdown.

#### How should this be manually tested?
 - Set AWS credentials for `open-learning-course-data-rc` (or production) and make sure you are set up to use `ocw_import` by having an `ocw-www` site based off of https://github.com/mitodl/ocw-hugo-projects/blob/main/ocw-www/ocw-studio.yaml
 - Spin up `ocw-studio` locally
 - Run `docker-compose run web ./manage.py import_ocw_course_sites -b ocw-to-hugo-output-rc --limit 1`
 - Verify that you have a site at http://localhost:8043/sites/1-00-introduction-to-computers-and-engineering-problem-solving-spring-2012/ successfully imported.  It should have pages, resources and instructors in the metadata.

PR #1191, opened 2022-03-31
Use first_published_to_production instead of publish_date for sorting new courses, get metadata from WebsiteContent

### LINKED ISSUE #1188: Home: list of "new" courses based on first published date instead of publish date
We use a studio API to provide a list of "new" courses for the home page. 

Currently this list is sorted by publish date, which means that every time a course is published -- even if its just to fix a typo -- it shows up in the list of new courses. 

Instead of using publish date, we should use first published date, which should surface courses that were recently published for the first time.

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #1188 

#### What's this PR do?
- Switches the sorting of the api query for new courses from `publish_date` to `first_published_to_production`
- In the websites API response, get the metadata from each site's associated `sitemetadata` `WebsiteContent` object instead of the `Website.metadata` field (because that field is null for all sites created in studio).
- Modify the instructors section of the metadata so that it contains the keys currently required by ocw-hugo-themes for the new courses carousel (see related issue [#599](https://github.com/mitodl/ocw-hugo-themes/issues/599))

In OCW Studio:
- Create 2 new sites.  Save metadata (including 1+ instructors) for one but not the other. 
- In django admin, manually set `first_published_to_production` for both to the current date/time.
-In an incognito browser window (not logged in to studio), go to http://localhost8043/api/websites and verify that:
  - the list of courses is in descending order by the `first_published_to_production` field
  - the new course with the metadata is present in the list and includes instructor info such as title
  - the new course without metadata is not present in the list.
  - imported sites are also in the list and have non-null metadata values.
  
For ocw-hugo-themes:  
- Set `OCW_STUDIO_BASE_URL=http://localhost:8043` in your `.env` file for `ocw-hugo-themes`, 
- run `npm run start:www`
- check that the carousel of new courses includes the first 10 courses as in the list above and looks/functions properly.

If/when [#599](https://github.com/mitodl/ocw-hugo-themes/issues/599) is implemented, the code introduced here to modify the `instructors` part of the metadata can be removed, though it is more informative than the original format.

### DIFF
--- a/websites/factories.py
+++ b/websites/factories.py
@@ -43,6 +43,7 @@ class WebsiteFactory(DjangoModelFactory):
     short_id = factory.Sequence(lambda n: "site-shortid-%s" % n)
     metadata = factory.Faker("json")
     publish_date = factory.Faker("date_time", tzinfo=pytz.utc)
+    first_published_to_production = factory.Faker("date_time", tzinfo=pytz.utc)
     draft_publish_date = factory.Faker("date_time", tzinfo=pytz.utc)
     starter = factory.SubFactory(WebsiteStarterFactory)
     owner = factory.SubFactory(UserFactory)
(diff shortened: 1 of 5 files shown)


ChristopherChudzicki (approved):
Thanks for the detailed testing instructions! I had overlooked using local studio for hugo-themes till now.

PR #1192, opened 2022-04-01
tolerate href, href_uid on resource shortcode

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#### What's this PR do?
We recently added a few new named parameters to the resource shortcode ("uuid", "href", and "href_uuid"---see https://github.com/mitodl/ocw-hugo-themes/pull/538). This commit makes CKEditor aware of these new parameters.

Additional changes:
- new Shortcode class for generic Typescript representation of Shortcodes.
- Added the CKEditor inspector (https://ckeditor.com/docs/ckeditor5/latest/framework/guides/development-tools.html) during development

0. Maybe need to rebuild to pick up the new npm package? I had some trouble with this, but I think it was specific to my setup... my docker complained about running out of space. I ended up deleting all my containers/images/volumes and rebuilding, and now everything is smooth. 
1. Open the course `18-04-complex-variables-with-applications-fall-1999` 
2. Using django admin, replace the markdown contents of some page (create a new one if you want) with the markdown below. *Note: You cannot add href and href_uuid params to resource shortcodes via CKEditor right now. You must add them via admin panel*.

    ```
    Currently OCW contains a bunch of link-wrapped images like the first one below.
    They get screwed up when you save. (Extra escapes)
    
    [{{< resource "d76bb2ff-8097-4548-ca14-2acc42d17b0c" >}}]({{< baseurl >}}/pages/study-materials/function4view1)  
    
    We are going to convert them to these things, which will not be screwed up when you save:

    {{< resource uuid="d76bb2ff-8097-4548-ca14-2acc42d17b0c" href_uuid="756fad96-a8b6-cb78-2c0f-1b8a186d2e8e" >}}
    
    {{< resource uuid="d76bb2ff-8097-4548-ca14-2acc42d17b0c" href="https://www.mit.edu" >}}
    ```

4. Open the page in studio's CKEditor
5. Make some edits and save
6. Verify that the href and href_uuid parameters are still there. NOTE: The first line above will get messed up when you save (extra escapes). That is expected. It's why we are doing this replacement.
7. You also might want to check out CKEditor's dev-tools. With CKEditor open, you can open the tools from bottom of your browser.
8. Try embedding new resources and verify that works, too.


static/js/components/StudioList.tsx
```diff
@@ -54,7 +54,7 @@ export function StudioListItem(props: ListItemProps): JSX.Element {
   )
 
   const closeMenu = useCallback(
-    (e: React.MouseEvent) => {
+    (e: MouseEvent) => {
```
ChristopherChudzicki, line 57: When I installed CKEditorInspector, some packages got minor/patch bumps (see yarn.lock). In particular, `@types/react-outside-click-handler` got bumped from v1.3.0 to v1.3.1, which changed the event type on these handlers from `React.MouseEvent` to `MouseEvent`. (The old type was incorrect. It has always been a MouseEvent. Nothing in the runtime code changes).

static/js/components/widgets/MarkdownEditor.tsx
```diff
@@ -67,7 +71,7 @@ export default function MarkdownEditor(props: Props): JSX.Element {
         editor.current.editing.view.focus()
       }
     },
-    [editor]
+    []
```
ChristopherChudzicki, line 74: This is unrelated to everything else. But `editor` is a ref—its reference never changes. So there's no point in including it as a dependency.  (And React's smart linting rules will not complain if we remove it)

static/js/lib/ckeditor/plugins/ResourceEmbed.ts
```diff
@@ -185,22 +220,12 @@ class ResourceEmbedEditing extends Plugin {
  * into the editor. These are rendered to Markdown as `{{< resource UUID >}}`
  * shortcodes.
  */
-export default class ResourceEmbed extends Plugin {
+export default class ResourceEmbed extends CKEPlugin {
   static get pluginName(): string {
     return "ResourceEmbed"
   }
 
-  static get requires(): Plugin[] {
```
ChristopherChudzicki, line 193: using `extends Plugin` was confusing Typescript. TS thought this was referring to the global, deprecated [Plugin](https://developer.mozilla.org/en-US/docs/Web/API/Plugin) interface.

mbertrand (approved):
Works great 👍

PR #1202, opened 2022-04-05
New sites API: Filter out sites without sitemetadata content instead of Website.metadata

### LINKED ISSUE #1188: Home: list of "new" courses based on first published date instead of publish date

mbertrand:
~~Blocked by https://github.com/mitodl/ocw-hugo-themes/issues/599~~

#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #1188

#### What's this PR do?
- Sorts the anonymous websites api response by reverse `first_published_to_production` field
- Instead of filtering out sites with null `Website.metadata` (excludes all new sites created in studio), filter out sites without a `sitemetadata` resource.

Check the output of `http://localhost:8043/api/websites` in an incognito browser, make sure it is sorted by `first_published_to_production` , most recent first and includes published sites created in studio.  If necessary, create a new site, add metadata to it, and set the `first_published_to_production` to the current date/time in django admin.

~~Should not be merged until after https://github.com/mitodl/ocw-hugo-themes/issues/599 is released, otherwise the new courses carousel on the home page will break.~~

### DIFF
--- a/websites/views.py
+++ b/websites/views.py
@@ -34,6 +34,7 @@ from users.models import User
 from websites import constants
 from websites.api import get_valid_new_filename, update_website_status
 from websites.constants import (
+    CONTENT_TYPE_METADATA,
     RESOURCE_TYPE_DOCUMENT,
     RESOURCE_TYPE_IMAGE,
     RESOURCE_TYPE_OTHER,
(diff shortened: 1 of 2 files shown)


abeglova (approved):
LGTM

Issue #1167, opened 2022-03-25
disable mass import


pdpinch:
Now that we have users in ocw-studio working on legacy content, it's not safe to run imports unfiltered. 

We should remove the code, to avoid any possibility of user error.

gumaerc:
So there are 2 management commands that can mass-import `ocw-to-hugo` data; `import_ocw_course_sites` which imports full courses and `overwrite_ocw_course_content` which overwrites single metadata properties in front matter.  I think for the time being, we will still need the latter, but with the former we could make `--filter` a required argument.  This would still allow someone to do something like `./manage.py import_ocw_course_sites -b ocw-to-hugo-output-production --filter 18-` and import an entire department though.  If we wanted to be more strict about it, we could change that to specify a course or list of courses like `mass_publish`, or we could just remove the command altogether.  I think we should keep it around for importing 1-off courses though, so I think making filter a required field and standardizing it to the way we filter on courses with our other commands would be the best approach.  What do you think @pdpinch ?

pdpinch:
In general, standardizing is good. 

In this particular case, what I had in mind was requiring the filter parameter. Does it have a minimum number of characters (perhaps 2)?

PR #1214, opened 2022-04-07
prevent mass import

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1167

#### What's this PR do?
After recently releasing the new OCW site live and going through a month long QA process getting the content into shape, it would be a big problem if `import_ocw_course_sites` were to be run unfiltered or with a vague filter.  The tasks that `import_ocw_course_sites` runs destructively updates content on imported legacy sites, and since a lot of manual work was done in production to get the site ready for launch, we want to avoid destroying this data.  This PR does a few things to prevent that:

 - Either `--filter` or `--filter-json` is now required when running the command
 - A confirmation step was added that lets you know how many sites will be imported, asking you to type `y` and hit enter before proceeding

 - Get `ocw-studio` up and running locally on this branch, making sure you've configured AWS access (okay to match RC) as well as a private Github org for publishing
 - Put the following in a `courses.json` file at the root of your `ocw-studio` folder:
```
[
  "8-591j-systems-biology-fall-2004",
  "15-057-systems-optimization-spring-2003",
  "6-776-high-speed-communication-circuits-spring-2005"
]
```
 - Once up and running, run the following commands:
   - `docker-compose run --rm web ./manage.py import_ocw_course_sites -b ocw-to-hugo-output-qa` - This should throw an error of `CommandError: This command cannot be run unfiltered.  Use the --filter or --filter-json argument to specify courses to import.`
   - `docker-compose run --rm web ./manage.py import_ocw_course_sites -b ocw-to-hugo-output-qa --filter "8-591j-systems-biology-fall-2004, 15-057-systems-optimization-spring-2003, 6-776-high-speed-communication-circuits-spring-2005"` - This should prompt you to confirm importing 3 courses, press y and then enter to confirm
   - `docker-compose run --rm web ./manage.py import_ocw_course_sites -b ocw-to-hugo-output-qa --filter-json courses.json` - This should also prompt you to confirm importing 3 courses, press y and then enter to confirm
   - `docker-compose run --rm web ./manage.py import_ocw_course_sites -b ocw-to-hugo-output-qa --filter "8-591j-systems-biology-fall-2004, 15-057-systems-optimization-spring-2003, 6-776-high-speed-communication-circuits-spring-2005"` - This should prompt you to confirm importing 3 courses, type anything other than `y` and hit enter and you should see `Aborting...`

### DIFF
--- a/ocw_import/management/commands/import_ocw_course_sites.py
+++ b/ocw_import/management/commands/import_ocw_course_sites.py
@@ -120,8 +120,6 @@ class Command(BaseCommand):
                 bucket_name, prefix=prefix, filter_list=filter_list
             )
         )
-        if limit is not None:
-            course_paths = course_paths[:limit]
 
         if options["list"] is True:
             pydoc.pager("\n".join(course_paths))
@@ -141,6 +139,7 @@ Would you like to proceed with the import? (y/n): """
             bucket_name=bucket_name,
             course_paths=course_paths,
             prefix=prefix,
+            limit=limit,
             delete_unpublished=delete_unpublished,
             chunk_size=options["chunks"],
         )
--- a/ocw_import/management/commands/import_ocw_course_sites.py
+++ b/ocw_import/management/commands/import_ocw_course_sites.py
@@ -5,7 +5,6 @@ import pydoc
 from django.conf import settings
 from django.core.management import BaseCommand, CommandError
 from mitol.common.utils.datetime import now_in_utc
-from six.moves import input
 
 from content_sync.tasks import sync_unsynced_websites
 from ocw_import.api import fetch_ocw2hugo_course_paths
--- a/ocw_import/management/commands/import_ocw_course_sites.py
+++ b/ocw_import/management/commands/import_ocw_course_sites.py
@@ -1,9 +1,11 @@
 """ Import OCW course sites and content via ocw2hugo output """
+import json
 import pydoc
 
 from django.conf import settings
-from django.core.management import BaseCommand
+from django.core.management import BaseCommand, CommandError
 from mitol.common.utils.datetime import now_in_utc
+from six.moves import input
 
 from content_sync.tasks import sync_unsynced_websites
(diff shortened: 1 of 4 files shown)


ocw_import/management/commands/import_ocw_course_sites.py
```diff
@@ -1,9 +1,11 @@
 """ Import OCW course sites and content via ocw2hugo output """
+import json
 import pydoc
 
 from django.conf import settings
-from django.core.management import BaseCommand
+from django.core.management import BaseCommand, CommandError
 from mitol.common.utils.datetime import now_in_utc
+from six.moves import input
```
ChristopherChudzicki, line 8: **Question:** Do we need six? We're not trying to support Python 2...

ChristopherChudzicki (commented):
Still running commands. Two questions

1. One about six... see comment
2. Maybe not worth the effort right now, and not really the point of this PR... but question: could we do more of the filtering on s3-server-side with with prefix? Right now we fetch all the paths then filter on our end, right?

ChristopherChudzicki (approved):
👍 ...guard rails 🎉.

still questions, but they not blocking.

ocw_import/management/commands/import_ocw_course_sites.py
gumaerc: 1.  I was confused about how `input` works in management commands, and it seems you don't need to use the `six` compatibility layer to get that to work.  I switched it to just use the native `input` function.

2.  The server side filtering is definitely a massive time sink and we should bake filtering into the actual use of the S3 API, but that's out of scope for this PR.  I made issues for that and another thing I noticed that we should change now that editing is turned off in Plone here:

https://github.com/mitodl/ocw-studio/issues/1217
https://github.com/mitodl/ocw-studio/issues/1218

PR #1210, opened 2022-04-06
add "other" tab to resource picker

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Screenshots and design review for any changes that affect layout or styling
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What's this PR do?
~This PR fixes broke links of the form...Just Kidding!~

This PR updates the resource picker dialog so that resources with `resourcetype="Other"` are now linkable. So, all resources should now be linkable. Also:
- fixed a bug where documents could be embedded. Checked with Peter/Ferdi...this was undesirable. 
- Added extension + "updated on" to help authors find relevant files. 

#### How should this be manually tested?
Using a new course or one you already have...
1. Upload some files to google drive and sync the content. Be sure to include a variety of file types: PDF, txt, json, zip files, whatever.

    _In particular, be sure to include some files whose MIME type does NOT start with 'text', 'image', or 'video'. These files will be classified as `resourcetype="Other"`. _
2. Go edit a page
3. Click "Add Link". Confirm that everything you uploaded is in one of the tabs, specifically json / docx, etc should be in "Other"
4. Confirm that extensions + updated on looks correct.
5. Confirm that Documents + Other are NOT available for embedding. 

Note: the Add Link -> "Document" and "Other" tabs are functionally equivalent right now, except one shows text documents and the other "Other" documents. 

Screen Shot 2022-04-06 at 4 41 50 PM

Issue #1168, opened 2022-03-25
Include more file types in add link widget


pdpinch:
We started imported a wide range of text file some time ago, in https://github.com/mitodl/ocw-data-parser/issues/169

Course authors would like to be able to link to these file types, but they don't appear in the add link widget. 

For example, in https://ocw-studio-rc.odl.mit.edu/sites/1-017-computing-and-data-analysis-for-environmental-applications-fall-2003/type/page/edit/dc1d597b-6acf-4e17-f390-0ffb0e32fdf5/ the user can't add a link to the file cdffit.m which does exist in the course (actually, it exists twice).

ChristopherChudzicki:
@pdpinch I want to double check on the desired behavior here regarding "Link" and "Embed".  From slack yesterday:

>I was talking to Ferdi about [this issue] & he suggested just adding a tab for “Other.” Do you see any problems with that?

I said "No". And this is in fact easy to do. But I want to double check on the desired behavior around "Add link" and "Embed" because I seem to have misunderstood how "Embed" currently works (in the case of documents). 

Currently it seems:
1. Add Link: In all cases, this creates a link to the "container page".
2. Embed, images and videos: really "embeds" images and videos in the current page.
3. Embed --> Document: Creates a link which is a direct download. *In particular,"Embedding" a PDF files does not embed a PDF viewer, which is sort of what I was expecting based on the name.*

    Also, for some reason I thought "Document" meant specifically PDF. But it seems to include other filetypes, too. (See screenshot from RC below). 

**Question A:** Is (3) the correct / expected behavior?

**Question B:** If Yes, then I think it might make more sense to just add all files to the "Documents" tab rather than creating a new "Other" tab. They would be functionally the same. 

- Whereas for all resource types, "Link" creates a link to the container page (not a direct download). 

So 


 I guess it turns out I've never tried embedding a

Ferdi:
I think we can only embed video/images.  Embed should not show on the rest.

ChristopherChudzicki:
Cool. After talking with Peter, I am going to remove "Documents" from the Embed picker.

PR #1216, opened 2022-04-08
Cc/default active tab

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#### What's this PR do?
This is a followup to #1210 .. That PR removed the "Document" tab from embed (cf https://github.com/mitodl/ocw-studio/issues/1168#issuecomment-1090676620), which introduced a bug where no tab would be initially visible when the picker was opened in "Embed mode".

So this PR ensures that the initially active tab is one of the tabs available in the current mode.

1. Open a page in editing mode, and click "Add link". The first tab should be visible and you should be able to switch tabs.
2. Close out of the resource picker and click "Embed Resource". The first tab should be visible and you should be able to switch tabs.
3. Close the editor and repeat, but this time open Embed first.

#### Also
This PR also makes a small change to `ShortcodeParam` in the markdown cleanup management command. This change is unrelated to the rest. Tests should still pass. If you really want to test that change, run `docker-compose run --rm web python manage.py markdown_cleanup shortcode_logging -o shortcodes.csv` to generate a CSV of all shortcodes in ocw. But it takes like an hour and the csv is 350MB.

### DIFF
--- a/websites/management/commands/markdown_cleaning/parsing_utils.py
+++ b/websites/management/commands/markdown_cleaning/parsing_utils.py
@@ -104,7 +104,7 @@ def unescape_string_quoted_with(text: str, single_quotes=False):
     if text.startswith(q) and text.endswith(q) and all_escaped:
         return escaped_quote_regex.sub(unescape, text[1:-1])
 
-    raise ValueError(f"{text} is not a valid single-quoted string")
+    raise ValueError(f"{text} is not a valid {q}-quoted string")
 
 
 def unescape_quoted_string(text: str):
@@ -127,7 +127,7 @@ class ShortcodeParam:
     name: Union[str, None] = None
 
     param_regex: ClassVar[re.Pattern] = re.compile(
-        r"^((?P<name>[0-9a-zA-Z_]+)=)?(?P<value>.*)$"
+        r"^((?P<name>[0-9a-zA-Z_\-]+)=)?(?P<value>.*)$"
     )
 
     @classmethod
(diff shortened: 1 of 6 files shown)


websites/management/commands/markdown_cleaning/parsing_utils.py
```diff
@@ -127,7 +127,7 @@ class ShortcodeParam:
     name: Union[str, None] = None
 
     param_regex: ClassVar[re.Pattern] = re.compile(
-        r"^((?P<name>[0-9a-zA-Z_]+)=)?(?P<value>.*)$"
+        r"^((?P<name>[0-9a-zA-Z_\-]+)=)?(?P<value>.*)$"
```
ChristopherChudzicki, line 130: In particular, the `data-ngdescrip` shortcode param in `< image-gallery-item >` necessitated this change.

websites/management/commands/markdown_cleaning/shortcode_grammar_test.py
```diff
@@ -89,3 +89,23 @@ def test_shortcode_grammar_with_nested_shortcodes():
     with pytest.raises(ValueError, match="nesting"):
         text_nested = R'{{< fake_shortcode uuid {{< sup 4 >}} "Cats and dogs" >}}'
         parser.parse_string(text_nested)
+
+
+def test_shortcode_parser_parses_named_parameter_and_dashes():
```
mbertrand, line 94: docstring

mbertrand (approved):
👍

Issue #1209, opened 2022-04-06
Pagination is broken when filters are applied


ChristopherChudzicki:
### Steps to Reproduce
1. Visit https://ocw-studio.odl.mit.edu/sites/18-04-complex-variables-with-applications-spring-2018/type/resource/?q=Topic _Or any filtered listing page with more than 10 list items_
2. Click the paginate buttons at bottom `>`

### Expected Behavior
Next page of results loads

### Actual Behavior
Next page of results loads and then you're quickly brought back to the first page.  😭

ChristopherChudzicki:
This went out in #1237

PR #1220, opened 2022-04-08
fix react-router / history typings

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

This is toward #1209, but there's no functionality here. I'd like to do it as a separate PR for git history visibility and since there is zero functionality change. 

#### What's this PR do?
Make typescript more useful with react-router. For example:

**Before / After**
Screen Shot 2022-04-08 at 12 32 49 PM / Screen Shot 2022-04-08 at 12 30 11 PM

See commit if you want all gory details.

Tests should pass. Routing should still work.


static/js/types/types.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { expectTypeOf } from "expect-type"
```
ChristopherChudzicki, line 1: I've used this in the past on some personal projects and when contributing typedefs to some open source repos (e.g.,  https://github.com/sequelize/sequelize). It's nice if you want to assert stuff about types in tests.

static/js/util/integration_test_helper.tsx
```diff
@@ -169,6 +169,7 @@ export default class IntegrationTestHelper {
       const wrapper = mount(
         <Provider store={store}>
           <ReduxQueryProvider queriesSelector={getQueries}>
+            {/* @ts-ignore */}
             <Router history={this.browserHistory}>
```
ChristopherChudzicki, line 173: This `ignore` was not previously necessary because previously `react-router` was using `v5` history types. But that was incorrect, since it really uses v4 internally. And react-router using v5 history types broke some typings elsewhere.

PR #1226, opened 2022-04-11
Cc/fix pagination

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#1209 

#### What's this PR do?
This refactors pagination for pages that are using the `useURLParamFilter` and fixes an issue where pagination was not preserving query parameters.

This also includes some tinkering with our types for react-router / history. See https://github.com/mitodl/ocw-studio/pull/1226/commits/ac6e7ff27399f268e12d90118f7aec83c0c59ace

On master:
1. On master...
2. Find any content listing that contains lots of item. For example http://yourname.ocw-studio.odl.local:8043/sites/8-591j-systems-biology-fall-2004/type/resource
3. Add a search query in the filter box such that you still have more content than fits on one page. For example, "s"
4. Try to go to "next page" with navigation at bottom. It'll work for a moment, then you'll be brought back to the first page. 😠 
5. Repeat on this branch, `cc/fix-pagination`
    - pagination should work and preserve query parameters
    - pagination should work for content listing and at /sites


static/js/hooks/search.ts
```diff
@@ -76,3 +79,50 @@ export function useURLParamFilter<LParams extends ListingParamsMinimum>(
...
+  count: number,
+  pageSize: number = WEBSITE_CONTENT_PAGE_SIZE
+): {
+  previous: RLocation | null
+  next: RLocation | null
+} => {
+  const location = useLocation()
+  const previous = offsetLocation(location, count, -pageSize)
+  const next = offsetLocation(location, count, +pageSize)
+  return { previous, next }
```
ChristopherChudzicki, line 127: **Comment:** The API responses from our listings APIs include ` { count, previous, next }`. Of that info, we're only using `count` and not previous/next for the frontend pagination.

I did consider basing this logic off of the API responses. One challenge with that is that the API routes do not map directly to our frontend routes. I decided logic of the form "keep frontend url exactly the same except for offset parameter" was preferable to trying to convert API routes to frontend routes since that varies about by scenario (content listing, site listing).

mbertrand (approved):
LGTM 👍

Issue #1161, opened 2022-03-25
required=true not working for WidgetVariant=relation


MAbdurrehman12:
### Steps to Reproduce
- Copy the starter config from `ocw-hugo-projects` or from any existing starter from studio admin.
- Change this config and set `required=true` for any relation field. 
e.g:
```
- label: Image
        name: image
        widget: relation
        **required: true**
        collection: resource
```

### Expected Behavior
- The field for which required is set to true, must be required now and an appropriate error message should be shown if a user tries to save the object without selecting any option for that field.

### Actual Behavior
- Object is successfully saved even without selecting any option for the relation field

### Related Issue
https://github.com/mitodl/ocw-hugo-projects/issues/147

### Issue Finding
https://github.com/mitodl/ocw-hugo-projects/issues/147#issuecomment-1077505751

PR #1240, opened 2022-04-13
fix: required=true for relation widgetvariant

MAbdurrehman12:
#### Pre-Flight checklist

- [ ] Migrations
- [x] Testing
  - [x] Code is tested
- [ ] Settings

Closes https://github.com/mitodl/ocw-studio/issues/1161

#### What's this PR do?
- Conditionally calls formik `required()` for fields having widget: `relation`, required: `true` and multiple: `false`
(More detail of this issue has been explained [here](https://github.com/mitodl/ocw-hugo-projects/issues/147#issuecomment-1077505751))

- For `widget=relation` fields, set `required=true` in config of website starter. (make sure multiple is false)
- Add/Edit object which has the field for which required has been set to true.
- Verify that an appropriate error is shown when you try to save the object without selecting any value for that field.
- Any other use-case, that needs to be tested with respect to this change.

image

image

### DIFF
--- a/static/js/components/forms/validation.ts
+++ b/static/js/components/forms/validation.ts
@@ -16,6 +16,7 @@ import {
   WidgetVariant
 } from "../../types/websites"
 import { FormSchema, SiteFormValues } from "../../types/forms"
+import { IS_A_REQUIRED_FIELD } from "../../constants"
 
 // This is added to properly handle file fields, which can have a "null" value
 setLocale({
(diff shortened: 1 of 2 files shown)


mbertrand (approved):
👍 , works great when `multiple=false`.   Do we need to eventually get `required=true` working with `multiple=true` too?

MAbdurrehman12:
> Do we need to eventually get required=true working with multiple=true too

I was not very sure about this scenario, as when `multiple=true`, `min` and `max` are the controlling attributes for validation (validation is being done based on their value), so I don't think we will be needing `required=true` in this case or will we need it?

PR #1244, opened 2022-04-13
Sync Website.title with the sitemetadata course_title

### LINKED ISSUE #1243: Website title should be synced with sitemetadata title value
### Steps to Reproduce

Create a new site with title "My New Course"
Open the site metadata edit page and enter "My New Course, Modified" as the course title.  Save.

### Expected Behavior

The site title is now displayed as "My New Course, Modified" on the page and in the sites list.

### Actual Behavior

The original site title is still displayed on the page and in the sites list.

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested
- [x] Settings

Closes #1243 

#### What's this PR do?
If a `WebsiteContent` object of type `sitemetadata` has a `course_site` field (specified in the new `settings.FIELD_METADATA_TITLE`), then update the related `Website.title` to have the same value.

- Go to the metadata edit page for a new site and enter a course title, then save.  The `Website` title should be updated on the page to the new value, and should appear in the sites list with the new value.
- Edit the metadata title again and and save, the website title on the page should update to reflect that.
- Edit some other content for the site - pages, resources, etc.  They should save normally without any errors.

### DIFF
--- a/app.json
+++ b/app.json
@@ -139,6 +139,10 @@
       "description": "Gdrive folder for video uploads",
       "required": false
     },
+    "FIELD_METADATA_TITLE": {
+      "description": "The site metadata field for title",
+      "required": false
+    },
     "FIELD_RESOURCETYPE": {
       "description": "The site config metadata field for the resource type",
       "required": false
--- a/main/settings.py
+++ b/main/settings.py
@@ -548,6 +548,11 @@ FIELD_RESOURCETYPE = get_string(
     default="resourcetype",
     description="The site config metadata field for the resource type",
 )
+FIELD_METADATA_TITLE = get_string(
+    name="FIELD_METADATA_TITLE",
+    default="course_title",
+    description="The site metadata field for title",
+)
 
 # YouTube OCW metadata fields
 YT_FIELD_CAPTIONS = get_string(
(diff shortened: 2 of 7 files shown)


pdpinch:
I'm still a bit concerned that this creates too strong a connection between ocw-studio and the internals of the ocw-course project structure. 

If we had a metadata field named simply "title" it would make sense to me that it should be displayed always in site listings. 

I'd like to get another opinion on this feedback.

mbertrand:
@pdpinch I tried to avoid too strong a connection by adding a new setting variable `FIELD_METADATA_TITLE` defaulting to `course_title`.  If someone uses a starter besides ocw-course that has a different field for title or has no sitemetadata content, this new code should have no effect and the behavior will be the same as before.  

It might make sense for that field to be simply `title` instead of `course_title` but that renaming would require updates to all the current sitemetadata resources and a change to ocw-hugo-themes as well.

Alternately, I could change this new setting to be a list of possible fields like `["course_title", "title"]` and if any of those are in the sitemetadata then it will be used to update the `Website.title`

abeglova (approved):
LGTM

PR #1257, opened 2022-04-20
Added a check to return as soon as filter_set has been gone over

### LINKED ISSUE #1217: fetch_ocw2hugo_course_paths always crawls the entire bucket
### Steps to Reproduce

Run an import management command like `import_ocw_course_sites` or `overwrite_ocw_course_content` that calls `fetch_ocw2hugo_course_paths` using the `--filter` or `--filter-json` argument to filter on a subset of courses

### Expected Behavior

The function only checks the paths for the courses specified in the filter

### Actual Behavior

The function crawls the entire S3 bucket, which takes a long time

Wassaf-Shahzad:
#### Pre-Flight checklist

- [x] Testing

Closes #1217 

#### What's this PR do?
Added check to immediately return when filter_set has been processed

- Verify fetch_ocw2hugo returns as soon as filter_set is processed
- Verify it doesn't crawl the whole bucket

### DIFF
--- a/ocw_import/api.py
+++ b/ocw_import/api.py
@@ -5,6 +5,7 @@ import re
 import uuid
 from copy import deepcopy
 from typing import Dict
+from os import path
 
 import dateutil
 import yaml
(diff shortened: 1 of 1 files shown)


ocw_import/api.py
```diff
@@ -74,13 +74,17 @@ def fetch_ocw2hugo_course_paths(bucket_name, prefix="", filter_list=None):
     s3 = get_s3_resource()
     bucket = s3.Bucket(bucket_name)
     paginator = bucket.meta.client.get_paginator("list_objects")
+    counter = 0
     for resp in paginator.paginate(Bucket=bucket.name, Prefix=f"{prefix}"):
```
gumaerc, line 78: Instead of keeping a counter and returning when we find all our results, it might be faster to take a different approach.  This command is primarily used by `import_ocw_course_sites` and can only be run with a filter.  Therefore, it might be better here to instead iterate the `filter_list` and build the `Prefix` argument here using the courses in the list.  That would ensure that we are only querying S3 for the courses in the filter.  We might want to still check that `filter_list` is set before doing this as I believe `overwrite_ocw_course_content`, another management command for updating individual metadata properties, also uses this function and can still be run un-filtered.  Either way, I think if `filter_list` is set, we should only be querying S3 for the courses in that list.

gumaerc (changes requested):
This is an improvement, but I think we can do better.  If the course in your filter is later in the list alphabetically, the code will still paginate the entire S3 bucket before it hits your course.  Maybe we can take a different approach:

ocw_import/api.py
Wassaf-Shahzad: @gumaerc Yeah, I actually did not want to change the pagination mechanism and was a bit confused about the prefix as I thought different courses will have different  pefixes. Thats why I added this naive solution but your above explanation cleared everything and I have updated the code as per required.

Issue #991, opened 2022-02-10
Prompt user to save work when opening the publish drawer 


pdpinch:
As a course author, I don't want to lose my work when I open the publish drawer and publish the course. 

We have a warning when the course author navigates away from the page without saving, but we don't have one when the course author opens the publish drawer, which can also be destructive. 

### To Do _updated by chris_
**Overall issue:** We don't want people to lose edits **and** we want to warn them when publishing with unsaved frontend edits (dirty forms).

**Steps to address:**
- [x] 1. If somehow "Publish" is initiated with a dirty form, data in the dirty form should not be lost.
- [x] 2. Warn user (prompt for confirmation) when navigating would lose information. E.g., navigating from "Metadata" to "Pages" after editing metadata without saving. 
- [x] 3. Clicking "Publish" with dirty forms (unsaved frontend edits) should warn user somehow. 

    *I lean toward an alert badge inside the publish drawer. I.e., still let people open the drawer. The drawer has useful info like links to ocw / staging*

ChristopherChudzicki:
@dseaton or @pdpinch Interested in your thoughts on below....*tldr: I think opening the publish drawer should NOT be destructive. There are a few other related usability issues we should fix.*

First...I want to make sure I understand the behavior in question. 

Currently, when a user does some actions such as
- edit metadata
- rearrange menus
then clicks "Publish --> Publish", the page content seems to reload and so any unsaved work is lost. 

I think that there are really three issues here:
1. Clicking publish causes page content to re-render, losing unsaved progress.

    **Resolution:** Don't re-render the page content when clicking publish. It's unnecessary. _And it should not happen. It only happens because of a quirk in how one of our components is constructed._
2. Making edits to "metadata", "menus", etc and navigating away without saving does NOT currently warn user. (As far as I can tell, only unsaved changes in the editing drawer warn user).

    **Resolution:** Warn using when navigating would lose unsaved changes.
3. Clicking "Publish" does not warn user that  they have unsaved changes.

    **Resolution:** Warn user "You have unsaved changes" before publishing.


(1) and (2) are very easy to do.

I suspect that (3) is a bit trickier, though I haven't looked into it too much. The difficulty with (3) is that the editing forms in studio all have deliberately local state, which makes it hard for the "Publish" button to be aware of their state. _But fixing (1) and (2) would be a big step... At worst, someone hits "Publish", publishes with unsaved changes, then tries to navigate, and realizes they have unsaved changes, and has to save and re-publish._

[...]

PR #1263, opened 2022-04-20
Cc/navigation prompts

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

This is two of the three tasks in #991 

#### What's this PR do?
This PR uses react-router's `<Prompt />` component to prompt users for confirmation when navigation would cause dirty form changes to be discarded. A few other things are included. See comments and "Background" below.

1. `docker-compose restart watch` to pickup changed frontend dependencies.
2. Check closing the editing drawer:

    - Open any editing drawer, e.g., in "Pages", "Resources", "Videos"...wherever.
	- Closing the drawer (by clicking off the drawer or with "x") should prompt you for convfirmation if the form is dirty (has unsaved edits).
	- If you cancel when prompted, the drawer should not close.
	- Similar behavior if navigating to an external site / reloading page with unsaved changes. (But it's a different warning via `beforeunload` event. We have less control over the warning in this case.) 
	
	Note: The "prompt for confirmation" dialog has changed and is now the browser builtin "confirm" prompt.

3. Check that navigating away from unsaved changes elsewhere in the app prompts for confirmation. For example, on "Metadata", "Menus", or "Collaborators" pages
	- Check internal navigation (to other routes within the single page app)
	- Check external navigation (to routes outside the SPA)
4. With a dirty "Metadata" form (or any other form for which the "Publish" button is still visible), click "Publish". The form changes should NOT be discarded.

	*There will be subsequent work to do something more informative here like alert user "Hey, do you really want to publish with some unsaved changes?". But with this PR, at least the changes are not lost.*
5. Do any other navigation / form editing you feel like checking!

On master we are only prompting users for dirty-form-discard-confirmation in two scenarios:
1. closing the editing drawer with unsaved changes *But not all form edits happen inside that drawer!*
2. when navigating to an external site 

Ie, *not* when discarding dirty metadata / menu forms.

(1) happens via custom logic (the `<ConfirmationModal />` component and associated `useConfirmation` hook) and (2) happens via `beforeunload` browser events. The `beforeunload` browser event is fired "when the window, the document and its resources are about to be unloaded."([mdn](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event)). So... not on navigation within the studio single-page app.

This PR replaces our custom `<ConfirmationModal />` with React Router's `<Prompt />` component. This change is enabled by the work Alice did tying editing to routes. Ie, before the editing drawer was tied to routes, we would not have been able to use <Prompt>, but now we can. So there's no need for a custom <ConfirmationModal />.


static/js/components/util/Prompt.tsx
```diff
@@ -0,0 +1,33 @@
...
+      event.returnValue = message
+      event.preventDefault()
+    }
+    window.addEventListener("beforeunload", listener, { capture: true })
+    const remove = () =>
+      window.removeEventListener("beforeunload", listener, { capture: true })
+    return remove
+  }, [onBeforeUnload, when, message])
+
+  return <ReactRouterPromp {...otherProps} />
```
ChristopherChudzicki, line 32: One consequence of using react-router's <Prompt /> is that confirmation is now requested via `window.confirm`, ie the browser's API. So the confirmation modal is now the browser's, not our customized version.

**Before**:
Screen Shot 2022-04-20 at 2 22 06 PM
**After**
Screen Shot 2022-04-20 at 2 23 07 PM

IMO the old one looks slightly better, but using <Prompt /> is more canonical react router approach and lets use do this more easily without custom code. And we can customize the confirmation dialog used by react-router if we want, but it is not currently a priority.

package.json
```diff
@@ -85,7 +84,7 @@
     "express": "^4.17.1",
     "fetch-mock": "^9.10.6",
     "formik": "^2.2.9",
-    "history": "^5.0.0",
+    "history": "^4.10.1",
```
ChristopherChudzicki, line 87: We use react-router v5 which uses history v4 internally. We should be using history 4, too. https://github.com/remix-run/history/issues/804#issuecomment-899124838

This also solves some typescript issues we've had.

static/js/pages/SitePage.tsx
```diff
@@ -23,36 +24,35 @@ export default function SitePage(props: SitePageProps): JSX.Element | null {
 
   const website = useWebsite()
 
-  if (isLoading) {
-    return <div className="site-page std-page-body container">Loading...</div>
-  }
```
ChristopherChudzicki, line 28: This early return was causing unnecessary component mount/unmounts lower in the component tree. For example:
1. Site is loaded, `<SiteContentListing />` renders
2. click publish, new site api request is sent

    **Before:** this `<div className="site-page std-page-body container">Loading...</div>` would replace the content listing, then a new content listing would be mounted after the API request is finished.

    **Now:** while the API request is in flight, the old content does not unmount (but a loading spinner shows atop it). When the API request finishes, the child component is updated (no unmount/remount, which was causing dirty form data to be lost).

gumaerc (approved):
:+1: LGTM!

PR #1274, opened 2022-04-25
Cc/publish alert

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Finishes #991 

#### What's this PR do?
A few things:
- adds a custom alert prompt instead of window.confirm, see screenshot below 
- control opening the publishing drawer via a search param `?publish=`. Its value is irrelevant. Its key present means "drawer is open".
- Use the `?publish=` search param along with react-router's prompt component to prompt user for confirmation when publishing with a dirty form (*this was the ticket goal ✅*)
- Adds a new frontend integration test helper to live alongside the old. See code comment.

1. Restart docker to pick up new dependencies. *Possibly only need to restart watch: `docker-compose restart watch`*.
2. Try publishing with unsaved changes on the metadata page (or menu page; any page with editable content where publish button is still visible). You should be prompted for confirmation if and only if you have unsaved changes.
3. You should still be prompted when navigating to other routes with a dirty for
4. You should not be prompted in other scenarios.

Screen Shot 2022-04-25 at 10 23 37 AM


static/js/util/IntegrationTestHelper.tsx
```diff
@@ -0,0 +1,156 @@
+import React from "react"
```
ChristopherChudzicki, line 1: This file, and its usage in `Header.test.tsx` is the biggest change in the PR.

Updating the publishing drawer alert broke some tests, so I used that as an opportunity to rewrite IntegrationTestHelper (or rather, create a new version). The biggest differences are:
- IntegrationTestHelper.tsx uses Jest for all mocking; the old file (now called integration_test_helper_old.tsx) used Sinon. Alice did a little work before she left toward replacing sinon mocks with Jest mocks in test files, but it was always a little awkward because the helper class used Sinon.
- IntegrationTestHelper.tsx uses `@testing-library/react`, whereas the old helper used `enzyme`.

See THIS ISSUE I'M GOING TO MAKE for why we should switch to @testing-library/react. The tldr is: enzyme doesn't seem to be actvitely developed (it does not support react 17 or react 18) and @testing-library/react is the officially-endorsed, canonical approach.

ChristopherChudzicki:
@gumaerc As discussed, this probably should have been at least two PRs. Possibly three:

These changes make the modal pretty instead of window.prompt 
- Router.tsx and its test file
- PromptConfirmationModal.tsx

These changes make the location change on publish, which triggers the prompt:
- Header.tsx
- ConfirmDiscardChanges.tsx has some meaningful related changes

Additionally:
- IntegrationTestHelper.tsx is the new test helper
- Header.test.tsx was completely re-written using the new helper

ChristopherChudzicki:
#1275 is tracking this.

Minor clarification:
> some...through Sinon and some running through the React testing library.
We have two inconsistencies right now:
- Enzyme vs React Testing Library (RTL)
- Sinon vs Jest...*We use Jest for test running, assertions, and mocking. Except mocking API calls. We use Sinon for mocking API calls.*

As we switch from Enzyme to RTL, we'll should switch to Jest exclusively for mocking.

PR #1276, opened 2022-04-25
add migration to move filetype to resourcetype

gumaerc:
#### Pre-Flight checklist

- [x] Migrations
- [ ] Testing

Prerequisite for https://github.com/mitodl/ocw-hugo-projects/issues/160

#### What's this PR do?
While https://github.com/mitodl/ocw-hugo-projects/pull/161 has been in review, it came to light that the `ocw-www` starter configuration has become a bit out of sync with some changes we've been making in `ocw-course`.  We were setting up linking / embedding of resources and realized that the filtering on resource type wasn't working properly.  This is because this field is currently called `filetype` in `ocw-www` when it should be called `resourcetype` because that's the field that `ocw-studio` looks at when using its filter logic.  This PR introduces a migration that moves `filetype` to `resourcetype` in resources that are part of `ocw-www` specifically.

 - Spin up `ocw-studio` locally and make sure you have an `ocw-www` site populated with some content, including resources
 - Run `docker-compose run --rm web ./manage.py migrate`
 - Go to Django admin and verify on a resource that's part of `ocw-www` that `filetype` has been migrated to `resourcetype`


websites/migrations/0046_filetype_to_resourcetype.py
```diff
@@ -0,0 +1,44 @@
...
+    migrate_field(apps, False)
+
+
+def migrate_field(apps, forward):
+    Website = apps.get_model("websites", "Website")
+    WebsiteContent = apps.get_model("websites", "WebsiteContent")
+    original_name = "filetype" if forward else "resourcetype"
+    updated_name = "resourcetype" if forward else "filetype"
+    try:
+        ocw_www = Website.objects.get(name="ocw-www")
```
MAbdurrehman12, line 22: Since this will only fetch a website having name `ocw-www`, so I was wondering whether there's a need to update the ocw-www test sites too on RC? 
image

MAbdurrehman12 (approved):
Looks good 👍 
I just have a thought mentioned in another comment.

websites/migrations/0046_filetype_to_resourcetype.py
gumaerc: This a good catch, probably not a big deal but we should migrate all the data if we can.  I changed it to iterate all sites that use the `ocw-www` starter instead and this is ready for another look.

websites/migrations/0046_filetype_to_resourcetype.py
```diff
@@ -0,0 +1,47 @@
...
+        return
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("websites", "0045_blank_publish_fields"),
+    ]
+
+    operations = [migrations.RunPython(migrate_fields_forward, migrate_fields_forward)]
```
MAbdurrehman12, line 47: Shouldn't one be `migrate_fields_backward`?

websites/migrations/0046_filetype_to_resourcetype.py
MAbdurrehman12: It looks like all ocw-www websites are not using the same starter, some of them have different starters with a different name but I agree that it's not a big deal so as far as we have `ocw-www` named starter on production and `ocw-www` website is using that starter, this should be fine. 

However, if you feel that we should migrate all data on RC too then maybe we can use `Like` query (equivalent to `contains` in django ORM), something like this:

`ocw_www_sites = Website.objects.filter(name__contains="ocw-www")` 

and I think it will be worth checking that this matching string: `ocw-www` is not there in any course name. (we don't have such a course name on RC though).

Apart from this, some tests are failing on this PR, I'm not sure why, I tried re-running them but it didn't work.

websites/migrations/0046_filetype_to_resourcetype.py
gumaerc: Wow, yes, I'm not sure how I missed that.

websites/migrations/0046_filetype_to_resourcetype.py
gumaerc: @MAbdurrehman12 Thanks for taking another look.  I fixed the issue with the tests and the backward migration argument.  I think for this case, only addressing sites with literally `ocw-www` as the starter makes sense, and all other starters are out of scope.  This PR is in response to changing the fields in the official `ocw-www` starter as it's defined in `ocw-hugo-projects`, and any other starters that have been modified outside of that context don't matter in this case.

MAbdurrehman12: Sounds good!

MAbdurrehman12 (approved):
I think all is good now! 👍

Issue #1177, opened 2022-03-29
No error message when user's session has expired


pdpinch:
### Steps to Reproduce

Not sure if this will work, but it's happened enough times that I'm confident there is a problem. 

1. Log into ocw-studio and 
2. Delete the session cookie
3. Click a button 

### Expected Behavior

Either: 
- user gets an error message instructing them to log in again
- OR, user is logged in again automatically with social auth 

### Actual Behavior

- the UI fails silently

ChristopherChudzicki:
This is the current redirect prompt, which triggers on all API responses that reject with an authentication error (not other errors, like permission errors).

<img width="1462" alt="Screen Shot 2022-05-05 at 11 45 33 AM" src="https://user-images.githubusercontent.com/9010790/166961676-97e91e16-653e-42ae-a689-39deb520f9ad.png">

ChristopherChudzicki:
@pdpinch @Ferdi This has been released.

PR #1297, opened 2022-05-04
Cc/user store

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Toward https://github.com/mitodl/ocw-studio/issues/1177

#### What's this PR do?
This PR moves user information from `window.SETTINGS` (provided by the backend in a Django template) to a redux store.

Alice once indicated to me that we've moved away from redux a bit in the recent past, so I was a little hesitant to introduce it here. That said,
- the user store is created using [redux toolkit](https://redux-toolkit.js.org/introduction/getting-started), the "official, opinionated, batteries-included toolset for efficient Redux development"
- user info seems like a very reasonable thing to keep in the a redux store (pretty global info)
- the immediate motivation for this is to access (and set) user information in middleware. Specifically...we are using `redux-query` for API calls. By moving the user information to the redux store, it will be easy to create redux middleware that triggers a "please go login again" alert when 401 Unauthenticated errors come from the backend.

You should still be able to log in and out.

### DIFF
--- a/static/js/pages/HomePage.test.tsx
+++ b/static/js/pages/HomePage.test.tsx
@@ -1,25 +1,35 @@
 import React from "react"
-import { shallow } from "enzyme"
+import { screen } from "@testing-library/react"
+import {
+  IntegrationTestHelper,
+  assertInstanceOf,
+  absoluteUrl
+} from "../testing_utils"
 
 import HomePage from "./HomePage"
 
-describe("HomePage", () => {
-  it("shows the Touchstone login button if the user is logged out", () => {
-    SETTINGS.user = null
-    const wrapper = shallow(<HomePage />)
-    const link = wrapper.find("a[href='/login/saml/?idp=default']")
-    expect(link.exists()).toBeTruthy()
-    expect(link.text()).toBe("Login with MIT Touchstone")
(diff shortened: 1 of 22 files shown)


static/js/pages/HomePage.test.tsx
```diff
@@ -1,25 +1,35 @@
...
+    const [result] = helper.render(<HomePage />)
+    const link = result.getByText(LOGIN_TEXT)
+    assertInstanceOf(link, HTMLAnchorElement)
+    expect(link.href).toBe(absoluteUrl("/login/saml/?idp=default"))
   })
 
-  it("should set the title", () => {
-    const wrapper = shallow(<HomePage />)
-    expect(wrapper.find("DocumentTitle").prop("title")).toBe("OCW Studio")
+  it("does NOT show Touchstone Login is NOT visible if user is already logged in", () => {
```
gumaerc, line 23: Since this statement is a double negative, couldn't we say something like "Doesn't show the Touchstone login if a user is already logged in?"

gumaerc (approved):
:+1: LGTM

Just one minor question about a test label, take it or leave it

Issue #1280, opened 2022-04-26
Studio should not silently fail to load markdown bodies


ChristopherChudzicki:
### Steps to Reproduce
1. View https://docs.google.com/spreadsheets/d/1q19lGfqbUrcC0CraBrA9YsY9q_8C3ejeeDFc2JX3XKo/edit?usp=sharing and click on any studio link in a row where `will_error=TRUE`.

All of these links have shortcodes inside shortcodes, e.g,

```
{{% resource_link 268b1b59-8ffd-96ff-8735-e48c21eaf2f9 "Stratospheric chemistry 1: O{{< sub \"X\" >}}, HO{{< sub \"X\" >}}, NO{{< sub \"X\" >}} (PDF)" %}}
{{< image-gallery-item href="f11329216387d91fe50b0e4d5e7856c1_lab1-12.jpg" data-ngdesc="Coarse granular magnetite crystals: Fe3O4. Courtesy of OCW." text="Coarse granular magnetite crystals: Fe{{< sub 3 >}}O{{< sub 4 >}}." >}}
{{< sup "{{< sub \"®\" >}}" >}}
```

None of them display well on Hugo, but the ones marked `will_error=TRUE` also fail to load in Studio. 

### Expected Behavior
Studio should either load the markdown without corrupting it, or at least provide the user with an informative error message that there was a problem loading the markdown.

### Actual Behavior

No visible error to user, just...CKEditor fails to load. (There is an error in the console.)
<img width="1859" alt="Screen Shot 2022-04-26 at 5 00 37 PM" src="https://user-images.githubusercontent.com/9010790/165391532-d2c6bf0f-395d-4b08-870c-114a5502e9a2.png">

### Notes
This is only an issue for legacy data. To my knowledge, "nested shortcodes" (which are invalid to Hugo) cannot be created in Studio.

ChristopherChudzicki:
Deployed in #1362

PR #1315, opened 2022-05-10
Cc/better shortcode parsing

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#1280 

#### What's this PR do?
This PR improves the frontend Markdown --> HTML converter to better handle shortocdes. In particular, this fixes an issue with shortcodes inside shortcodes. 

Shortcodes inside shortcodes will not display well in Hugo. Currently they crash studio. With this PR they will at least not crash studio.

*If you'd prefer to test with real data / have a copy of prod courses locally, this spreadsheet lists a bunch (all) pages affected by the bug being fixed here: https://docs.google.com/spreadsheets/d/1q19lGfqbUrcC0CraBrA9YsY9q_8C3ejeeDFc2JX3XKo/edit?usp=sharing*

1. On master, edit a page in studio and create a resource link.
2. In the admin panel, edit the resource link so it looks like
    ```
    {{% resource_link uuid "superscripts! {{< sup 2 >}} oh no" %}}
    ```
3. View that content in Studio and observe that the editor fails to load.
4. Repeat (3) on this branch. Editor should load fine.

### DIFF
--- a/static/js/lib/ckeditor/plugins/ResourceEmbed.ts
+++ b/static/js/lib/ckeditor/plugins/ResourceEmbed.ts
@@ -14,10 +14,8 @@ import {
   RESOURCE_EMBED,
   RESOURCE_EMBED_COMMAND
 } from "./constants"
-import { Shortcode, makeHtmlString } from "./util"
+import { Shortcode, makeHtmlString, replaceShortcodes } from "./util"
 import { isNotNil } from "../../../util"
-
-export const RESOURCE_SHORTCODE_REGEX = /{{< resource .*? >}}/g
 
 /**
  * Class for defining Markdown conversion rules for ResourceEmbed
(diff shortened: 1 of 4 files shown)


static/js/lib/ckeditor/plugins/ResourceEmbed.ts
```diff
@@ -14,11 +14,9 @@ import {
   RESOURCE_EMBED,
   RESOURCE_EMBED_COMMAND
 } from "./constants"
-import { Shortcode, makeHtmlString } from "./util"
+import { Shortcode, makeHtmlString, replaceShortcodes } from "./util"
 import { isNotNil } from "../../../util"
 
-export const RESOURCE_SHORTCODE_REGEX = /{{< resource .*? >}}/g
```
ChristopherChudzicki, line 20: This is the regexp we were previously using to find `resource_link` text. The text would then get turned into shortcode objects via `Shortcode.fromString(matchingText)`.

One weakness of using regexes for this is that the regexp finds the **first** instance of `%}}` after the opening `{{% resource_link`. Ideally, it would find the **matching / closing** instance of `%}}`. So there's some danger that the match will end **too soon**. The "match ends too soon" scenario is really only a problem for legacy data, or for people who type "%}}" / ">}}% in their shortcode text. The shortcode constructor had some logic intended to catch this. Two issues:

1. It was catching slightly too many things, e.g., it would error on `{{% resource_link uuid "some text {{< sup 2 >}}" %}}` because the content included angle-shortcode delimiters.
2. We do have legacy data that triggered the error

Anyway... switching to a non-regex approach that can properly find "the matching closing delimiter" is not that hard. So that's the main change in this PR: switch to "find matching delimiter" approach.

static/js/lib/ckeditor/plugins/util.test.ts
```diff
@@ -1,4 +1,10 @@
-import { unescapeStringQuotedWith, Shortcode, ShortcodeParam } from "./util"
+import {
+  unescapeStringQuotedWith,
+  Shortcode,
+  ShortcodeParam,
+  replaceShortcodes,
+  findNestedExpressions
+} from "./util"
```
ChristopherChudzicki, line 7: These frontend shortcode parsing utilities are now, IMO, pretty good. But if we need to do anything much more with the shortcodes (e.g., support opening/closing shortcodes in CKEditor) it may be worth using a more formal grammar / parsing approach. The backend is using PyParsing for this. A similar frontend library (that I have less experience with) is https://www.npmjs.com/package/nearley

mbertrand (commented):
Code looks great, the fix works well on the example markdown in the description.  I experimented with different variations and came up with this which breaks the link but at least does not make the markdown editor blank.  Do you think this is worth handling or will it realistically not happen without deliberately causing it in django admin?

```
Bad link: {{% resource_link "a265b519-8b99-48b2-ae89-be4a38db88d5" "file_in_file_folder.png {{% sup2 }} oops " %}} 
```

ChristopherChudzicki:
@mbertrand Ah, good catch. Hugo allows this and you can actually create it in studio, so I'll fix it. Funny that before this PR `{{% resource_link uuid "false %}} closer" %}}` (which you can also create in studio) broke the editor, and now  `{{% resource_link uuid "false {{% opener" %}}` breaks it.

I guess this is a common issue with parsing nested stuff. PyParsing's `nested_expr` helper has an for ignoring some openeners/closers, which defaults to ignoring ones in quoted strings.

ChristopherChudzicki:
Closing this in favor of https://github.com/mitodl/ocw-studio/pull/1347

PR #1331, opened 2022-05-11
Added a github action for production releases.

rhysyngsun:
N/A

#### What's this PR do?
This adds another github action workflow for production, similar to the RC one, that triggers production deploys either manually or when a new commit on the `release` branch passes.

We'll have to test this when we release to production, but review-wise double check that the branch references have been updated correctly.

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -1,0 +1,34 @@
+name: Production Deploy
+
+on:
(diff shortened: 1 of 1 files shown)


mbertrand (approved):
LGTM, I noticed that the RC file has a typo though (`release-candiate.yml`, missing a `d`).

PR #1333, opened 2022-05-12
Update the prod deploy script to point to prod

rhysyngsun:
N/A

#### What's this PR do?
Correct the `production.yml` script to use the production heroku app instead of RC.

We'll test on the next deploy, but give it a good code review.


gumaerc (approved):
:+1:

Issue #1265, opened 2022-04-21
Differentiate sites in /sites


pdpinch:
Site names are not unique in studio, so when there are multiple sites with the same name (typically courses), it is difficult to differentiate them. 

For example, https://ocw-studio.odl.mit.edu/sites/?q=linear+algebra

<img width="1332" alt="image" src="https://user-images.githubusercontent.com/430126/164454678-03be4984-91fa-4822-aa50-f62f8204cc8b.png">

pdpinch:
Although term and year would be the most familiar way for OCW course authors to differentiate sites, those fields are specific to the ocw-course starter.  

Instead, we can display the `short_id` for the site, which all sites will have, no matter the starter. It is also unique across sites.

ChristopherChudzicki:
`name` might be more readable than `short_id`, and it's also unique
```
title                                       ,name                                                        ,short_id
Single Variable Calculus                    ,18-01sc-single-variable-calculus-fall-2010                  ,18.01sc-fall-2010
Multivariable Calculus                      ,18-02sc-multivariable-calculus-fall-2010                    ,18.02sc-fall-2010
Computation Structures                      ,6-004-computation-structures-spring-2017                    ,6.004-spring-2017
The Analytics Edge                          ,15-071-the-analytics-edge-spring-2017                       ,15.071-spring-2017
Religious Architecture and Islamic Cultures ,4-614-religious-architecture-and-islamic-cultures-fall-2002 ,4.614-fall-2002
Classical Mechanics                         ,8-01sc-classical-mechanics-fall-2016                        ,8.01sc-fall-2016
"Unified Engineering I, II, III, & IV"      ,16-01-unified-engineering-i-ii-iii-iv-fall-2005-spring-2006 ,16.01-fall-2005
Differential Equations                      ,18-03sc-differential-equations-fall-2011                    ,18.03sc-fall-2011
Introduction to Probability                 ,res-6-012-introduction-to-probability-spring-2018           ,res.6-012-spring-2018
```

ChristopherChudzicki:
May be better to stick with `short_id`, it's not clear to me that `name` will stick around, and if it does, whether it will continue being unique. (Per discussion in https://github.com/mitodl/hq/discussions/1).

MAbdurrehman12:
> Instead, we can display the short_id for the site

@pdpinch Should we only display `short ID` or both? Something like `Title - short ID`?

e.g:

Only short ID:
[...]

PR #1346, opened 2022-05-16
short ID added with title in sites list

MAbdurrehman12:
#### Pre-Flight checklist

- [ ] Migrations
- [x] Testing
  - [x] Code is tested
- [ ] Settings

Closes https://github.com/mitodl/ocw-studio/issues/1265

#### What's this PR do?
- Adds `short ID` along with title in sites list in `studio-url/sites` with `format: title (short ID)`. Format is still yet to be confirmed 

- Checkout this branch
- Go to `/sites`, verify that short ID is also shown along with title for every site listed.

image

### DIFF
--- a/static/js/pages/SitesDashboard.tsx
+++ b/static/js/pages/SitesDashboard.tsx
@@ -73,7 +73,7 @@ export default function SitesDashboard(): JSX.Element {
         <StudioList>
           {listing.results.map((site: Website) => (
             <StudioListItem
-              title={site.title}
+              title={`${site.title} (${site.short_id})`}
               subtitle={siteDescription(site) ?? ""}
               to={siteDetailUrl.param({ name: site.name }).toString()}
               key={site.uuid}
--- a/static/js/components/AuthenticationAlert.test.tsx
+++ b/static/js/components/AuthenticationAlert.test.tsx
@@ -58,7 +58,9 @@ describe("Prompting for authentication", () => {
 
       setMockWebsiteResponse(authRejectionBody, 403)
 
-      const siteLink = await waitFor(() => result.getByText(website.title))
+      const siteLink = await waitFor(() =>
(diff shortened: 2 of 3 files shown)


static/js/pages/SitesDashboard.tsx
```diff
@@ -73,7 +73,7 @@ export default function SitesDashboard(): JSX.Element {
         <StudioList>
           {listing.results.map((site: Website) => (
             <StudioListItem
-              title={site.title}
+              title={`${site.title} (${site.short_id})`}
               subtitle={siteDescription(site) ?? ""}
```
ChristopherChudzicki, line 77: I think it would be good to visually differentiate the title / short id somehow. An easy way to do this would be to use the subtitle prop on `StudioListItem`:
```suggestion
              title={site.title}
              subtitle={site.short_id}
```

Interesting... I just took a quick look at `siteDescription(site)` which is supposed to display the course number and term, one of the goals of this short_id feature. But `siteDescription(site)` uses outdated metadata properties, so it never does anything. Safe to remove.

Screen Shot 2022-05-16 at 1 09 50 PM

ChristopherChudzicki (changes requested):
This will be super useful... I think would be good to use subtitle for this, though. That seems to have been our original intent.

pdpinch:
I agree with Chris’s comments

static/js/pages/SitesDashboard.tsx
MAbdurrehman12: I agree. Subtitle is certainly more appropriate and a good visual differentiator. I was just a bit unsure whether it's fine to remove the existing metadata or not but thanks for confirming. I'll update the PR.

ChristopherChudzicki (approved):
So excited for this to be live 😀  much nicer to see the course numbers

Issue #1277, opened 2022-04-26
`minimal=true` should be the default for markdown widget


ChristopherChudzicki:
As an engineer, I'd like it to be easy to create course configs that function properly.

#### Acceptance Criteria:
- [ ] `minimal=true` should be the default for the Markdown widget. Only body fields can support non-minimal markdown.
- [ ] update existing configs to use `minimal=false` on markdown body fields (that don't currently specify a value).

ChristopherChudzicki:
Deployed in https://github.com/mitodl/ocw-studio/pull/1362

PR #1350, opened 2022-05-16
make markdown editor minimal by default

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#1277 

#### What's this PR do?
1. Updates the frontend code to use `minimal=true` the default for markdown widget.
2. Syncs the studio configs with https://github.com/mitodl/ocw-hugo-projects/pull/182 _doing this manually was no fun and it seems like maybe we haven't done it recently. I think we've talked about just pulling in the configs for running tests? I will start a discussion_

1. Use the course config from https://github.com/mitodl/ocw-hugo-projects/pull/182
2. View any widget markdown widget for which `minimal` is not specified. A good example is Resource description. Verify that it is minimal (i.e., no tables, no code blocks, etc.)
3. Check that widgets for which minimal=false are still full, e.g., page bodies or resource bodies.

### DIFF
--- a/localdev/configs/ocw-course-site-config.yml
+++ b/localdev/configs/ocw-course-site-config.yml
@@ -6,6 +6,12 @@ collections:
     label: Pages
     name: page
     fields:
+      - label: Draft
+        name: draft
+        required: true
+        widget: boolean
+        help: Enable this setting to prevent publishing in production
+
       - label: Title
         name: title
         widget: string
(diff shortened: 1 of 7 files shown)


ChristopherChudzicki:
I'm going to close this and re-open without https://github.com/mitodl/ocw-studio/pull/1350/commits/7191784b0893154967ae1ee55407c2ad26d9477a ... we can do that separately if we want.

Issue #1354, opened 2022-05-17
Limit site starter options when creating new site


pdpinch:
As a site author, I'd like to like to have the most common site starters appear as the default when I create a new site. 

As a site author, I'd like to have a short list of site starters to choose from, so I don't pick the wrong one by accident. 

#### Approach 

Based on [this discussion](https://github.com/mitodl/hq/discussions/29#discussioncomment-2742880), we can add a status field the `websitestarter` model with (possible) values `default`, `active` and `inactive`. 

- Only show the starters with a status of `active` or `default`. 
- Set the default value to the default starter.

MAbdurrehman12:
Currently, multiple starters can have `default` status but only 1 of them will be shown by default. Do we need to handle this? Maybe we can restrict that only 1 starter can have a default status so if a user tries to set `default` status for another starter, we can show an error?

Ferdi:
> Maybe we can restrict that only 1 starter can have a default status so if a user tries to set default status for another starter, we can show an error?

Good idea.

Tbh, this is probably not the most normalized way to represent the information.  A simpler way would have been a field `active` which can be `true` or `false` 
and have "default_starter" be an environment variable. 

Whatever is easier for you to implement , I would say

pdpinch:
Well, I guess it's easier to leave it as is. 

I don't need a filter in the django admin

PR #1355, opened 2022-05-17
feat: Limit site starter options when creating new site

MAbdurrehman12:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested
- [ ] Settings

Closes https://github.com/mitodl/ocw-studio/issues/1354

#### What's this PR do?
- Adds a new dropdown/choices field named `status` in `websitestarter` model with values `Default`, `Active`, `Inactive`
- Updates the `websitestarter` list API to only include those starters which have status either `Default` or `Active` only.
- Shows `status` in website starters list display in Django admin. 
- While adding a new site,  shows the starter, by default, which has `status='Default'`. 
    - Note: Currently, multiple starters can have default status and only 1 of them (the top one) will be shown. Maybe we can restrict that only 1 starter can have a default status. https://github.com/mitodl/ocw-studio/issues/1354#issuecomment-1129829857

- Checkout this branch
- Apply migrations: `docker-compose run web python manage.py migrate`
- Verify that all your `websitestarters`, now have a status field, having `Active` as the default.
- Verify that this status is being shown in `websitestarters` list-display in Django admin.
- Go to the page of adding a new site:
    - Verify that all website starters are being shown in the starters dropdown.
    - Set a few statuses to `inactive` and now verify that these starters are now not showing in the list.
    - For any starter, set its status to `default`. Now verify that this starter is always selected/shown by default, whenever you add a new site, also verify that this site is created successfully.
- Any other use-case which needs to be tested.

image

image

image

Starters with `default` and `active` status are shown only. Starter with `default` status is shown/selected by default when we open the add site page.

image

image


websites/models.py
```diff
@@ -342,6 +344,12 @@ class WebsiteStarter(TimestampedModel):
         null=False,
         help_text="Human-friendly name of the starter project.",
     )
+    status = models.CharField(
+        max_length=30,
+        choices=WEBSITE_STARTER_STATUS_CHOICES,
```
MAbdurrehman12, line 349: I haven't followed the conventional approach for this choice field (i.e: `zip(choices_list, choices_list)`) as I think keys should have a different list than the values. As values might change, or have spaces, or have upper/lower case stuff etc which keys can not have, so a class named `WebsiteStarterStatus` has been created to store the keys only, and choices util from `django-model-utils` is being used to construct the choices structure, and this choices obj/structure is used in model. 

To me, this approach looks a bit clean and smooth but I'm open to suggestions. 
And if this approach is fine, then I think, for future choices fields, we can follow this approach and for previous choices fields, maybe we can change/improve them with time. (if required)

mbertrand (approved):
👍

PR #1360, opened 2022-05-17
remove two unused dependencies

ChristopherChudzicki:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

None

#### What's this PR do?
Removes two unused dependencies from package.json. I noticed `react-addons-shallow-compare` and found the other via [depcheck](https://www.npmjs.com/package/depcheck). _Most of depcheck's output is false positives, dependencies that are actually used. A reliable tool for finding unused dependencies would be nice... _

Tests should pass, studio should work.


yarn.lock
```diff
@@ -11417,7 +11391,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"react-addons-shallow-compare@npm:^15.6.2, react-addons-shallow-compare@npm:^15.6.3":
+"react-addons-shallow-compare@npm:^15.6.3":
```
ChristopherChudzicki, line 11394: Evidently this is still required (via `react-nestable`) but it doesn't need to be listed in package.json, `react-nestable` can install it itself.

MAbdurrehman12 (approved):
👍

Issue #1343, opened 2022-05-13
No default value for license, despite being set in project starter


pdpinch:
This is particularly confusing because there is a default value set in the ocw-course projects 

### Steps to Reproduce

1. Add a new resource to a course by syncing it from gdrive
2. Edit the resource
3. Save the resource without changing anything

### Expected Behavior

- The resource should save and the drawer should close

### Actual Behavior

- validation error: `License is a required field`

### Related Issues

There's a similar [issue with the default value for `draft`](https://github.com/mitodl/ocw-hugo-projects/issues/173) 

I presume this is a bug on ocw-studio, because it appears that a default value is set in the course starter, at https://github.com/mitodl/ocw-hugo-projects/blob/540afebbe9b18a826ea79ee10817920e910674b1/ocw-course/ocw-studio.yaml#L175

pdpinch:
There's a similar issue with the default setting for `draft` on resources. @MAbdurrehman12 thought it might have something to do with the way resources are created via google drive: https://github.com/mitodl/ocw-hugo-projects/pull/174#issuecomment-1126026752

abeglova:
The default in the theme just affects the ui form in ocw-studio and is why you can't edit and save a resource without setting the license and draft value. License and draft are currently not set when resources are created from google drive and were not set for resources that were imported from legacy ocw. We need to update the code to set it when creating resources through the gdrive process and add a backpopulate command for existing resources

ChristopherChudzicki:
> The default in the theme just affects the ui

And the default is only used for creating new resources, not updating existing resources (e.g., updating resources created by gdrive).

ChristopherChudzicki:
@pdpinch or @abeglova Re

>  License and draft are currently not set when resources are created from google drive and were not set for resources that were imported from legacy ocw. We need to update the code to set it when creating resources through the gdrive process and add a backpopulate command for existing resources

Does the legacy data have license information that should be imported, or should we use the default license (https://creativecommons.org/licenses/by-nc-sa/4.0/) for all legacy content?

pdpinch:
> Does the legacy data have license information that should be imported, or should we use the default license (https://creativecommons.org/licenses/by-nc-sa/4.0/) for all legacy content?

There is no license information in the legacy data. We should assume that all legacy data is https://creativecommons.org/licenses/by-nc-sa/4.0/

ChristopherChudzicki:
This is going out in https://github.com/mitodl/ocw-studio/pull/1401

[...]

PR #1363, opened 2022-05-17
update generate_item_metadata to optionally use config default value

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#1343 

#### What's this PR do?
This PR changes `create_gdrive_resource_content` to set default metadata values based on default field values in the config; these field values would be populated in a form when creating a new item on the frontend.

Create a gdrive resource and verify that the license is set. _No other resource fields currently have a default._

### DIFF
--- a/websites/site_config_api.py
+++ b/websites/site_config_api.py
@@ -116,10 +116,25 @@ class SiteConfig:
                 return config_item
         return None
 
-    def generate_item_metadata(self, name: str, cls: object = None, **kwargs) -> Dict:
-        """Generate a metadata dict with blank keys for the specified item"""
+    def generate_item_metadata(
+        self, name: str, cls: object = None, use_defaults=False, values: Dict = {}
+    ) -> Dict:
+        """Generate a metadata dict with blank keys for the specified item. If
+        use_defaults is True, fill the keys with default values from config.
+        """
         item_dict = {}
         item = self.find_item_by_name(name)
+
+        def get_leaf_field_value(config_field: ConfigField):
+            key = config_field.field["name"]
+            if values.get(key) is not None:
+                return values.get(key)
+            if use_defaults and config_field.field.get("default") is not None:
+                return config_field.field.get("default")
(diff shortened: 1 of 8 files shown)


websites/site_config_api.py
```diff
@@ -116,10 +116,25 @@ def find_item_by_name(self, name: str) -> Optional[ConfigItem]:
                 return config_item
         return None
 
-    def generate_item_metadata(self, name: str, cls: object = None, **kwargs) -> Dict:
-        """Generate a metadata dict with blank keys for the specified item"""
+    def generate_item_metadata(
+        self, name: str, cls: object = None, use_defaults=False, values: Dict = {}
+    ) -> Dict:
```
ChristopherChudzicki, line 121: The only two places where `generate_item_metadata` is used are:
- `create_gdrive_resource_content`, creating gdrive resources
- the `update_content_metadata` management command.

It seems to me that we probably always want `use_defaults=True`, but I am not entirely sure.

What do others think? If so, I could remove the kwarg and just always use the defaults.

websites/site_config_api.py
```diff
@@ -116,10 +116,30 @@ def find_item_by_name(self, name: str) -> Optional[ConfigItem]:
                 return config_item
         return None
 
-    def generate_item_metadata(self, name: str, cls: object = None, **kwargs) -> Dict:
-        """Generate a metadata dict with blank keys for the specified item"""
+    def generate_item_metadata(
+        self,
+        name: str,
+        cls: object = None,
+        use_defaults=False,
```
ChristopherChudzicki, line 123: The only two places where `generate_item_metadata` is used are:
- `create_gdrive_resource_content`, creating gdrive resources
- the `update_content_metadata` management command.

It seems to me that we probably always want `use_defaults=True`, but I am not entirely sure.

What do others think? If so, I could remove the kwarg and just always use the defaults.

PR #560, opened 2021-09-07
add support for linking to resources in the Markdown editor

alicewriteswrongs:
#### Pre-Flight checklist

- [x] Screenshots and design review for any changes that affect layout or styling
  - [x] Desktop screenshots
  - [ ] Mobile width screenshots
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?

closes #541  
closes #542 

#### What's this PR do?

this adds basic support for adding links to resources (not embeds) in the Markdown editor. These are rendered to markdown via the `resource_link` shortcode. In order to support this I added a new CKEditor node type plugin and a new React component to actually render the things when they're added to the editor. I also made some modifications to the resource picker to allow the user to decide whether they want to embed or link to a resource.

#### How should this be manually tested?

Create some resources for one of your courses. Then go to create a new page and check out the markdown editor. If you click 'add resource' you should get a popup as before. The difference is two things:

- if you click on one of the resources, instead of immediately adding it to the editor it will display two buttons along the bottom of the dialog, 'embed' and 'link', and highlight the resource you selected to make it clear it's 'focused'
- if you click those they should do what they say on the tin

I also added a display component for showing the resource links in the editor (see screenshots below).

#### Screenshots (if appropriate)

resource picker with nothing selected:

Screen Shot 2021-09-07 at 5 05 32 PM

with one of the resources selected:

Screen Shot 2021-09-07 at 5 05 37 PM

a variety of things linked and embedded in the editor:

Screen Shot 2021-09-07 at 5 05 46 PM

PR #1364, opened 2022-05-17
allow setting link / embed on minimal markdown editor

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Prerequisite for https://github.com/mitodl/ocw-hugo-projects/issues/185

#### What's this PR do?
In https://github.com/mitodl/ocw-studio/pull/112, we added a "minimal" configuration for CKEditor.  Markdown fields in starter configurations that specify `minimal: true` utilize this config and it hides things like the table controls.  In https://github.com/mitodl/ocw-studio/pull/560, we added the ability to link to resources in the markdown editor.  Currently, you cannot set `minimal: true` and also set the `link` or `embed` properties to allow linking / embedding of pages / resources / whatever inside a markdown editor with `minimal: true`.  This PR changes that behavior so that if both `minimal: true` and `link` or `embed` are set, the markdown editor is initialized with the minimal configuration, but also the link and embed controls specified in the configuration.

 - Clone `ocw-hugo-projects` locally on the `cg/body-description` branch
 - Spin up `ocw-studio` locally on this branch, making sure you are set up to publish to a test organization of your choice
 - If you do not already have a superuser, create one and go to Django admin
 - Inside Django admin, pull up the `ocw-www` and `ocw-course` starters
 - Paste in the contents of the respective starters from your locally cloned `ocw-hugo-projects` on the `cg/body-description` branch
 - If you do not already have test sites created for `ocw-www` and `ocw-course`, create them and add at least one resource to each
 - Open up each resource for editing, and in the description markdown editor you should **not** see the insert table button, but you **should** see the "Add link" button
 - Add a link to a resource or page to your test resource descriptions and save
 - Re-open the resources and ensure that the links are still there
 - Publish your test course to your test org and view the output, make sure you see the `resource_link` written into the markdown for the resource you added the link to in the markdown body

### DIFF
--- a/static/js/components/widgets/MarkdownEditor.tsx
+++ b/static/js/components/widgets/MarkdownEditor.tsx
@@ -92,8 +92,24 @@ export default function MarkdownEditor(props: Props): JSX.Element {
   )
 
   const editorConfig = useMemo(() => {
+    const toolbarItemsFilter = (item: string): boolean => {
+      if (item === ADD_RESOURCE_LINK) {
+        return link.length > 0
+      }
+      if (item === ADD_RESOURCE_EMBED) {
+        return embed.length > 0
+      }
+      return true
+    }
+
     if (minimal) {
-      return MinimalEditorConfig
+      return {
+        ...MinimalEditorConfig,
+        toolbar: {
+          ...MinimalEditorConfig.toolbar,
+          items: MinimalEditorConfig.toolbar.items.filter(toolbarItemsFilter)
+        }
+      }
     } else {
       // this render function is stuck into the editor config
       // our ResourceEmbed plugin can pull the callback out,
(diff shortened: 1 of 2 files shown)


MAbdurrehman12:
Sorry, I assigned this to myself for review but I got stuck in some other things and hence unassigned myself.

mbertrand (commented):
For the ocw-www site, if the link is to a page instead of a resource, the link is not rendered by ocw-hugo-themes.  Is this a known issue that will be addressed in a related PR?  On a ocw-course site, the page link works fine.

Screen Shot 2022-05-18 at 10 24 36 AM

gumaerc:
@mbertrand I tried replicating what you were describing and was unable to do so.  Linking to a page from the description of a resource in `ocw-www` automatically fills the label part of the link with the page title for me:

image

Maybe your page you were testing with didn't have a title set?  Something worth thinking about here that's outside of the context of this PR is that the `ocw-www` theme doesn't currently have a layout for displaying standalone resource pages.  When you visit the rendered page, you get basically a blank page with the links:

image

So, it's worth thinking about this I think and we should open another issue to port the single resource view over to the `www` theme.  I think it's outside the scope of these changes though, who's purpose is to simply enable linking of pages and resources in the description.  If we don't want to enable it in `ocw-www`, I could turn that off in my upcoming `ocw-hugo-projects` PR.

mbertrand:
@gumaerc the issue turned out to be that I had inadvertently set `draft=true` for the page. After setting it to false the link/title showed up.

mbertrand (approved):
👍

Issue #1366, opened 2022-05-19
Show error when site url is not unique


ChristopherChudzicki:
As an author, I want to understand why more course failed to publish.

#### Acceptance Criteria:
- [ ] When the publishing API fails because of a form error (e.g., "site url is already in use") we should show that to the author. 

#### Out of Scope
Failure errors related to the pipelines, not the publishing API calls.

ChristopherChudzicki:
I started this yesterday as Matt was finishing other stuff related to #1316 ...

Before / After
<img width="300" alt="Screen Shot 2022-05-19 at 1 48 04 PM" src="https://user-images.githubusercontent.com/9010790/169367878-82d0b620-a1bd-447b-b461-1198ce03b0b7.png"> <img width="300" alt="Screen Shot 2022-05-19 at 1 46 54 PM" src="https://user-images.githubusercontent.com/9010790/169367904-d87510e0-533e-4f38-9606-8e396b30d722.png">

PR #1367, opened 2022-05-19
Show publishing api errors in publishing drawer

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#1366 

#### What's this PR do?
This PR adds field-specific error messages to the publishing drawer, building on #1316 

1. Create a new site
2. Try to publish the site using a url that is already in use
3. You should see a specific error near the error field.
4. If you want to check that generic / unexpected errors only show the "overall" error message... I am not sure the best way to trigger that. You could change the publishing URL in `websitePublishAction` code from `"preview"` to something invalid.

Before / After
Screen Shot 2022-05-19 at 1 48 04 PM Screen Shot 2022-05-19 at 1 46 54 PM

### DIFF
--- a/static/js/components/PublishDrawer.test.tsx
+++ b/static/js/components/PublishDrawer.test.tsx
@@ -252,50 +252,6 @@ describe("PublishDrawer", () => {
           })
         })
 
-        it("renders an error message if the publish didn't work", async () => {
-          const actionStub = helper.mockPostRequest(
-            siteApiActionUrl
-              .param({
-                name:   website.name,
-                action: api
-              })
-              .toString(),
-            {},
-            500
-          )
-          const { wrapper } = await render()
-          await act(async () => {
-            const onChange = wrapper.find(`#publish-${action}`).prop("onChange")
-            // @ts-expect-error
-            onChange({ target: { checked: true } })
-          })
-          wrapper.update()
-          await act(async () => {
-            wrapper
-              .find("PublishForm")
-              .find(".btn-publish")
-              .simulate("submit")
-          })
-          wrapper.update()
-          expect(wrapper.find(".publish-option-description").text()).toContain(
-            "There was an error publishing the site."
-          )
-          sinon.assert.calledOnceWithExactly(
-            actionStub,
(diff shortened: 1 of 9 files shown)


static/js/components/PublishDrawer.tsx
```diff
@@ -1,151 +1,160 @@
...
+      envLabel:              "Production",
+      date:                  website.publish_date,
+      status:                website.live_publish_status,
+      hasUnpublishedChanges: website.has_unpublished_live
+    }
+  }
+  throw new Error("Invalid PublishingEnv")
+}
+
+const PublishingOption: React.FC<PublishingOptionProps> = props => {
```
ChristopherChudzicki, line 61: This was previously a function defined in `PublishingDrawer`, `renderOption`, now a real component.

static/js/components/PublishDrawer.test.tsx
```diff
@@ -1,27 +1,36 @@
...
 import sinon, { SinonStub } from "sinon"
 import { act } from "react-dom/test-utils"
 import { isEmpty } from "ramda"
 
 import { siteApiActionUrl, siteApiDetailUrl } from "../lib/urls"
-import { shouldIf } from "../test_util"
+import { assertInstanceOf, shouldIf } from "../test_util"
 import { makeWebsiteDetail } from "../util/factories/websites"
-import IntegrationTestHelper, {
+import IntegrationTestHelperOld, {
```
mbertrand, line 10: Is there something that `IntegrationTestHelperOld` can test that the new one can't?

ChristopherChudzicki: No. But the old one uses Enzyme, and the new one uses `@testing-library/react` ("RTL") (and other packges in the `@testing-library` family). RTL has a somewhat different API than Enzyme, so converting tests will be a bit of effort. 

I've been trying to write new tests with RTL + the new helper, and have been converting old tests when refactoring significantly. (Which is natural, since the Enzyme tests tend to be a bit brittle, so refactoring tends to break them).

mbertrand (approved):
👍

Issue #393, opened 2021-07-16
Add term and year to course titles in collection UI to disambiguate courses with the same title


noisecapella:
Related to #391

We should disambiguate courses so we don't see duplicate listings, perhaps by adding a date to the title:

![Screenshot from 2021-07-16 15-56-09](https://user-images.githubusercontent.com/863262/126004008-f7eafea7-c72b-48ee-8432-1a19e7e51771.png)

pdpinch:
The OCW team is used to disambiguating courses using the term and the year.

ChristopherChudzicki:
Do we want to use short ids here instead of course+year, like in https://github.com/mitodl/ocw-studio/issues/1265 ?

pdpinch:
Good idea. Will it fit?

ChristopherChudzicki:
I think so; we'll see. @pdpinch did you mean to move this out of to-do?

pdpinch:
I did, just because I want to focus on publishing blockers today (Friday). 

It’s OK to work on this, it’s just low priority.

MAbdurrehman12:
@ChristopherChudzicki Are you working on this?

ChristopherChudzicki:
I am not. Thanks for checking... I'll unassign myself from this and other stuff for the moment.

ChristopherChudzicki:
I've been reviewing @MAbdurrehman12 PR for this (https://github.com/mitodl/ocw-studio/pull/1375). This turns out to be kind of hard to do well in OCW right now. Right now, it's feasible to get the short id into the "select" dropdown (since that's populated by OCW studio), but rather hard to get them in the sortable list:

<img width="1097" alt="Screenshot 2023-03-03 at 10 55 53 AM" src="https://user-images.githubusercontent.com/9010790/222772210-0dda2ac7-e8d5-4e27-a131-7d22e365b7c9.png">

The sortable area is populated by course list resource metadata. Fundamentally, I think our course lists aren't storing the right data.

WebsiteContent course list metadata look like this:
```yaml
content_type: course-lists
courses:
- id: courses/11-165j-urban-energy-systems-and-policy-fall-2021
  title: Urban Energy Systems and Policy
- id: courses/4-401-environmental-technologies-in-buildings-fall-2018
  title: Environmental Technologies in Buildings
# ...
```

- `id` is really `website.url_path`. This is used by Hugo to fetch website data from the static `data.json` files. (E.g., for making course cards in course lists).
[...]

PR #1375, opened 2022-05-25
courses short_id shown in course list

MAbdurrehman12:
#### Pre-Flight checklist

- [ ] Migrations
- [x] Testing
  - [x] Code is tested
- [ ] Settings

Closes https://github.com/mitodl/ocw-studio/issues/393

#### What's this PR do?
- Adds `short_id` in brackets next to `title` in the courses dropdown while adding/editing course lists so the user doesn't see duplicate listings. 

- Checkout this branch.
- In `/sites`, go to `ocw-www`
- Add a new course list or edit an existing one.
- Open the courses dropdown and verify that the format is `title (short_id)`
- Select course(s).
- Save the course list and verify that the course list is successfully saved.

#### Screenshots 
image

image

#### Related Issue:
https://github.com/mitodl/ocw-studio/issues/1265

### DIFF
--- a/static/js/components/widgets/WebsiteCollectionField.tsx
+++ b/static/js/components/widgets/WebsiteCollectionField.tsx
@@ -36,22 +36,29 @@ export default function WebsiteCollectionField(props: Props): JSX.Element {
     [onChange, websiteMap, name]
   )
 
+  const optionsWithoutShortID = useWebsiteSelectOptions(
+    "title",
+    "",
+    "url_path",
+    true
+  ).options
+
+  useEffect(() => {
+    setWebsiteMap(cur => {
+      const newMap = new Map(cur)
+      optionsWithoutShortID.forEach(({ value, label }) => {
+        newMap.set(value, label)
+      })
+      return newMap
+    })
+  }, [optionsWithoutShortID, setWebsiteMap])
+
   const { options, loadOptions } = useWebsiteSelectOptions(
     "title",
     "short_id",
     "url_path",
     true
   )
-
-  useEffect(() => {
-    setWebsiteMap(cur => {
-      const newMap = new Map(cur)
-      options.forEach(({ value, label }) => {
-        newMap.set(value, label)
-      })
-      return newMap
-    })
-  }, [options, setWebsiteMap])
 
   const isOptionDisabled = useCallback(
     (option: Option) => {
--- a/static/js/components/widgets/WebsiteCollectionField.tsx
+++ b/static/js/components/widgets/WebsiteCollectionField.tsx
@@ -37,6 +37,7 @@ export default function WebsiteCollectionField(props: Props): JSX.Element {
   )
 
   const { options, loadOptions } = useWebsiteSelectOptions(
+    "title",
     "short_id",
     "url_path",
     true
--- a/static/js/components/widgets/WebsiteCollectionField.tsx
+++ b/static/js/components/widgets/WebsiteCollectionField.tsx
@@ -36,7 +36,11 @@ export default function WebsiteCollectionField(props: Props): JSX.Element {
     [onChange, websiteMap, name]
   )
 
-  const { options, loadOptions } = useWebsiteSelectOptions("url_path", true)
+  const { options, loadOptions } = useWebsiteSelectOptions(
+    "short_id",
+    "url_path",
+    true
+  )
 
   useEffect(() => {
     setWebsiteMap(cur => {
--- a/static/js/components/widgets/RelationField.test.tsx
+++ b/static/js/components/widgets/RelationField.test.tsx
@@ -103,7 +103,7 @@ describe("RelationField", () => {
 
     websites = makeWebsites()
     useWebsiteSelectOptions.mockReturnValue({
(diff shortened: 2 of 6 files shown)


MAbdurrehman12:
> I assumed we would do this the same way we did https://github.com/mitodl/ocw-studio/pull/1346 -- displaying the title and the short_id

@pdpinch  
We are using `react-select` component here and currently, I don't think it supports anything like `subtitle`, it only displays an option value. So in order to show something similar to `#1346` (i.e: both, title and short_id), we will have to customize the `default option label`, after which we can have something like this (or any other design since we are customizing it):

image

Since this might not be a small change and this issue is also of lower priority so I think we can put this on hold for a bit and focus on something of high priority.

Whenever you want me to work on this, let me know.

pdpinch:
Thanks. Let’s put this on hold for now.

pdpinch:
Instead of display using different styling for each string, can we do something like `title (short_id)` ?

MAbdurrehman12:
> Instead of display using different styling for each string, can we do something like `title (short_id)` ?

Sure

fakhar-ud-din (approved):
LGTM 👍

ChristopherChudzicki (commented):
One thing I am noticing is that the shortId is ending up in OCW github content. See for example, https://github.com/cchudzicki-ocw-testing/ocw-www/commit/1e7a91d909f0a1ec89e773bf6fa140a6d755c7af#diff-07e9bf991c2562b4777c372cccbf3af338b14cbb8b14d62cd845f8d42bbfea5cR21:

```yaml
content_type: course-lists
courses:
- id: courses/9-40-introduction-to-neural-computation-spring-2018
  title: Introduction to Neural Computation (9.40-spring-2018)
description: "These energy courses\_on OCW are not officially part of the Energy Minor\
  \ program, but may be of interest."
```

I think it would be better if the shortId did NOT end up in the github content.

It turns out this doesn't currently matter for the published site. The only place we ever use `website-collection` widget is Course Lists on ocw-www, and the hugo templates for course lists don't use the `courses[index].title` property.

I think it would be best if we just dropped the `.title` property from the values stored by WebsiteCollection—we get the title from the course id. The only reason to store the title is if we wanted to be able to use the title in Hugo without making a `getRemote` call to fetch the course data.

ChristopherChudzicki (changes requested):
@MAbdurrehman12 Could you take a look at ensuring we don't save the short_id as part of the website title in git? I understand Hugo doesn't currently use the `.title` property on course list items, but I think it would be better not to include it.

One idea would be to change `useWebsiteSelectOptions` so that the options it returns are more like `{ value: Website, label: <customized label here> }`. `value` doesn't necessarily need to be the whole website object, could just be `{ id, title }`.  *Edit: I'm not sure this idea would work well with the Select component library; it seems to expect string values =/*

MAbdurrehman12:
@ChristopherChudzicki @pdpinch FYI: I have also discussed this PR and issue with @HussainTaj-arbisoft so incase it is not wrapped up/completed by Monday, Hussain can take over it.

static/js/components/widgets/WebsiteCollectionField.tsx
```diff
@@ -36,17 +36,29 @@ export default function WebsiteCollectionField(props: Props): JSX.Element {
     [onChange, websiteMap, name]
   )
 
-  const { options, loadOptions } = useWebsiteSelectOptions("url_path", true)
+  const optionsWithoutShortID = useWebsiteSelectOptions(
+    "title",
+    "",
+    "url_path",
+    true
+  ).options
```
ChristopherChudzicki, line 44: Slight issue here: The sortable select is a searchable field:
- Initially, 10 websites are fetched
- if you type, new websites are fetched to match your search filter.

`options` below will be updated when you type. But `optionsWithoutShortID` is never updated.

So I don't think this approach works. 😦 Searching for a course and selecting it, and you'll see its title is not used.

ChristopherChudzicki (changes requested):
@MAbdurrehman12 I don't think this approach quite works—there's an issue for sorting courses.

I'm going to write a longer post on the issue https://github.com/mitodl/ocw-studio/issues/393 about why this is hard and what changes I think we'd need to make in order for this to work well.

static/js/components/widgets/WebsiteCollectionField.tsx
ChristopherChudzicki: `WebsiteCollectionField` maintains a map `websiteMap` of id -> title. Let's call it `intialWebsitesMap`. One idea for addressing this would be to make `useWebsiteSelectOptions` return its `websiteMap` of the loaded objects. 

That's the only idea I can think of without changing the data as discussed in https://github.com/mitodl/ocw-studio/issues/393

HussainTaj-arbisoft:
I read this PR and its discussion, I could not think of a way to make it work. So I created an alternative solution for this issue. It has the same effect as intended by this PR.

@ChristopherChudzicki, please look at https://github.com/mitodl/ocw-studio/pull/1715 and let me know your thoughts about that approach.

If the new PR is accepted, I'll close this one. Putting this on hold for now.

HussainTaj-arbisoft:
Closing this PR as it has been superseded.

PR #1383, opened 2022-05-27
Always set publish fields in publish_website function

### LINKED ISSUE #1372: Course is published, but studio says there are unpublished changes
### Steps to Reproduce

1. Publish the course `res-6-013-ai-101-fall-2021` to staging on production

### Expected Behavior

- When publishing has succeeded, the text "You have unpublished changes." should be removed. 

### Actual Behavior

- Even though the course is published, the UI still says "You have unpublished changes."

### Screenshot or Screencast

image

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #1372

#### What's this PR do?
Always set the `publish_status_*` and `has_unpublished_*` fields when running the `publish_website` function.  

Tests should pass.
Run this in a shell for a site with recent changes, then verify that `has_unpublished_draft` is `False`:
```python
from content_sync.api import *
publish_website(<website.name>, "draft", prepublish=False, trigger_pipeline=False)
```

### DIFF
--- a/content_sync/api.py
+++ b/content_sync/api.py
@@ -141,18 +141,13 @@ def publish_website(  # pylint: disable=too-many-arguments
             }
         else:
             update_kwargs = {}
-        if (
-            getattr(website, f"{version}_publish_status") != PUBLISH_STATUS_NOT_STARTED
-            or getattr(website, f"{version}_publish_status_updated_on") is None
-        ):
-            # Need to update additional fields
-            update_kwargs = {
-                f"{version}_publish_status": PUBLISH_STATUS_NOT_STARTED,
-                f"{version}_publish_status_updated_on": now_in_utc(),
-                f"{version}_last_published_by": None,
-                f"has_unpublished_{version}": False,
-                **update_kwargs,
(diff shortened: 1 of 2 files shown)


pdpinch:
I heard about another course with a similar problem, https://ocw-studio.odl.mit.edu/sites/6-801-machine-vision-fall-2020/type/video_gallery/?publish=

Do you think this will fix it as well?

mbertrand:
> Do you think this will fix it as well?
I believe so

gumaerc (approved):
:+1: LGTM

A minor nitpick about the PR itself though, it would be good to have a bit of a description as to why the publish fields were only being updated if the publish status was `PUBLISH_STATUS_NOT_STARTED` or if the updated on date was `None`.  Is there a reason this was the case?

mbertrand:
If a user clicks on the publish button, the view endpoint would also update those fields, but that won't necessarily be the case if a mass publish is done.  So now worst case scenario now should be that they just get set to the same values twice.

PR #1389, opened 2022-06-02
Added permission hook for conditional rendering 

### LINKED ISSUE #1292: Users see "Add Site" button even if they lack permission
### Steps to Reproduce
1. Log in as a user (not a superuser) that does not have permission to add a website.
2. Visit https://ocw-studio-rc.odl.mit.edu/sites/

### Expected Behavior
Add site button is not available

### Actual Behavior
Add site button is available, and 403s if you attempt to add a site

### Other comments
The frontend current has no knowledge of what permissions a user has/

Wassaf-Shahzad:
#### Pre-Flight checklist

- [ ] Migrations
- [x] Testing
  - [ ] Code is tested
- [ ] Settings

Closes #1292 

#### What's this PR do?
- Adds a Permission hook.
- Adds conditional rendering based on permission hook.

- Login as a super user.
- Verify Add Sites button is visible and active.
- Login as a standard user.
- Verify Add Site button does not render.

Screenshot 2022-06-02 at 12 28 12 PM

### DIFF
--- a/static/js/constants.ts
+++ b/static/js/constants.ts
@@ -15,9 +15,9 @@ export const ROLE_OWNER = "owner"
 
 export const ROLE_LABELS = {
   [ROLE_GLOBAL]: "Administrator",
-  [ROLE_ADMIN]:  "Administrator",
+  [ROLE_ADMIN]: "Administrator",
   [ROLE_EDITOR]: "Editor",
-  [ROLE_OWNER]:  "Owner"
+  [ROLE_OWNER]: "Owner"
 }
 
 export const EDITABLE_ROLES = [ROLE_ADMIN, ROLE_EDITOR]
@@ -208,4 +208,8 @@ export enum WebsiteStarterStatus {
 export enum PublishingEnv {
   Staging = "staging",
   Production = "production"
+}
+
+export enum Permissions {
+  canCreateSites = "canCreateSites"
 }
--- a/static/js/constants.ts
+++ b/static/js/constants.ts
@@ -211,5 +211,5 @@ export enum PublishingEnv {
 }
 
 export enum Permissions {
-  canCreateSites = "canCreateSites"
+  canAddWebsite = "canAddWebsite"
 }
--- a/static/js/constants.ts
+++ b/static/js/constants.ts
@@ -15,9 +15,9 @@ export const ROLE_OWNER = "owner"
 
 export const ROLE_LABELS = {
   [ROLE_GLOBAL]: "Administrator",
-  [ROLE_ADMIN]: "Administrator",
+  [ROLE_ADMIN]:  "Administrator",
   [ROLE_EDITOR]: "Editor",
-  [ROLE_OWNER]: "Owner"
+  [ROLE_OWNER]:  "Owner"
 }
 
 export const EDITABLE_ROLES = [ROLE_ADMIN, ROLE_EDITOR]
--- a/static/js/constants.ts
+++ b/static/js/constants.ts
@@ -210,6 +210,6 @@ export enum PublishingEnv {
   Production = "production"
 }
 
-export enum Permissions {
-  canAddWebsite = "canAddWebsite"
+export enum Permission {
+  CanAddWebsite = "canAddWebsite"
 }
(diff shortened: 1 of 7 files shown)


static/js/constants.ts
```diff
@@ -207,3 +207,7 @@ export enum WebsiteStarterStatus {
   Active = "active",
   Inactive = "inactive"
 }
+
+export enum Permissions {
+  canCreateSites = "canCreateSites"
```
MAbdurrehman12, line 212:
```suggestion
  canAddWebsite = "canAddWebsite"
```
I think `canAddWebsite` instead of `canCreateSites` would be more suitable and similar to the django permission.
or maybe we can make permissions something like this: 
(just a thought)
```
permissions = {
    website: {
        view: "value", 
        add: "value", 
        change: "value", 
        delete: "value" 
    }
}
```

MAbdurrehman12, line 212:
```suggestion
  canAddWebsite = "canAddWebsite"
```
I think `canAddWebsite` instead of `canCreateSites` would be more suitable and similar to the django permission.
or maybe we can make permissions something like this: 
(just a thought)
```
permissions = {
    website: {
        view: "value", 
        add: "value", 
        change: "value", 
        delete: "value" 
    }
}
```

static/js/hooks/permissions.ts
```diff
@@ -0,0 +1,15 @@
+import { useAppSelector } from "../hooks/redux"
+import { Permissions } from "../constants"
+
+// The main idea of this hook is to be able to define multiple validators which
+// grant access based on user permissions
+export function usePermission(permission: Permissions): boolean {
+  const { user } = useAppSelector(state => state.user)
+  const validators = { canCreateSites: _canCreateSites }
```
MAbdurrehman12, line 8:
```suggestion
  const validators = { [Permissions.canCreateSites]: _canCreateSites }
```

MAbdurrehman12 (changes requested):
Looks good!

Just a few things.
It also seems like that the permission to create site is only granted to `superuser`. 
So if I create a staff user with permission to add a website, I'm able to add website via django admin but I'm unable to add website via React side. 

So any user having permission to add a website but not being the `superuser` will not be able to add a website. Maybe we should handle the permissions the way are handled in django side? 

ps: I'm also not sure about the priority of this task so not sure if all this effort will be worth it.
Maybe @pdpinch has some thoughts on this?

static/js/constants.ts
Wassaf-Shahzad: Good Catch, I have updated the code as required and as for your thought for creating a permission object. it depends on the scope.

gumaerc:
I tried this out and since I don't have touchstone set up locally, I had to log in as a staff user and then revoke their privileges after the fact.  I'm not sure if this is because of the way I created the user, but viewing any course gives me a 404 as the normal user for some reason.  Are there additional permissions I need to manually grant this user to see the UI?

chrome_2022-06-08_17-46-53

Wassaf-Shahzad:
> I tried this out and since I don't have touchstone set up locally, I had to log in as a staff user and then revoke their privileges after the fact. I'm not sure if this is because of the way I created the user, but viewing any course gives me a 404 as the normal user for some reason. Are there additional permissions I need to manually grant this user to see the UI?
> 
> chrome_2022-06-08_17-46-53

Does the user have,
- View websites
- Create and update websites permission.

static/js/hooks/permissions.ts
```diff
@@ -0,0 +1,15 @@
...
+import { Permissions } from "../constants"
+
+// The main idea of this hook is to be able to define multiple validators which
+// grant access based on user permissions
+export function usePermission(permission: Permissions): boolean {
+  const { user } = useAppSelector(state => state.user)
+  const validators = { canAddWebsite: _canAddWebsite }
+
+  function _canAddWebsite(): boolean {
+    return user?.canAddWebsite || false
```
ChristopherChudzicki, line 11:
```suggestion
    return user?.canAddWebsite ?? false
```

ChristopherChudzicki (changes requested):
Looks good, just needs to resolve the merge conflicts 👍

gumaerc (approved):
👍 

I was able to test this after modifying a user and logging in as them in admin as described, and the permissions work properly.

ChristopherChudzicki (approved):

PR #1392, opened 2022-06-03
Option to delete existing pipelines in management commands

### LINKED ISSUE #1391: Provide option to delete all individual site pipelines in backpopulate_pipelines command
As a developer, I'd like a way to remove all the existing concourse pipelines (specifically on RC) to make sure there are no obsolete old ones, unaffiliated with any existing RC studio sites, that might be lingering around causing trouble - such as the still unexplained deletion of files from the RC draft bucket.

#### Designs and Mockups
Provide a `--delete` flag that when set to true, will make an API call to get all individual draft/live pipelines on concourse and delete every one, before inserting new draft/live pipelines for each studio site.

#### Acceptance Criteria:

- [ ] `--delete` optional flag for `backpopulate_pipelines` that destroys all draft/live site pipelines.

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #1391 

#### What's this PR do?
- Adds an optional parameter to all pipeline commands to delete existing pipelines first
- Modifies the concourse api, pipeline classes to support the above.

- Set up and run concourse locally as described in the README
- Run the following with and without the new `--delete` option, they all should run without errors.
`backpopulate_pipelines`
`upsert_site_removal_pipeline`
`upsert_theme_assets_pipeline`
`upsert_mass_build_pipeline`

It might be hard to notice if the pipelines are actually deleted before getting recreated, so if you want you can add a `return` line within the `if delete:` block of each command.

There are lots of old orphaned site pipelines on RC.  One of them might be causing the draft S3 publish content to get deleted, not sure, but it would be good to remove these pipelines anyway.

### DIFF
--- a/content_sync/management/commands/backpopulate_pipelines.py
+++ b/content_sync/management/commands/backpopulate_pipelines.py
@@ -60,8 +60,8 @@ class Command(BaseCommand):
         )
         parser.add_argument(
             "-d",
-            "--delete",
-            dest="delete",
+            "--delete-all",
+            dest="delete_all",
             action="store_true",
             help="Delete all existing site pipelines first",
         )
@@ -81,9 +81,9 @@ class Command(BaseCommand):
         create_backend = options["create_backend"]
         is_verbose = options["verbosity"] > 1
         unpause = options["unpause"]
-        delete = options["delete"]
+        delete_all = options["delete_all"]
 
-        if delete:
+        if delete_all:
             self.stdout.write("Deleting all existing site pipelines first")
             api = get_pipeline_api()
             if api:
--- a/content_sync/management/commands/backpopulate_pipelines.py
+++ b/content_sync/management/commands/backpopulate_pipelines.py
@@ -1,9 +1,11 @@
 """ Backpopulate website pipelines"""
+from content_sync.constants import VERSION_LIVE, VERSION_DRAFT
 from django.conf import settings
 from django.core.management import BaseCommand, CommandError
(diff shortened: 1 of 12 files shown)


content_sync/management/commands/backpopulate_pipelines.py
```diff
@@ -56,6 +58,13 @@ def add_arguments(self, parser):
             action="store_true",
             help="Unpause the pipelines after creating/updating them",
         )
+        parser.add_argument(
+            "-d",
+            "--delete",
+            dest="delete",
+            action="store_true",
+            help="Delete all existing site pipelines first",
```
ChristopherChudzicki, line 66: Could we name this `--delete-all`?

Originally, I was surprised when `backpopulate_pipelines --filter "es-255-physics-of-rock-climbing-spring-2006 --delete-all" deleted all my other pipelines, ie, was unaffected by filter. But that behavior makes sense...otherwise the delete flag wouldn't really do anything (right?).

This is my requested changed, unless `--delete-all` would be inconsistent with our naming for similar behavior elsewhere.

ChristopherChudzicki (changes requested):
👍 Looks good, one small request around naming of the arg

content_sync/management/commands/backpopulate_pipelines.py
mbertrand: Sure, will do

ChristopherChudzicki (approved):

PR #1402, opened 2022-06-10
Add management command to assist in populating default license values

ChristopherChudzicki:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

#1343 

#### What's this PR do?
This PR adds a management command `set_content_metadata_to_default` that updates content metadata to the default value specified in the config, only affecting content that has null or empty value for that metadata field. 

The present use-case is to set resources to have the default license (CC-BY-SA). Currently on prod:
```
license                                            ,count  ,text_id

NULL                                               ,110660 ,70518f87-dfd5-c723-ee3c-d5d912001b51
""                                                 ,   296 ,8f80e269-30e5-4db7-86d9-bd10e17b2136
https://creativecommons.org/licenses/by-nc-sa/4.0/ ,   270 ,ac106d0a-b0e2-6e50-51d6-b47de9445f3d
https://creativecommons.org/licenses/by-nc/4.0/    ,    20 ,43dd1e03-1b2e-4122-67e4-4b5e6d9d7a1f
https://creativecommons.org/licenses/by/4.0/       ,    11 ,bca02d3a-7d36-b448-fcde-27950d22f93b
https://creativecommons.org/publicdomain/zero/1.0/ ,     8 ,b18f7e34-a663-4eb4-b228-0d433102f5ad
```

1. Make sure your local `ocw-course` starter has a default value for resource licenses. (It should)
2. Find example resources with
    - A) null license
    - B) empty string license
    - C) license set to some value besides the default (cc-by-sa)

    *Create the data in django admin, or if you have a local copy of prod data, the table above can help you find examples.*
4. Run `docker compose run --rm web ./manage.py set_content_metadata_to_default --start ocw-course --type resource --field license`. *Optionally filter by websites, if you like.*
5. Check that the license field was updated for (A) and (B), but not (C).

### DIFF
--- a/main/management/commands/filter.py
+++ b/main/management/commands/filter.py
@@ -2,6 +2,9 @@
 import json
 
 from django.core.management import BaseCommand
+from django.db.models import Q
+
+from websites.models import WebsiteContentQuerySet
 
 
 class WebsiteFilterCommand(BaseCommand):
@@ -36,3 +39,14 @@ class WebsiteFilterCommand(BaseCommand):
             ]
         if self.filter_list and options["verbosity"] > 1:
             self.stdout.write(f"Filtering by website: {self.filter_list}")
+
+    def filter_website_contents(
+        self, website_contents: WebsiteContentQuerySet
+    ) -> WebsiteContentQuerySet:
+        """Filter website_contents based on CLI arguments."""
+        if not self.filter_list:
+            return website_contents
+        return website_contents.filter(
+            Q(website__name__in=self.filter_list)
+            | Q(website__short_id__in=self.filter_list)
+        )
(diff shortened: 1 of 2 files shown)


mbertrand (commented):
Works great.  The new command seems somewhat similar to the pre-existing `update_content_metadata`, do you think it might be worth combining the functionality into one command, or is the use case different enough to keep them separate?

ChristopherChudzicki:
> The new command seems somewhat similar to the pre-existing update_content_metadata, do you think it might be worth combining the functionality into one command, or is the use case different enough to keep them separate?

I thought about putting it there, and I'm still waffling a bit. The reason I left `update_content_metadata` as-is is that `update_content_metadata` currently never alters existing metadata values. At most, it adds new keys. The `set_defaults...`  command **does** alter existing values (only from `None` to the default). I wasn't sure that this was always appropriate for `update_content_metadata`, hence I made a new command. It could be part of `update_content_metadata` with a flag, but then it feels less combined anyway.

Anyway... I am waffling on this, but going to merge and leave as-is for now.

PR #1405, opened 2022-06-15
Added a check for site status

### LINKED ISSUE #1404: unpublished site still listed as published on /sites
### Steps to Reproduce

1. Visit /sites and choose a published course to unpublish, like [the 2006 version of 8.022](https://ocw-studio.odl.mit.edu/sites/?q=8.022). Note that in the list view, it is marked "Published" 
1. Unpublish a site, like https://ocw.mit.edu/courses/8-022-physics-ii-electricity-and-magnetism-fall-2006/
2. Confirm that it has been removed. 
3. Return to the list view. 

### Expected Behavior

- The site should no longer have the "Published" label. 

### Actual Behavior

- The listing still indicates that it is published

### Screenshot

image

Wassaf-Shahzad:
#### Pre-Flight checklist

- [x] Testing

closes #1404

#### What's this PR do?
Adds a check on live-publish-status

- Publish a site and verify its status is displayed as published on SiteDashboard
- Unpublish a site and verify status is displayed as Draft on SiteDashboard

### DIFF
--- a/static/js/pages/SitesDashboard.tsx
+++ b/static/js/pages/SitesDashboard.tsx
@@ -74,7 +74,7 @@ export default function SitesDashboard(): JSX.Element {
               to={siteDetailUrl.param({ name: site.name }).toString()}
               key={site.uuid}
             >
-              {site.publish_date && site.live_publish_status ? (
+              {site.publish_date && !site.unpublished ? (
                 <div className="text-success">Published</div>
               ) : (
                 <div className="text-dark">Draft</div>
--- a/static/js/pages/SitesDashboard.tsx
+++ b/static/js/pages/SitesDashboard.tsx
@@ -74,7 +74,7 @@ export default function SitesDashboard(): JSX.Element {
               to={siteDetailUrl.param({ name: site.name }).toString()}
               key={site.uuid}
             >
-              {site.publish_date ? (
+              {site.publish_date && site.live_publish_status ? (
                 <div className="text-success">Published</div>
               ) : (
                 <div className="text-dark">Draft</div>
--- a/static/js/types/websites.ts
+++ b/static/js/types/websites.ts
@@ -234,6 +234,7 @@ export interface WebsiteStatus {
   sync_status: GoogleDriveSyncStatuses | null
   synced_on: string | null
   sync_errors: Array<string> | null
+  unpublished: boolean
 }
 
 export type Website = WebsiteStatus & {
@@ -256,7 +257,6 @@ export type Website = WebsiteStatus & {
   url_path: string | null
   url_suggestion: string
   s3_path: string | null
-  unpublished: boolean | undefined
 }
 
 type WebsiteRoleEditable = typeof ROLE_ADMIN | typeof ROLE_EDITOR
--- a/static/js/types/websites.ts
+++ b/static/js/types/websites.ts
@@ -256,7 +256,7 @@ export type Website = WebsiteStatus & {
   url_path: string | null
   url_suggestion: string
   s3_path: string | null
-  unpublished: boolean
+  unpublished: boolean | undefined
 }
 
 type WebsiteRoleEditable = typeof ROLE_ADMIN | typeof ROLE_EDITOR
--- a/static/js/types/websites.ts
+++ b/static/js/types/websites.ts
@@ -256,6 +256,7 @@ export type Website = WebsiteStatus & {
   url_path: string | null
   url_suggestion: string
   s3_path: string | null
+  unpublished: boolean
 }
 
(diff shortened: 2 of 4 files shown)


pt2302:
@Wassaf-Shahzad Could you please provide guidance for how to test this locally? On a local version of OCW Studio, this change appears to make all of the courses' status into `Draft`.

Wassaf-Shahzad:
> @Wassaf-Shahzad Could you please provide guidance for how to test this locally? On a local version of OCW Studio, this change appears to make all of the courses' status into `Draft`.
> 
> In particular, it looks like the variable `site.live_publish_status` may not be being set properly. Querying the backend of a local copy of the prod database, most sites have `live_publish_status` as `succeeded`. Also, should this check
> 
> https://github.com/mitodl/ocw-studio/blob/eb793b4f3bc1264e631cf6efbd1efef9b07a79c1/static/js/pages/SitesDashboard.tsx#L77
> 
> be looking for `live_publish_status === "succeeded"` rather than not `null`?

- For local testing I was actually following this [code](https://github.com/mitodl/ocw-studio/blob/master/websites/api.py#L339) and setting values from the admin panel
- The reason I did not check for `live_publish_status` as `succeeded` is  since we can republish a course. Now in the case of a republish if the publish fails or is aborted the publish_date remains unchanged while  live_publish_status is changed to `not None`. In this case the user should see published (Since the course was never unpublished, just republished and failed) on the course card. In case of and unpublished event live_publish_status becomes none and we Draft is shown irrespective of publish_date.

pdpinch:
What are the possible values of the status label?

ChristopherChudzicki:
@Wassaf-Shahzad Looking at the API response for `GET /api/websites/?limit=10&offset=0`, I think @pt2302 is correct that `live_publish_status` by `WebsiteSerializer` https://github.com/mitodl/ocw-studio/blob/09ef8c6f1972061704e933e0d0c226372dfafc0e/websites/serializers.py#L37-L47 (Looks like it's only included in the detail response).

I think we should just add the two publishing statuses to the list serializer.

ChristopherChudzicki:
It looks like the frontend Website type is the "detailed" response for `/api/websites<site_name>`, and a quick look makes me think we do not have a separate type for the less detailed paginated `/api/websites` response. (In contrast, we do have separate detail vs list types for WebsiteContent objects on the frontend.)

IMO we should probably just make the "detailed" and "less detailed" responses as similar as possible...at least include all the non-FK fields.

ChristopherChudzicki:
@Wassaf-Shahzad I'm also having trouble with this, as indicated above. It seems that all sites display "Draft" even when they are published.  I think the issue is with `WebsiteSerializer` as mentioned above (and the related issue that the typescript definition for the API response was incorrect. Sorry about that!)

Screen Shot 2022-06-24 at 11 33 11 AM

pt2302:
@Wassaf-Shahzad Why did you add a new `unpublished` variable instead of passing the value of the existing `live_publish_status` variable to the frontend? To me, that would seem like a simpler route.

Wassaf-Shahzad:
> @Wassaf-Shahzad Why did you add a new `unpublished` variable instead of passing the value of the existing `live_publish_status` variable to the frontend? To me, that would seem like a simpler route.

Yes, since if you look at the description of the model property in `websites/models.py`, The unpublished property is specifically calculated to check whether  the course has been unpublished or not.
Also its not exactly new, since to pass the `live_publish_status` to the frontend would need the same amount of changes.

Also there are some things that i might need clarification on @ChristopherChudzicki .
- We are using `Websites.Meta.fields` property in CreateWebsiteSerializer and WebsiteDetailSerializer, without inheriting WebsiteSerializer. Why ? 
- This results in some issue where serialisers fields (not directly coming from the model) need to be described in both of these models.
- Would it not be better to just inherit the serialiser ?
- Also if we cant inherit them for some reason or the other. We should make the common fields in Webiste.Meta.fields as a list of constants and add them in meta when needed.


        class Meta:
            fields  = [custom_fields] + common_fields

ChristopherChudzicki (approved):
@Wassaf-Shahzad I agree that using `!unpublished` seems better than checking the status.

Re
> We are using Websites.Meta.fields property in CreateWebsiteSerializer and WebsiteDetailSerializer, without inheriting WebsiteSerializer. Why ? ... Would it not be better to just inherit the serialiser ?

I don't know. Sounds like maybe it should? Matt (on vacation), @gumaerc or @abeglova might have better insight.

But, this looks good to me now. 👍

pt2302 (approved):
With the latest commit, this looks good to me too.

Issue #1434, opened 2022-08-05
Code blocks should default to "Plain Text" instead of "C"


pdpinch:
As a course author I'd like to have code block formatting be more generic. 

Currently it defaults to `c` formatting, but a simpler formatting, that works with python, HTML and shell code would be preferred. 

#### Acceptance Criteria:

- [ ] markdown saved by studio should specify plaint text code formatting
- [ ] confirm that it looks correct in the rendered site, including some screenshots

PR #1437, opened 2022-08-09
Setting default language for code blocks to plain text

pt2302:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1434.

#### What's this PR do?
The default language for code blocks in OCW Studio is now "plain text"; users can still select a different language as before from the dropdown menu next to the code block icon.

Create a code block within a page in OCW Studio, and verify that `plain text` is the default.

The code previously sorted all of the languages available for code blocks alphabetically, resulting in C appearing first. Since CKEditor sets the first language listed as the default, the previous default was C.


mbertrand (approved):
👍

PR #1442, opened 2022-08-10
clean up mass build sites logging

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested
  - [ ] Changes have been manually tested

#### What are the relevant tickets?
Closes https://github.com/mitodl/ocw-studio/issues/1440

#### What's this PR do?
In https://github.com/mitodl/ocw-studio/pull/950, we added the `mass-build-sites` pipeline and the associated management command for pushing it up to Concourse. This pipeline runs a bash script that queries the `ocw-studio` API for a list of publishable courses, iterating them and building them one by one. One major issue with this has been that it produces so much log output that the resulting pipeline output is unreadable.  There is so much in it that it will eventually crash the Concourse web UI when trying to view the pipeline run. This PR sets a "quiet" option for `curl`,`hugo` and `aws` commands and produces intentional logging where applicable. Errors will still be logged, so looking back at the log it should be easy to tell where and why a given error occurred. The volume of logging will be significantly reduced, so the log should actually load in the Concourse UI without crashing the browser.

#### How should this be manually tested?
 - Spin up `ocw-studio` locally with Concourse and Minio enabled; see the readme for more details on this if you are unsure how to do so
 - Restore a production backup to your local `ocw-studio` database or make sure you have `ocw-www` as well as some test course sites in your database
 - If you have many sites in your local database, you may want to edit `websites/views.py` and on line 292 change it to be: `sites = sites.prefetch_related("starter").order_by("name")[:10]` to limit to 10 sites
 - Upsert the mass build pipeline to your local Concourse instance with `docker-compose exec web ./manage.py upsert_mass_build_pipeline`
 - Browse to http://concourse:8080, login and find the mass build pipeline then trigger a build
 - View the log output of the `get-repo-build-course-publish-course` step and make sure you see only 4 lines of logging per course and the build should succeed:
```
STARTING PUBLISH OF 10-01-ethics-for-engineers-artificial-intelligence-spring-2020
RUNNING HUGO BUILD FOR 10-01-ethics-for-engineers-artificial-intelligence-spring-2020
STARTING S3 SYNC FOR 10-01-ethics-for-engineers-artificial-intelligence-spring-2020
PUBLISH OF 10-01-ethics-for-engineers-artificial-intelligence-spring-2020 COMPLETE
```
[...]

PR #1445, opened 2022-08-12
don't retry the mass build automatically

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Follow up for https://github.com/mitodl/ocw-studio/pull/1442

#### What's this PR do?
In https://github.com/mitodl/ocw-studio/pull/1442, we made some changes to the mass build pipeline that cleans up the logging and also will set the build as failed if any errors are detected during the build. With `attempts: 3` set on the job, the pipeline will retry the ~2 hour mass build step 3 times if any of the sites fail.  This is obviously not great and should only be done manually after the issues with the first run are addressed.  This sets it to `attempts: 1`

Follow the same instructions as https://github.com/mitodl/ocw-studio/pull/1442, except skip straight to the failure mode. Ensure that when the pipeline fails, it isn't retried.


mbertrand (approved):
👍

PR #1447, opened 2022-08-16
resolve contentNames discrepancy

ChristopherChudzicki:
PR for discussion

### DIFF
--- a/static/js/components/widgets/ResourcePickerDialog.test.tsx
+++ b/static/js/components/widgets/ResourcePickerDialog.test.tsx
@@ -72,7 +72,7 @@ describe("ResourcePickerDialog", () => {
       ),
       {
         mode:         RESOURCE_EMBED,
-        contentNames: ["resource", "pages"],
+        contentNames: ["resource", "page"],
         isOpen:       true,
         closeDialog:  closeDialogStub,
         insertEmbed:  insertEmbedStub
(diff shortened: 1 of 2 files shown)


static/js/components/widgets/ResourcePickerDialog.test.tsx
```diff
@@ -72,7 +72,7 @@ describe("ResourcePickerDialog", () => {
       ),
       {
         mode:         RESOURCE_EMBED,
-        contentNames: ["resource", "pages"],
+        contentNames: ["resource", "page"],
```
ChristopherChudzicki, line 75: So I believe what was happening was...

1. `contentNames` should be `CollectionName[]`
2. `<ResourcePickerDialog />` filters the contentNames based on what collections actually exist in the config

In the test file, there was a mismatch between `contentNames` and the actual collection names.

Note:
    - in the test file, `contentNames` is passed manually
    - in "real" ocw, `contentNames` is calculated in a parent component, `<MarkdownEditor />` based on the config

The calculation of `contentNames` could be moved into ResourcePickerDialog (nothing else uses it) then the prop could be gotten rid of all together. (It'd still be possible to have a mismatch in the config, though).

static/js/components/widgets/ResourcePickerDialog.test.tsx
```diff
@@ -102,12 +102,12 @@ describe("ResourcePickerDialog", () => {
     },
     {
       mode:         RESOURCE_LINK,
-      contentNames: ["pages"],
+      contentNames: ["page"],
```
ChristopherChudzicki, line 105: I think in this case:
- `"pages"` was not a valid content name
- so after filtering the array was empty, so you got that error about "Cannot read property of undefined"

static/js/components/widgets/ResourcePickerDialog.tsx
```diff
@@ -113,6 +113,13 @@ export default function ResourcePickerDialog(props: Props): JSX.Element {
       .filter(entry => entry.category === "Content")
       .flatMap(elt => [elt.name, elt.label])
```
ChristopherChudzicki, line 114: Not really related to the bug, but IMO  we should just use the name, not name or label.

Issue #1419, opened 2022-07-14
studio config file should specify resource picker tabs by itself


pdpinch:
The underlying issue is an incomplete separation of concerns between Studio frontend and OCW configs... When you insert links / resources into markdown, there’s this picker dialog with tabs (below). What tabs show up is determined by the config. (So ocw-www and ocw-courses can show different tabs, for example).

<img width="711" alt="Screen Shot 2022-06-30 at 12 13 39 PM" src="https://user-images.githubusercontent.com/9010790/176726793-870f48f4-a5da-46df-a003-140ba58d524d.png">

But—as currently written—studio needs know the set of possible tabs. The possible tabs and labels are stored in studio here:  https://github.com/mitodl/ocw-studio/blob/f00c392adbd827ea9c68051ef61a406c881a9b82/static/js/components/widgets/ResourcePickerDialog.tsx#L52

And the original commit specified a tab `resource-list` that studio was not prepared to deal with https://github.com/mitodl/ocw-hugo-projects/commit/9fe2ed0bb5d057535339274cb83d39f052af0337#diff-4e28d8dc0f3e2776c95d55c22426e08abadbb81b90355b2d10de8772fb7fcde1R37

The tabIds are hard-coded content type values. It would be good if we avoided this hard-coding of "possible tab" content types all together, and config handled it all

_Originally posted by @ChristopherChudzicki in https://github.com/mitodl/ocw-hugo-projects/issues/206#issuecomment-1171415424_

pt2302:
Per discussion with @Ferdi and @pdpinch, this is the current plan for this issue:
- Leave the current config files in `ocw-hugo-projects` unchanged.
- The frontend should handle "known" types with hardcoded values (resources) and for anything else, use the `label` property from the config file. For example,
```
  - category: Content
    folder: content/video_galleries
    label: "Video Gallery"
    name: video_gallery
```
and 
```
link:
   - video_gallery
```
would create a `TabSettings` object with the following properties:
```
     title: "Video Gallery" #from the label, as above
     id: "video_gallery" #from the name, as above
     contentType: "video_gallery" #from the name, as above
     resourcetype: null #default
     embeddable: false #default
     singleColumn: true #default
```     
- If there is no associated content type defined in the config file, ignore the name provided in the link field and don't create a tab in the ResourcePicker.

pt2302:
[...]

PR #1448, opened 2022-08-17
Pt/resource picker tabs

pt2302:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1419.

#### What's this PR do?
Instead of being hardcoded in the frontend, OCW Studio now lets users define which new Content Types should appear in the Resource Picker. Any Content Type that is defined in the config file (website starter) can now be added to the `link` menu in the Resource Picker by adding it in the config file. Previously, if the new Content Type was not of the types hardcoded in the frontend, this would be ignored and it would not be possible to add a tab for it. 

See https://github.com/mitodl/ocw-studio/issues/1419#issuecomment-1206813572 and https://github.com/mitodl/ocw-studio/issues/1419#issuecomment-1209309554 for additional context.

To test this locally, open Django admin within a locally-running instance of OCW Studio. If a content type is listed under `link` or `embed` in the config file but is not defined in the config as a Content Type, it should be ignored and not show up as a tab in the Resource Picker. If a content type is added to the config file and is also listed under `link` in the config file, a new tab in the Resource Picker should appear for that Content Type.

### DIFF
--- a/static/js/components/widgets/ResourcePickerDialog.tsx
+++ b/static/js/components/widgets/ResourcePickerDialog.tsx
@@ -19,6 +19,7 @@ import {
 } from "../../lib/ckeditor/plugins/constants"
 import { WebsiteContent } from "../../types/websites"
 import { useWebsite } from "../../context/Website"
+import _, { keyBy } from "lodash"
 
 interface Props {
   mode: ResourceDialogMode
@@ -37,11 +38,14 @@ interface TabSettings {
   singleColumn: boolean
 }
 
-export function toTabSettings(elt: string[]) {
+export function toTabSettings(
+  contentType: string,
+  contentTitle: string
+): TabSettings {
   const newTabSettings: TabSettings = {
-    title:        elt[1],
-    id:           elt[0],
-    contentType:  elt[0],
+    title:        contentTitle,
+    id:           contentType,
+    contentType:  contentType,
     resourcetype: null,
     embeddable:   false,
     singleColumn: true
@@ -109,9 +113,11 @@ export default function ResourcePickerDialog(props: Props): JSX.Element {
   const { mode, isOpen, closeDialog, insertEmbed, contentNames } = props
   const website = useWebsite()
   const definedCategories = useMemo(() => {
-    return website.starter?.config?.collections
-      .filter(entry => entry.category === "Content")
-      .flatMap(elt => [elt.name, elt.label])
+    const contentCollections =
+      website.starter?.config?.collections.filter(
+        entry => entry.category === "Content"
+      ) ?? []
+    return keyBy(contentCollections, collection => collection.name)
   }, [website.starter?.config?.collections])
   const tabs = useMemo(
     () =>
@@ -119,12 +125,13 @@ export default function ResourcePickerDialog(props: Props): JSX.Element {
         .flatMap(name => {
           if (name && name === "resource") {
             return [documentTab, videoTab, imageTab, otherFilesTab]
-          } else if (
-            definedCategories &&
-            definedCategories?.indexOf(name) !== -1
-          ) {
-            const idx = definedCategories?.indexOf(name)
-            return [toTabSettings(definedCategories?.slice(idx, idx + 2))]
+          } else if (definedCategories && _.has(definedCategories, name)) {
+            return [
+              toTabSettings(
+                definedCategories[name]["name"],
+                definedCategories[name]["label"]
+              )
+            ]
           } else return []
         })
         .filter(tab => tab && (mode !== RESOURCE_EMBED || tab.embeddable)),
--- a/static/js/components/widgets/ResourcePickerDialog.tsx
+++ b/static/js/components/widgets/ResourcePickerDialog.tsx
@@ -38,10 +38,7 @@ interface TabSettings {
   singleColumn: boolean
 }
 
-export function toTabSettings(
-  contentType: string,
-  contentTitle: string
-): TabSettings {
+const toTabSettings = (contentType: string, contentTitle: string) => {
   const newTabSettings: TabSettings = {
     title:        contentTitle,
     id:           contentType,
@@ -125,7 +122,7 @@ export default function ResourcePickerDialog(props: Props): JSX.Element {
         .flatMap(name => {
           if (name && name === "resource") {
             return [documentTab, videoTab, imageTab, otherFilesTab]
-          } else if (definedCategories && _.has(definedCategories, name)) {
+          } else if (_.has(definedCategories, name)) {
             return [
               toTabSettings(
                 definedCategories[name]["name"],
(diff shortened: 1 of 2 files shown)


static/js/components/widgets/ResourcePickerDialog.tsx
```diff
@@ -29,12 +30,23 @@ interface Props {
...
-  contentType: ContentType
+  id: string
+  contentType: string
   resourcetype: string | null
   embeddable: boolean
   singleColumn: boolean
-  sourceWebsiteName?: string
+}
+
+export function toTabSettings(elt: string[]) {
```
ChristopherChudzicki, line 40: The argument `elt` should always have exactly two elements, right? 

**Requested Change:** If yes, let's just use two arguments with meaningful names, e.g., `toTabSettings(contentType, title)`.

static/js/components/widgets/ResourcePickerDialog.tsx
```diff
@@ -151,13 +107,30 @@ const modeText = {
 
 export default function ResourcePickerDialog(props: Props): JSX.Element {
   const { mode, isOpen, closeDialog, insertEmbed, contentNames } = props
+  const website = useWebsite()
+  const definedCategories = useMemo(() => {
+    return website.starter?.config?.collections
+      .filter(entry => entry.category === "Content")
+      .flatMap(elt => [elt.name, elt.label])
+  }, [website.starter?.config?.collections])
```
ChristopherChudzicki, line 115: This has structure like
```
["pages", "Pages", "bananas", "Bananas", "video_galleries", "Video Galleries", ...]
```
so implicitly, indices 0, 1 are a single item, 2, 3 are a single item, etc.

**Requested change:** Make that explicit: group the things together.

**Suggestion:** I think a structure like
```ts
const contentCategires = {
  pages: { name: page, label: "Page" }, 
  bananas: { name: bananas, label: "Bananas" },
  ...
}
```
would be helpful (more explicit, and also easier lookups below... you won't have to fuss with `indexOf`.)

lodash's `[keyBy](https://lodash.com/docs/4.17.15#keyBy)` is very helpful for this sort of thing...something like
```ts
  const definedCategories = useMemo(() => {
    const contentCollections = website.starter?.config?.collections
      .filter(entry => entry.category === "Content") ?? []
    return keyBy(contentCollections, collection => collection.name)
  }, [website.starter?.config?.collections])
```

static/js/components/widgets/ResourcePickerDialog.tsx
```diff
@@ -151,13 +107,30 @@ const modeText = {
...
-        .flatMap(name => TABS[name])
+        .flatMap(name => {
+          if (name && name === "resource") {
+            return [documentTab, videoTab, imageTab, otherFilesTab]
+          } else if (
+            definedCategories &&
+            definedCategories?.indexOf(name) !== -1
+          ) {
+            const idx = definedCategories?.indexOf(name)
+            return [toTabSettings(definedCategories?.slice(idx, idx + 2))]
```
ChristopherChudzicki, line 127: With the above suggestion, you should be able to just lookup the category by name: `definedCategories[name]`.

ChristopherChudzicki (changes requested):
Works great! A couple small requests to hopefully improve readability.

static/js/components/widgets/ResourcePickerDialog.tsx
pt2302: Updated in https://github.com/mitodl/ocw-studio/pull/1448/commits/b46ad80309a35e46f7487f9c547179a8bcbe6fe2.

static/js/components/widgets/ResourcePickerDialog.tsx
pt2302: Updated in https://github.com/mitodl/ocw-studio/pull/1448/commits/b46ad80309a35e46f7487f9c547179a8bcbe6fe2.

static/js/components/widgets/ResourcePickerDialog.tsx
pt2302: Updated in https://github.com/mitodl/ocw-studio/pull/1448/commits/b46ad80309a35e46f7487f9c547179a8bcbe6fe2.

static/js/components/widgets/ResourcePickerDialog.tsx
```diff
@@ -29,12 +31,26 @@ interface Props {
...
-  contentType: ContentType
+  id: string
+  contentType: string
   resourcetype: string | null
   embeddable: boolean
   singleColumn: boolean
-  sourceWebsiteName?: string
+}
+
+export function toTabSettings(
```
ChristopherChudzicki, line 41: **Request:** don't export this, it's not used anywhere else. Sorry if I missed that in initial review.

**Very Minor Suggestion:** Personally, I prefer [arrow function expressions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions), i.e., 
```ts
const toTabSettings = (contentType: string, contentTitle: string) => {...}
```
The primary differences are:
- functions declared with `function` keyword are [hoisted](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting)
- functions declared with the `function` keyword have their own `this` context, whereas arrow functions **always** use the `this` context in which they were defined. .... [`this`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this) in js kinda quirky.

Both of the things above sometimes cause confusion... so unless I really need a function with its own `this` context, I tend to avoid functions declared with function keyword.

ChristopherChudzicki (approved):
👍 Looks good to me!

I left a request and a suggestion, feel free to make them...neither warrant a re-review imo

static/js/components/widgets/ResourcePickerDialog.tsx
```diff
@@ -151,13 +111,33 @@ const modeText = {
...
+    return keyBy(contentCollections, collection => collection.name)
+  }, [website.starter?.config?.collections])
   const tabs = useMemo(
     () =>
       contentNames
-        .flatMap(name => TABS[name])
+        .flatMap(name => {
+          if (name && name === "resource") {
+            return [documentTab, videoTab, imageTab, otherFilesTab]
+          } else if (definedCategories && _.has(definedCategories, name)) {
```
ChristopherChudzicki, line 128:
```suggestion
          } else if (_.has(definedCategories, name)) {
```
`definedCategories` might be empty, but it's never falsy. (Empty objects/arrays in js are truthy, unlike empty dictionaries/lists in python)

static/js/components/widgets/ResourcePickerDialog.tsx
pt2302: Updated in a1f7c9758490ba9de2953f293f058057e134db9f.

static/js/components/widgets/ResourcePickerDialog.tsx
pt2302: Updated in a1f7c9758490ba9de2953f293f058057e134db9f.

Issue #1438, opened 2022-08-09
Warn user when publishing a site with draft content


pdpinch:
As a course author, I'd like to be reminded when I publish a course with draft content in it. We had an issue recently where a course author published a course and couldn't figure out why their content displayed in staging, but not on the production site. 

Currently, we display warnings in the publish drawer for a variety of course issues, like missing caption files. 

#### Acceptance Criteria:

- [ ] When publishing a site with draft content, display a warning in the publish drawer below the status message. 

#### screenshot of current design for warnings

<img width="524" alt="image" src="https://user-images.githubusercontent.com/430126/183651797-58a978d7-be3c-469e-88f9-987bca826ec7.png">

pdpinch:
If it's easier, we could display the warning when the drawer is opened, instead of waiting for when the site is published.

pt2302:
It appears that one option is to add this as a warning to this function:
https://github.com/mitodl/ocw-studio/blob/3c70b1044d110bf716a88f9b65dbd7f10037dfe9/websites/api.py#L367 and this would be then get written to the Publish Drawer via this function:
https://github.com/mitodl/ocw-studio/blob/04bd4425cd945a1ede3f759656cae54c497348b9/static/js/components/PublishDrawer.tsx#L176-L184.

ChristopherChudzicki:
IMO it makes sense to put the draft warning as part of the content warnings. @pt2302 from slack you said:

> although draft content isn't exactly an example of "incomplete content"

Maybe we could change the Python function's name to `get_content_warnings` or something. (I.e., any relevant warnings, not just incompleteness.) Looks like the response body property is already has the more generic name "content_warnings".

PR #1456, opened 2022-08-29
Pt/draft content warnings

pt2302:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1438.

#### What's this PR do?
Adds a warning to the Publish Drawer listing the title of any content that has been set to Draft.

Change some content (such as Resources, Pages, etc.) to `Draft` by selecting `True` in the Edit drawer. Then, click the `Publish` button to open the Publish Drawer. There should now be a warning listing the title of the content. Change `Draft` back to `False`, and the warning should disappear.

DraftContentWarning

### DIFF
--- a/websites/api.py
+++ b/websites/api.py
@@ -270,9 +270,7 @@ def videos_missing_captions(website: Website) -> List[WebsiteContent]:
 
 
 def draft_content(website: Website) -> List[WebsiteContent]:
-    """Return a list of WebsiteContent objects that are"""
-    if not is_ocw_site(website):
-        return []
+    """Return a list of WebsiteContent objects that are set to Draft"""
     query_draft_field = get_dict_query_field("metadata", "draft")
     return WebsiteContent.objects.filter(
         Q(website=website) & Q(**{query_draft_field: True})
--- a/websites/api.py
+++ b/websites/api.py
@@ -192,7 +192,7 @@ def is_ocw_site(website: Website) -> bool:
 
 
 def update_youtube_thumbnail(website_id: str, metadata: Dict, overwrite=False):
-    """ Assign a youtube thumbnail url if appropriate to a website's metadata"""
+    """Assign a youtube thumbnail url if appropriate to a website's metadata"""
     website = Website.objects.get(uuid=website_id)
     if is_ocw_site(website):
         youtube_id = get_dict_field(metadata, settings.YT_FIELD_ID)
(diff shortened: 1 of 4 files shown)


websites/api.py
```diff
@@ -269,6 +269,16 @@ def videos_missing_captions(website: Website) -> List[WebsiteContent]:
     )
 
 
+def draft_content(website: Website) -> List[WebsiteContent]:
+    """Return a list of WebsiteContent objects that are"""
+    if not is_ocw_site(website):
+        return []
```
ChristopherChudzicki, line 275: I think you should check with Peter/Ferdi about this. IMO, it makes sense to always issue warnings about drafts.

BTW, re drafts... 
- **in reference to websites**: in the codebase, we refer to "Publish to staging" as publishing the "draft" version of a website
- **in reference to websitecontent:**
     1.  Yaml formatted frontmatter at the start of a markdown file is "metadata" to hugo. Our websitecontent.metadata becomes this yaml-formatted metadata when we publish.
     2. "draft" is a special hugo metadatafield. When you run hugo, it converts all the markdown files to HTML. By default, hugo ignores stuff with `draft: true` in its frontmatter


Screen Shot 2022-08-29 at 11 41 32 AM

ChristopherChudzicki (approved):
👍 Looks good, but I do think you should check regarding the one comment I made.

websites/api.py
pdpinch: > I think you should check with Peter/Ferdi about this. IMO, it makes sense to always issue warnings about drafts.

I agree with Chris.

pdpinch:
Would it be a lot of work to add a bit more formatting to the message, such as:
 
- The following content is still set to **Draft** 
  - Page 1 
  - Page 2
  - Resource 3

pt2302:
@pdpinch I agree that this would be a cleaner display format, but for consistency I think that this should be a change we should make to all warnings (and so, would probably make sense as a separate PR).

websites/api.py
pt2302: Updated in 6be55ac78b9854334a8c6614ef51965733c893ea.

ChristopherChudzicki:
@pdpinch I don't think it would be a lot of work, but I agree with @pt2302 that it makes sense to be done in a separate pr.

@pt2302 Re "would this involve frontend change"...

This would probably be a frontend + backend change: frontend to do the formatting, backend to provide more structured data. Right now the `GET websites/` api response includes a `content_warnings` property which is an array of strings:
```ts
// currently:
[
    "This website has unpublished stuff: thing1, thing2, thing3...",
    "This website is missing captions on: video1, video2, video3, ..."
    "Some other warning"
  ]
```
To facilitate nicer formatting, I think we'd want to tweak the API response to something like
So I think we'd want to change the API response to be something like
```
[
  {
    description: "This website has unpublished stuff",
    items: ["thing1", "thing2", "thing3"]
    plaintext: "maybe a plaintext version?"
  },
  ...
]
```

Issue #1457, opened 2022-08-29
Warnings when unpublishing courses


pdpinch:
As an admin who is unpublishing a course, I'd like to get a message about any site dependencies that might break when the course is removed. I would use this information to make list of sites/courses that need to be updated after the course is unpublished. 

#### Acceptance Criteria:

- [ ] since there's no UI for unpublishing, it would be OK to implement this as text returned from the unpublish API (i.e. `/api/websites/<Website.name>/unpublish/`)
- [ ] ideally, the API would return warnings about:
  - [ ] course lists in ocw-www that reference the unpublished course
  - [ ] resource lists in ocw-www that reference resources in the unpublished course
  - [ ] links in course content to the unpublished course (least important and possibly harder to do)

MAbdurrehman12:
@pdpinch Should we proceed with unpublishing right away and a warning text returned in response? 
I wonder if a better approach will be to take the user to a `confirm unpublish` page where we have this warning text along with a button to proceed/unpublish. So the side effects can be taken care of beforehand.

What do you think?

pdpinch:
Is there a way to implement that kind of UX when we're just using an API? There's a lot of complications with adding a web UI for this feature that I'd like to avoid for now.

MAbdurrehman12:
Maybe what we can do is: Whenever the user hits `/api/websites/<Website.name>/unpublish/)`: This is a `GET` request, we can fetch the dependencies and show them as a text. 
Currently, since method for `GET` is not defined so we are seeing this: 
<img width="405" alt="image" src="https://user-images.githubusercontent.com/93309234/187908637-cce61377-0271-4512-b47b-d7030c41b448.png">
So if we define it and return the dependencies as a text, it would be fine too. Post button to unpublish will still be there and we won't be having any redirection to `confirm unpublish` too which I stated above.

MAbdurrehman12:
I'm talking about something like this: 
Whenever we go to `/api/websites/<Website.name>/unpublish` for unpublishing, we will see a message something like this: 
<img width="822" alt="image" src="https://user-images.githubusercontent.com/93309234/187912948-cef61817-f949-4b1a-bb37-370819b617cf.png">

and when we click the `POST` button, we will see a success msg, something like this:
<img width="697" alt="image" src="https://user-images.githubusercontent.com/93309234/187913236-cdf4490b-d825-4adc-a968-597e9e08095c.png">


Is this fine or you'd recommend something else?

[...]

PR #1468, opened 2022-09-06
feat: site dependencies shown while course unpublishing

MAbdurrehman12:
#### Pre-Flight checklist

- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested
- [ ] Settings

Closes https://github.com/mitodl/ocw-studio/issues/1457

#### What's this PR do?
- While unpublishing a course, returns: (in response of `GET` request)
    - course lists in ocw-www that reference the unpublished course
    - resource lists in ocw-www that reference resources in the unpublished course
    - courses that have links to the unpublished course in content

- Go to `/api/websites/<Website.name>/unpublish/`
- Use any course which has dependencies in ocw-www `course lists` and `resource collections` or in any course
- Verify that they (all the dependencies) are sent in response when you visit the above URL.
- Unpublish the course
- Verify the course is unpublished successfully.

image

image

### DIFF
--- a/websites/views.py
+++ b/websites/views.py
@@ -271,14 +271,14 @@ class WebsiteViewSet(
                 return Response(
                     status=200,
                     data={
-                        "Site dependencies": {
-                            "ocw www": WebsiteContentSerializer(
+                        "site_dependencies": {
+                            "ocw_www": WebsiteContentSerializer(
                                 instance=ocw_www_dependencies, many=True
                             ).data,
                             "course": WebsiteBasicSerializer(
                                 instance=course_dependencies, many=True
                             ).data,
-                            "course content": WebsiteContentSerializer(
+                            "course_content": WebsiteContentSerializer(
                                 instance=course_content_dependencies, many=True
                             ).data,
                         }
--- a/websites/views.py
+++ b/websites/views.py
@@ -58,6 +58,7 @@ from websites.permissions import (
     is_global_admin,
 )
 from websites.serializers import (
+    WebsiteBasicSerializer,
     WebsiteCollaboratorSerializer,
     WebsiteContentCreateSerializer,
     WebsiteContentDetailSerializer,
@@ -65,7 +66,6 @@ from websites.serializers import (
     WebsiteDetailSerializer,
     WebsiteMassBuildSerializer,
     WebsiteSerializer,
-    WebsiteBasicSerializer,
     WebsiteStarterDetailSerializer,
     WebsiteStarterSerializer,
     WebsiteStatusSerializer,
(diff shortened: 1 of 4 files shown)


pt2302:
@MAbdurrehman12 This works for me! One thing that I think would make this better would be to include the course name that has a link to the course being unpublished in the response for courses. While it's possible to find the page by the UUID, I think that it would be easier/clearer to include the course name as well.

(@pdpinch, do you have thoughts on this?)

websites/views.py
```diff
@@ -236,19 +239,53 @@ def publish(self, request, name=None):
...
+                    (
+                        Q(metadata__courses__icontains=website.name)
+                        | Q(metadata__resources__content__icontains=website.name)
+                    ),
+                    website__name=settings.ROOT_WEBSITE_NAME,
+                )
+                course_dependencies = WebsiteContent.objects.filter(
+                    type=CONTENT_TYPE_PAGE,
+                    markdown__contains=website.name,
+                )
```
ChristopherChudzicki, line 265: One other place (outside of markdown) where cross-course links show up fairly often is the overall course description. For exmaple, in 8.06
- on ocw https://ocw.mit.edu/courses/8-06-quantum-physics-iii-spring-2005/
- on github: https://github.mit.edu/mitocwcontent/8.06-spring-2018/blob/main/data/course.json#L31

It would be nice if we could do a `metadata__contains=website.name` treating metadata as a string instead of json.

Might not be worth worrying about since @pdpinch indicated that this was the lowest priority of the 3 warnings.

MAbdurrehman12:
> This works for me! One thing that I think would make this better would be to include the course name that has a link to the course being unpublished in the response for courses. While it's possible to find the page by the UUID, I think that it would be easier/clearer to include the course name as well.

I agree

websites/views.py
MAbdurrehman12: yeah, let's do it

MAbdurrehman12:
@pt2302 @ChristopherChudzicki Thanks for your feedback.
I've made some changes, can you please have another look?

pt2302 (approved):
This works well for me! I think that having the course name as part of the results makes it much clearer and easier to use.

websites/views.py
```diff
@@ -236,19 +240,60 @@ def publish(self, request, name=None):
...
+                    status=200,
+                    data={
+                        "Site dependencies": {
+                            "ocw www": WebsiteContentSerializer(
+                                instance=ocw_www_dependencies, many=True
+                            ).data,
+                            "course": WebsiteBasicSerializer(
+                                instance=course_dependencies, many=True
+                            ).data,
+                            "course content": WebsiteContentSerializer(
```
ChristopherChudzicki, line 281: Any reason for whitespace in the response property names? Could we use `ocw_www`, etc?

MAbdurrehman12: @ChristopherChudzicki Sure.
can you have another look?

ChristopherChudzicki (approved):

PR #1469, opened 2022-09-06
added initial code for Subscripts

### LINKED ISSUE #1061: create and edit sub/superscript text in studio
As a course author, I need to be able to add subscript and superscript text in the wysiwyg editor, for math and chemistry courses. 

#### Designs and Mockups

TBD

#### Acceptance Criteria:

- [ ] write a proposal for how to approach this problem, in studio, in markdown and in hugo
- [ ] Add subscript or superscript text 
- [ ] publish subscript and superscript text

#### Out of Scope

- editing would be nice, but if its difficult we can instruct authors to delete and re-add the sub/superscript when they need to

Wassaf-Shahzad:
#### Pre-Flight checklist

- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested
- [ ] Settings

Closes #1061 

#### What's this PR do?
This PR adds a Markdown Syntax for subscript and superscript. 

This PR can be manually tested by the following steps
- Create a new site.
- Add a page to new site.
- Add a subscript to the site.
- Publish the site
and verify this works as per requirements 

(Optional)

Screenshot 2022-09-06 at 4 17 30 PM

#### Link to Published Course
Link to local published course https://github.com/OCW-LOCAL-COURSES/test-sub

### DIFF
--- a/static/js/lib/ckeditor/plugins/SubscriptSyntaxPlugin.tsx
+++ b/static/js/lib/ckeditor/plugins/SubscriptSyntaxPlugin.tsx
@@ -3,7 +3,6 @@ import Showdown from "showdown"
 
 import MarkdownSyntaxPlugin from "./MarkdownSyntaxPlugin"
 import { TurndownRule } from "../../../types/ckeditor_markdown"
-import { Shortcode } from "./util"
 
 
 export default class SubscriptSyntaxPlugin extends MarkdownSyntaxPlugin {
--- a/static/js/lib/ckeditor/plugins/SubscriptSyntaxPlugin.tsx
+++ b/static/js/lib/ckeditor/plugins/SubscriptSyntaxPlugin.tsx
@@ -6,8 +6,6 @@ import { TurndownRule } from "../../../types/ckeditor_markdown"
 import { Shortcode } from "./util"
 
 
-const SUBSCRIPT_REGEX = Shortcode.regex("subscript", true)
-
 export default class SubscriptSyntaxPlugin extends MarkdownSyntaxPlugin {
   static get pluginName(): string {
     return "SubscriptSyntaxPlugin"
@@ -19,11 +17,12 @@ export default class SubscriptSyntaxPlugin extends MarkdownSyntaxPlugin {
       return [
         {
           type:    "lang",
-          regex:   SUBSCRIPT_REGEX,
+          // eslint-disable-next-line no-useless-escape
+          regex:   new RegExp(/\\\(\s+_{\w+}_\s+\\\)/),
           replace: (s :string) => {
-            const value = Shortcode.fromString(s).get("content")
+            const value = s.match("(?<=_{)(.*)(?=}_)")
             // eslint-disable-next-line no-debugger
-            return `<sub class='subscript'>${value}</sub>`
+            return `<sub class='subscript'>${value ? value[0] : ""}</sub>`
           }
         }
       ]
@@ -39,8 +38,7 @@ export default class SubscriptSyntaxPlugin extends MarkdownSyntaxPlugin {
             return (node.nodeName === "SUB")
           },
           replacement: (content: string, _: Turndown.Node): string => {
-            const text = content
-            const v = Shortcode.toSubscript(text).toHugo()
+            const v = `\\( _{${content}}_ \\)`
             return v
           }
         }
--- a/static/js/lib/ckeditor/plugins/SubscriptSyntaxPlugin.tsx
+++ b/static/js/lib/ckeditor/plugins/SubscriptSyntaxPlugin.tsx
@@ -4,12 +4,10 @@ import Showdown from "showdown"
 import MarkdownSyntaxPlugin from "./MarkdownSyntaxPlugin"
 import { TurndownRule } from "../../../types/ckeditor_markdown"
 
-
 export default class SubscriptSyntaxPlugin extends MarkdownSyntaxPlugin {
   static get pluginName(): string {
     return "SubscriptSyntaxPlugin"
   }
-
 
   get showdownExtension() {
     return function subscriptExtension(): Showdown.ShowdownExtension[] {
@@ -18,8 +16,8 @@ export default class SubscriptSyntaxPlugin extends MarkdownSyntaxPlugin {
           type:    "lang",
           // eslint-disable-next-line no-useless-escape
           regex:   new RegExp(/\\\(\s+_{\w+}_\s+\\\)/),
-          replace: (s :string) => {
-            const value = s.match("(?<=_{)(.*)(?=}_)")
+          replace: (s: string) => {
+            const value = s.match("(?<=_{)(.*)(?=})")
             // eslint-disable-next-line no-debugger
             return `<sub class='subscript'>${value ? value[0] : ""}</sub>`
           }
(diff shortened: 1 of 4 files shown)


ChristopherChudzicki (changes requested):
@Wassaf-Shahzad The code looks good, I do think we should consider other possible ways to achieve superscripts/subscript sin OCW Next, though.

**Current Situation:**
- We already support generic subscripts and superscripts in legacy courses via the `{{% sub %}}` and `{{% sup %}}` shortcodes.
- We do not support creating new `{{% sub %}}` and `{{% sup %}}` shortcodes via CKeditor.

**This PR:** Moves forward by enabling CKEditor to edit/create new sub/sup shortcodes.

I'm not entirely sure that the sub/sup shortcodes are the way to move forward with superscripts in OCW Next. My concern around sub/sup shortocdes is mostly around using subscripts/superscripts inside resource links. Something like this:

Screen Shot 2022-09-06 at 10 38 55 AM

Which after saving renders as:
Screen Shot 2022-09-06 at 10 39 09 AM

I.e., the sub/sup-scripts aren't saved.

_Note:_ Subscripts/superscripts in resource link titles is a use case that comes up fairly frequently, I believe. (E.g., chemistry courses linking to scientific articles with chemical formulas in the name; something similar for math courses.)

As long as we use a simple shortcode for resource links, then using shortcodes for sup/sub-scripts would mean we can't use them in resource_link titles.¹

**Question:** For @pdpinch @Ferdi: Do we want a general-purpose "make any text superscript" functionality? I remember there may have been some hesitation about that. Want to check since https://github.com/mitodl/ocw-studio/issues/1061 hasn't been touched in a while.

If we are particularly interested in subscripts for math/chemistry, maybe MathJax would be a good choice? (I believe MathJax has good support for chemistry via [mhchem](https://mhchem.github.io/MathJax-mhchem/))

----
¹ Possibly if we changed resource_link to be an open+closing shortcode, e.g., `{{% resource_link uuid %}} title goes here {{ /% resource_link %}}`, then we could use other shortcodes in the title. I'm not sure.

pdpinch:
Can we customize the subscript button to render mathjax? 

I do think we should be moving in the direction of mathjax. However, it would be helpful for most course authors if there were still a simple way to write a subscript or a superscript without having to know Latex or MathML.

Wassaf-Shahzad:
> Can we customize the subscript button to render mathjax?
> 
> I do think we should be moving in the direction of mathjax. However, it would be helpful for most course authors if there were still a simple way to write a subscript or a superscript without having to know Latex or MathML.

I can look into it

Wassaf-Shahzad:
> Can we customize the subscript button to render mathjax?
> 
> I do think we should be moving in the direction of mathjax. However, it would be helpful for most course authors if there were still a simple way to write a subscript or a superscript without having to know Latex or MathML.

@pdpinch @ChristopherChudzicki 



https://user-images.githubusercontent.com/87968618/189668217-adccdfe2-a731-4345-a75e-e9763b110a1c.mov

static/js/lib/ckeditor/plugins/SubscriptSyntaxPlugin.tsx
```diff
@@ -0,0 +1,47 @@
...
+  get turndownRules(): TurndownRule[] {
+    return [
+      {
+        name: "SubscriptSyntaxPlugin",
+        rule: {
+          filter: function(node) {
+            return (node.nodeName === "SUB")
+          },
+          replacement: (content: string, _: Turndown.Node): string => {
+            const v = `\\( _{${content}}_ \\)`
```
ChristopherChudzicki, line 40: I don't think there should be a trailing underscore. Just `_{$content}}`. (This version works fine in studio, but to get the shortcodes to work in Hugo I had to remove the trailing underscore.)

ChristopherChudzicki (changes requested):
I'm glad to know Mathjax does work in shortcode parameters like resource links. I left one comment about an implementation issue, which is the "requested change". Wider thoughts below.

@pdpinch Re your question:
> Can we customize the subscript button to render mathjax?

This (updated) PR successfully does that. But I think there are some product downsides to this approach. 

Mathjax is nice for subscripts because it integrates well with shortcodes (e.g., we can use it inside shortcode parameters, which we can't do with a shortcode-based subscript).

But I do not think MathJax is suitable for a "general purpose" subscript: the font is a little different (mathy) and authors (unbeknownst the themselves) are writing LaTeX in Studio, which could have unintended consequences (e.g., ` ^{ 2, 3, & 4 }` is going to render fine in studio, but is going to display a MathJax error in Hugo because & is a special character in LaTeX):

Screen Shot 2022-09-13 at 11 35 42 AM


So:
- I do think MathJax is a good solution **for math and chemistry**. 
- but a subscript / superscript button that actually renders MathJax is a "general purpose" subscript button. There's no hint to authors that this is "for math" and "for chemistry"

@pdpinch Re
> However, it would be helpful for most course authors if there were still a simple way to write a subscript or a superscript without having to know Latex or MathML.

Would we be open to a general-purpose LaTeX implementation (which would support subscripts / superscripts, of course)? Two options
- [ckeditor5-math](https://github.com/isaul32/ckeditor5-math#screenshots): probably low-effort, but requires writing TeX
- Some customized version of the above using a WYSIWYG latex editor like [Mathlive](https://cortexjs.io/mathlive/demo/)

static/js/lib/ckeditor/plugins/SubscriptSyntaxPlugin.tsx
Wassaf-Shahzad: done

pdpinch:
@ChristopherChudzicki did you say there was a "text" mode in LaTeX/MathJax that would render strings like 4th differently?

Wassaf-Shahzad:
@ChristopherChudzicki @pdpinch 
Continues working on this PR and ran into a couple of issues,
- The way I was rendering translating the <sub> tags to markdown did not work in ocw-hugo-themes  since they were rendering as `E=mc\( _2 \)`
- To remedy this situation, I redesigned the turndown rule to actually include the parents tags content as well into the markdown but that causes the following issues
    - You cant really filter the <sub> nodes on the bases of the parent node.
    - One <p> tag can have multiple <sub> tags
    - We only need the sibling #text node to the <sub> node and but currently we get the whole <p> tags text content
- To resolve the above issues tried to create a custom Subscript button plugin which could add subscripts in a form we could easily filter but that will result is the same amount of effort or more as adding the math plugin.

I think we should move ahead with adding the math plugin rather then customising  the subscripts plugins.

ChristopherChudzicki:
Re

> I think we should move ahead with adding the math plugin rather then customising the subscripts plugins.

My opinion is:
- we should move ahead with the math plugin; it should **only be used for math + chemistry**
- we may also want to move ahead withs text (non-math) subscripts/superscripts. Whatever approach we use for that should not use mathjax / latex. 

for text-based subscripts/superscripts, I think this is promising: https://github.com/mitodl/ocw-studio/issues/1061#issuecomment-1262837136 *@Wassaf-Shahzad I initially forgot to tag you on that comment; sorry!*

ChristopherChudzicki:
Closing this since subscripts/superscripts were released in https://github.com/mitodl/ocw-studio/pull/1631

Issue #1486, opened 2022-09-20
Disappearing Tables


ChristopherChudzicki:
We initialize our HTML --> Markdown editor once globally, and if it is initialized in an area that does not support table editing, the table editing plugin never gets added. This bug is all about "on what page did the converter get initialized", which is why we've had trouble reproducing it.

### Steps to Reproduce
1. Reload OCW studio.
2. Visit a page that supports markdown editing but does NOT support tables. For example, a resource list. 

    E.g., https://ocw-studio-rc.odl.mit.edu/sites/chris-test/type/resource-list/edit/70093336-a677-47ce-806f-355bfd7c83f4/
3. Close the resource list and navigate (without reloading) to edit a page containing a table

    E.g., in that same site, pages --> "Tables!"
4. Make any trivial edit to the page markdown.
5. Save, and reopen the page you just edited.

### Expected Behavior

Tables still exist

### Actual Behavior

Tables are gone.

pdpinch:
😳🤦🏻‍♂️

ChristopherChudzicki:
Well, comfortingly I was wrong about it being "we" who initialized one global html -> md converter.

It's [@ckeditor/ckeditor-markdown-gfm](https://www.npmjs.com/package/@ckeditor/ckeditor5-markdown-gfm) initalizes a single turndown instance for all of their editors. Which doesn't seem great, but isn't really a problem for them, since their public API doesn't support changing the turndown instance / configuring markdown at all. We've hacked that a bit.

pdpinch:
I believe this was fixed by #1490

PR #1490, opened 2022-09-20
fix table bug ... emulate separate turndown instances

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#1486 

#### What's this PR do?
This PR fixes our issue with disappearing tables.

**Do this on RC:**
1. Create a page with a table.
2. Reload studio, and start creating a new Resource List. (You do not need to actually create it, just load the form.)
3. Close the resource list and navigate to your table page.
4. Make a trivial edit to the markdown, and save.
5. Re-open the page you just edited. The table should be gone.

Repeat ☝️ on this branch, and check that the table is still there.

### DIFF
--- a/static/js/lib/ckeditor/plugins/Markdown.test.ts
+++ b/static/js/lib/ckeditor/plugins/Markdown.test.ts
@@ -1,24 +1,20 @@
 import Markdown, { MarkdownDataProcessor } from "./Markdown"
 
-import { createTestEditor, markdownTest } from "./test_util"
-import { turndownService } from "../turndown"
+import { createTestEditor, getConverters, markdownTest } from "./test_util"
 
 import { TEST_MARKDOWN, TEST_HTML } from "../../../test_constants"
-
-const getEditor = createTestEditor([Markdown])
+import MarkdownSyntaxPlugin from "./MarkdownSyntaxPlugin"
+import { ShowdownExtension } from "showdown"
+import { TurndownRule } from "../../../types/ckeditor_markdown"
 
 describe("Markdown CKEditor plugin", () => {
+  const getEditor = createTestEditor([Markdown])
+
   it("should have a name", () => {
     expect(Markdown.pluginName).toBe("Markdown")
   })
 
   describe("basic Markdown support", () => {
-    it("should set custom rules flag after instantiation", async () => {
-      await getEditor("")
-      // @ts-ignore
-      expect(turndownService._customRulesSet).toBeTruthy()
-    })
-
     it("should set editor.data.processor", async () => {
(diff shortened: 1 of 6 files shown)


static/js/lib/ckeditor/plugins/Markdown.test.ts
```diff
@@ -30,3 +26,71 @@ describe("Markdown CKEditor plugin", () => {
...
+  test("Separate editors can use conflciting turndown plugins", async () => {
+    const catEditor = await getCatEditor()
+    const dogEditor = await getDogEditor()
+    const cat = getConverters(catEditor)
+    const dog = getConverters(dogEditor)
+
+    const paragraph = "<p>1234567890</p>"
+
+    expect(cat.html2md(paragraph)).toBe("meowmeowmeow!")
+    expect(dog.html2md(paragraph)).toBe("woofwoofwoof!")
```
ChristopherChudzicki, line 94: This test would fail on master. The dog markdown would be "meowmeowmeow!". Poor dog.

static/js/lib/ckeditor/plugins/Markdown.ts
```diff
@@ -132,6 +132,25 @@ export default class Markdown extends MarkdownConfigPlugin {
...
+   */
+  turndown = (html: string) => {
+    try {
+      turndownService.rules.array = this.turndownRules
+      return turndownService.turndown(html)
+    } finally {
+      turndownService.rules.array = BASE_TURNDOWN_RULES
+    }
+  }
+
```
ChristopherChudzicki, line 153: It's not entirely clear to me that overriding `rules.array` in this manner is part of Turndown's public API. But we're already doing it https://github.com/mitodl/ocw-studio/blob/8d992c7f28d051e8235636a57a41c4f096db09ac/static/js/lib/ckeditor/turndown.ts#L26

gumaerc (approved):
:+1:

static/js/lib/ckeditor/plugins/Markdown.ts
gumaerc: The functions in Turndown for adding rules manipulate the `rules` array, so I don't see any issues with directly manipulating it ourselves.

ChristopherChudzicki: Yeah, but it always makes me nervous... Does Turndown consider this an implementation detail that might change in a minor release? Whereas if they change how `addRule` behaves, that's clearly a breaking change for their public API.

But... if it idid break in a minor release, our tests would catch it.

PR #1494, opened 2022-09-21
Provide a more helpful error log message when something goes wrong w/git authentication

### LINKED ISSUE #1493: Obtuse error when git app authentication fails
### Steps to Reproduce

Edit and/or publish a site when the git app credential settings are bad for whatever reason (incorrectly formatted, expired, etc).

### Expected Behavior

A useful error message is logged indicating that git authentication via app failed.

### Actual Behavior

An obtuse TypeError is logged: "string indices must be integers" at `content_sync/apis/github.py in get_app_installation_id at line 133`

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #1493 

#### What's this PR do?
If anything goes wrong authenticating with github, raise an `ImproperlyConfigured` exception with the message "Could not initialize github app, check credentials"

- Assuming you already have  valid `GITHUB_APP_ID` and `GITHUB_APP_PRIVATE_KEY` .env values for your own github organization or the same as RC, edit and save any page/resource.  There should be no console errors and the changes should be saved to the git repo.
- Set `GITHUB_APP_PRIVATE_KEY` to "bad_key", restart containers.
- Edit any page/resource and save, there should be an `ImproperlyConfigured` exception with the message "Could not initialize github app, check credentials" in the console.
- Set `GITHUB_APP_PRIVATE_KEY` to your original valid value, but set  `GITHUB_APP_ID` to a different integer, then restart containers.
- Edit any page/resource and save, there should be an `ImproperlyConfigured` exception with the message "Could not initialize github app, check credentials" in the console.

### DIFF
--- a/content_sync/apis/github.py
+++ b/content_sync/apis/github.py
@@ -124,10 +124,12 @@ def get_app_installation_id(app: GithubIntegration) -> Optional[str]:
         "User-Agent": "PyGithub/Python",
     }
 
-    response_dict = requests.get(
+    response = requests.get(
         f"{app.base_url}/app/installations",
         headers=headers,
-    ).json()
+    )
+    response.raise_for_status()
+    response_dict = response.json()
     if response_dict:
         for git_app in response_dict:
             if git_app["app_id"] == settings.GITHUB_APP_ID:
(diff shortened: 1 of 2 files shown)


content_sync/apis/github.py
```diff
@@ -137,18 +139,23 @@ def get_app_installation_id(app: GithubIntegration) -> Optional[str]:
...
+                ),
+                **(
+                    {"base_url": settings.GIT_API_URL}
+                    if settings.GIT_API_URL is not None
+                    else {}
+                ),
+            )
+            return app.get_access_token(get_app_installation_id(app)).token
+        except Exception as exc:
+            raise ImproperlyConfigured(
```
MAbdurrehman12, line 156: I wonder if there's an exception other than bad credentials, the error message will still be to check credentials?
Though I'm not sure if this is something to be taken care of as we are already raising `ImproperlyConfigured`

MAbdurrehman12 (approved):
Looks good 👍 
Just a thought mentioned above.

content_sync/apis/github.py
mbertrand: If anything goes wrong here it's most likely going to be due to wrong settings used to authenticate with git (`GITHUB_APP_ID`, `GITHUB_APP_PRIVATE_KEY`, `GIT_API_URL`).  How about if I change the message to `Could not initialize github app, check the relevant settings`?

mbertrand:
@MAbdurrehman12 I adjusted the error message and narrowed down the types of exceptions to catch if you want to take another look.

Issue #1482, opened 2022-09-15
CKEditor LaTeX support


ChristopherChudzicki:
As an OCW course author, I'd like to be able to edit existing LaTeX in courses and create new LaTeX expressions, including mathematical expressions and chemical equations.

#### Designs and Mockups

From https://github.com/isaul32/ckeditor5-math

<img width="1471" alt="Screen Shot 2022-09-15 at 9 08 18 AM" src="https://user-images.githubusercontent.com/9010790/190412002-051fc79f-b1bd-4865-8809-0a3837e16947.png">

#### Acceptance Criteria:

- [ ] Be able to edit and add new math expressions

#### Context

_This issue is the result of discussions following #1469._

ChristopherChudzicki:
**Note:** The latest version of https://github.com/isaul32/ckeditor5-math uses CKEditor5 v34. We probably need to use a slightly older version of that plugin since we are on v31 iirc.

ChristopherChudzicki:
@Wassaf-Shahzad I looked at the mathjax plugin a bit more after your slack question. I modified the one of the examples on their github github: https://jsfiddle.net/a4zesopf/

There you can see the HTML generated by their plugin is:
```html
<p>Hello world.&nbsp;&nbsp;
    <script type="math/tex">E = mc^2</script>…And also
</p>
<script
    type="math/tex; mode=display">e^{x} = 1 + x + \frac{x^2}{2!} + \cdots</script>
<p>And some chemistry:&nbsp;
    <script type="math/tex">\ce{N2 + 3H2 -> 2 NH3}</script>
</p>
```

And the corresponding Markdown that we would want would be 
```
Hello world \(E = mc^2\) and also

\[e^{x} = 1 + x + \frac{x^2}{2!} + \cdots\]

And some chemistry \(\ce{N2 + 3H2 -> 2 NH3}\)
```

So I think we just need a markdown plugin that:
- converts `<script type="math/tex">LaTeX</script>` to `\(LaTeX\)` (w/ round parens)and vice versa
- converts `<script type="math/tex; mode=display">LaTeX</script>` to `\[LaTeX\]` (w/ square brackets) and vice versa

ChristopherChudzicki:
Closed by #1522

PR #1512, opened 2022-09-28
Math plugin on ckeditor, inline

ChristopherChudzicki:
For discussion around #1482.

@Wassaf-Shahzad I took a stab at the markdown syntax plugin this morning, here's the result. (I only did the inline math version, not display math).

https://user-images.githubusercontent.com/9010790/192853214-b2051d15-7820-457f-b847-edf42b789984.mov


ChristopherChudzicki:
@Wassaf-Shahzad Re next steps, I think:
- Modify the markdown plugin to handle display math, too. (`\[ ... \]` in the markdown should become `<script type="math/tex; mode=display">e^{x} = 1 + x + \frac{x^2}{2!} + \cdots</script>` in the HTML.
- Write some tests for display and inline math. The other tests in that folder are a good template. I would make sure to include things like:
    - inline math at beginning/middle/end of a line,
    - display math, too
    - slashed-stuff in the LaTeX (like `\frac{1}{2}`
- any other cleanup... Do we need KaTeX? (Maybe, I'm not sure. I'm not sure.)

ChristopherChudzicki (commented):
This is working pretty well but I've noticed a few thingss that don't behave right. See below. Since Wassaf is gone next week, I may make these changes. Will check with Peter. 
**Requesting Changes:**
- Let's use mathjax 3 like OCW does
- Inequalities don't work right now; they're getting escaped by markdown. We definitely want to add some tests for this, too. 
- Ditto with ampersands. 
- Need to write a real PR description.

ChristopherChudzicki:
Closing in favor of #1522

Issue #1514, opened 2022-09-29
ocw-www fails to publish on RC


gumaerc:
### Steps to Reproduce

Trigger a publish of `ocw-www` in QA Concourse: https://cicd-qa.odl.mit.edu/teams/ocw/pipelines/live?vars.site=%22ocw-www%22

### Expected Behavior

The build completes successfully

### Actual Behavior

The `copy-s3-buckets` step times out

PR #1429, opened 2022-07-29
alternate theme rendering - mass build pipeline

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested
  - [ ] Changes have been manually tested

#### What are the relevant tickets?
Closes https://github.com/mitodl/ocw-studio/issues/1428

#### What's this PR do?
This PR adds 3 optional arugments to the `upsert_mass_build_pipeline` management command:

 - `--prefix` - Adds an optional prefix to the path of sites deployed by the mass build pipeline
 - `--themes-branch` - Adds an optional override for the branch of `ocw-hugo-themes` to be used when building sites in the mass build pipeline
 - `--projects-branch` - Adds an optional override for the branch of `ocw-hugo-projects` to be used when building sites in the mass build pipeline

Each unique configuration will create a new set of pipelines for `draft` and `live` in the Concourse `mass-build-sites` group.  The only set of pipelines with automatic triggering enabled, however, will be the pipeline with the default branches of `ocw-hugo-themes` and `ocw-hugo-projects` matching `settings.GITHUB_WEBHOOK_BRANCH`.  Other mass build pipelines will need to be triggered manually.

#### How should this be manually tested?
 - Testing this PR will require setting up Concourse and Minio containers locally as well as a Github test organization, which only requires that you set some environment variables before running `docker-compose up`.  Check the readme for details.  You will also want to probably have a recent backup of the production database restored to your local instance by doing the following (replace `PROD_DB_URL` with the URL of the production database, `TODAY_DATE` with today's date):
```
pg_dump PROD_DB_URL > ./dumps/prod_TODAY_DATE.sql
docker-compose rm -sv db
docker-compose up -d db
docker-compose exec -T -e PGPASSWORD=postgres db psql -h db -U postgres < ./dumps/prod_TODAY_DATE.sql
docker-compose exec web ./manage.py createsuperuser
```
 -  Once your instance of `ocw-studio` is running, run the following to push up the theme assets pipeline and `ocw-www` pipeline:
```
docker-compose exec web ./manage.py upsert_theme_assets_pipeline
docker-compose exec web ./manage.py backpopulate_pipelines --filter ocw-www
```
- Once this is complete, browse to http://concourse:8080, login with test / test and go to the theme assets pipeline and trigger a build
- Once the theme assets build is complete, go to the live `ocw-www` pipeline and trigger a build
- Once both of these are complete, verify that you see the OCW home page at http://localhost:8045
[...]

PR #1515, opened 2022-09-29
increase timeout on copy-s3-buckets

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1514

#### What's this PR do?
Recently we added the ability in PR's related to https://github.com/mitodl/ocw-studio/pull/1429 to upsert additional `mass-build-sites` pipelines to the Concourse instance attached to `ocw-studio` with different prefixes, essentially creating an entire other copy of every site at that prefix. The purpose of this functionality is testing alternate theme rendering with all sites, like `course-v2` and `offline`. This creates a lot of extra files in the s3 bucket. When publishing `ocw-www`, the site is published to the root of the bucket. The S3 sync that pushes the files up now has to deal with scanning all of the extra files created by essentially duplicating the entire site, significantly increasing the sync time. This PR simply increases the timeout on teh `copy-s3-buckets` step from 20 minutes to 40 minutes. This is a stopgap to make sure `ocw-www` will publish in RC. In the long term we should come up with a better solution because `ocw-www` should not take as long to publish as it does. The amount of instructor static JSON combined with the fact that it's syncing to the root makes for a very long sync time in general, and an alternate strategy for `ocw-www` should likely be pursued.

This needs to be tested on RC.

### DIFF
--- a/content_sync/pipelines/definitions/concourse/site-pipeline.yml
+++ b/content_sync/pipelines/definitions/concourse/site-pipeline.yml
@@ -407,7 +407,7 @@ jobs:
                 text: "User Aborted during build-course-task : ((pipeline_name))/((site-name))"
             # END NON-DEV
       - task: copy-s3-buckets
-        timeout: 20m
+        timeout: 40m
         attempts: 3
         # START DEV-ONLY
         params:


mbertrand (approved):
👍

Issue #1495, opened 2022-09-21
Extract metadata when importing resources


pdpinch:
As a course author, it would save me time and reduce errors if studio would extract metadata from files when they are imported. In my authoring, I already add this metadata to the PDF files before uploading to gdrive. It's duplicate work to have to enter that metadata again into ocw-studio. 

All of the PDFs added to OCW in the past few years have had "title" and "author" metadata added to them. For example: https://ocw.mit.edu/courses/6-006-introduction-to-algorithms-spring-2020/resources/mit6_006s20_lec1/

The metadata can be viewed by opening the file in Acrobat Reader and typing command-D (on a Mac). 

<img width="724" alt="image (22)" src="https://user-images.githubusercontent.com/430126/192044353-a660d6de-6805-4ac6-9549-756a247f636f.png">

#### Acceptance Criteria:

- [ ] research options for extracting metadata from PDFs, our most common file format
- [ ] when creating new resources from files on gdrive, extract the metadata and update it in studio
- [ ] when updating existing resources from files on gdrive, extract the metadata and replace values in studio

#### Out of Scope

- Adding new metadata fields is out of scope. I don't want to add fields without knowing that they will be used, and for what.

Ferdi:
* What are some metadata that we'd import from the file ?
* We built studio as the place to enter metadata.  Is there a reason they like to enter metadata by editing the pdf file instead ?

pdpinch:
> What are some metadata that we'd import from the file ?

- Title, which makes sense to me. 
- The OCW team also adds author, and in Plone that was mapped to contributor. No one has been able to tell me yet what that metadata is used for. Maybe archiving. Possibly a legal obligation when the document author isn't the instructor. 
- similarly, there is a field for language (e.g. `us-en`), but I haven't seen it used anywhere yet. 

> Is there a reason they like to enter metadata by editing the pdf file instead ?

I asked this question. Here was the response (lightly edited for clarity): 

> Right now many of us use the PDF Wizard [a tool in Adobe Acrobat Pro] to [edit the metadata in] multiple files at the same time, which is much faster than opening each individual resource to add metadata. There is still some cutting & pasting involved, but it is very efficient.

> Could it function so that when you add/update the metadata in Studio and it is DIFFERENT (like an update or correction) than the existing metadata in the file OR if there is no metadata in the file, it makes the update?

Since I wasn't sure what tooling was available, I thought we would start with reading metadata and then consider writing metadata in the future.

[...]

PR #1516, opened 2022-09-30
Pt/read title from pdf metadata

pt2302:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1495.

#### What's this PR do?
If a PDF is uploaded to Google Drive and synced, and if the PDF contains `Title` in its metadata, the update in this PR populates the `Title` field in OCW Studio. If an existing resource has its title changed (and replaced in Google Drive), then its `Title` field will be updated.

Upload a test PDF via Google Drive and sync to Studio. The `Title` in metadata can be set via Adobe Acrobat or via the `exiftool` command line tool (available via most package managers depending on OS, including `apt` and `homebrew`). For example, 
```
exiftool -Title="TestTitle" mypdf.pdf
``` 
Then, change the `Title` field to something else, for example,
```
exiftool -Title="NewTestTitle" mypdf.pdf
``` 
replace the resource in Google Drive, sync again, and check that the resource's `Title` has been updated in OCW Studio.

### DIFF
--- a/gdrive_sync/api.py
+++ b/gdrive_sync/api.py
@@ -342,12 +342,12 @@ def walk_gdrive_folder(folder_id: str, fields: str) -> Iterable[Dict]:
 
 def get_pdf_title(drive_file: DriveFile) -> str:
     """Get the title of a PDF from its metadata, if available"""
-    pdf_file = io.BytesIO(GDriveStreamReader(drive_file).read())
-    pdf_reader = PyPDF2.PdfReader(pdf_file)
-    pdf_metadata = pdf_reader.metadata
-    if "/Title" in pdf_metadata and pdf_metadata["/Title"] != "":
-        return pdf_metadata["/Title"]
-    return drive_file.name
+    with io.BytesIO(GDriveStreamReader(drive_file).read()) as pdf_file:
+        pdf_reader = PyPDF2.PdfReader(pdf_file)
+        pdf_metadata = pdf_reader.metadata
+        if "/Title" in pdf_metadata and pdf_metadata["/Title"] != "":
+            return pdf_metadata["/Title"]
+        return drive_file.name
 
 
 @transaction.atomic
--- a/gdrive_sync/api.py
+++ b/gdrive_sync/api.py
@@ -6,6 +6,7 @@ import os
 from datetime import datetime
 from typing import Dict, Iterable, Optional
 
+import PyPDF2
 from botocore.exceptions import ClientError
 from django.conf import settings
 from django.db import transaction
@@ -75,7 +76,7 @@ class GDriveStreamReader:
 
 
 def get_drive_service() -> Resource:
-    """ Return a Google Drive service Resource"""
+    """Return a Google Drive service Resource"""
     key = json.loads(settings.DRIVE_SERVICE_ACCOUNT_CREDS)
     creds = ServiceAccountCredentials.from_service_account_info(
         key, scopes=["https://www.googleapis.com/auth/drive"]
@@ -116,7 +117,7 @@ def query_files(query: str, fields: str) -> Iterable[Dict]:
 
 
 def get_parent_tree(parents):
-    """ Return a list of parent folders """
+    """Return a list of parent folders"""
     service = get_drive_service()
     tree = []  # Result
     while True:
@@ -199,7 +200,7 @@ def process_file_result(
 
 @retry_on_failure
 def stream_to_s3(drive_file: DriveFile):
-    """ Stream a Google Drive file to S3 """
+    """Stream a Google Drive file to S3"""
     try:
         s3 = get_boto3_resource("s3")
         bucket_name = settings.AWS_STORAGE_BUCKET_NAME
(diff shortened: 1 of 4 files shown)


Ferdi:
Can you try to run some experiment about the performance penalty we are introducing here.  How much more time it takes to sync 10 pdf files from gdrive. How about 100 ?

mbertrand (changes requested):

gdrive_sync/api.py
```diff
@@ -383,6 +392,12 @@ def create_gdrive_resource_content(drive_file: DriveFile):
             )
         else:
             resource.file = drive_file.s3_key
+            if extension.lower() == ".pdf":
+                pdf_file = io.BytesIO(GDriveStreamReader(drive_file).read())
+                pdf_reader = PyPDF2.PdfReader(pdf_file)
+                pdf_metadata = pdf_reader.metadata
+                if "/Title" in pdf_metadata and pdf_metadata["/Title"] != "":
+                    resource.title = pdf_metadata["/Title"]
```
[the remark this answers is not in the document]
pt2302: Updated in e0e49b863001e22325a4d890580dec2245d92c50.

pdpinch:
> My assumption is that the file should be synced anyway even if it isn't a valid pdf, but it might be worth checking with Ferdi and/or Peter about that.

I disagree. If the extension says .pdf and it's not a PDF, then the course author needs to know, and do something to correct it. Can we fail in a way that the course author knows what happened?

mbertrand:
> Can we fail in a way that the course author knows what happened?

Currently, if you click the "Sync error details" link, it will show "Could not create a resource from google drive file fakepdf.pdf" - so the user will at least know the file which failed, but it won't explain why.


Screen Shot 2022-10-04 at 2 18 23 PM

pt2302:
This is a bash script to create 100 PDFs that have different title metadata, where the command line argument to this script is the PDF to generate copies of (for example, `changetitle.sh sample_file.pdf`).  I will upload the files that this generates to Google Drive and benchmark syncing for different file sizes with vs. without reading in the title metadata.

`changetitle.sh`:
```
#!/bin/bash
for i in {0..100}
do
  exiftool -Title="TestTitle_"$i $1
  cp $1 "${1%%.*}_$i.pdf"
done
```

pt2302:
Ran the following timing test by adding 
```start_time = timeit.default_timer()``` 
and 
```log.exception("TIME FOR SYNC: %s", timeit.default_timer() - start_time)``` 
to the beginning and end of the following function https://github.com/mitodl/ocw-studio/blob/9be645c08d65c615d87a734a59df798ec819ae68/gdrive_sync/api.py#L343 and processed the resulting Docker log with this Python script:
```
FIN = open("metadata_log.txt")
times = []
line = FIN.readline()
while line:
  line = line.split("TIME FOR SYNC: ")
  if len(line) > 1:
    times.append(float(line[-1]))
  line = FIN.readline()
FIN.close()
```
The results for creating a resource from 100 sample 242 KB PDFs, in seconds, are below (per file):
Version | Average | Std. Dev. | Min. | Max.
--- | --- | --- | --- | --- | 
Reading Metadata |  1.56 | 0.27 | 1.08 | 2.11
Not Reading Metadata | 0.37 | 0.03 | 0.32 | 0.64

Note that this is for 100 copies of essentially the same file (just with slightly different metadata titles).

So there's definitely some performance overhead for this. For 100 PDF files of this size, that's a total of about 2 minutes longer to complete the sync.

File: gdrive_sync/api.py [hunk 2]
```diff
@@ -339,17 +340,27 @@ def walk_gdrive_folder(folder_id: str, fields: str) -> Iterable[Dict]:
                 yield sub_result
 
 
+def get_pdf_title(drive_file: DriveFile) -> str:
+    """Get the title of a PDF from its metadata, if available"""
+    pdf_file = io.BytesIO(GDriveStreamReader(drive_file).read())
```
mbertrand, line 345: I used `docker stats` to monitor memory usage on celery, and synced 40 pdfs (22 MB each), results are:

with current code: memory usage increased from ~500 MB to 2 GB
with no pdf reading: from ~500 MB to 850 MB
with the code change below: from ~500 MB to 1.4 GB

```python
    with io.BytesIO(GDriveStreamReader(drive_file).read()) as pdf_file:
        pdf_reader = PyPDF2.PdfReader(pdf_file)
        pdf_metadata = pdf_reader.metadata
        pdf_file.close()
        if "/Title" in pdf_metadata and pdf_metadata["/Title"] != "":
            return pdf_metadata["/Title"]
        return drive_file.name
```

1.4 GB. is still high, but better than 2 GB, so I think it might be worth making the change above.

mbertrand (changes requested):
Works well, but docker memory usage is high, a small code change seemed to help lower that a bit.

gdrive_sync/api.py
mbertrand: just tweaked the code snippet above a bit

mbertrand (approved):
👍

PR #1520, opened 2022-10-06
Replace `@ts-ignore` with `@ts-expect-error`, and remove most of them

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Toward https://github.com/mitodl/eslint-config/issues/14, updating our eslint config for typescript

#### What's this PR do?
This PR:
1. Replaces all `@ts-ignore` with `@ts-expect-error`. The latter has the advantage that if it is used unnecessarily, TS will yell at you.
2. Additionally, I made some fairly minor code changes to reduce the number of `@ts-expect-error`s needed. For the most part, these were in test files, things like
    - using `const thing = jest.mocked(originalThing)` which does nothing but change the object's type to a jest mock (with signature matching `originalThing`).
    - changed a bunch of enzyme calls like `wrapper.find("StringComponentName")` to `wrapper.find(ComponentClass)`. That way TS knows the types of props, etc
    - Adding `as const` to our test case arrays (or rearranging them to objects)
3. I updated our CI's node version to 16.15, to match docker. My motivation was that CI said [Array.prototype.at](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at) was not a function, though I ended up switching to `lodash.last` since `at` only has 90% browser support.

    But, still seems good for CI to match docker.

**All in all:** We had 176 ignore/expect-errors. Now we have 44 expect-errors.

There should be no behavior change, but if you want, check ... that Studio still works.

⚠️ ⚠️ ⚠️ Strongly suggest reviewing this commit-by-commit. https://github.com/mitodl/ocw-studio/pull/1520/commits/d79aa320418286a1a581d4ec08eb7fbde76afa4d is the only commit that could affect app behavior; the other commits are all about linting or tests.

### DIFF
--- a/static/js/lib/ckeditor/plugins/LegacyShortcodes.ts
+++ b/static/js/lib/ckeditor/plugins/LegacyShortcodes.ts
@@ -1,5 +1,5 @@
 import { ShowdownExtension } from "showdown"
-import Plugin from "@ckeditor/ckeditor5-core/src/plugin"
+import CKPlugin from "@ckeditor/ckeditor5-core/src/plugin"
 import { editor } from "@ckeditor/ckeditor5-core"
 
 import MarkdownSyntaxPlugin from "./MarkdownSyntaxPlugin"
@@ -79,7 +79,7 @@ class LegacyShortcodeSyntax extends MarkdownSyntaxPlugin {
 const shortcodeModelName = (shortcode: string) =>
   `legacy-shortcode-${shortcode}`
 
-class LegacyShortcodeEditing extends Plugin {
+class LegacyShortcodeEditing extends CKPlugin {
   static get pluginName(): string {
     return "LegacyShortcodeEditing"
   }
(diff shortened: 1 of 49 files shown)


static/js/lib/ckeditor/plugins/LegacyShortcodes.ts
```diff
@@ -185,13 +185,8 @@ class LegacyShortcodeEditing extends Plugin {
   }
 }
 
-export default class LegacyShortcodes extends Plugin {
-  static get requires(): Plugin[] {
```
ChristopherChudzicki, line 189: I believe TS was treating this as the deprecated Plugin interface. https://developer.mozilla.org/en-US/docs/Web/API/Plugin

pt2302 (approved):
With the latest commit (removing TODOs), everything looks good to me!

Issue #1527, opened 2022-10-14
Fix up the resource filenames in a couple courses


pdpinch:
### Steps to Reproduce

1. Compare the file names in google drive to the published site for

course | gdrive ƒ | draft site 
-- | -- | --
[17.810](https://ocw-studio.odl.mit.edu/sites/game-theory-and-political-theory/type/resource/) | [17.810-spring-2021](https://drive.google.com/drive/folders/1J8UeFJtpjn_2H8rMiGNmAoPnl_BiqY3K) | [game-theory-and-political-theory]()
[IDS.333](https://draft.ocw.mit.edu/courses/17-810-game-theory-and-political-theory-spring-2021) | [ids.333-fall-2021](https://drive.google.com/drive/folders/1yge0_BjLyy49vO9bdBg9Ro70ebtyBMxY) | [Risk and Decision Analysis](https://draft.ocw.mit.edu/courses/ids-333-risk-and-decision-analysis-fall-2021)

### Expected Behavior

- file names should be the same on gdrive and on the published site. 

### Actual Behavior

- additional characters are added to file names, possibly to avoid a name collision. 
- This is problematic for files that end with `lec1` for "lecture 1" and they get changed to `lec12`

### Related Issues

- https://github.com/mitodl/hq/discussions/268
- In the short term, fixing up the data would be best, so these courses can be published properly
- In the long run, we should consider changing the file naming behavior

pt2302:
Specific resources with this issue:
For https://ocw-studio.odl.mit.edu/sites/game-theory-and-political-theory:
1. https://ocw-studio.odl.mit.edu/sites/game-theory-and-political-theory/type/resource/edit/97b5960e-068c-483c-b25d-ec0820cd7cb4
2. https://ocw-studio.odl.mit.edu/sites/game-theory-and-political-theory/type/resource/edit/10f2771d-44f2-4f74-bec0-54321c942d66
3. https://ocw-studio.odl.mit.edu/sites/game-theory-and-political-theory/type/resource/edit/36f3cf8f-fb85-43bf-b99a-b06287ba41a4
4. https://ocw-studio.odl.mit.edu/sites/game-theory-and-political-theory/type/resource/edit/7d65f82b-b789-469f-b1e1-6b65f3dee49f
5. https://ocw-studio.odl.mit.edu/sites/game-theory-and-political-theory/type/resource/edit/2831c124-3787-498f-9927-631e24912d81
[...]

PR #1538, opened 2022-10-24
Management Command for Renaming Files on S3

pt2302:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Once this command is run on the list of relevant resources, this PR closes https://github.com/mitodl/ocw-studio/issues/1527.

#### What's this PR do?
Given the `text_id` of a resource (which can be obtained either from Django admin or from the URL for the Edit Resource Drawer), this adds a management command for renaming the associated file on S3.

For example, given this resource:
`https://ocw-studio.odl.mit.edu/sites/game-theory-and-political-theory/type/resource/edit/97b5960e-068c-483c-b25d-ec0820cd7cb4`, if we want to rename the associated file from `mit17_810s21_lec72.pdf` to `mit17_810s21_lec7.pdf`, we would run:

`heroku run -a ocw-studio python manage.py  rename_file_by_text_id --text_id 97b5960e-068c-483c-b25d-ec0820cd7cb4 --new_filename mit17_810s21_lec7.pdf`.

This creates a copy of the file `mit17_810s21_lec72.pdf` with the new name `mit17_810s21_lec7.pdf` and then deletes `mit17_810s21_lec72.pdf` (since there is no direct renaming command on S3).

This can be tested locally or on RC, by specifying the appropriate `text_id` for a resource in the Django database. Some things to note:
- The `new_filename` argument is passed to `slugify` before the renaming operation, so something like `MiT17_810s.PDf` will become `mit17_810s.pdf`.
- This handles multiple extensions, so a `new_filename` argument like `MYFILE.ABC.xyz` will become `myfile.abc.xyz`.
- The change in the filename will also be reflected in the Edit Resource drawer (filename was added to the Edit Resource drawer in https://github.com/mitodl/ocw-studio/pull/1406).


pt2302:
I suppose that a more sophisticated version of this management command could take, for example, a JSON of `text_id`:`new_filename` pairs and update all of them at once. This may be useful if this type of update is anticipated to be run more frequently.

websites/management/commands/rename_by_text_id.py
```diff
@@ -0,0 +1,55 @@
...
+        df_path = df.s3_key.split("/")
+        df_path[-1] = new_filename
+        new_key = "/".join(df_path)
+        s3.Object(settings.AWS_STORAGE_BUCKET_NAME, new_key).copy_from(
+            CopySource=settings.AWS_STORAGE_BUCKET_NAME + "/" + df.s3_key
+        )
+        s3.Object(settings.AWS_STORAGE_BUCKET_NAME, df.s3_key).delete()
+        df.s3_key = new_key
+        obj.file = new_key
+        obj.filename = get_dirpath_and_filename(new_filename)[1]
```
mbertrand, line 53: I got an integrity error when running this a certain way, which I think might match the scenario under which the problem it's trying to fix happened in production:

- Upload a file to google drive ("word_sample_1.doc")
- Sync, creating `Gdrive` and `WebsiteContent` objects.
- Delete the file from google drive, then upload it again.
- Sync again.  Now I have two resources, with filenames "word_sample_1.doc" (original) and "word_sample_12.doc" (reuploaded)

I created links to both resources in a page.  I also added links to both resources in the site menu.

Run `docker-compose run web python manage.py rename_by_text_id --text_id  <text_id of reuploaded content> --new_filename word_sample_1.doc`

Because the original resource still exists, and has a `filename` of `word_sample_1` which must be unique within each site, I get an `IntegrityError`.

```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "unique_page_content_destination"
DETAIL:  Key (website_id, dirpath, filename)=(538d4c8b-3914-4685-95b0-b5fa48db9c5f, content/resources, word_sample_1) already exists.
```

So either the rename can't be the same as that of original (which I think might defeat the purpose?) or the old `WebsiteContent` object also needs a rename, or the old one needs to be deleted, which requires checking for links/embeds and changing those to the new resource instead.  Might be good to get some feedback from Peter on what's the most ideal here from the user perspective.

I tried again using a slightly different rename (`word_sample__1`) just to proceed further, republished, and noticed another issue with site menu.  [The menu is still pointing to the old url](https://github.com/mbtestorg2022/course9/blob/main/config/_default/menus.yaml#L12).  However, the link within the page is still pointing to the new, correct url.  So I think there might need to be a check added to see if the resource is linked to from the site menu, and if so update it.

mbertrand (changes requested):

websites/management/commands/rename_by_text_id.py
pt2302: Possible approach to this:
- check if a WebsiteContent object with the `new_filename` exists in that course.
- if it does not, proceed as currently in the PR.
- if it does, check whether it exists in any course content (links or embedded content).
    - if it exists in any course content, print the name of the course content and abort
    - else: print a warning that the old object is being deleted and then proceed as currently in the PR.

pt2302: Note that there should be at most 1 existing `WebsiteContent` object with this filename already in the database, because otherwise there would have already been an `IntegrityError`, as above.

websites/management/commands/rename_by_text_id.py
```diff
@@ -0,0 +1,55 @@
...
+        df.save()
+        obj.save()
```
mbertrand, line 55: The bulk of this `handle` function code could be put in a separate function, (ie, `gdrive_sync.api.rename_file`) which would allow you to add 1 or more unit tests to make sure it does what's expected.  Also would be good to provide some feedback at the end.

```python
def handle(self, *args, **options):
    rename_file(options["text_id"], options["new_filename"])
    sys.stdout.write("File successfully renamed.\n")
```

mbertrand (changes requested):

websites/management/commands/rename_by_text_id.py
pdpinch: Thanks Paul. I think deleting the resource fits the use case. Do we have a method for deleting resources? I've gotten requests for that as well, but I haven't been prioritizing them.

pt2302: We can directly delete the `WebsiteContent` object `obj` with `obj.delete()`. The file on S3 would not need to be deleted, since it's being written over by the renaming command.

Also, just documenting that we are not going to currently support issues with broken links due to a file rename either in navigation menus or in pages, because:

- Menus should be linking to pages, not resources.
- Linking to an internal URL in a page instead of using a resource link is discouraged.

websites/management/commands/rename_by_text_id.py
```diff
@@ -0,0 +1,55 @@
...
+            "--new_filename",
+            dest="new_filename",
+            help="New filename to be associated with the WebsiteContent object",
+            required=True,
+        )
+
+    def handle(self, *args, **options):
+        obj = WebsiteContent.objects.get(text_id=options["text_id"])
+        df = DriveFile.objects.get(resource=obj)
+        s3 = boto3.resource("s3")
```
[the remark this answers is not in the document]
pt2302: Updated in 1b06530b50510c319223255519936a00280aa741.

gdrive_sync/api.py
```diff
@@ -485,3 +490,58 @@ def update_sync_status(website: Website, sync_datetime: datetime):
     website.sync_status = new_status
     website.sync_errors = (website.sync_errors or []) + errors
     website.save()
+
+
+@transaction.atomic
+def rename_file(obj_text_id, obj_new_filename):
```
mbertrand, line 496: add a unit test?

websites/management/commands/rename_by_text_id.py
```diff
@@ -0,0 +1,30 @@
...
+
+        parser.add_argument(
+            "--new_filename",
+            dest="new_filename",
+            help="New filename to be associated with the WebsiteContent object",
+            required=True,
+        )
+
+    def handle(self, *args, **options):
+        rename_file(options["text_id"], options["new_filename"])
```
mbertrand, line 30: `sys.stdout.write("File successfully renamed")`

gdrive_sync/api.py
```diff
@@ -485,3 +490,58 @@ def update_sync_status(website: Website, sync_datetime: datetime):
...
+    obj.filename = get_dirpath_and_filename(new_filename)[1]
+    df.save()
+    obj.save()
+
+    s3.Object(settings.AWS_STORAGE_BUCKET_NAME, new_key).copy_from(
+        CopySource=settings.AWS_STORAGE_BUCKET_NAME + "/" + old_key
+    )
+    s3.Object(settings.AWS_STORAGE_BUCKET_NAME, old_key).delete()
+
+    log.info("File successfully renamed.\n")
```
mbertrand, line 547: Move this line to the command as `sys.stdout.write`

gdrive_sync/api.py
```diff
@@ -485,3 +490,58 @@ def update_sync_status(website: Website, sync_datetime: datetime):
...
+        if old_obj == obj:
+            log.error("New filename is the same as the existing filename.")
+            return
+        dependencies = WebsiteContent.objects.filter(
+            Q(website=site)
+            & Q(type=CONTENT_TYPE_PAGE)
+            & Q(markdown__icontains=old_obj.text_id),
+        )
+        if dependencies:
+            log.error(
```
mbertrand, line 524: `raise Exception("Not renaming file .... etc`

mbertrand (changes requested):

websites/management/commands/rename_by_text_id.py
pt2302: Updated in 8eee92a8b1ec916f8fd918512194b6dcb7152963.

gdrive_sync/api.py
pt2302: Updated in 8eee92a8b1ec916f8fd918512194b6dcb7152963.

gdrive_sync/api.py
pt2302: Updated in d4623bd275e9e79b46059736e36cdddb7e916fe4.

gdrive_sync/api.py
pt2302: Added in 3fa3378092e68fa98a2403d76fd3fb79eeae9743.

mbertrand (approved):
Looks good! 👍

PR #1548, opened 2022-10-27
ocw hugo themes sentry dsn added 

MAbdurrehman12:
#### Pre-Flight checklist

- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested
- [ ] Settings

- None

#### What's this PR do?
- Sets `OCW_HUGO_THEMES_SENTRY_DSN` env so it can be templated into the pipelines

- Checkout this branch
- Set any value for `OCW_HUGO_THEMES_SENTRY_DSN` in your `.env` file.
- Run `backpopulate_pipelines` or `upsert_mass_build_pipeline` and verify that the env value is propagated into your local pipelines.

- This PR has a dependency on https://github.com/mitodl/ocw-hugo-themes/pull/903

### DIFF
--- a/content_sync/pipelines/definitions/concourse/mass-build-sites.yml
+++ b/content_sync/pipelines/definitions/concourse/mass-build-sites.yml
@@ -113,7 +113,7 @@ jobs:
       GIT_KEY: ((git-private-key-var))
       SITEMAP_DOMAIN: ((sitemap-domain))
       PREFIX: ((prefix))
-      OCW_HUGO_THEMES_SENTRY_DSN: ((ocw-hugo-themes-sentry-dsn))
+      SENTRY_DSN: ((ocw-hugo-themes-sentry-dsn))
       # START DEV-ONLY
       AWS_ACCESS_KEY_ID: ((minio-root-user))
       AWS_SECRET_ACCESS_KEY: ((minio-root-password))
--- a/content_sync/pipelines/definitions/concourse/mass-build-sites.yml
+++ b/content_sync/pipelines/definitions/concourse/mass-build-sites.yml
@@ -113,6 +113,7 @@ jobs:
       GIT_KEY: ((git-private-key-var))
       SITEMAP_DOMAIN: ((sitemap-domain))
       PREFIX: ((prefix))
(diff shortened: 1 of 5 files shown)


content_sync/pipelines/definitions/concourse/mass-build-sites.yml
```diff
@@ -113,6 +113,7 @@ jobs:
       GIT_KEY: ((git-private-key-var))
       SITEMAP_DOMAIN: ((sitemap-domain))
       PREFIX: ((prefix))
+      OCW_HUGO_THEMES_SENTRY_DSN: ((ocw-hugo-themes-sentry-dsn))
```
gumaerc, line 116: It's good to be specific when it comes to the Django setting as there could be other Sentry DSN's potentially in the `ocw-studio` environment, but in the pipelines I think we could just call it `SENTRY_DSN` for simplicity.

content_sync/pipelines/definitions/concourse/site-pipeline.yml
```diff
@@ -327,6 +327,7 @@ jobs:
           STATIC_API_BASE_URL: ((static-api-base-url))
           OCW_IMPORT_STARTER_SLUG: ((ocw-import-starter-slug))
           SITEMAP_DOMAIN: ((sitemap-domain))
+          OCW_HUGO_THEMES_SENTRY_DSN: ((ocw-hugo-themes-sentry-dsn))
```
gumaerc, line 330: Here too, I think we can just call the variable `SENTRY_DSN`.

gumaerc (approved):
This looks good and could be merged as-is, I just have a naming suggestion. As a reminder, before this merges you're going to want to do the following:

 - PR in `salt-ops` to add the proper values that should be assigned to this env variable in the different Heroku environments, defined in this file: https://github.com/mitodl/salt-ops/blob/main/pillar/heroku/ocw-studio.sls
 - PR in `ocw-hugo-projects` to add the env variable (whatever you decide to name it) to every `config.yaml` and `config-offline.yaml` file to the `security.funcs.getenv` section so Hugo allows you to fetch the env variable during the build. This operates on a whitelist in case you have shortcodes that return values using the environment, to reduce the risk of exposing environment variables that you don't want exposed.

Issue #1558, opened 2022-11-02
Add a slack alert when theme pipeline fails


ChristopherChudzicki:
As a developer working on OCW, I want to be alerted when pipelines fail

#### Acceptance Criteria:

- [ ] When the theme-assets pipeline fails, we should get a slack alert in slack channel `notifications-ocw-publish`

#### Similar work
@Ardiea did something similar for the individual site publishing pipelines in https://github.com/mitodl/ocw-studio/pull/1503

ChristopherChudzicki:
More context... When theme-assets pipeline finishes, it uploads a new `webpack.json` file to s3, which triggers a mass publish. So if we release ocw-hugo-themes, and theme-assets pipeline fails, a mass publish never gets triggered.

This happened in https://github.com/mitodl/ocw-hugo-themes/pull/923 (which was released about a week ago). The theme-assets pipeline originally when ocw-hugo-themes was deployed. When that failed, we did fix the pipeline, but it never got upserted and triggered. (that was my mistake). Consequently, no hugo-themes releases within the last week have triggered a mass publish.

**Suggestion:** The error message should possibly include instructions about what needs to be done, e.g., "(1) fix the pipeline definition AND (2) upsert it anew AND (3) start the job". We did (1), but apparently never did (2) and (3).

PR #1576, opened 2022-11-17
Issue slack alert when theme assets pipeline fails

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing

#1558 

#### What's this PR do?
This PR uses concourse's [job hooks](https://concourse-ci.org/hooks-example.html#hooks-example) to issue a slack alert when a job is aborted or fails. 

Two options: either don't test it locally—we can test it on rc using our actual slack channel—or test on your own slack organization.

If you would like to test locally:
1. [Create a slack workspace](https://slack.com/help/articles/206845317-Create-a-Slack-workspace) _or use a workspace you already own_
2. In the workspace you own, [create a slack app and enable incoming webhooks](https://api.slack.com/messaging/webhooks#getting_started__1.-create-a-slack-app-if-you-dont-have-one-already). Use whatever channel you want.
3. Copy the webhook URL
4. Checkout this branch, and modify `theme-assets-pipeline.yml`:
    - replace `((slack-webhook))` with your webhook url
    - remove the `START NON-DEV` and `END NON-DEV` comments surrounding the webhook stuff. (Or checkout  https://github.com/mitodl/ocw-studio/commit/a1f061567d2a75b7341ee9c2a36176c32570d393, the first commit in this two-commit PR, which does not have those comments.)
5. `docker compose run --rm web ./manage.py upsert_theme_assets_pipeline` 
6. In your local concourse ui http://concourse:8080/:
    - trigger a theme assets build
    - abort it
    - check your slack channel for notification
7. Insert some artificial error into the pipeline definition, re-upsert the pipeline, and start it. The pipeline should fail, and you should get a slack notification.

### DIFF
--- a/content_sync/pipelines/definitions/concourse/theme-assets-pipeline.yml
+++ b/content_sync/pipelines/definitions/concourse/theme-assets-pipeline.yml
@@ -1,21 +1,21 @@
 ---
 resource_types:
- #
+ # START NON-DEV
   - name: slack-alert
     type: registry-image
     source:
       repository: arbourd/concourse-slack-alert-resource
       tag: v0.15.0
-  #
+  # END NON-DEV
 resources:
-#
+  # START NON-DEV
 - name: slack-webhook
   type: slack-alert
   check_every: never
   source:
     url: ((slack-webhook))
     disabled: false
-#
+  # END NON-DEV
 - name: ocw-hugo-themes
   type: git
   source:
@@ -110,7 +110,7 @@ jobs:
           - 'Fastly-Key: ((fastly_live.api_token))'((purge_header))
           - https://api.fastly.com/service/((fastly_live.service_id))/purge/ocw-hugo-themes
   # END NON-DEV
-  #
+  # START NON-DEV
   on_failure:
     try:
       do:
@@ -133,4 +133,4 @@ jobs:
             User aborted while building theme assets.
             
             Append `((atc-search-params))` to the url below for more details.
-#
+   # END NON-DEV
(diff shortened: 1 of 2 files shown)


content_sync/pipelines/definitions/concourse/theme-assets-pipeline.yml
```diff
@@ -94,3 +110,27 @@ jobs:
...
+    try:
+      do:
+      - put: slack-webhook
+        timeout: 1m
+        params:
+          alert_type: failed
+          text: |
+            Failed to build theme assets.
+            
+            Append `((atc-search-params))` to the url below for more details.
```
ChristopherChudzicki, line 124: This is...maybe not worth it.

The slack alert includes a URL, cool!

Screen Shot 2022-11-17 at 4 20 48 PM

But our pipeline names include some URL search parameters at the end, e.g.,
```
http://concourse:8080/teams/main/pipelines/ocw-theme-assets/jobs/build-theme-assets/builds/4?vars.branch=%22main%22
```
So the slack notification's URL doesn't actually work, which is annoying.

There doesn't appear to be a way to tell `slack-alert` resource to add query params. https://www.google.com/search?q=concourse+slack+alert&oq=concourse+slack+&aqs=chrome.0.69i59l2j69i57j0i512l2j69i60l3.1603j0j7&sourceid=chrome&ie=UTF-8

gumaerc: These are called ["instance vars"](https://concourse-ci.org/instanced-pipelines.html) and yea, it doesn't seem that `slack-alert` supports them. It doesn't seem like there's a whole lot of activity over there, but I think if we think this important enough we could put up a PR there to add support for them, or at least create an issue.

gumaerc (approved):
Screenshot_20221128_153433

I did end up testing this locally with a private Slack org and everything works as expected. LGTM! :+1:

PR #1585, opened 2022-11-29
change slack-webhook to slack-url var

ChristopherChudzicki:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

#1558 

#### What's this PR do?
This PR is a followup to #1576. In that PR, I made a mistake naming the variable. Now the variable is `slack-url`, just like in the site pipelines.

It should be tested on RC: All other testing of this functionality was done with #1576. This just changes how concourse gets the var from vault.

### DIFF
--- a/content_sync/pipelines/definitions/concourse/theme-assets-pipeline.yml
+++ b/content_sync/pipelines/definitions/concourse/theme-assets-pipeline.yml
@@ -13,7 +13,7 @@ resources:
   type: slack-alert
   check_every: never
   source:
-    url: ((slack-webhook))
+    url: ((slack-url))
     disabled: false
   # END NON-DEV
 - name: ocw-hugo-themes


content_sync/pipelines/definitions/concourse/theme-assets-pipeline.yml
```diff
@@ -13,7 +13,7 @@ resources:
   type: slack-alert
   check_every: never
   source:
-    url: ((slack-webhook))
+    url: ((slack-url))
```
ChristopherChudzicki, line 16: cf https://github.com/mitodl/ocw-studio/blob/bc3099caf9553ace1160b05ebdc472882ff9c775/content_sync/pipelines/definitions/concourse/site-pipeline.yml#L74

gumaerc (approved):
:+1: 

I should have looked at vault / the existing pipeline that uses this before approving, but this looks correct now.

PR #1588, opened 2022-11-30
Subscripts and superscripts

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#1061 

#### What's this PR do?
This PR adds support for subscripts and superscripts in markdown via inclusion of raw `<sub>` and `<sup>` tags. Arbitrary HTML is not permitted: only HTML tags whitelisted in the markdown widget config's `allow_html` list will be permitted. Additionally, the only non-markdown tags that our CKEditor setup will generate are sub/sup, and only if the tags are allowed.

This approach turned out to not be as quite "works out of the box" as I originally thought. Two issues came up, both having to do with the fact that while MD --> HTML conversion is well specified by CommonMark, there is no specification for  HTML --> MD conversion (which we need when extracting data from CKEditor). The two issues were:

1. **`<sup>` at start of paragraph:** Markdown literally cannot handle this. HTML at the beginning of a line creates an HTML block, and the remainder of the line is parsed as raw HTML rather than Markdown. So the two lines

    ```md
    <sup>1</sup>:  This renders raw html with **asterisks**

    x<sup>1</sup> This renders as markdown with **bold**.
    ```

    end up behaving very differently.
2. **MD inside html tags**: The default behavior for Turndown (our html->md converter) when including raw HTML in markdown is to not process the content of the HTML *at all*. But HTML tags are generally allowed to contain markdown[^1] content, and it's important to convert that content to markdown.

[^1]: This is absolutely true for [markdown inline html](https://spec.commonmark.org/0.30/#raw-html). The story is a little more complicated for [markdown block html](https://spec.commonmark.org/0.30/#html-blocks).

1. Load the course-v2 config from https://github.com/mitodl/ocw-hugo-projects/pull/234 into your local studio
2. Create a page and add subscripts/superscripts. Some things that are worth trying are subscripts/superscripts
    - at the beginning of a line
    - in links and resource links
    - in tables
    - bold inside the subscript/superscript *though we do not encourage authors to do this, ckeditor will not prevent it.*
3. Check the markdown in admin panel. It should look correct.
4. Check that the markdown renders correctly in ocw-hugo-themes using branch from https://github.com/mitodl/ocw-hugo-themes/pull/1004

#### Screenshots

Screen Shot 2022-11-30 at 9 05 59 AM

#### Other Context

HQ discussion: https://github.com/mitodl/hq/discussions/303

### DIFF
--- a/static/js/lib/ckeditor/turndown.ts
+++ b/static/js/lib/ckeditor/turndown.ts
@@ -114,7 +114,7 @@ itemRule.replacement = (
 
 const BASE_TURNDOWN_RULES = [...turndownService.rules.array]
 const BASE_TURNDOWN_KEEP = [
-  // @ts-expect-error `_keep` is not part of Turndown's public API, see `resetTurndownKeep` for more.
+  // @ts-expect-error `_keep` is not part of Turndown's public API, see `resetTurndownService` for more.
   ...turndownService.rules._keep
 ]
 const BASE_TURNDOWN_KEEP_REPLACEMENT = turndownService.rules.keepReplacement
@@ -127,17 +127,12 @@ export const resetTurndownService = () => {
   turndownService.rules.array = [...BASE_TURNDOWN_RULES]
 
   /**
-   * By default, Turndown does not keep any HTML tags when converting from html -> md.
-   * We can tell turndown to keep some tags via a `turndown.keep(tags)`. Subsequent
-   * calls to `keep` ADD more tags that turndown should keep.
-   *
-   * Howevver, Turndown has no public API to remove tags from the keep list after
-   * they have been added.
-   *
-   * All of these machinations are only necessary because CKEditor's GFM plugin
-   * uses a single global turndown instance.
+   * We need to access `_keep` in order to "reset" what HTML tags Turndown will
+   * keep when converting HTML to Markdown. Turndown has a public API for
+   * adding tags to its "keep" list, but not for removing tags or resetting its
+   * list.
    */
-  // @ts-expect-error `_keep` is not part of Turndown's public API, see `resetTurndownKeep` for more.
+  // @ts-expect-error `_keep` is not part of Turndown's public API, see above
   turndownService.rules._keep = [...BASE_TURNDOWN_KEEP]
   turndownService.rules.keepReplacement = BASE_TURNDOWN_KEEP_REPLACEMENT
 }
--- a/static/js/lib/ckeditor/turndown.ts
+++ b/static/js/lib/ckeditor/turndown.ts
@@ -211,7 +211,7 @@ class TurndownHtmlHelpers {
      * To address Issue #1, we'll mark places where HTML is inserted. Later, we
      * can remove all the markings except the ones that begin a new line.
      */
-    return `<raw_html_follows/>${clone.outerHTML}`
+    return `<raw_inline_html/>${clone.outerHTML}`
   }
 
   /**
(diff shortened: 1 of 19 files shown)


static/js/lib/ckeditor/turndown.ts
```diff
@@ -111,3 +111,121 @@ itemRule.replacement = (
...
+   * tag that begins what would otherwise be a Markdown paragraph. Should
+   * be used alongside {@link TurndownHtmlHelpers.keepReplacer}.
+   *
+   * See {@link TurndownHtmlHelpers} for more details.
+   */
+  turndown = (html: string) =>
+    this.turndownInstance
+      .turndown(html)
+      .replace(/^[ ]*<raw_inline_html\/>/g, "\u200b")
+      .replace(/<raw_inline_html\/>/g, "")
```
ChristopherChudzicki, line 228: Markdown treats HTML tags at the beginning of a line and elsewhere in a line very differently: html at the beginning starts an [*html block*](https://spec.commonmark.org/0.30/#html-blocks), which disables markdown for the remainder of that line. HTML tags within the line form [inline html](https://spec.commonmark.org/0.30/#raw-html), which allows markdown syntax elsewhere in the line and within the html content.

So if we have an inline html element that would start a line, insert a zero-width space so that it doesn't actually start the line.

I am fairly happy with the whole HTML tag approach to subsup, except for this aspect. And while subscripts at the beginning of the line probably won't occur, superscripts could, e.g., for footnotes[^1].

[^1]: although there are much better ways to do footnotes in markdown, though we don't support them right now.

static/js/lib/ckeditor/turndown.ts
```diff
@@ -111,3 +111,121 @@ itemRule.replacement = (
...
+   * Convert HTML to markdown and insert a zero-width space before any HTML
+   * tag that begins what would otherwise be a Markdown paragraph. Should
+   * be used alongside {@link TurndownHtmlHelpers.keepReplacer}.
+   *
+   * See {@link TurndownHtmlHelpers} for more details.
+   */
+  turndown = (html: string) =>
+    this.turndownInstance
+      .turndown(html)
+      .replace(/^[ ]*<raw_inline><\/raw_inline>/g, "\u200b")
```
gumaerc, line 227: We've used techniques similar to this with our Turndown code in `ocw-to-hugo`, before we switched to shortcode tables. We would leave markers to later replace with pipes to handle constructing tables in markdown. This is a bit different of an issue, but the concept remains the same: leave a specific marker that can be replaced later in the process.

As far as the zero width space, I think it's probably fine as long as it doesn't introduce any practical formatting / word wrapping issues with CSS on the built site. Since it's being inserted at the beginning of the line, I don't see it being a problem. I was trying to reproduce this by putting a subscript at the beginning of a line and saving it, but I didn't even see the character in the raw markdown pushed up to Github. Either way, I don't think Hugo will render it right?

gumaerc (approved):
:+1: 

This is thoroughly tested and documented, and everything went well practically testing this in my local setup. Just a comment typo that I found as well as some comments on the "zero width space" situation.

static/js/lib/ckeditor/turndown.ts
ChristopherChudzicki: This was an error on my part and fixed in https://github.com/mitodl/ocw-studio/pull/1588/commits/1539f1942931827eff2f590466595a791efba831. The zwsp was only being inserted at the beginning of the text, not at the beginning of a line.

In the example you shared on slack (https://github.com/gumaerc-testorg/5.13-fall-2003/blob/main/content/pages/syllabus.md) ... If you had done `x_2 blah blah **foo**` I expect that `**foo**` would have shown up with literal asterisks, not bold. Now it should show bold.

gumaerc (approved):
@ChristopherChudzicki Thanks for pointing out that I should have been seeing the zero width space with subscript at the beginning of the line. The changes look good and I do see the zero width space now. As I said above, my opinion on it personally is somewhat neutral. Practically, it's not really that big of a deal. It could be confusing for someone trying to manually edit the markdown wondering why they have to backspace on certain lines twice, but it doesn't really seem like a problem formatting wise.

PR #1597, opened 2022-12-02
Revert "Subscripts and superscripts"

ChristopherChudzicki:
Reverts mitodl/ocw-studio#1588

See https://github.com/mitodl/ocw-studio/issues/1061#issuecomment-1335839283

### How should this be manually tested

I don't think you should manually test this. It's a revert to known-state. But you could check that subscripts/superscripts are not available.


gumaerc (approved):
:+1:

PR #1617, opened 2022-12-14
Subscript and Superscript support in CKEditor

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#1061 

#### What's this PR do?
This is a new version of #1588, which we reverted because of a CKEditor issue. We believe that updating from v31 to v35 of CKEditor resolved that issue.

This PR adds support for subscripts and superscripts in markdown via inclusion of raw `<sub>` and `<sup>` tags. Arbitrary HTML is not permitted: only HTML tags whitelisted in the markdown widget config's `allow_html` list will be permitted. Additionally, the only non-markdown tags that our CKEditor setup will generate are sub/sup, and only if the tags are allowed.

This approach turned out to not be as quite "works out of the box" as I originally thought. Two issues came up, both having to do with the fact that while MD --> HTML conversion is well specified by CommonMark, there is no specification for  HTML --> MD conversion (which we need when extracting data from CKEditor). The two issues were:

1. **`<sup>` at start of paragraph:** Markdown literally cannot handle this. HTML at the beginning of a line creates an HTML block, and the remainder of the line is parsed as raw HTML rather than Markdown. So the two lines

    ```md
    <sup>1</sup>:  This renders raw html with **asterisks**

    x<sup>1</sup> This renders as markdown with **bold**.
    ```

    end up behaving very differently.
2. **MD inside html tags**: The default behavior for Turndown (our html->md converter) when including raw HTML in markdown is to not process the content of the HTML *at all*. But HTML tags are generally allowed to contain markdown[^1] content, and it's important to convert that content to markdown.

[^1]: This is absolutely true for [markdown inline html](https://spec.commonmark.org/0.30/#raw-html). The story is a little more complicated for [markdown block html](https://spec.commonmark.org/0.30/#html-blocks).

1. Load the course-v2 config from https://github.com/mitodl/ocw-hugo-projects/pull/234 into your local studio
2. Create a page and add subscripts/superscripts. Some things that are worth trying are subscripts/superscripts
    - at the beginning of a line
    - in links and resource links
    - in tables
    - bold inside the subscript/superscript *though we do not encourage authors to do this, ckeditor will not prevent it.*
3. Check the markdown in admin panel. It should look correct.
4. Check that the markdown renders correctly in ocw-hugo-themes using branch from https://github.com/mitodl/ocw-hugo-themes/pull/1004
5. ⚠️ Try to reproduce  the bug shown here https://github.com/mitodl/ocw-studio/issues/1061#issuecomment-1335839283 ... you'll need to watch the dev console locally. The bug should not be reproducible now that we've updated to CKEditor 35.

#### Screenshots

Screen Shot 2022-11-30 at 9 05 59 AM

#### Other Context

HQ discussion: https://github.com/mitodl/hq/discussions/303

### DIFF
--- a/static/js/components/widgets/MarkdownEditor.test.tsx
+++ b/static/js/components/widgets/MarkdownEditor.test.tsx
@@ -15,6 +15,7 @@ import {
   ADD_RESOURCE_EMBED,
   ADD_RESOURCE_LINK,
   CKEDITOR_RESOURCE_UTILS,
+  MARKDOWN_CONFIG_KEY,
   RESOURCE_EMBED,
   RESOURCE_LINK
 } from "../../lib/ckeditor/plugins/constants"
@@ -38,7 +39,7 @@ jest.mock("@ckeditor/ckeditor5-react", () => ({
 }))
 
 const render = (props = {}) =>
-  shallow(<MarkdownEditor link={[]} embed={[]} {...props} />)
+  shallow(<MarkdownEditor allowedHtml={[]} link={[]} embed={[]} {...props} />)
 
 describe("MarkdownEditor", () => {
   let sandbox: SinonSandbox
@@ -55,7 +56,7 @@ describe("MarkdownEditor", () => {
   ;[
     [true, MinimalEditorConfig],
     [false, FullEditorConfig]
-  ].forEach(([minimal, expectedComponent]) => {
+  ].forEach(([minimal, expectedConfig]) => {
     [
       ["value", "value"],
       [null, ""]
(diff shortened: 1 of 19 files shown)


pdpinch:
> Currently, the UI in OCW Studio allows for both subscript and superscript to be selected at the same time. Is this the desired behavior?

All things being equal, no. But I do trust our users to make sensible authoring decisions. If the risk / effort ratio is low, we can allow this. If not, I would be fine with a follow up issue to address it.

ChristopherChudzicki:
I'll look into disabling one tool based on another being currently active. I agree it's undesirable but not blocking.

*Aside: There are some legitimate use cases for subscripts and superscripts at the same time (e.g., `e^{x_1 + x_2}`), but those should be handled byLaTeX (mathjax) not the general-purpose subscript/superscript of this PR. *

ChristopherChudzicki:
@pdpinch Here's a concrete example of what I meant by somewhat legitimate use cases. _Somewhat, because I still think LaTeX is better for this sort of thing._

Screenshot 2022-12-21 at 9 19 35 AM

Issue #1621, opened 2022-12-16
Mass build error reporting, cache purging


ChristopherChudzicki:
As a developer, I want to be alerted when some sites fail to publish during a mass build. Additionally, sites that were successfully built should have their fastly caches purged so we can see the updates.

#### Acceptance Criteria:

- Partial Success:
    - [ ] If some but not all sites fail to build, we should still clear fastly's cache
    - [ ] If some but not all sites fail to build, we should be alerted in slack
- Failure:
    - [ ] if the mass build pipeline fails, we should be alerted in slack.

PR #1623, opened 2022-12-16
purge cache on build failures; add alerts

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing

#1621 

#### What's this PR do?
This PR changes alerting and cache purging for the mass build pipeline:
- we now get an alert if the overall mass build job fails or is aborted
- we now purge the cdn cache if EITHER the job is successful, OR the hugo build step fails. *The point is: if 2500 courses succeed and 10 fail, we want to purge the cache. Currently there is no checking as to how many failed. Basically, the cache is always purged if the job gets to the hugo build step. (unless the job is aborted during that step).*

This PR involves a slack webhook. If you want to see that webhook actually work, set up a slack app as described in https://github.com/mitodl/ocw-studio/pull/1576#issue-1454008273. **But:** imo, you don't really need to check this functionality. you can still see the slack webhook fire, it will just fail since there's no valid URL. _on the other hand, it's not htat hard to set up._

1. You will be running the mass build. You don't want to build all the courses, so manually edit `WebsiteMassBuildViewSet` in websites/views to a filter, something like (modify as you see fit):
    ```py
    sites = sites.filter(Q(short_id="1.00-spring-2012") | Q(short_id="17.810-spring-2001"))
    sites = sites.prefetch_related("starter").order_by("name")
    ```
2. Re-upsert the mass build pipeline: `docker compose exec web ./manage.py upsert_mass_build_pipeline`
3. Run the pipeline in local concourse (http://concourse:8080/). It should succeed.
    - check that the cdn is purged *It won't actually be purged, but there will be a message about it.*
4. Make one of the courses fail...somehow. One way to do this is to manually commit a bad instructor uuid to the course repo.
5.  Run the pipeline in local concourse (http://concourse:8080/). It should fail, AND:
    - cdn purging message should still appear
    - a slack alert should be issued _if you didn't set up slack alerting, you'll just see some message about an invalid slack url_

### DIFF
--- a/content_sync/pipelines/definitions/concourse/mass-build-sites.yml
+++ b/content_sync/pipelines/definitions/concourse/mass-build-sites.yml
@@ -11,6 +11,12 @@ resource_types:
       repository: governmentpaas/s3-resource
       tag: latest
 resources:
+  - name: slack-webhook
+    type: slack-alert
+    check_every: never
+    source:
+      url: ((slack-url))
+      disabled: false
   - name: ocw-hugo-themes
     type: git
     source:
@@ -33,6 +39,7 @@ resources:
       bucket: ((artifacts-bucket))
       versioned_file: ocw-hugo-themes/((ocw-hugo-themes-branch))/webpack.json
   # END ONLINE-ONLY
+
 # START ONLINE-ONLY
 task-config: &webhook-config
   platform: linux
@@ -40,6 +47,34 @@ task-config: &webhook-config
     type: registry-image
     source: {repository: curlimages/curl}
 # END ONLINE-ONLY
+
+task-purge-cdn-cache: &task-purge-cdn-cache
+  task: clear-cdn-cache
+  timeout: 1m
+  attempts: 3
+  config:
+    platform: linux
+    image_resource:
+      type: registry-image
+      source: {repository: curlimages/curl}
+    # START NON-DEV
+    run:
+      path: curl
+      args:
+        - -f
+        - -X
+        - POST
+        - -H
+        - 'Fastly-Key: ((fastly_((version)).api_token))'
+        - -H
+        - 'Fastly-Soft-Purge: 1'
+        - https://api.fastly.com/service/((fastly_((version)).service_id))/purge_all
+    # END NON-DEV
+    # START DEV-ONLY
+      path: echo
+      args: [Skipping cdn cache purge (this is a dev environment)]
+    # END DEV-ONLY
+
 jobs:
 - name: mass-publish
   serial: true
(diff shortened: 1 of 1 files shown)


content_sync/pipelines/definitions/concourse/mass-build-sites.yml
```diff
@@ -33,13 +39,42 @@ resources:
...
+
 # START ONLINE-ONLY
 task-config: &webhook-config
   platform: linux
   image_resource:
     type: registry-image
     source: {repository: curlimages/curl}
 # END ONLINE-ONLY
+
+task-purge-cdn-cache: &task-purge-cdn-cache
```
ChristopherChudzicki, line 51: Yaml anchor. See https://yaml.org/spec/1.2.2/#3222-anchors-and-aliases, https://medium.com/@kinghuang/docker-compose-anchors-aliases-extensions-a1e4105d70bd, or your favorite google result

content_sync/pipelines/definitions/concourse/mass-build-sites.yml
```diff
@@ -33,13 +39,42 @@ resources:
...
+        - POST
+        - -H
+        - 'Fastly-Key: ((fastly_((version)).api_token))'
+        - -H
+        - 'Fastly-Soft-Purge: 1'
+        - https://api.fastly.com/service/((fastly_((version)).service_id))/purge_all
+    # END NON-DEV
+    # START DEV-ONLY
+      path: echo
+      args: [Skipping cdn cache purge (this is a dev environment)]
```
ChristopherChudzicki, line 75: This is new. I wanted some dev-mode evidence of this happening.

content_sync/pipelines/definitions/concourse/mass-build-sites.yml
```diff
@@ -11,6 +11,12 @@ resource_types:
       repository: governmentpaas/s3-resource
       tag: latest
 resources:
+  - name: slack-webhook
+    type: slack-alert
+    check_every: never
+    source:
+      url: ((slack-url))
+      disabled: false
```
ChristopherChudzicki, line 19: Elsewhere we wrap this in NON-DEV only. But I don't think there's really any need. If we include this in dev and a failure happens, you'll just see an error about the url being bad.

The case for not using this in dev might be if we sent alerts on success, because then it would fail an otherwise succeeding pipeline.

gumaerc (approved):
:+1: 

These changes look good to me. In my testing though, the only way I could get a course build to fail was by manually adding something like this to the course build loop:

```
              if [[ "$NAME" == "1-010-uncertainty-in-engineering-fall-2008" ]]
              then
                echo "INTENTIONALLY FAILING BUILD AS A TEST" > /dev/tty
                return 1
              fi
```

Manually committing a bad instructor did not fail the build, which I imagine will change after https://github.com/mitodl/ocw-hugo-themes/pull/1010 is merged? Either way, introducing the failure did mark the build as failed, but at least I did see the cache clear step attempt to run, as well as the Slack webhooks.

ChristopherChudzicki:
@gumaerc 

> Manually committing a bad instructor did not fail the build, which I imagine will change after https://github.com/mitodl/ocw-hugo-themes/pull/1010 is merged?

ah yeah...I had been testing using that themes branch and was waffling on whether to merge that themes PR before this, and I guess I wrote the insturctions assuming i would. Anyway, merging them both now.

Issue #1626, opened 2022-12-21
Disallow simultaneous subscript and superscript


ChristopherChudzicki:
We don't want to allow simultaneous subscript and superscript. So:

Acceptance Criteria:
- [ ] modify CKEditor to disallow adding subscripts to superscripts, and vice versa

See discussion in https://github.com/mitodl/ocw-studio/pull/1617#issuecomment-1361313736

ChristopherChudzicki:
This turns out to be fairly straightforward with https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_model_schema-Schema.html#function-addAttributeCheck

PR #1627, opened 2022-12-21
Cc/disallow simul subsup

ChristopherChudzicki:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested ❌  not until we add a browser-based testing environment like playwright, :sob:

#1626 

#### What's this PR do?
This PR adds an [attribute check](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_model_schema-Schema.html#function-addAttributeCheck) to the CKEditor schema to disallow subscripts/superscripts.

I also updated the typedefs for ckeditor.

Try things like:
- entering subscript mode inside a superscript, and vice versa ... the button should be disabled.
- select a region of text that contains normal text AND a superscript, and try applying a subscript. *NOTE: The subscript button will **not** be disabled in this case, but if you press it, the subscript is only applied to the normal text. This seems to be CKEditor's expected behavior.*

#### Where should reviewer start
Start with commit https://github.com/mitodl/ocw-studio/pull/1627/commits/e79c72788986d8bb4d3afb932fbca116889da7b4 ... the other commit should be looked at, but don't focus on it. It is just a bunch of changes because of how the ckeditor types were structured.

Superscript is disabled inside subscript:
Screenshot 2022-12-21 at 12 04 05 PM

Subscript is allowed if a portion of the selection can be subscripted. But if clicked, it only applies to the allowable portion.
Screenshot 2022-12-21 at 11 56 55 AM

### DIFF
--- a/static/js/components/widgets/MarkdownEditor.tsx
+++ b/static/js/components/widgets/MarkdownEditor.tsx
@@ -9,6 +9,7 @@ import {
   MinimalEditorConfig,
   insertResourceLink
 } from "../../lib/ckeditor/CKEditor"
+import { checkNotSubAndSup } from "../../lib/ckeditor/attributeChecks"
 import EmbeddedResource from "./EmbeddedResource"
 import {
   ADD_RESOURCE_LINK,
@@ -52,6 +53,7 @@ export default function MarkdownEditor(props: Props): JSX.Element {
     if (process.env.NODE_ENV === "development" && editor.current) {
       CKEditorInspector.attach(editor)
     }
+    editor.current.model.schema.addAttributeCheck(checkNotSubAndSup)
   }, [])
 
   const [resourcePickerMode, setResourcePickerMode] = useState<
@@ -174,7 +176,7 @@ export default function MarkdownEditor(props: Props): JSX.Element {
         editor={ClassicEditor}
         config={editorConfig}
         data={value ?? ""}
-        onReady={setEditorRef}
+        onReady={onReady}
         onChange={onChangeCB}
         onError={throwSynchronously}
       />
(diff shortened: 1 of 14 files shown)


static/js/components/widgets/MarkdownEditor.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { useCallback, useMemo, useRef, useState } from "react"
 import { CKEditor } from "@ckeditor/ckeditor5-react"
-import { editor } from "@ckeditor/ckeditor5-core"
+import { Editor } from "@ckeditor/ckeditor5-core"
```
ChristopherChudzicki, line 3: The new version of CKEditor types got rid of `editor` namespace, so now `editor.Editor` becomes `Editor`.

static/js/lib/ckeditor/attributeChecks.ts
```diff
@@ -0,0 +1,48 @@
...
+  context: SchemaContext,
+  attributeName: string
+): boolean => {
+  for (const item of context) {
+    for (const attribute of item.getAttributeKeys()) {
+      if (attribute === attributeName) {
+        return true
+      }
+    }
+  }
```
ChristopherChudzicki, line 26: Above, `context` and `item.getAttributeKeys()` are non-array iterables, hence the for loops as opposed to `Array.prototype.some`.

ChristopherChudzicki:
> If it would be possible to make the other one not selectable when one is selected,

That is...strange. I would not say "selected" here, but rather "focused": Apparently CKEditor does allow focusing disabled tools, which makes them partially appear active. (The toolbar item gets a focus ring, but the tool is not really selected as far as CKEditor is concerned...).

I couldn't find an open issue over there for this, so created https://github.com/ckeditor/ckeditor5/issues/13133

PR #1629, opened 2022-12-21
abort onReady if no editor

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [ ] Code is tested

Small followup to #1627 

#### What's this PR do?
See comment.

When you open the editor, you should not see the 'reading "model"' error below
Screenshot 2022-12-21 at 2 05 17 PM

### DIFF
--- a/static/js/components/widgets/MarkdownEditor.tsx
+++ b/static/js/components/widgets/MarkdownEditor.tsx
@@ -50,7 +50,16 @@ export default function MarkdownEditor(props: Props): JSX.Element {
   const editor = useRef<Editor>()
   const onReady = useCallback((editorInstance: Editor) => {
     editor.current = editorInstance
-    if (process.env.NODE_ENV === "development" && editor.current) {
+    if (!editor.current) {
+      /**
+       * It is unclear to me why this happens.
+       * It seems like when our MarkdownEditor opens, an editor is created,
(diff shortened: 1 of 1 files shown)


static/js/components/widgets/MarkdownEditor.tsx
```diff
@@ -50,7 +50,16 @@ export default function MarkdownEditor(props: Props): JSX.Element {
...
+       * It seems like when our MarkdownEditor opens, an editor is created,
+       * immediately destroyed, onReady is called (with null), and then
+       * re-created, and onReady is called again (with real editor)
+       */
+      return
+    }
+    if (process.env.NODE_ENV === "development") {
       CKEditorInspector.attach(editor)
     }
     editor.current.model.schema.addAttributeCheck(checkNotSubAndSup)
```
ChristopherChudzicki, line 65: We've had a mysterious `TypeError: Cannot read properties of null (reading 'destroy')` console.error for a long time (the second console.error below).

I've never really understood the cause of it, but it has never caused problems that I am aware of. This line of code introduced a similar error (presumably related, since they are always logged together as far as I can tell). I'd like to avoid this since, unlike the other error, this really is coming from our code.

Screenshot 2022-12-21 at 2 05 17 PM

Issue #1632, opened 2022-12-23
CKEditor error on drawer open


ChristopherChudzicki:
### Steps to Reproduce

1. Visit https://ocw-studio-rc.odl.mit.edu/sites/what-is-capitalism/type/page/
2. Open dev console
3. Open any page for editing

### Expected Behavior

No errors in dev console

### Actual Behavior

There is an error in dev console:
```
ckeditor.js:5 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'destroy')
    at Ot._destructor (ckeditor.js:5:20995)
    at ckeditor.js:5:22255
```

This has been happening for a long, long time (my guess is since well before OCW launched in March 2022). It seems to be innocuous. But it's confusing and useless noise. Let's fix it.

PR #1633, opened 2022-12-23
Fix ckeditor console error

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing

#1632 

#### What's this PR do?
- Prevents SiteContentForm from loading until its data has been loaded.
- Removes redux-logger, because it is annoying and you can get the same info from redux dev tools

1. Navigate to the page listing
2. Open any page for editing
3. Check that no error is logged in console
4. Check that editing works as expected.


static/js/store/configureStore.ts
```diff
@@ -20,7 +19,7 @@ export default function configureStore(initialState?: ReduxState) {
   let createStoreWithMiddleware
   if (process.env.NODE_ENV !== "production" && !global._testing) {
     createStoreWithMiddleware = compose(
-      applyMiddleware(...COMMON_MIDDLEWARE, createLogger()),
```
ChristopherChudzicki, line 23: I've always found our redux logging annoying and not helpful...it just clutters the dev console. In general, the same information is available in redux dev tools ([chrome](https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd?hl=en); [firefox](https://addons.mozilla.org/en-US/firefox/addon/reduxdevtools/)

Screenshot 2022-12-23 at 11 02 52 AM

It was sufficiently annoying while debugging this issue that I just got rid of it, and I think that should be a permanent change.

PR #1637, opened 2023-01-03
skip syncing offline build if no offline config exists

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Follow-up hotfix for https://github.com/mitodl/ocw-studio/pull/1630

#### What's this PR do?
In https://github.com/mitodl/ocw-studio/pull/1630, we added functionality to the single site pipeline that builds the offline version of a site, zips it up and uploads it before the online build is run for the live site. An oversight was made in this PR. Before the offline build is run, a check is run to see if an offline configuration exists for the type of site being built. This is mostly for `ocw-www`, which does not have an offline configuration as it is the root website for the online version of OCW. This check was not added to the step that runs an `aws s3 sync` to push up the offline ZIP file. This causes the build to try and run a sync against the root of the site, which ends up taking a very long time as it has to scan the entire bucket. This PR adds the same check to see if the offline configuration actually exists before attempting to sync the ZIP file.

Follow the steps in https://github.com/mitodl/ocw-studio/pull/1630 specifically for `ocw-www`, noting that the `upload-offline-build` step should now be skipped with the message "Offline configuration not found for site type ocw-www."

### DIFF
--- a/content_sync/pipelines/definitions/concourse/site-pipeline.yml
+++ b/content_sync/pipelines/definitions/concourse/site-pipeline.yml
@@ -463,6 +463,7 @@ jobs:
         config:
           inputs:
             - name: build-course-offline
+            - name: ocw-hugo-projects
           platform: linux
           image_resource:
             type: registry-image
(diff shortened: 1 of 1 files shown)


ChristopherChudzicki (approved):
LGTM after the path fix:
- ocw-www does not sync offline build
- courses do sync it

Issue #1619, opened 2022-12-14
Resource links **without** forking CKEditor


ChristopherChudzicki:
Currently, the resource link UI is created via a fork of ckeditor's link plugin. (https://www.npmjs.com/package/@mitodl/ckeditor5-resource-link, https://github.com/mitodl/ckeditor5). Maintaining this fork is a bit of a pain, and is part of the reason that we so rarely update CKEditor.

#### Acceptance Criteria:

- [ ] Implement a proof of concept, no-fork, resource link OR decide that the fork is actually our best option

#### Notes

- The resource link plugin is different from the regular link plugin in two ways:
   -  First, it stores a uuid on `data-uuid` attribute rather than a URL on a href attribute. So a builtin-link plugin alternative is to use  idea is to use a query parameter to encode the UUID
   - Second: The default link text is the resource title, whereas for links the default link text is the url. This would be more complicated to re-implement without forking the plugin, but we probably can ... https://ckeditor.com/docs/ckeditor5/latest/framework/guides/architecture/editing-engine.html#changing-the-model would be helpful in this regard.
- Resource links actually use a few ckeditor plugins:
    - ResourcePicker: is a ckeditor plugin responsible for opening the ResourcePickerDialog (where users pick the relevant resource)
    - ResourceLink: CKEditor logic for displaying a resource link, editing its title text, and deleting the resource link. **This is the fork.**
    - ResourceLinkMarkdownSyntax: converts the HTML output of ResourceLink to markdown, and vice versa. 
- Alice's original PR about resource links in the markdown editor https://github.com/mitodl/ocw-studio/pull/560

PR #1643, opened 2023-01-12
Resource links without forking CKEditor's Link Plugin

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#1619 

#### What's this PR do?
This PR changes how we implement CKEditor support for resource links. Resource links are very similar to regular links in CKEditor. The main differences are markdown conversion and the fact that the default text for a regular link is its `href`, but the default text for a resource link is the resource title.

- **Previously**: We [forked CKEditor](https://github.com/mitodl/ckeditor5) and edited its link plugin to accommodate the differences. Resource link uuid was stored as `data-uuid` attribute on the anchor.
- **Now:** We just use the regular link plugin and store the uuid as a query parameter value. The custom behavior we want is achieved by new plugins rather than editing a forked plugin.

**Two behavior changes:**
- Previously, if you clicked on a resource link it would show "no href". Now it shows an href that links to that specific resource's page in studio. *Using the actual link plugin, we need to show some href, so this was a natural choice*.
- Previously, you could edit (click link, click ✏️ button) the href of a resource link which...broke stuff. That was really a bug. It's fixed now. 

Try things like:
- inserting a new resource link
    - with no text highlighted... link title should be resource title
    - with text highlighted ... text should be converted to a resource link
- editing the text of a resource link
    - adding characters to the end of a resource link either as part of the link or as part of the following non-link text. 
    - adding/removing characters mid-link
- subscripts/superscripts in a resource link
- removing a resource link by clicking it and hitting the "X" button
- try to edit the resource link href by clicking the link and hitting ✏️. (It should be disabled)

Start with docstrings on `ResourceLink`, `ResourceLinkMarkdownSyntax`, and `ResourceLinkUI` plugins.

Screenshot 2023-01-12 at 10 54 21 AM


static/js/lib/ckeditor/plugins/ResourceLinkMarkdownSyntax.ts
```diff
@@ -1,64 +1,105 @@
...
+  isResourceLinkHref = (href?: string): boolean => {
+    if (!href) return false
+    const url = new URL(href)
+    return url.searchParams.has(queryKeys.uuid)
+  }
+
+  makeResourceLinkHref = (uuid: string, suffix = "") => {
+    const href = new URL(this.hrefTemplate.replace(/:uuid/g, uuid))
+    href.searchParams.set(queryKeys.uuid, uuid)
+    href.searchParams.set(queryKeys.suffix, suffix)
```
ChristopherChudzicki, line 76: I don't love storing these on query parameters, but I do think it's better than forking the Link plugin.

I would prefer to store these on data-attributes, which is what our fork was doing. I investigated whether we could do this without forking as described here: https://stackoverflow.com/a/51365516/2747370. That approach does work to add data attributes to the link. However, it has some issues:
1. deleting the link does not delete the data attributes
2. inserting text at the end of the link can be problematic

(2) might sound especially weird. But CKEditor's Link plugin treats adding text at the end of a link as a special case because there is ambiguous add 
- new link text at the end of the link?
- new non-link text at the end of the link?
In CKEditor-speak, when you insert text at the end of the link, should it have the `linkHref` attribute?

Anyway, CKEditor jumps through some hoops to support both of the cases above, which severely complicates adding "extra" link attributes. *CKEditor does have a mechanism, [decorators](https://ckeditor.com/docs/ckeditor5/latest/features/link.html#custom-link-attributes-decorators), for custom link attributes. But it's not flexible enough for what we want. E.g., the attribute values cannot be specified by the user; since we want the attribute value to be a uuid, that's problematic.)

static/js/store/configureStore.ts
```diff
@@ -20,7 +19,7 @@ export default function configureStore(initialState?: ReduxState) {
   let createStoreWithMiddleware
   if (process.env.NODE_ENV !== "production" && !global._testing) {
     createStoreWithMiddleware = compose(
-      applyMiddleware(...COMMON_MIDDLEWARE, createLogger()),
+      applyMiddleware(...COMMON_MIDDLEWARE),
```
ChristopherChudzicki, line 22: I've always found our redux logging annoying and not helpful...it just clutters the dev console. In general, the same information is available in redux dev tools ([chrome](https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd?hl=en); [firefox](https://addons.mozilla.org/en-US/firefox/addon/reduxdevtools/)

Screenshot 2022-12-23 at 11 02 52 AM

It was sufficiently annoying while debugging this issue that I just got rid of it, and I think that should be a permanent change.

Issue #1635, opened 2023-01-03
Adding Sortable UI


Wassaf-Shahzad:
As per discussion on [post](https://github.com/mitodl/ocw-hugo-projects/issues/183). We need to add sortable UI based on isSortable Config.

Issues and PRs related to this issue
- https://github.com/mitodl/ocw-hugo-projects/issues/237
- https://github.com/mitodl/ocw-hugo-projects/pull/238
- https://github.com/mitodl/ocw-studio/pull/1634

#### Designs and Mockups

![173149425-af9de0bb-183a-4fd1-8c80-000ee79f1365](https://user-images.githubusercontent.com/87968618/210341656-35a58a44-d417-4ac3-95c6-b7290d45b97b.png)

#### Acceptance Criteria:

- [ ] Repeatable Content with isSortable Should be drag and dropable

ChristopherChudzicki:
Have we decided how to handle sorting and pagination? Some options that seem reasonable to me:
1. Do not support sorting across pages, at least initially.
2. On page X, add context menu options for "Move to page X-1", "Move to page X+1"?

(2) would be similar to the GIF shown here https://django-admin-sortable2.readthedocs.io/en/latest/ (whether or not that package is actually useful to us).

pdpinch:
@Wassaf-Shahzad it's useful to have a link to the original discussion, but we also need a more complete description for this issue: what are we doing, and how are you going about doing it.

PR #1644, opened 2023-01-12
Added sorting studio list and sorting list item

Wassaf-Shahzad:
#### Pre-Flight checklist

- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested
- [ ] Settings

(Required)

#### What's this PR do?
(Required)

(Required)

(Optional)

(Optional)

(Optional)

### DIFF
--- a/static/js/components/RepeatableContentListing.tsx
+++ b/static/js/components/RepeatableContentListing.tsx
@@ -150,15 +150,28 @@ export default function RepeatableContentListing(props: {
   const pages = usePagination(listing.count ?? 0)
 
   const DynamicStudioListName = isSortable ? SortableStudioList : StudioList
-  const wrapSortableItem = (isSortable: boolean, child: React.ReactNode, item: WebsiteContentListItem) => {
+  const dynamicStudioListProps = isSortable ?
+    {
+      handleDragEnd: handleDragEnd,
+      attribute:     "text_id"
+    } :
+    {}
+  const wrapSortableItem = (
+    isSortable: boolean,
+    child: React.ReactNode,
+    item: WebsiteContentListItem
+  ) => {
     if (isSortable) {
-      return <SortableItem
-        deleteItem={() => null}
-        key={item.text_id}
-        id={item.text_id}
-        item={item.text_id}>
-        {child}
-      </SortableItem>
+      return (
+        <SortableItem
+          deleteItem={() => null}
+          key={item.text_id}
+          id={item.text_id}
+          item={item.text_id}
+        >
+          {child}
+        </SortableItem>
+      )
     }
     return child
   }
@@ -201,7 +214,7 @@ export default function RepeatableContentListing(props: {
               className="btn cyan-button add flex-shrink-0"
               to={siteContentNewUrl
                 .param({
-                  name: website.name,
+                  name:        website.name,
                   contentType: configItem.name
                 })
                 .query(searchParams)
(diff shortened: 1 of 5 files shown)


Wassaf-Shahzad:
@ChristopherChudzicki @gumaerc Would love your thoughts on the basic implementation.
Also would like your help in the following issue .

Currently repeatable content List is being fetched from redux store and is used to render the content list. The issue is after sorting we need to call a callback (handleDragEnd ) and I am currently a bit stuck in how we use it in Sortable Select.

pdpinch:
@Wassaf-Shahzad please fill in the PR description. Is this related to https://github.com/mitodl/ocw-studio/issues/1635 ?

ChristopherChudzicki:
*Probably would be good to put some high-level answers to questions below in https://github.com/mitodl/ocw-studio/issues/1635.*

This looks like a good start on the sorting UI, but I think there is a little backend work and some UI design decisions to make before we do more UI implementation. 

**Model:** First,`WebsiteContentModel` needs some sort of `weight` field to record the user-specified sort order

**UI Questions:** Also As @gumaerc mentioned [here](https://github.com/mitodl/ocw-hugo-projects/isasues/183#issuecomment-1152734173) there are some issues with regards to searching and pagination. In my opinion:
1. sorting needs to be disabled when there's a search filter applied

    One approach would be to use [Edit], [Order] tabs that [@gumaerc's mock](https://github.com/mitodl/ocw-hugo-projects/issues/183#issuecomment-1152734173) showed. Search would be disabled when the "Order" tab is open. And having a specific "Order" tab would make it more natural to put drag handles on the title cards, indicating that they are sortable.
2. sorting across pages is tricky. We should decide how to handle this, even if we call it out-of-scope for the first iteration of this feature. 

**A new API, maybe?**: Regarding your question
> ...we need to call a callback (handleDragEnd ) and I am currently a bit stuck in how we use it in Sortable Select.
The callback should update the `weight` property of several records. E.g., if we have promos `[P1 ... P10]` then I would expect dragging P6 above P3 to result in 
```
P1       P1
P2       P2
P3       P6     <---\
P4       P3     <---| ... all affected
P5       P4     <---|
P6       P5     <---/
P7       P7
P8       P8
P9       P9
```

We could just make N api requests, but a dedicated API might be better. For one thing, then the swapping logic could live on the backend. (And maybe Django has libraries for this already, e.g., https://django-admin-sortable2.readthedocs.io/en/latest/ ... that seems focused on the admin panel, but maybe is more general purpose than the GIF makes me think.)

Wassaf-Shahzad:
@ChristopherChudzicki @pdpinch  Thanks for looking into this. I do agree we might need to hold off the UI and move to the backend side first. I did some RnD on the handleDragEnd and did figure out that the best approach would be to  change the listing.result order based on moved array.

## Pagination 
In cases like ours where items are being manually sorted (given order by drag and drop) the number of items per collection would be (should be) relatively small. Manually sorting through 30 for 40 items would be bad UX (as per my opinion). However even with this mind, I dont think that we can limit the number of items in an `isSortable` collection. We could implement an infinite scroll view and disable pagination for sortable collections.

## Searching and UI
We could just disable sorting for searching as you recommended but I have recommendation is terms of UI. Rather then having separate buttons for edit and sort modes, collections which have the `isSortable` property could be rendered in the sortable UI (Title card rendered with drag handles) in default and we remove them when searching for pages. This could intuitively give the user context on which items are draggable and which are not.

ChristopherChudzicki:
Closing because we went a slightly different direction. See https://github.com/mitodl/ocw-hugo-projects/issues/183#issuecomment-1402012595

Issue #1664, opened 2023-02-01
current --baseUrl setting is incompatible with relativeURLs: true


gumaerc:
### Steps to Reproduce

 - Publish a site using the `ocw-course-v3` project from `ocw-hugo-projects` as your starter
 - Build the site in Concourse

For example, say you have a nav url of `/pages/syllabus` and your course is called `course-1`. Hugo will be run with `--baseUrl /courses/course-1` with `relativeUrls: true`. During the build process, Hugo will generate URLs to this page as `/courses/course-1/pages/syllabus`.

### Actual Behavior

`relativeUrls: true` is a [*post processing*](https://gohugo.io/content-management/urls/#relative-urls) step, so upon exporting the site as HTML the URL becomes `./courses/course-1/pages/syllabus`. Clicking this link on the course home page would bring you to `http://localhost:8045/courses/course-1/courses/course-1/pages/syllabus`

### Expected Behavior

On the course home page, the URL should be `./pages/syllabus`, and clicking it should bring you to `http://localhost:8045/courses/course-1/pages/syllabus`.

### Additional Info

 In our pipelines upserted by `ocw-studio`, we set the `--baseUrl` argument to be a root-relative URL such as `--baseUrl /courses/course-1`. Whether you set it to a fully qualified or root-relative URL, the effect is the same. When writing our nav URLs into a course content repo here in `ocw-studio`, we generate a menu configuration file at `config/_default/menus.yaml` using [this format](https://gohugo.io/content-management/menus/#add-non-content-entries-to-a-menu). This file contains menu entries with root-relative URLs, i.e. `/pages/syllabus`. According to the Hugo documentation:

> "The URLs must be relative to the context root. If the baseURL is https://example.com/mysite/, then the URLs in the menu must not include the context root mysite. Using an absolute URL will override the baseURL. If the value used for URL in the above example is https://subdomain.example.com/, the output will be https://subdomain.example.com."

In `ocw-hugo-themes`, our nav URL is rendered from a partial called `nav_url.html` In the `course-v2` theme, this partial's code is:

```
{{ return .menuItem.URL }}
```

We are not doing any manipulation of the URL, simply calling for it and rendering it on the page. According to the documentation above, Hugo considers root-relative URLs to be "relative" and fully qualified URLs to be "absolute." The [Hugo documentation](https://gohugo.io/getting-started/configuration/#baseurl) does not say anything specific about how `--baseUrl` is used, but it does say:

[...]

PR #1674, opened 2023-02-07
allow overriding hugo build arguments

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1664

#### What's this PR do?
This PR adds an argument to the management commands `backpopulate_pipelines` as well as `upsert_mass_build_pipeline` called `hugo_args` that allows you to pass a string of arguments to append / override in the Hugo commands used by pipelines. While this PR gives you a method of bypassing the issues faced in https://github.com/mitodl/ocw-studio/issues/1664, it is more generally useful in that while developing locally and testing different things with Hugo builds in Concourse pipelines you can arbitrarily change the arguments passed to `hugo`.

Since we have decided to not pursue unifying the online and offline builds, we reverted the `course-v3` and `course-offline-v2` themes in https://github.com/mitodl/ocw-hugo-themes/pull/1081 and https://github.com/mitodl/ocw-hugo-projects/pull/245, therefore the issue in https://github.com/mitodl/ocw-studio/issues/1664 cannot be reproduced on the `main` branches of `ocw-hugo-themes` and `ocw-hugo-projects`.

Since https://github.com/mitodl/ocw-hugo-themes/pull/1081 and https://github.com/mitodl/ocw-hugo-projects/pull/245 have been reverted, we cannot exactly reproduce the issue from https://github.com/mitodl/ocw-studio/issues/1664 so we will verify that everything is working by enabling the `--templateMetrics` flag in a single course build:

 - Spin up `ocw-studio` with `docker compose up`, making sure you have set up Concourse and Minio as described in the readme
 - Run `docker compose exec web ./manage.py backpopulate_pipelines --filter 10-302-transport-processes-fall-2004 --hugo_args="--templateMetrics"`, replacing the filter string with the course of your choice if desired
 - Publish the course from the `ocw-studio` UI
 - Go to the Concourse web UI at http://concourse:8080 and look for the pipeline matching the site you published by putting the ID in the search box
 - Open the `build-course-offline` or `build-course-online` steps and verify that you see template metrics in the terminal output

### DIFF
--- a/content_sync/api.py
+++ b/content_sync/api.py
@@ -72,6 +72,7 @@ def get_mass_build_sites_pipeline(  # pylint:disable=too-many-arguments
     projects_branch: Optional[str] = None,
     starter: Optional[str] = None,
     offline: Optional[bool] = None,
+    hugo_args: Optional[str] = None,
 ) -> object:
     """Get the mass build sites pipeline if the backend has one"""
     return import_string(
@@ -84,6 +85,7 @@ def get_mass_build_sites_pipeline(  # pylint:disable=too-many-arguments
         projects_branch=projects_branch,
         starter=starter,
         offline=offline,
+        hugo_args=hugo_args,
     )
 
 
--- a/content_sync/api.py
+++ b/content_sync/api.py
@@ -44,7 +44,9 @@ def get_pipeline_api() -> BasePipeline:
 
 
 @is_publish_pipeline_enabled
-def get_site_pipeline(website: Website, hugo_args: Optional[str] = "", api: Optional[object] = None) -> BasePipeline:
+def get_site_pipeline(
+    website: Website, hugo_args: Optional[str] = "", api: Optional[object] = None
+) -> BasePipeline:
     """ Get the configured sync publishing pipeline """
     return import_string(
         f"content_sync.pipelines.{settings.CONTENT_SYNC_PIPELINE_BACKEND}.SitePipeline"
--- a/content_sync/api.py
+++ b/content_sync/api.py
@@ -44,11 +44,11 @@ def get_pipeline_api() -> BasePipeline:
 
 
 @is_publish_pipeline_enabled
-def get_site_pipeline(website: Website, api: Optional[object] = None) -> BasePipeline:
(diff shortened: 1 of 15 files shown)


mbertrand (commented):
When I update the pipeline with the mgmt command and run directly from concourse, the template metrics show up.  If I publish from the UI - for a site that hasn’t been published to production yet - the pipeline gets updated again (no options) and so the metrics disappear.  

The reason for updating the pipeline every time if it hasn't been published to production yet is because the publish path might have changed.

```
        if trigger_pipeline and settings.CONTENT_SYNC_PIPELINE_BACKEND:
            pipeline = get_site_pipeline(website, api=pipeline_api)
            # Always upsert pipeline in case the site url path changed,
            # unless it's been published to production (url path permanent).
            if not website.publish_date:
                pipeline.upsert_pipeline()
```

gumaerc:
> When I update the pipeline with the mgmt command and run directly from concourse, the template metrics show up. If I publish from the UI - for a site that hasn’t been published to production yet - the pipeline gets updated again (no options) and so the metrics disappear.
> 
> The reason for updating the pipeline every time if it hasn't been published to production yet is because the publish path might have changed.
> 
> ```
>         if trigger_pipeline and settings.CONTENT_SYNC_PIPELINE_BACKEND:
>             pipeline = get_site_pipeline(website, api=pipeline_api)
>             # Always upsert pipeline in case the site url path changed,
>             # unless it's been published to production (url path permanent).
>             if not website.publish_date:
>                 pipeline.upsert_pipeline()
> ```

Hm, so that is a good point, but I don't think it's necessarily a show stopper. This functionality isn't going to be used to set permanent arguments on the Hugo command on a pipeline, it's more for debugging / testing purposes. For a little bit of context, I was originally working on this as part of the research into consolidating the offline and online themes. We eventually decided to go another direction on that and keep them separate instead, but I was going to use it basically to verify PR's in RC that needed to change the baseUrl setting without having to permanently change the default setting.

mbertrand (approved):
👍

Issue #1642, opened 2023-01-10
Allow authors to unpublish sites 


pdpinch:
As an author, I need a way to remove course sites from the OCW site. 

#### Acceptance Criteria:

- [x] We already have an API and a management command to do this, but neither of those is self-service for course authors. 
- [ ] Unpublishing a site shouldn't be taken lightly, so it shouldn't be too easy. 
   - [ ] add a ⋮ kebob menu to the site listing view, with the only option being "unpublish" 
   - [ ] display a "are you sure" pop-up

#### Out of Scope

- the API for unpublishing can return a list of courses with links that would be broken after the course is unpublished. If there's a super-easy way to include those links in the "are you sure" pop-up, that would be nice. But if it's difficult, we shouldn't bother -- experience has shown that the course authors don't worry too much about the links.

PR #1684, opened 2023-02-16
feat: unpublish sites

MAbdurrehman12:
#### Pre-Flight checklist

- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested
- [ ] Settings

closes https://github.com/mitodl/ocw-studio/issues/1642

#### What's this PR do?
- Allow course authors to unpublish websites
- Introduces a dropdown to the site listing view
    - Since this dropdown has only one `Unpublish` option so the dropdown is not shown is the website is `draft`  
- Upon clicking `Unpublish` button, a modal is opened with confirmation.
- Upon confirming, the site is unpublished (unpublish pipeline is triggered)

- Checkout this branch
- Go to the site listing view.
- Verify that the dropdown with a `Unpublish` option is being shown only with sites that are published, and dropdown is not being shown for draft websites. 
- Verify that the dropdown works fine.
- Verify that upon clicking `unpublish`, a modal is opened asking the user for a confirmation.
- Verify that the modal works fine. Closes upon clicking close/x button
- Verify that, upon clicking `Confirm` button, the site is successfully unpublished (unpublish pipeline is triggered)

image

image

image

image


pdpinch:
I agree with Paul. 
- Let’s remove the warnings for now
- “The site has been submitted for unpublishing”

Ferdi:
@MAbdurrehman12 what does "dependency" mean.   What makes a site depend on another site or another piece of content ?

pdpinch:
@MAbdurrehman12 can you update the screenshots, now that you're removed the "dependencies" warning and changed the message?

gumaerc (approved):
:+1: 

This looks good to me, although I did encounter a few issues while testing that are not necessarily blockers:

 - Locally, if you haven't yet run the `upsert_site_removal_pipeline` management command, the pipeline will not exist yet. If you try and unpublish a site without it, the site is still marked as draft and the dialog will tell you it's been queued for unpublishing. However, the site will not be unpublished because the pipeline doesn't exist.
 - After unpublishing, I was able to click the context menu again and trigger a second unpublish job. If I refreshed the page, the listing would say "Draft" and not allow me to unpublish it again, but before you do that you can infinitely trigger more jobs by doing it again and again. This might be confusing for end users, so maybe we should refresh the page after triggering an unpublish? There isn't necessarily any harm in running the unpublish job more than once, but a confused user might trigger a lot of them and they could pile up.

MAbdurrehman12:
> what does "dependency" mean. What makes a site depend on another site or another piece of content ?

@Ferdi dependency is basically the reference/link of 1 course into other courses/course content etc. 
So in this case, the references of a course in other courses/content breaks as soon as this course is unpublished.
FYI: This was the issue ticket https://github.com/mitodl/ocw-studio/issues/1457 to show warnings when unpublishing courses and this was its PR https://github.com/mitodl/ocw-studio/pull/1468

Does this answer your question?

> can you update the screenshots, now that you're removed the "dependencies" warning and changed the message?

@pdpinch Yes, sure. I added the new/updated screenshots but didn't remove the old ones (kept them for reference)

image

but sorry about that. I'll remove the old screenshots now

MAbdurrehman12:
Thanks for the review @gumaerc :)

> Locally, if you haven't yet run the upsert_site_removal_pipeline management command, the pipeline will not exist yet. If you try and unpublish a site without it, the site is still marked as draft and the dialog will tell you it's been queued for unpublishing. However, the site will not be unpublished because the pipeline doesn't exist.

yeah, and we have an exception regarding the missing pipeline in the logs but is this out of scope here or do we need to worry about it? 
Maybe we can update the backend API to check whether the pipeline exists or not? Do you think this is something worth doing? Since this is something only on local setup and will not occur on RC/Prod.

> After unpublishing, I was able to click the context menu again and trigger a second unpublish job. If I refreshed the page, the listing would say "Draft" and not allow me to unpublish it again, but before you do that you can infinitely trigger more jobs by doing it again and again. This might be confusing for end users, so maybe we should refresh the page after triggering an unpublish? There isn't necessarily any harm in running the unpublish job more than once, but a confused user might trigger a lot of them and they could pile up.

ahh.. thanks for catching this. Yes, this is something we should definitely resolve. I'll update the PR accordingly.

MAbdurrehman12:
@gumaerc Can you please have another look?

gumaerc (approved):
:+1: LGTM

The issue with the page refresh on unpublish is fixed, and the success callback seems like a good way to do that. As far as the pipeline issue, no I don't think that's something that needs to be fixed here. However, if an issue like that is known it's usually good to include that in the testing instructions, so people reviewing the PR can avoid pitfalls like that. It would be something you might typically label as "out of scope."

MAbdurrehman12:
> However, if an issue like that is known it's usually good to include that in the testing instructions, so people reviewing the PR can avoid pitfalls like that. It would be something you might typically label as "out of scope."

Ahh yes. I should have done that. Apologies for any inconvenience.

Issue #1689, opened 2023-02-17
copy offline builds to S3


gumaerc:
part of https://github.com/mitodl/hq/issues/785

As an admin in a geography with poor public Internet access, I'd like to be able to get a hard drive with a local copy of OCW content. I might attach the drive to a file server or to a basic web server so that my users on the local intranet might access it.

#### Acceptance Criteria:

- [ ] During the publishing of a single course, copy the offline site build artifacts to the appropriate bucket
- [ ] During a run of `mass-build-sites` with the `--offline` flag set, copy the offline site build artifacts to the appropriate bucket

pdpinch:
The title "archive" is confusing to me. I changed this to "copy offline builds to S3"

I also updated the description to reference the epic, https://github.com/mitodl/hq/issues/785

gumaerc:
This was mistakenly marked closed by merging https://github.com/mitodl/salt-ops/pull/1556, but it will only be closed when https://github.com/mitodl/ocw-studio/pull/1690 merges.

PR #1690, opened 2023-02-18
mirror drive s3 upload

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested
- [ ] Settings
  - [ ] Opened issue for DevOps regarding necessary configuration changes to deployed environments - https://github.com/mitodl/salt-ops/pull/1556

Closes https://github.com/mitodl/ocw-studio/issues/1689

#### What's this PR do?
This PR adds code to Concourse pipelines that, if an offline build is being run, uploads the output to a separate set of S3 buckets defined by `AWS_OFFLINE_PREVIEW_BUCKET_NAME` and `AWS_OFFLINE_PUBLISH_BUCKET_NAME`. This is being done as part of https://github.com/mitodl/hq/issues/785 to support creating a mirror drive copy of the entire OCW site.

 - Make a temporary edit to `websites/views.py`, changing line 341 to `sites = sites.prefetch_related("starter").order_by("name")[:10]` to limit the mass build to 10 sites
 - Spin up `ocw-studio`, making sure you have Concourse and Minio configured and the following in your `.env`:
```
OCW_HUGO_THEMES_BRANCH=main
OCW_HUGO_PROJECTS_BRANCH=main
AWS_OFFLINE_PREVIEW_BUCKET_NAME=ocw-content-offline-draft
AWS_OFFLINE_PUBLISH_BUCKET_NAME=ocw-content-offline-live
```
 - Run `docker compose exec web ./manage.py upsert_theme_assets_pipeline` to ensure you have the latest version of the theme assets pipeline pushed up
 - Browse to the Concourse UI at http://localhost:3000 and go to the `ocw-theme-assets` group, finding the `main` branch version that we just pushed up
 - Unpause and trigger a build of the theme assets pipeline, waiting for it to complete
 - Run `docker compose exec web ./manage.py upsert_mass_build_pipeline --offline --starter ocw-course-v2`
 - Log into the Concourse UI at http://localhost:8080
 - Find the mass build pipeline you just pushed up with offline set to true, version set to live and starter set to ocw-course-v2
 - Trigger a build and wait for it to finish
 - Log into the Minio web interface at http://localhost:9001
 - Browse to `ocw-content-offline-live` and open the `courses` folder, then select a course folder, then click download in the right side context menu
 - Extract the ZIP file and double click on `index.html`
 - Double click on the `index.html` file to open the site and make sure:
   - There are no errors in the console
   - JS / CSS / Images load properly
   - Nav links work
   - Links to resources work


ChristopherChudzicki (changes requested):
When I run the mass build, resource links seem not to work in the offline site. They do work when I view the offline zip from a single-site build, though.

**Single-site build** (offline version)
Screen Shot 2023-02-23 at 12 05 14 PM

**Mass build, offline**
Screen Shot 2023-02-23 at 12 04 49 PM

Can you check the resource links in your offline mass builds?

I don't see any obvious changes that would cause something like this.

gumaerc:
@ChristopherChudzicki I tested this again and all looks good on my end as far as resource links go. Here's One of the pages you posted, loading with the `file://` protocol from my local drive after downloading from the offline bucket in Minio:

image

Maybe check the markdown in your database? I did a restore fairly recently so my data is pretty close to production.

content_sync/pipelines/concourse.py
```diff
@@ -726,7 +735,8 @@ def upsert_pipeline(self):  # pylint:disable=too-many-locals
             .replace(
                 "((artifacts-bucket))", template_vars["artifacts_bucket_name"] or ""
             )
-            .replace("((ocw-bucket))", template_vars["destination_bucket"] or "")
+            .replace("((web-bucket))", template_vars["web_bucket"] or "")
+            .replace("((offline-bucket))", template_vars["offline_bucket"] or "")
```
ChristopherChudzicki, line 739: Why do we default to `""`? Shouldn't we error?

(Even if we should error, I don't think we should address it in this PR, since it seems to be a common pattern in this file.)

ChristopherChudzicki (approved):
After updating course data I could not reproduce the issue I encountered yesterday. Not sure what was going on with that.

**One request:** Could you add default values for the new env vars to `.env.example`?

content_sync/pipelines/concourse.py
gumaerc: This is a good point, although the dict of vars is built from env variables in [this function](https://github.com/mitodl/ocw-studio/blob/8895d67e50f9ebb558b97d32239cdfeaccfdda78/content_sync/utils.py#L106). Each pipeline class in `concourse.py` has a `self.MANDATORY_SETTINGS` property [like this](https://github.com/mitodl/ocw-studio/blob/8895d67e50f9ebb558b97d32239cdfeaccfdda78/content_sync/pipelines/concourse.py#L605) that is checked [here](https://github.com/mitodl/ocw-studio/blob/8895d67e50f9ebb558b97d32239cdfeaccfdda78/content_sync/utils.py#L80). So, if any of those settings are not set, an error will be thrown. I added the offline bucket name settings to the `MANDATORY_SETTINGS` list there applicable in this PR.

gumaerc:
> **One request:** Could you add default values for the new env vars to `.env.example`?

I'm going to make an issue to make a more general update to this in a separate PR, as the example env here is barren of most of the settings we've added and it needs a lot more than just the offline buckets.

Issue #1695, opened 2023-02-24
update example env file


gumaerc:
Our `.env.example` file is out of date and doesn't accurately reflect an env file you could get started with developing `ocw-studio`. It should be updated to reflect the current env settings you might have if you're just getting started spinning up `ocw--studio` for the first time.

PR #1697, opened 2023-02-24
update example env file

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1695

#### What's this PR do?
This PR updates the example `.env` file, `.env.example`, with appropriate values for getting started with development of `ocw-studio`. A value of `changeme` is set on values that should be filled with a user specific configuration, as noted in the comments.

 - Backup any existing `.env` file you may have
 - Copy and paste the contents of `.env.example` into your `.env` file
 - Verify that all containers spin up successfully when running `docker-compose up`
 - Log into Minio at http://localhost:9001 with changeme/changeme and verify that the local S3 buckets were successfully created
 - Look through the example env and verify that all the settings you would expect to be there are present, have sane defaults and are categorized in a way that makes sense


MAbdurrehman12 (approved):
👍

Issue #1207, opened 2022-04-06
Fix flaky frontend test


ChristopherChudzicki:
Twice recently I've had frontend tests fail on CI in a PR that did not touch frontend at all. For example, https://github.com/mitodl/ocw-studio/runs/5854057454?check_suite_focus=true

Acceptance Criteria:
- [ ] Figure out why and fix it.

pt2302:
Another possible example: https://github.com/mitodl/ocw-studio/runs/7197287993?check_suite_focus=true

pt2302:
Possibly related: https://github.com/mitodl/ocw-studio/issues/269.

pt2302:
Another example: https://github.com/mitodl/ocw-studio/runs/7362418653?check_suite_focus=true

pdpinch:
any update on this @Wassaf-Shahzad ?

Wassaf-Shahzad:
Sorry no @pdpinch I could not reproduce this locally

HussainTaj-arbisoft:
To determine this issue's impact and existence, I ran the `SingletonsContentListing.test` multiple times. (Thanks to https://github.com/mitodl/ocw-studio/issues/1207#issuecomment-1185785648, I did not have to run the entire suite).

| Test | Count |
| ------- | ----- |
| Passed | 159 |
| Failed | 4 |
| Total | 163 |

Although it could use more runs for a more accurate number, we can say this issue **impacts 2.4% of the PRs**.

I ran a few experiments but unfortunately, I could not determine the cause of the problem yet.

Here's a sample of the error message
```
  ● SingletonsContentListing › loads content detail from the API if needed without content context

    AssertError: expected stub to not have been called but was called once
        stub('/api/websites/suscipit-autem-enim-sapiente-qui/content/eos/', 'GET', { body: undefined, headers: undefined, credentials: undefined }) => {
      body: {
        text_id: 'ec6675ae-91bd-4bdd-b7a2-1579864907ce',
        title: 'Quia facere veritatis',
        type: 'molestiae',
        updated_on: '2006-03-25T18:23:21.117Z',
        markdown: 'Quasi alias minima harum voluptatem dolor aut odit corporis. Consequatur magni aut officia quis quia. Doloribus est non quo quibusdam in ab cumque sed. Amet enim perspiciatis. Voluptas earum temporibus voluptates animi dolorum.',
        metadata: { description: 'Eum quidem architecto accusantium.' },
        content_context: [],
[...]

Issue #269, opened 2021-05-12
Fix flaky JS test (SingletonsContentListing)


gsidebo:
This JS test failed during the build for a backend-only PR

```
FAIL static/js/components/SingletonsContentListing.test.tsx
  ● SingletonsContentListing › loads content detail from the API if needed

    AssertError: expected stub to be called once but was called 0 times

      165 |       tabLink.prop("onClick")({ preventDefault: helper.sandbox.stub() })
      166 |     })
    > 167 |     sinon.assert.calledOnce(contentDetailStub)
          |                  ^
      168 |   })
      169 |
      170 |   it("should render the SiteContentEditor component", async () => {

      at Object.fail (node_modules/sinon/lib/sinon/assert.js:110:25)
      at failAssertion (node_modules/sinon/lib/sinon/assert.js:67:20)
      at Object.assert.<computed> [as calledOnce] (node_modules/sinon/lib/sinon/assert.js:93:17)
      at static/js/components/SingletonsContentListing.test.tsx:167:18
      at fulfilled (static/js/components/SingletonsContentListing.test.tsx:5:58)
```

mbertrand:
Failed again today

PR #1700, opened 2023-02-28
Fix: Flaky Frontend Test

HussainTaj-arbisoft:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1207 
Closes https://github.com/mitodl/ocw-studio/issues/269

#### What's this PR do?

This PR populates singletonConfigItems with a fixed `name` text instead of generating it by `casual`.

1. Open the terminal and navigate to your local [ocw-studio](https://github.com/mitodl/ocw-studio) setup.
2. Run `docker-compose run -it watch bash`
3. Run `for run in {1..100}; do yarn test >>1207-pr.log 2>>1207-pr.log; done`. This runs the test suite 100 times.
4. Wait about 2 to 4 hours for the command to finish. Let it run in the background.
5. Open `1207-pr.log` file. Search for the keyword `FAIL`.
     **Expected Result**: You should find no matches indicating that a test failed. Other operational failures may occur, but they do not concern us.

tenor-1150284905

### DIFF
--- a/static/js/components/SingletonsContentListing.test.tsx
+++ b/static/js/components/SingletonsContentListing.test.tsx
@@ -56,9 +56,9 @@ describe("SingletonsContentListing", () => {
     })
     website = makeWebsiteDetail()
     singletonConfigItems = [
-      makeSingletonConfigItem('textId-1'),
-      makeSingletonConfigItem('textId-2'),
-      makeSingletonConfigItem('textId-3')
+      makeSingletonConfigItem("textId-1"),
+      makeSingletonConfigItem("textId-2"),
+      makeSingletonConfigItem("textId-3")
     ]
     configItem = {
       ...makeSingletonsConfigItem(),
--- a/static/js/components/SingletonsContentListing.test.tsx
+++ b/static/js/components/SingletonsContentListing.test.tsx
@@ -56,9 +56,9 @@ describe("SingletonsContentListing", () => {
     })
     website = makeWebsiteDetail()
     singletonConfigItems = [
-      makeSingletonConfigItem("textId-1"),
-      makeSingletonConfigItem("textId-2"),
-      makeSingletonConfigItem("textId-3")
+      makeSingletonConfigItem(),
+      makeSingletonConfigItem(),
+      makeSingletonConfigItem()
     ]
     configItem = {
       ...makeSingletonsConfigItem(),
--- a/static/js/components/SingletonsContentListing.test.tsx
+++ b/static/js/components/SingletonsContentListing.test.tsx
@@ -56,9 +56,9 @@ describe("SingletonsContentListing", () => {
     })
     website = makeWebsiteDetail()
     singletonConfigItems = [
-      makeSingletonConfigItem(),
-      makeSingletonConfigItem(),
-      makeSingletonConfigItem()
+      makeSingletonConfigItem('textId-1'),
+      makeSingletonConfigItem('textId-2'),
+      makeSingletonConfigItem('textId-3')
     ]
     configItem = {
       ...makeSingletonsConfigItem(),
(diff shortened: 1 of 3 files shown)


static/js/components/SingletonsContentListing.test.tsx
```diff
@@ -56,9 +56,9 @@ describe("SingletonsContentListing", () => {
     })
     website = makeWebsiteDetail()
     singletonConfigItems = [
-      makeSingletonConfigItem(),
-      makeSingletonConfigItem(),
-      makeSingletonConfigItem()
+      makeSingletonConfigItem("textId-1"),
+      makeSingletonConfigItem("textId-2"),
+      makeSingletonConfigItem("textId-3")
```
ChristopherChudzicki, line 61: Nice sleuthing figuring out that this was caused by duplicate names.

This change fixes the issue *for this test*, but it seems to me that makeSingletonConfigItem could create duplicate names else. Could we change `makeSingletonConfigItem` so that it always uses a unique name?

Maybe we could add a `uniqueWord` helper function, like
```ts
const uniqueWord = () => casual.word + _.uniqueId()
```
doesn't have to be [`_.uniqueId`](https://lodash.com/docs/3.10.1#uniqueId)...could be anything to guarantee uniqueness.

_Random aside: I've recently been prefering `@faker-js/faker` for generating test data, largely for this reason. It has a [`faker.helpers.unique`](https://fakerjs.dev/api/helpers.html#unique) method. Not sure that's worth adding to studio just for this, though._

ChristopherChudzicki (changes requested):

static/js/components/SingletonsContentListing.test.tsx
HussainTaj-arbisoft: @ChristopherChudzicki Thank you. I agree. I did think about this option.

I chose to keep it simple for two reasons:
1. This change impacts the least amount of code. Makes it easier to test, and track bugs caused by it.
2. If other tests already pass with duplicates, we should let them be. In this case, removing duplicates only reduces our test data space.

I don't know what a "SingletonConfigItem" really is. If it is meant to always have a unique name property then my opinion changes and I think your suggestion is better.

Let me know what you think.

PS: Yeah, faker is great, it's my first choice too.

ChristopherChudzicki: Re "what a "SingletonConfigItem" really is": OCW Studio's behaves differently for different types of sites (example: courses and ocw-www). The behavior is controlled via config files which are stored in [ocw-hugo-projects](https://github.com/mitodl/ocw-hugo-projects).

RepeatableConfigItem is the configuration for "repeatable" content, like pages or resources. (example: pages https://github.com/mitodl/ocw-hugo-projects/blob/main/ocw-course-v2/ocw-studio.yaml#L6).

SingletonConfigItem is configuration for singleton content, like the "Menu" (https://github.com/mitodl/ocw-hugo-projects/blob/main/ocw-course-v2/ocw-studio.yaml#L777). *"Menu" is the links that are on the left-hand side of OCW courses*.

--- 

Re
> I chose to keep it simple for two reasons...

Makes sense. IMO, reason (2) is not a concern: `name` should be unique. If it's not unique, we will have lots of problems.

Re (1), since `makeSingletonConfigItem` is only used in tests, if the tests pass, I think you're safe. Especially if they pass 100 times with your loop :).

ChristopherChudzicki (approved):
@HussainTaj-arbisoft IMO, it's worth changing `makeSingletonConfigItem`, but you decide.  I have approved, so if you'd rather leave as-is, go ahead and merge. If you decide to change and want a re-review, let me know!

HussainTaj-arbisoft:
@ChristopherChudzicki thank you for the explanation. I appreciate it. I've changed `makeSingletonConfigItem` as you suggested.

ChristopherChudzicki (approved):

Issue #1698, opened 2023-02-24
automatic publishing of site listing


gumaerc:
As a content author, I'd like my course to automatically appear in a list of courses in OCW mirror drives

#### Acceptance Criteria:

- [ ] A scheduled celery task runs every x minutes that does the following for both draft and live:
  - [ ] Fetch the root website via `ROOT_WEBSITE_NAME`
  - [ ] Fetch all other sites that have been published at least once in the given context (draft / live)
  - [ ] For each site found in step 2, create or update a `WebsiteContent` with `website` set to the root website and a `dirpath` of `content/websites` containing the following metadata set in the `metadata` field:
    - [ ] The site's metadata found in the `sitemetadata` `WebsiteContent` object associated with that `Website`
    - [ ] `headless: true`
  - [ ] Push the `WebsiteContent` objects that were just created / updated to Github and publish to the appropriate branch (draft / live) without publishing other content
  - [ ] Update the sync states of these `WebsiteContent` objects so that studio knows they have been synced to Github

PR #1705, opened 2023-03-03
publish websites in root website

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1698

#### What's this PR do?
This PR automatically creates or updates `WebsiteContent` objects of type "website" associated with the root website, denoted by `settings.ROOT_WEBSITE_NAME`. This content will be used to build an "offline index" of courses for mirror drives. There are a few components:

 - The `update_websites_in_root_website` management command, which triggers a task of the same name. The task fetches every `Website` that has been published at least once and creates a `WebsiteContent` of type "website" for it, associated with the `Website` denoted by `settings.ROOT_WEBSITE_NAME`. In order to better facilitate this, the [`django-bulk-update-or-create`](https://github.com/fopina/django-bulk-update-or-create) package was added. This allows us to bulk update or create large numbers of objects, saving significant time as opposed to just looping and running `update_or_create`.
 - The `update_website_in_root_website` task, which is meant to be set in `PREPUBLISH_ACTIONS`. This assures that when a website is published, a `WebsiteContent` object of type "website" associated with the root website is created or updated. If the root website does not have other pending content to be published, it is published as well.
 - If a site is unpublished, the `remove_website_in_root_website` task is triggered, which will remove the above described `WebsiteContent` object from the root website.

 - Spin up `ocw-studio` on this branch, ensuring you have Minio, Concourse and a custom Github org configured
 - Run `./manage.py update_websites_in_root_website`
 - Once the command has completed successfully, log in to the `ocw-studio` UI at http://localhost:8043/ publish your root website (`ocw-www` by default)
 - Once the root website is published, visit its git repository in your custom Github org and verify that you have a folder at `content/websites` on the `main`, `preview` and `live` branches that contains an entry for each published site in your database
 - Visit any course site and make an edit to the site metadata, then save your changes and publish the site live
 - Back in Github, once the course publish has finished, make sure that its Github repo has been updated as well as its entry in the root website`s Github repo
 - Unpublish the course site, either through the UI or the `unpublish_sites` management command
 - Afterwards, publish the root website again
 - Verify in the site has been unpublished and that the site's `content/websites` entry has been removed from the root website's Github repo


pdpinch:
Is there someplace I can see a sample of this output?

gumaerc:
@pdpinch https://github.com/gumaerc-testorg/ocw-www/tree/main/content/websites

content_sync/tasks.py
```diff
@@ -392,3 +393,48 @@ def check_incomplete_publish_build_statuses():
...
+            title=site.title,
+            dirpath=dirpath,
+            filename=site.short_id,
+            is_page_content=True,
+            defaults={"title": site.title, "metadata": site_metadata},
+        )
+    website_content = WebsiteContent.objects.filter(
+        website=root_website, type="website"
+    )
+    github_api = github.GithubApiWrapper(website=root_website)
```
mbertrand, line 439: Ideally it would be better to do this instead, in order to avoid having code explicitly linked to github as much as possible:

```python
backend = get_sync_backend(website)
backend.sync_all_content_to_backend(query_set=website_content)
```
And you would need to change the function definition for `BaseSyncBackend` and `GithubBackend` to include that optional parameter and pass it on to `GithubApiWrapper`.

However, the default way the github wrapper here works is to commit to main, then merge from main to draft or publish branch when published, and you need to commit directly to either draft or publish branch, which will need to be different all the time, right?  You could add an additional parameter for that.  Maybe something like:

BaseSyncBackend:
```python
def sync_all_content_to_backend(
  self, 
  query_set: Optional[WebsiteContentQuerySet] = None, 
  **kwargs
):
```

GithubBackend:
```python
def sync_all_content_to_backend(
  self, 
  query_set: Optional[WebsiteContentQuerySet] = None, 
  branch=setttings.GIT_BRANCH_MAIN
):
```

in code:
```
kwargs = {"branch": <preview/release>}
backend.sync_all_content_to_backend(query_set=website_content, **kwargs)
```

and then pass those on to the GithubApiWrapper, though that's where things might get real messy.  I think there's some assumptions in that code that everything gets committed to main.  When there's a standard publish triggered, main gets merged to draft or live.  The `WebsiteContent` objects autogenerated in this PR *might* get picked up with that and included in the resources to commit to main and then merged to preview/release, which would not be good - so might need to always include a queryset that excludes those resources (by `type="website"`?) when committing to git anywhere outside of this specific task, to ensure they stay out of the main branch.

My hunch is that as long as these new resources are saved only to the preview/release branches and not main, there won't be any conflicts when merging from main, but I'm not positive, probably will require some testing to find out for sure.

mbertrand: Will also need to remove these from git if a site is unpublished

gumaerc: Thanks for the suggestions. I switched from directly calling the Github backend to calling the generic one. I experimented with pushing the content directly up to `preview` / `live`, and the solution ends up being very messy. I wasn't even able to quite get it to work, I'm not sure exactly why but it had something to do with the `ContentSyncState` not evaluating in a way that the content would be included to commit, even after clearing said `ContentSyncState` objects. I'm somewhat stuck on this, but I'm thinking it might be a better idea to create a new branch in each content repo for syncing these changes and always push the content up there. In this case, the branch would be called something like `websites`. Then, we would add a new optional argument to the `publish_website` function to specify a source branch. I'm thinking it would go something like this:

 - Hard reset `websites` branch to `preview`
 - Commit the "website" content directly to the `websites` branch based on the current contents of the DB
 - Merge `websites` branch to `preview`
 - Repeat the process for `live`

In theory, setting `draft: true` on websites that have only been published in draft should allow us to exclude those websites from the index if `--buildDrafts` is not enabled.

mbertrand: I don't recall exactly how the draft toggle works, but If we use `draft: <true/false>` to differentiate between `website` content objects that have been published to draft vs live, can we just push to the main branch for this type of content like all the others?  

`draft: true`: include in the draft list of courses only
`draft: false`: include in the draft and live list of courses

gumaerc:
@pdpinch @Ferdi I reworked this PR based on a conversation Ferdi and I had yesterday. To summarize:

 - The bulk task, `update_websites_in_root_website` is now not set to be run automatically, rather manually through a management command. This will update all websites that have been published at least once in the root website's website listing. We'll mostly need it to backpopulate the content after this PR is released.
 - Another task for updating a single website in the root website's listing was created, `update_website_in_root_website`, which is intended to be set in `PREPUBLISH_ACTIONS`. This way, before any site is published that isn't the root website, the "website" type `WebsiteContent` object in the root website will be updated.

This avoids us having the need for a constantly running task, as it's mostly unnecessary with the frequency new OCW sites are added. Having it set as a `PREPUBLISH_ACTION` means that it will just happen as part of the publish process. Still left to do is to make sure that sites are removed when unpublished.

mbertrand (changes requested):
Some initial comments, still going through it, was delayed by my local theme pipeline being ornery again.

content_sync/tasks.py
```diff
@@ -392,3 +394,119 @@ def check_incomplete_publish_build_statuses():
...
+                        filename=website.short_id,
+                        is_page_content=True,
+                        metadata=site_metadata,
+                    )
+                )
+        website_content = WebsiteContent.objects.filter(
+            website=root_website, type="website"
+        )
+        backend = api.get_sync_backend(website=root_website)
+        backend.sync_all_content_to_backend(query_set=website_content)
```
mbertrand, line 453: I ran `docker-compose run web python manage.py update_websites_in_root_website` after modifying a site's metadata (title), and the corresponding `website` ContentFile was updated in the database, but no new commit was made to git.  I suspect it's because the signal to update ContentSyncState might not have been triggered by the bulk update?

content_sync/tasks.py
```diff
@@ -392,3 +394,119 @@ def check_incomplete_publish_build_statuses():
...
+        root_website = Website.objects.get(name=settings.ROOT_WEBSITE_NAME)
+        website_content = WebsiteContent.objects.get(
+            website=root_website,
+            type="website",
+            title=website.title,
+            dirpath=WEBSITE_LISTING_DIRPATH,
+            filename=website.short_id,
+            is_page_content=True,
+        )
+        website_content.delete()
```
mbertrand, line 512: Do a sync here? The content is not deleted from git.

```python
        backend = api.get_sync_backend(website=root_website)
        backend.sync_all_content_to_backend(
            query_set=WebsiteContent.objects.filter(text_id=website_content.text_id)
        )
        ```

File: content_sync/tasks.py [hunk 4]
```diff
@@ -392,3 +394,119 @@ def check_incomplete_publish_build_statuses():
...
+                )
+        website_content = WebsiteContent.objects.filter(
+            website=root_website, type="website"
+        )
+        backend = api.get_sync_backend(website=root_website)
+        backend.sync_all_content_to_backend(query_set=website_content)
+
+
+@app.task(acks_late=True)
+def update_website_in_root_website(website, version):
```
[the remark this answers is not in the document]
gumaerc: I tried various ways of accomplishing this all day today and hit a brick wall of circular imports at every turn. `content_sync.api` calls a number of functions from `websites.api` already, so importing anything from `content_sync` causes a circular import error. I tried creating a new `tasks.py` file in `websites`, adding the `update_website_in_root_website` and `remove_website_in_root_website` as simple functions rather than tasks inside `websites/api.py`. The only way I could get around the error was to modify `update_website_in_root_website` to get the sync backend and the `publish_website` function manually with `import_string`, but then the function became untestable because at that point you can't mock getting the sync backend. Overall, I don't think it's worth the trouble to try and call it during `update_website_status` as even if the pipeline fails and the `WebsiteContent` object is created, the pipeline will likely be fixed eventually and the site published. If the site is unpublished, the `WebsiteContent` object will be removed anyway. If you can think of something stupid I might be doing wrong, or a better way to accomplish this while avoiding the circular import please let me know.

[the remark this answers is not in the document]
mbertrand: I'm fine with leaving it as is then, shouldn't happen often and as you say the pipeline should successfully complete eventually.

gumaerc:
@mbertrand This is ready for another look. As discussed, we will continue running `update_website_in_root_website` as a prepublish action due to the circular import issues with calling the `content_sync` api from the `websites` api. I think I was able to address everything else you mentioned and did some more extensive testing with unpublishing a site, republishing it to draft, then back to production. I found that I had issues with `IntegrityError` on the `update_or_create` call unless I limited the keyword args to `website`, `dirpath` and `filename`. This was due to the fact that `WebsiteContent` uses `django-safedelete` as well as the fact that it has this `UniqueConstraint`:

```
            UniqueConstraint(
                name=CONTENT_FILEPATH_UNIQUE_CONSTRAINT,
                fields=("website", "dirpath", "filename"),
                condition=Q(is_page_content=True),
            ),
```

Calling `update_or_create` with any more parameters beyond what's specified here would cause `update_or_create` to try and create a new object, but if that object had been deleted the constraint would still throw an error. This is because `safedelete` prevents the object from being hard deleted. `update_or_create` properly chooses to "create", but the unique constraint is still triggered even though the object is soft deleted. Changing the keyword args to exactly match the constraint solved this.

content_sync/tasks.py
```diff
@@ -392,3 +398,120 @@ def check_incomplete_publish_build_statuses():
...
+                "is_page_content": True,
+                "metadata": get_website_in_root_website_metadata(website, version),
+            },
+        )
+        backend = api.get_sync_backend(website=root_website)
+        backend.sync_all_content_to_backend(
+            query_set=WebsiteContent.objects.filter(text_id=website_content.text_id)
+        )
+        if not root_has_unpublished:
+            api.publish_website(root_website.name, version, trigger_pipeline=False)
```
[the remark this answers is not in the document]
mbertrand: One potential complication if the root site pipeline should trigger, but the pipeline for a new individual site hasn't completed yet (or fails), is that the root site pipeline might fail because it can't find the json for that site?  I got this error after publishing a new local site then manually publishing the root site:

```
ERROR 2023/03/16 14:59:39 Failed to fetch new course info from https://ocw.mit.edu/courses/3-16-2-march-16-site-2-summer-2027/data.json
```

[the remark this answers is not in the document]
gumaerc: > with `trigger_pipeline=False`, the git repo for the root site will be synced but it won't actually be published. Is that the desired behavior? The issue seems to infer that it should be published if there are no other changes.

The idea here is to get the markdown into Github, but the pipeline to build the live root website does not need to be triggered. Since the markdown files generated during this process have:

```
_build:
  list: true
  render: false
```

This means they will show up in lists during the Hugo build such as `site.Pages`, but no HTML or JSON will show up in the output based on them. This content will be used to build a paginated list of courses in an offline `ocw-www` site.



> One potential complication if the root site pipeline should trigger, but the pipeline for a new individual site hasn't completed yet (or fails), is that the root site pipeline might fail because it can't find the json for that site? I got this error after publishing a new local site then manually publishing the root site:
> 
> ```
> ERROR 2023/03/16 14:59:39 Failed to fetch new course info from https://ocw.mit.edu/courses/3-16-2-march-16-site-2-summer-2027/data.json
> ```

As far as this issue, can you try reproducing it on `master`? I don't believe it is related to this work. What I described above has nothing to do with the JSON that's fetched at build time by the live `ocw-www` site. When an individual course site is built, `data.json` files are generated and deployed as part of the build and the `www` theme queries these during its build based on an API response from `ocw-studio` about which courses have been published. What you're experiencing here is that you likely have `STATIC_API_URL` set to https://ocw.mit.edu/ in your env file. This gets pushed up to pipelines in your local Concourse and when building `ocw-www`, this is where these JSON files will be requested from. Since this course only exists in your local database, the request will fail every time. If you set `STATIC_API_URL` to http://10.1.0.102:8045/ then in theory you could have a pipeline hit your locally running `ocw-studio` nginx for this data, but unless you have published every site that it would be looking for then you will get errors like you posted above.

mbertrand (approved):
LGTM 👍

Issue #1706, opened 2023-03-03
Stop scheduled sync of gdrive to studio


pdpinch:
Currently, we have a scheduled task to sync gdrive to Studio, even if the user hasn't initiated a sync. 

This is surprising and unpredictable to course authors who are trying to carefully edit files, and filenames and get them into the correct place. 

We should only sync gdrive to studio on request.

PR #1709, opened 2023-03-07
Remove automatic sync of Google Drive to Studio

pt2302:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1706.

#### What's this PR do?

The environment variable `DRIVE_IMPORT_RECENT_FILES_SECONDS` controls how frequently the scheduled sync of Google Drive to OCW Studio occurs; the previous default was 3600 (once per hour). This PR checks if the value of the variable is `None`, and does not run the task in that case. The functionality to run this task periodically is off by default, but can still be enabled by the user in the `.env` file.

Set `DRIVE_IMPORT_RECENT_FILES_SECONDS` to a small value (for example, 60) in the `.env` file, change some files in a course's Google Drive folder without manually syncing, and check that they have been updated in Studio. Now, remove the `DRIVE_IMPORT_RECENT_FILES_SECONDS` variable from the `.env` file, restart the Studio containers, change some files in a course's Google Drive folder without manually syncing, and verify that they have not been updated.

### DIFF
--- a/gdrive_sync/tasks_test.py
+++ b/gdrive_sync/tasks_test.py
@@ -235,10 +235,13 @@ def test_import_recent_files_videos(
                 )
         else:  # chained tasks should be run
             mock_process_func.assert_any_call(LIST_VIDEO_RESPONSES[i]["files"][0]["id"])
-            assert tracker.last_dt == datetime.strptime(
-                LIST_VIDEO_RESPONSES[0]["files"][0]["modifiedTime"],
-                "%Y-%m-%dT%H:%M:%S.%fZ",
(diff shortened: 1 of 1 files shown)


mbertrand (commented):
If @pdpinch is confident that we don't ever want to do this, it might be simpler to just remove the `import_recent_files` task altogether along with the associated setting and beat schedule entry.

pdpinch:
We should remove this. It's completely unpredictable for users, and we don't want to surprise them.

Issue #1703, opened 2023-03-02
Limit pre-existing captions to compatible file types


pdpinch:
As a course author, it's very confusing that I can supply a pre-existing caption in the .srt format, and it will be added to my video, but it doesn't work with our video player. 

In #1656 we added a method for syncing pre-existing caption files from gdrive and adding them to the appropriate video resource. This code appears to work with any file extension, which gives course authors the impression that .srt files will work. Unfortunately, our video player will only work with .vtt files. 

We should limit the pre-existing captions to compatible file formats, which for now are .vtt and .webvtt

PR #1711, opened 2023-03-08
Allow only vtt and webvtt extensions for pre-existing captions

pt2302:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1703.

#### What's this PR do?
This PR limits pre-existing captions to `.vtt` and `.webvtt` formats; any other file types will not be associated with a given video. Note that we do not currently check that the captions are actually a valid VTT file.

Follow the same testing instructions as for https://github.com/mitodl/ocw-studio/pull/1683, but now try with both `.srt` extension (or another format) and `.vtt` extension captions. The `.srt` extension captions should not be associated with the video, but the `.vtt` extension captions should be.

### DIFF
--- a/videos/models.py
+++ b/videos/models.py
@@ -6,6 +6,7 @@ from django.db.models import CASCADE
 from mitol.common.models import TimestampedModel
 
 from main import settings
+from main.utils import get_file_extension
 from videos.constants import (
     DESTINATION_YOUTUBE,
     VideoFileStatus,
@@ -18,7 +19,7 @@ from websites.utils import get_dict_query_field
 
 
 class Video(TimestampedModel):
-    """ Video object"""
+    """Video object"""
 
     def upload_file_to(self, filename):
         """Return the appropriate filepath for an upload"""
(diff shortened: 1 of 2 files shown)


gumaerc (approved):
:+1: 

This does what it says on the tin, but I did notice something during testing. I'm not sure if we need to fix this or not, but this is what happened:

 - Uploaded Video with PDF transcript and SRT captions, synced and spoofed YouTube upload to get studio to associate the captions & transcript
 - SRT subtitles were not associated with the video, as expected
 - VTT subtitles matching the video filename uploaded into `files_final`, gdrive sync run
 - YouTube upload spoofed again, waited for video processing task to run
 - VTT subtitles not associated with video
 
 I tried again from scratch with just VTT subtitles and everything seemed to work, and I think in this scenario you can likely upload / sync the VTT subtitles and then edit the path manually on the video. I'm not sure if the VTT subs should be associated after the fact if you try again like that, but I figured I'd share my experience.

PR #1713, opened 2023-03-10
Fakhar/1646 exclude assests offline

### LINKED ISSUE #1646: Exclude unnecessary assets in offline build
The offline course builds currently all webpack assets, including `www`, `course`-(v1), and, `course-v2`. If https://github.com/mitodl/ocw-hugo-themes/pull/1042 is merged, source maps for the js/css files will be included, too, which adds about 9MB (uncompressed).

It would be nice to not download unnecessary assets, especially the source maps.

_Comment: 9MB is comparable in size some PDFs, so maybe not worth worrying about. On the other hand, this does affect all courses._

fakhar-ud-din:
#### Pre-Flight checklist

- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested

Closes #1646

#### What's this PR do?
It helps to filter .js.map extension files to be excluded from the offline build of a course

Will add the steps for testing after a meeting with @ChristopherChudzicki 
Getting some other issues while publishing a course

### DIFF
--- a/content_sync/pipelines/concourse_test.py
+++ b/content_sync/pipelines/concourse_test.py
@@ -690,7 +690,7 @@ def test_upsert_mass_build_pipeline(
         assert "$SHORT_ID.zip" in config_str
         if settings.ENVIRONMENT == "dev":
             assert (
-                f"STUDIO_S3_RESULT=$(aws s3{endpoint_url} sync s3://{settings.AWS_STORAGE_BUCKET_NAME}/$S3_PATH ./content/static_resources --exclude *.mp4 --only-show-errors) || STUDIO_S3_RESULT=1"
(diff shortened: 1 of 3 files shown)


gumaerc (changes requested):
This will have the effect of filtering out sourcemaps in single course offline builds, but in order for this to be effective we also need to make this change in the `mass-build-sites.yml` pipeline: https://github.com/mitodl/ocw-studio/blob/50d7c85808eac8c6f7f258e85fe72cdc4e58dbdd/content_sync/pipelines/definitions/concourse/mass-build-sites.yml#L218

To give some background, when we initially were testing using Concourse to build OCW sites, all we had were the individual pipelines and a "mass publish" would simply trigger all of these pipelines at the same time. This had the effect of overloading various parts of Concourse. Each pipeline contains several steps, and each step runs inside its own Docker container, which means that triggering the 2600+ OCW courses to build all at once means that 2600 * n number of steps Docker containers are spun up simultaneously in Concourse. Our solution to this was to create a separate pipeline specifically for doing a mass build of courses, which builds all of the courses in a loop inside a single Docker container. The pipeline definition for this is in the file linked above, `mass-build-sites.yml` and will also need to be updated to filter sourcemaps as you've already done in `site-pipeline.yml`.

gumaerc (approved):
:+1: LGTM!

PR #1670, opened 2023-02-06
Update Missing Captions/Transcripts

pt2302:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested
  - [X] Changes have been manually tested
  
#### What are the relevant tickets?
Follow up to https://github.com/mitodl/ocw-studio/pull/1612 and https://github.com/mitodl/ocw-studio/pull/1665.

#### What's this PR do?
Updates the management command for syncing captions and transcripts for any videos missing them from one course (`from_course`) to another (`to_course`). The original management command assumed that there was a single YouTube ID <-> video pair in each course; this update handles the case when there might be multiple such pairs and avoids generating duplicate files in the course's S3 bucket. The management command now first checks whether there are any existing videos in the course that match the YouTube ID of the video that is missing captions or a transcript, and only creates new captions/transcript files if there is no such match.

#### How should this be manually tested?
For testing locally, make sure that Minio is properly configured. Also, make sure that `RESOURCE_BASE_URL_LIVE` is not set in your `.env` file for OCW Studio. Copy the RC or production data for `18-02-multivariable-calculus-fall-2007` and `18-02sc-multivariable-calculus-fall-2010` from S3 to your local Minio instance, and then run `docker compose run --rm web ./manage.py sync_transcripts --from_course 18-02-multivariable-calculus-fall-2007 --to_course 18-02sc-multivariable-calculus-fall-2010`.

Then, run `docker compose run --rm web ./manage.py reset_sync_states --filter 18-02sc-multivariable-calculus-fall-2010 --skip_sync` and `docker compose run --rm web ./manage.py sync_website_to_backend --filter 18-02sc-multivariable-calculus-fall-2010`. Then, publish the course site using the OCW Studio UI. Verify that the captions and transcript are properly linked in the published course, for example http://localhost:8045/courses/18-02sc-multivariable-calculus-fall-2010/resources/clip-total-differentials-and-chain-rule should have both captions and transcript.

PR #1717, opened 2023-03-20
Sync missing captions and transcripts

### LINKED ISSUE #1518: No captions, no transcript on 3.091 fall 2018
### Steps to Reproduce

See, for example, https://ocw.mit.edu/courses/3-091-introduction-to-solid-state-chemistry-fall-2018/resources/lecture-1/

or https://ocw.mit.edu/courses/3-091-introduction-to-solid-state-chemistry-fall-2018/resources/carbon-dioxide-concentration-lec28/

### Expected Behavior

- There should be an interactive transcript player. 
- There should be a button to turn on captions 

### Actual Behavior

- no transcript
- no captions 

### Screenshot or Screencast

image

fakhar-ud-din:
#### Pre-Flight checklist

- [x] Testing
  - [ ] Code is tested

Closes #1518

#### What's this PR do?
Syncs captions and transcripts for missing videos
- Filter videos not having captions or transcripts
- Fetches respective file from 3play
- Create new WebsiteContent 
- Update existing video with new content object

   - **Threeplay variables** 
   - **AWS variables**   
-  Checkout to this branch
- **Testing all courses**
   - Run command `docker exec -it ocw-studio-web-1 python3 manage.py sync_management_command`
- **Testing single course**
   - Run command `docker exec -it ocw-studio-web-1 python3 manage.py sync_management_command --filter="name-of-the-course-year,"`
- You can stop the running command after a few minutes (Optional)
- Head to `http://localhost:8043/sites/`
- Search for the course name you provided
- Under Course Resources, and new transcripts/ captions resources should be present
- Open any resource, and you should see captions/transcripts file paths (previously empty)

### Example of captions/ transcript conent objects and paths
Screenshot 2023-04-06 at 5 09 33 PM
Screenshot 2023-04-06 at 5 09 46 PM


fakhar-ud-din:
I think we should not merge both due to a couple of factors,
- Underlying logic for methods is different, Syncing through 3play and Syncing through another course
   - Aforementioned uses different metadatas to be stored for the new WebsiteContent objects
   - Both uses different object defaults for the new WebsiteContent objects
- Merging both would unnecessarily complicate the code
- The practical benefit would be very minimal over the changes required to integrate both

pdpinch:
@fakhar-ud-din you mention the differences but not the similarities. Are there any functions that should be refactored into a utility library?

Also, can you explain the differences more? You say "different metadata" and "different object defaults"

fakhar-ud-din:
The only similarities between both are, both loop over some course videos contents and update existing video content with new file paths using a method "create_new_content". Not sure if we can add any method in the utility library, I will check to see if there could be one.

Sure I will explain the differences in more details @pdpinch,
- Different metadata
   - We make new WebsiteContent object (Video file content, PDF file content, SRT file content), which essentially stores a field "metadata" containing information about the content file. So for example, a pdf file will have the field "file_type" in "metadata" as "application/pdf" and an srt file would have the same file type as "application/x-subrip"
   - Whereas the other PR, #1670, utilises the  "metadata" field of another "Video file" and copies it into a new object. Which has different "metadata" fields than what a PDF or SRT file would have.
- Different object defaults
   - When a new WebsiteContent object is made, it has a few other fields other than "metadata". For example, "title", "type", "filename", "text_id"
   - The pdf and srt files stores titles as "3play transcript file" and "3play captions file" respectively, where as the other PR, uses the title used for source Video object (It uses a source and destination video objects to essentially copy legacy data to new object)
   - Similarly, the #1670 fix uses existing "filepath" for the new WebsiteContent obj "file" field, whereas we now make and store a new File object for the "file" field

pdpinch:
> For object defaults, it might make sense to standardize these in general. For example, do we want to use "3play transcript file" or use a more informative title for the transcript resource? (@pdpinch)

This is a good point. "3play transcript file" is a horrible title. Ideally the title for the caption and transcript resources would be based off of the video resource title. And there's no need to mention "3play" in the title.

fakhar-ud-din:
I think my previous comment might have caused a bit of confusion, and I am sorry about that.

- By SRT files, I meant the VTT files, since they both are related to subtitles, I used a term which was more known.
And I completely agree to move s3 part to utils @pt2302, and I will incorporate that change here.

- Whereas, the WebsiteContent objects have different type of resources pdf, vtt, video.  #1670, <-- This utilises an existing WebsiteContent object, of Video metadata, and copies it to a new WebsiteContent. Whereas, the current PR uses entirely different metadata having different fields.
- Hence, merging the 2 logics, would complicate the code.

- I was under the impression to keep the naming consistent for the objects defaults, therefore, I used the legacy titles, i.e "3play transcript file"


I will standardise the things we discussed and re-evaluate if merging the 2 logics would make sense.

Thank you for your suggestions @pt2302 and @pdpinch

fakhar-ud-din:
Hey @pt2302 

I have moved a few methods in the utils and made necessary changes.
But, I am still not quite confident about merging the 2 commands. Despite the similarities, the differences would  still require some code to be duplicated.
For example, check the the metadata being used in both,

### WebsiteContent (Video Resource Metadata)
Screenshot 2023-03-28 at 6 43 14 PM

### WebsiteContent (VTT/PDF Resource Metadata)
Screenshot 2023-03-28 at 6 44 39 PM

Let me know if you think I am missing something.

Thank You

fakhar-ud-din:
Hey @pt2302 
I have added the testing step in the description. 
Let me know if I missed anything.

Thank You

fakhar-ud-din:
Hey @pt2302, testing steps have been updated for a single course.
Thank You

fakhar-ud-din:
@pt2302 I have updated the instructions. I missed those since I already had those in the .env prior to this ticket while testing some other things.
I have also added a message in a course does not exist.

Regarding the error you mentioned, I couldn't replicate that locally and the command seems to run fine on my end. Could you confirm if you are using a fresh database dump?

fakhar-ud-din:
That's strange. I will make a fresh env copy and work with that and will see if that works.
I will provide an update after trying that out

gumaerc:
@fakhar-ud-din @pt2302 

So, there is an underlying issue here. As you're both aware, in local dev we override interactions with S3 to instead go to a locally running Minio server. This is accomplished by having the utility functions we use for getting an S3 resource, which while processing the options does this:

```
    if settings.ENVIRONMENT == "dev":
        options.update({"endpoint_url": "http://10.1.0.100:9000"})
```

This means that if your environment is `dev`, any interactions with S3 that you perform on the object returned by `get_boto3_resource` will hit Minio instead. The issue here is with this code in `videos.management.commands.sync_missing_captions.create_new_content`:

```
        defaults = {
            "metadata": new_obj_metadata,
            "title": title,
            "type": "resource",
            "file": file_content,
            "text_id": new_text_id,
        }
```

Here the `file` property is being set on the `WebsiteContent` object. The `file` property is a `FileField`, and setting a value on it will cause this data to be uploaded to whatever is defined as the default strorage backend in Django settings. For us, this is the S3 backend provided by `django-storages`. If you do not get an error with the code as-is, that means that whatever you have your AWS info set to in your `.env` is pointing to a real S3 bucket and the uploads are completing successfully. We want these uploads to go to Minio instead when working locally. In another PR, we should make a change so that the `AWS_S3_ENDPOINT_URL` setting is pointing to Minio when running locally. For now, in this PR, we should omit the `file` property in the defaults above and instead use `get_boto3_resource` and upload the file to S3 directly with boto.

gumaerc:
@fakhar-ud-din This PR could also use a rebase as it was created before we merged all the dependabot PR's. It still wants to use `django-server-status` which has been removed.

fakhar-ud-din:
Thank you for the detailed feedback @gumaerc and @pt2302 
I have made minor code refactors, including the boto3 resource handling as suggested by @gumaerc 
Moreover, I have added checks and messages to ensure we download and upload required files from 3play and s3/minio respectively.



> * if a course is provided to the `filter` argument but isn't in the database, something like `Course with short ID <short id> not found`, rather than only providing an error message if _all_ courses are not found.

Regarding this part of your comment @pt2302 , We are using Django queries to filter required `WebsiteContent` objects using the filter list provided. We would have to include an extra loop to check against all provided entries in the filters, if we want to print relevant course existence messages. 
Since the current loop will work only for courses with missing captions and transcripts, and will skip any ambiguous courses, I think we should avoid adding additional loop.

fakhar-ud-din:
@pt2302 I made some changes to add display summary, I have also attached a sample output for reference below,

Screenshot 2023-04-26 at 3 16 48 PM

pdpinch:
> More fundamentally, there appears to be a potential data issue with 3Play where captions are not being tagged with the correct YouTube ID. We should figure out how prevalent this data issue is.

I asked the OCW team about this and they said this is an occasional error that can happen when they manually update 3Play. They have fixed this video in 3Play. I don't think we should let this stop us from proceeding with this PR.

pdpinch:
Now that this has merged, can we try running it on 3.091 2018 to see if it works?

pdpinch:
And/or 8.06, which has two videos missing captions: https://github.com/mitodl/ocw-studio/issues/1732

fakhar-ud-din:
@pdpinch I will run and test this in production!

PR #1720, opened 2023-03-22
Exclude JS Map assets from builds

### LINKED ISSUE #1646: Exclude unnecessary assets in offline build

fakhar-ud-din:
#### Pre-Flight checklist

- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested

Closes #1646

#### What's this PR do?
It helps to filter .js.map extension files to be excluded from the offline build of a course

- Publish a course
- Run the concourse pipelines
- Download the course from localhost:3000
- Navigate to course files, /static_shared folder
- .js.map files should not exist

### DIFF
--- a/content_sync/pipelines/definitions/concourse/mass-build-sites.yml
+++ b/content_sync/pipelines/definitions/concourse/mass-build-sites.yml
@@ -118,7 +118,8 @@ jobs:
           npm run build:webpack
           npm run build:githash
           mkdir -p ./base-theme/static/static_shared/
-          cp -r ./base-theme/dist/static_shared/* ./base-theme/static/static_shared/
+          find ./base-theme/dist/static_shared/* ! -name '*.js.map' -type f | xargs cp -t ./base-theme/static/static_shared/
+          
   # END OFFLINE-ONLY
   - task: get-sites
     timeout: 2m
@@ -215,7 +216,7 @@ jobs:
               # END ONLINE-ONLY
               # START OFFLINE-ONLY
               echo "PULLING IN STATIC RESOURCES FOR $NAME" > /dev/tty
-              STUDIO_S3_RESULT=$(aws s3((cli-endpoint-url)) sync s3://((ocw-studio-bucket))/$S3_PATH ./content/static_resources --exclude *.mp4 --exclude *.js.map --only-show-errors) || STUDIO_S3_RESULT=1
+              STUDIO_S3_RESULT=$(aws s3((cli-endpoint-url)) sync s3://((ocw-studio-bucket))/$S3_PATH ./content/static_resources --exclude *.mp4 --only-show-errors) || STUDIO_S3_RESULT=1
               if [[ $STUDIO_S3_RESULT == 1 ]]
               then
                 echo "SYNCING s3://((ocw-studio-bucket))/$S3_PATH to ./content/static_resources FAILED FOR $NAME" > /dev/tty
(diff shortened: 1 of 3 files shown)


content_sync/pipelines/definitions/concourse/site-pipeline.yml
```diff
@@ -383,15 +383,15 @@ jobs:
...
+                  find ../static-resources ! -name '*.mp4' -type f | xargs cp -t ./content/static_resources
                 fi
                 HTML_COUNT="$(ls -1 ./content/static_resources/*.html 2>/dev/null | wc -l)"
                 if [ $HTML_COUNT != 0 ];
                 then
                   mv ./content/static_resources/*.html ./static/static_resources
                 fi
                 touch ./content/static_resources/_index.md
-                aws s3((cli-endpoint-url)) sync s3://((web-bucket))/static_shared ./static/static_shared
+                aws s3((cli-endpoint-url)) sync s3://((web-bucket))/static_shared --exclude *.js.map ./static/static_shared
```
gumaerc, line 394: This `--exclude` argument needs to go at the end of the command. The `aws s3 sync` command expects 2 positional arguments after `sync`, the source and the destination. You can't place keyword arguments in the middle like this, as it produces an error.

gumaerc (changes requested):
I believe this PR needs a rebase, as it's showing that it's reverting changes that have already been reverted. Also, I noticed an error with the way the exclude argument is written in the single site pipeline:

HussainTaj-arbisoft (approved):
LGTM :+1:

content_sync/pipelines/definitions/concourse/mass-build-sites.yml
```diff
@@ -118,7 +118,8 @@ jobs:
           npm run build:webpack
           npm run build:githash
           mkdir -p ./base-theme/static/static_shared/
-          cp -r ./base-theme/dist/static_shared/* ./base-theme/static/static_shared/
+          find ./base-theme/dist/static_shared/* ! -name '*.js.map' -type f | xargs cp -t ./base-theme/static/static_shared/
```
gumaerc, line 121: So, I don't think this `find` / `xargs cp` strategy is going to work here. I tested this and ended up getting some errors:

```
cp: will not overwrite just-created './base-theme/static/static_shared/tex.js' with './base-theme/dist/static_shared/mathjax/output/svg/fonts/tex.js'
cp: will not overwrite just-created './base-theme/static/static_shared/tex.js' with './base-theme/dist/static_shared/mathjax/input/tex.js'
```

This is happening because this will only work with a flat directory structure. When we filter out the mp4 files, the `static_resources` folder is a flat structure so this works. One approach that I've considered in the past would be to add `rsync` to our `ocw-course-publisher` docker image. This would allow easy filtering with something like:

```
rsync -av --exclude='*.js.map' ./base-theme/dist/static_shared ./base-theme/static/static_shared
```

However, `rsync` is not part of our docker image. We could add it and revisit this, or alternatively for now we could do something like:

```
cp -r ./base-theme/dist/static_shared/* ./base-theme/static/static_shared/
find ./base-theme/static/static_shared -name "*.js.map" -type f -delete
```

gumaerc (changes requested):
The changes to the single course pipeline look good, I tested them and the map files were excluded. There were some issues with the mass build pipeline, however:

PR #1724, opened 2023-03-27
feat: update drive sync and allow file deletion

HussainTaj-arbisoft:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1555
Relevant https://github.com/mitodl/hq/discussions/359

#### What's this PR do?

This PR:
- Updates the website GDrive sync task
- Deletes files that no longer exist in GDrive and aren't being used on any page
  - It was necessary to delete existing files to achieve the effect expected in the linked ticket. Since I was writing the delete logic, I also considered https://github.com/mitodl/hq/discussions/359 in the implementation.
  - Note: Files that are used in content other than `page` will also be deleted.
- The rest of the details can be read from the issue comments.

Might be irrelevant but in case you wonder, I experimented with the [changes API](https://developers.google.com/drive/api/v3/reference/changes) and decided not to use it for two reasons.
1. The API does not provide any filtering options.
2. A Change can be removed from the list because of a user action. The documentation does not specify a strict set of reasons.
3. In the end, even if possible, it would've been more complicated to implement.

1. Download the folder [here](https://drive.google.com/drive/folders/1fJ0Sv5ZKy-NKCdfnLEOAcM-jrEZuMU0P?usp=sharing)
2. Head to your local ocw-studio setup
4. Checkout this branch `hussaintaj/1555-gdrive-sync`
5. Start ocw-studio and create a test course.
6. Open the folder you downloaded in step 1. Open the `step01` subdirectory. Upload these files to the associated course's GDrive folder `final_files`.
7. In ocw-studio, open the resources tab of your test course and hit `Sync w/ GDrive`.
8. **Expected Behaviour**: You should see 5 files after a successful sync.
9. Click on `image3.png` and add something in the description field. This will act as a marker that we will use later.
10. Create a page in your test course. In the body, embed `image5.png`. Save the page.
11. Finish adding the metadata for the course and then publish it to draft or live, or both.
12. Open your local instance of MinIO. Log in and navigate to your course files.
13. **Expected Result**: You should see the synced files in the `static_resources`, and content in the `resources` directories.
14. Open the GDrive folder from step 5. Delete the following files `image3.png`, `image4.png`, `image5.png`.
15. Open the folder you downloaded in step 1. Open the `step02` subdirectory.
    * Upload `image2.png`. Gdrive will ask you if you want to replace it. Say yes.
    * Upload `image3.png`.
16. Repeat step 6
17. After sync. **Expected Result**
    * `image4.png` is removed from the resources file.
    * `image5.png` is not removed. Because it was used in step 9.
    * Click `image3.png`, the description should have the content you added in step 8.
18. Go to your GDrive trash. Restore `image3.png` from step 13.
19. Repeat step 6
20. After sync. **Expected Result**: You should see two `image3.png`.
21. Publish the course again.
22. Repeat step 11 and 12. Make sure all the changes in the files are reflected properly. Note: One file will be named `image32.png`
23. That's it for this systematic testing. Feel free to go crazy and try other things that you can think of.


ibrahimjaved12:
Accidentally put `on hold`, so removed it

ibrahimjaved12 (approved):
LGTM 👍🏻

mbertrand (commented):
When videos are deleted, should the corresponding Youtube videos be deleted or made private?  Should the video files (including transcodes) and transcripts on S3 be deleted?

After I deleted a video from gdrive and resynced, the sync process status seemed to get stuck on "Processing", and the video resource did not disappear from the page until I reloaded.    Ditto for other deleted files.  Might have something to do with the changes to how the tasks are chained/grouped, but not sure.  Haven't looked closely at the code yet, will do that tomorrow.

HussainTaj-arbisoft:
Thank you for testing this.

> When videos are deleted, should the corresponding Youtube videos be deleted or made private? Should the video files (including transcodes) and transcripts on S3 be deleted?

Ideally, yes, they should be deleted since we would have no way to track them after a resource is deleted. I'm triggering the delete logic that already existed, I'll take a look at the implementation to verify if youtube videos and transcripts are deleted.

HussainTaj-arbisoft:
> > When videos are deleted, should the corresponding Youtube videos be deleted or made private? Should the video files (including transcodes) and transcripts on S3 be deleted?
> 
> Ideally, yes, they should be deleted since we would have no way to track them after a resource is deleted. I'm triggering the delete logic that already existed, I'll take a look at the implementation to verify if youtube videos and transcripts are deleted.

I did a bit of digging, the youtube video should be deleted with the video object thanks to the:

https://github.com/mitodl/ocw-studio/blob/b6c2a4df416d18afdfd45456892b850d582c292b/videos/signals.py#L11

If captions or transcripts are added manually through drive sync, those will be deleted when the user deletes the files from the drive.

For 3play generated transcripts, I'm not sure if they are being deleted. I'll look into it a bit more.

gdrive_sync/api.py
```diff
@@ -121,6 +121,80 @@ def query_files(query: str, fields: str) -> Iterable[Dict]:
             yield file_obj
 
 
+def _get_or_create_drive_file(
+    file_obj: Dict,
+    drive_path: str,
+    website: Website,
+    sync_date: Optional[datetime],
+    replace_file_with_the_same_name: bool = True,
```
mbertrand, line 129: Maybe shorten to just `replace_file`?

gdrive_sync/tasks.py
```diff
@@ -100,18 +112,17 @@ def import_recent_files(self, last_dt: str = None):  # pylint: disable=too-many-
...
+    """
+    Returns a tuple (files, errors).
+
+    `files` is a dict where keys are subfolder names and value is a
+    list of file objects.
+    `errors` is a list of errors while fetching files.
+    """
     errors = []
-    tasks = []
+    gDriveSubfolderFiles = {}
```
mbertrand, line 124: `gdrive_subfolder_files`

gdrive_sync/api.py
```diff
@@ -121,6 +121,80 @@ def query_files(query: str, fields: str) -> Iterable[Dict]:
...
+        "sync_dt": sync_date,
+    }
+
+    if existing_file:
+        for k, v in file_data.items():
+            setattr(existing_file, k, v)
+        existing_file.save()
+        drive_file = existing_file
+    else:
+        existing_file_same_path = DriveFile.objects.filter(
```
mbertrand, line 173: It might be good to come up with some different names to clarify what the difference is between `existing_file` and `existing_file_same_path`.  A couple ideas:

`existing_file_same_id` and `existing_file_same_path`
`existing_file` and `other_file_same_path`

HussainTaj-arbisoft:
I've added a receiver to delete transcript files from Video file fields.

I don't think we can delete entities from 3play since it provides no API from what I can tell from the [docs](https://docs.3playmedia.com/apiv3/cheat_sheet).

I don't see any other location a transaction/caption might be kept.

HussainTaj-arbisoft:
Thank you for the review @mbertrand. I've implemented your suggestions and transcript deletion (as described here https://github.com/mitodl/ocw-studio/pull/1724#issuecomment-1511177607). Please have another look.

I've tried to cover as many cases as I could think of. It's a complex piece of code and we want to make sure we don't miss something. I'd appreciate it if you could specifically test the video files to the best of your knowledge.

mbertrand (approved):
Code looks great and functionality works (including w/videos) - 👍

Issue #1755, opened 2023-04-19
backpopulate_archive_videos doesn't handle archive_url that starts with http://


gumaerc:
### Steps to Reproduce

 - Find a site with videos that have an `archive_url` that starts with `http://` and not `https://`
 - Run `./manage.py backpopulate_archive_videos --bucket ocw-website-mirror --prefix OcwExport/InternetArchive --filter <site.name here>`

### Expected Behavior

The archive URL prefix is properly removed before trying to copy the archive video

### Actual Behavior

The prefix is not removed and the copy fails

pdpinch:
Should we just update the data for these? There's no excuse for using `http` anymore.

PR #1743, opened 2023-04-11
backpopulate archive videos

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested
  - [ ] Changes have been manually tested

#### What are the relevant tickets?
Closes https://github.com/mitodl/ocw-studio/issues/1739

#### What's this PR do?
This PR adds a new management command, `backpopulate_archive_videos`, along with a batchable task behind it that does the actual work involved. The management command takes all `Website` objects, optionally filtered with `--filter` and does the following:

 - For each `Website`, find `WebsiteContent` objects where `metadata.video_files.archive_url` is not null
 - For each `WebsiteContent` object, find the associated video in the bucket specified by the `--bucket` argument and optionally the `--prefix` argument based on the path portion of `metadata.video_files.archive_url`
 - Sync the video we found to the buckets stored in these settings:
   - `AWS_STORAGE_BUCKET_NAME`
   - `AWS_PREVIEW_BUCKET_NAME`
   - `AWS_PUBLISH_BUCKET_NAME`
   - `AWS_OFFLINE_PREVIEW_BUCKET_NAME`
   - `AWS_OFFLINE_PUBLISH_BUCKET_NAME`

This work is being done as part of https://github.com/mitodl/hq/discussions/1197 and the reason we are moving the videos into all these different buckets is to truly import them into the `ocw-studio` system, as if the videos had been uploaded by content authors. Putting them into the storage and preview / publish buckets prevents a large sync from having to happen on each site after this command is run for the first time, as all the videos will already be in place. This allows us to deliver videos from legacy sites in mirror drives without having to conditionally pull the videos from a different bucket at build time depending on if the sites is considered "legacy" or not.

#### How should this be manually tested?
 - Make sure you have the following prerequisites set up (if unsure, check the readme):
   - Recent DB restore from production or RC with sync states and gdrive folders reset
   - Local Concourse integration configured and running
   - Local Minio integration configured and running
   - Theme assets pipeline pushed up to Concourse and run at least once
   - In your `.env`, these are the settings I used minus the minio user / password and aws access keys:
```
AWS_STORAGE_BUCKET_NAME=ol-ocw-studio-app
AWS_PREVIEW_BUCKET_NAME=ocw-content-draft
AWS_PUBLISH_BUCKET_NAME=ocw-content-live
AWS_ARTIFACTS_BUCKET_NAME=ol-eng-artifacts
RESOURCE_BASE_URL_DRAFT=https://draft.ocw.mit.edu
[...]

PR #1756, opened 2023-04-19
support irregular values in archive_url in backpopulate_archive_videos

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1755

#### What's this PR do?
This PR fixes an issue with `backpopulate_archive_videos` that improperly handles an `archive_url` that doesn't start with `https://archive.org/download/`. It was found after https://github.com/mitodl/ocw-studio/pull/1743 was merged that there are some alternative prefixes used in `archive_url` that make simple string replacement / regex matching impratcial. This PR revises that with a strategy that uses `urllib.parse` and `os.path` to fetch the filename and immediate parent folder and then join them, that way it works regardless of whatever the rest of the URL is.

Follow the steps in https://github.com/mitodl/ocw-studio/pull/1743, but in the second step use `MIT6.041SCF13` in the sync command and `6-041sc-probabilistic-systems-analysis-and-applied-probability-fall-2013` for your `--filter` argument in the 4th step.


mbertrand (approved):
👍

PR #1770, opened 2023-05-03
fix: IntegrityError - duplicate key while creating resources

HussainTaj-arbisoft:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1702

#### What's this PR do?

This PR:
- Changes how files are named. Now files will be named `filename_extX` instead of `filenameX` (where X is an optional collision prevention counter). As discussed here: https://github.com/mitodl/hq/discussions/391
- Changes the order in which WebsiteContent i.e. resources are created. Ealier resources for drive_files were created in parallel. Now we create all the resources in a separate task. Reason here: https://github.com/mitodl/ocw-studio/issues/1702#issuecomment-1453574957

(Required)

### DIFF
--- a/gdrive_sync/api.py
+++ b/gdrive_sync/api.py
@@ -426,6 +426,7 @@ def create_gdrive_resource_content(drive_file: DriveFile):
         resource_type = get_resource_type(drive_file)
         resource = drive_file.resource
         basename, extension = os.path.splitext(drive_file.name)
+        basename = f"{basename}_{extension.lstrip('.')}"
         if not resource:
             site_config = SiteConfig(drive_file.website.starter.config)
             config_item = site_config.find_item_by_name(name=CONTENT_TYPE_RESOURCE)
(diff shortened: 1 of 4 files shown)


HussainTaj-arbisoft:
@mbertrand thank you for the review.

> ... I'm assuming this is okay but wanted to mention it just in case.

Yes, this is the behavior that currently exists in production. So it does not fall under the scope of this ticket.

Also, I believe this is the correct behavior, or at least it used to be. Before we implemented the deletion of resource if a user manually deleted and replaced a file in GDrive, the new resource would get a counter appended to the filename in Studio. And the UI rightly showed the original name `file.txt`.

It is true that it might be confusing for the user if the downloaded filename is different than the one shown on the screen. However, it is equally confusing for the author if the filename on the UI doesn't match the one they uploaded.

I guess this behavior needs some user feedback and requirement engineering. (@pdpinch perhaps you might be interested.)

pdpinch:
@mbertrand from your screenshot, it looks like you're commenting on the resource title, which is user editable. 

If you open up the resource in Studio, you'll see that the title and the file name are pretty clear. And the course author can change them if they want to. [For example](https://ocw-studio-rc.odl.mit.edu/sites/peters-test-site-for-unpbulishing/type/resource/edit/87ea1d9a-e631-4882-b10e-8d8cf12e2cf8/?q=duplicate): 

image

I think this is fine as is, as long as it didn't require @HussainTaj-arbisoft to write a lot of logic.

PR #1772, opened 2023-05-04
Sync missing captions - File seek(0)

fakhar-ud-din:
#### Pre-Flight checklist

- [x] Testing
  - [ ] Code is tested

#1717

#### What's this PR do?
We seek BytesIO file object to start

   - **Threeplay variables** 
   - **AWS variables**   
-  Checkout to this branch
- **Testing all courses**
   - Run command `docker exec -it ocw-studio-web-1 python3 manage.py sync_management_command`
- **Testing single course**
   - Run command `docker exec -it ocw-studio-web-1 python3 manage.py sync_management_command --filter="name-of-the-course-year,"`
- Head to `http://localhost:9001` (Local Minio)
- Go to `ol-ocw-studio-app/courses/<course-name>`
- You should see new captions and transcripts file with some `Size` value (previously empty **-**)

Screenshot 2023-05-04 at 12 58 02 PM

### DIFF
--- a/videos/threeplay_api.py
+++ b/videos/threeplay_api.py
@@ -147,6 +147,8 @@ def fetch_file(source_url: str) -> BytesIO:
 
     file = BytesIO()
     file.write(response.content)
+    file.seek(0)
+
     return file
 
 


pdpinch:
How/can we test this with S3, so we can know this works?

fakhar-ud-din:
When we check out local storage for newly created captions and transcripts file, their size should be some value instead of 0/ or empty.
When we were using `fetch_file` method earlier, it wrote the response content to the file, and moved the "file cursor" at the end, causing other operations to read "empty" content and eventually generating empty file.

With the current fix, we move back the cursor to the start of the file, so when make a File object now, it would have the data which was not the case prior to this.

gumaerc (approved):
After personally debugging this issue yesterday and coming to the same conclusion that the `BytesIO` object needs to be seeked back to 0, I'll approve this PR. However, the testing instructions have some inaccuracies that would make it hard to review this PR for someone without such context:

 - The management command in question is `sync_missing_captions` not `sync_management_command`
 - If you run the command unfiltered, it will sync transcripts for every course in your database. Running a filtered command afterward would do nothing, because all the missing transcripts would have been pulled in
 - What files am I looking for in Minio, specifically? What I did was watch the terminal output of the command with my log level set to `DEBUG`, watching for transcript / caption files that were being updated, then checking for those in Minio. Someone without context would not know what files they're looking for, especially since this is a follow-up PR to https://github.com/mitodl/ocw-studio/pull/1717. You can see in your screenshot that you're looking at a `_captions.webvtt` file, but what about the `_transcript.pdf` file?
 - On that note, you should link PR's that provide previous context into follow-up PR's like this, so someone without context can gain some by reading what was worked on previously

fakhar-ud-din:
@pt2302, I added the updated instruction for testing this in the PR's description. I have verified the fix against already published course `3.091-fall-2018` and also included a screenshot regarding this.

The old captions and transcript files had no size (empty), whereas, when we run the management command after this fix, the new captions and transcript files have a size value, implying the files are not empty.

If I misunderstood, please let me know, and I will update the PR accordingly.

Thanks

gumaerc:
@fakhar-ud-din You can go ahead and merge this. @pt2302 made that comment before you updated the instructions, and my comment gave you more specific direction on how the instructions were confusing. Another thing to note is than running `docker exec -it etc.` you can use Docker Compose shorthand:

```
Instead of:

docker exec -it ocw-studio-web-1 python3 manage.py sync_missing_captions

Use:

docker-compose exec web ./manage.py sync_missing_captions
```

Notice how I'm also using the `./` shorthand to execute the python file in the shell directly rather than manually calling `python3`. This ensures that the default Python interpreter on any given container you might be running this on will run the command. There's nothing wrong with the way you're executing it, it's just a little unnecessarily verbose.

Issue #1779, opened 2023-05-05
AccessDenied error when checking webpack-json


gumaerc:
### Steps to Reproduce

Push a pipeline up with `./manage.py backpopulate_pipelines`

### Expected Behavior

The pipeline runs successfully

### Actual Behavior

An error is triggered when checking `webpack-json` in Concourse QA

### Stacktrace

```
selected worker: ip-10-1-2-100
error finding versions: AccessDenied: Access Denied
	status code: 403, request id: 6MWSS5T9YMSKTS4C, host id: uYVloxUILEEPpcMXB+TYzxuW5HCtPMnUhPvUG41nH8kQGUpGiMIlt/tSsZgNMSDLmvOZPJ4gM7k=
```

PR #1763, opened 2023-04-26
separate online and offline parts of site-pipeline into separate jobs

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested
  - [ ] Changes have been manually tested

#### What are the relevant tickets?
Closes https://github.com/mitodl/ocw-studio/issues/1764

#### What's this PR do?
This PR takes the online and offline parts of `site-pipeline.yaml` and separates them into two separate jobs. A resource of type `keyval` was added, which is used as a simple gate between the jobs. At the end of the online build, the `timestamp` value in `offline-build-gate` is updated, which triggers the offline build.

There is also a change to how the mp4 filtering works. In the online build, mp4 files are filtered out the same way they always have been by removing them before the s3 sync takes place. In the offline build, the mp4 files are synced to the mirror drive buckets, but are excluded from the ZIP download that is uploaded to the live site.

#### How should this be manually tested?

 - Spin up `ocw-studio` on this branch, making sure you have Concourse, Minio, Google Drive and Github configured properly for publishing websites, as well as a recent database restore from RC
 - Create a folder somewhere on your computer and pull down the static resources for the website we'll be using from S3 with `aws s3 sync s3://ol-ocw-studio-app-qa/courses/6-041sc-probabilistic-systems-analysis-and-applied-probability-fall-2013/ ./` in a bash terminal in the folder you want to download them in, using the credentials from `ocw-studio` RC
 - Browse to http://localhost:9001 and log in to the Minio web UI
 - In your `ol-ocw-studio-app` bucket, create a folder structure matching what we synced above, creating a `courses` folder with a `6-041sc-probabilistic-systems-analysis-and-applied-probability-fall-2013` folder under that
 - Inside this folder, upload the static resources we downloaded above
 - Refresh the pipeline definition for our test course by running `docker compose exec web ./manage.py backpopulate_pipelines --filter 6-041sc-probabilistic-systems-analysis-and-applied-probability-fall-2013`
 - In the `ocw-studio` UI at http://localhost:8043/sites/6-041sc-probabilistic-systems-analysis-and-applied-probability-fall-2013 publish our test site live
 - Browse to http://localhost:8080/teams/main/pipelines/live?vars.site=%226-041sc-probabilistic-systems-analysis-and-applied-probability-fall-2013%22 and confirm that the `build-online-ocw-site` job is running and that the `build-offline-ocw-site` job has not yet started
 - After the online job finishes, ensure that:
[...]

PR #1780, opened 2023-05-05
move webpack-json resource back to forked s3 resource type

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1779

#### What's this PR do?
In https://github.com/mitodl/ocw-studio/pull/1763, the `governmentpaas/s3-resource` resource type was refactored out for the official S3 resource. Unfortunately, the official Concourse S3 resource still does not support IAM instance profiles, so the check for credentials fails in QA where these instance profiles are used. This PR moves back to using the forked version of the resource (https://github.com/alphagov/paas-s3-resource) for the time being to unblock publishing in RC / production.

I would say that the instructions in https://github.com/mitodl/ocw-studio/pull/1763 should be repeated, but that would be mostly pointless as the problem this solves only happens in our cloud environments where IAM instance profiles are used for authenticating with AWS. We should still verify that this works locally with Minio by at least doing:

 - Push up a pipeline with `./manage.py backpopulate_pipelines --filter <course_id>`
 - Go to `http://localhost:8080`, find the pipeline and check that the `webpack-json` resource successfully runs its check

### DIFF
--- a/content_sync/pipelines/definitions/concourse/site-pipeline.yml
+++ b/content_sync/pipelines/definitions/concourse/site-pipeline.yml
@@ -9,6 +9,11 @@ resource_types:
     type: registry-image
     source:
       repository: ghcr.io/cludden/concourse-keyval-resource
+  - name: s3-resource-iam
+    type: registry-image
+    source:
+      repository: governmentpaas/s3-resource
+      tag: latest
   # START NON-DEV
   - name: slack-alert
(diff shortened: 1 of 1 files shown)


HussainTaj-arbisoft (approved):
LGTM 👍

Issue #1791, opened 2023-05-10
Sync missing captions management command saved files in incorrect path


fakhar-ud-din:
Steps to Reproduce

The [management command to fetch captions/transcripts](https://github.com/mitodl/ocw-studio/blob/a6f019a8d347c153ec643907d5a7f795a75aebd9/videos/utils.py#L25) has a bug where it appears to fetch the relevant files saves them in incorrect path with an additional `/` after the bucket name. The extra slash causes 404 on captions and transcripts request from a video page.

## Example from RC
![Screenshot 2023-05-10 at 9 41 04 PM](https://github.com/mitodl/ocw-studio/assets/68846543/a0547f2f-7c59-464c-8bc4-b56a8e4e19de)

PR #1796, opened 2023-05-11
Sync missing captions, filepath issue, existing captions issue fixed

fakhar-ud-din:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1791

#### What's this PR do?
The PR adds an a flag `--delete-existing` 
Testing this PR has been very troublesome since once the captions and transcripts references are updated on Video resources, a celery syncs that data to the Github.
So to test the sync command, `sync_missing_captions`, reviewer had to run manual queries to remove those references before testing, resulting in increased `turnaround time`

Moreover, when testing the prior fix, https://github.com/mitodl/ocw-studio/pull/1772, we noticed the incorrect paths being formed which were not clear when testing locally. That has also been addressed in this PR.

   - **Threeplay variables** 
   - **AWS variables**   
-  Checkout to this branch
- **Running the command to:**
  - **Test a list of courses, also deleting existing captions**
    - `docker-compose run web ./manage.py sync_missing_captions --filter <website_short_id> --delete-existing`
    - This will clear `video_captions_url` and `video_transcript_file` and delete existing caption and transcript objects
  - **Test all courses without deleting previous captions**
     - Run command `docker-compose run web  ./manage.py sync_missing_captions`
  - **Test single course**
     - Run command `docker-compose run web  ./manage.py sync_missing_captions --filter <course-short_id>`
- Head to `http://localhost:9001` (Local Minio)
- Go to `ol-ocw-studio-app/courses/<course-name>`
- You should see new captions (\*_captions.webvtt) and transcripts (\*_transcript.pdf) file with some `Size` value (previously empty **-**)


HussainTaj-arbisoft:
> So we should not be deleting existing captions introduced erroneously via the testing/validation process.

Perhaps then we can let those captions sit there and create new ones to link to the videos. So instead of `--delete-existing` we can say `--replace-existing` and it would update the existing captions. Additionally, we can make it so that we only replace captions once we find another copy on 3play to ensure that we don't lose anything. @pt2302, is this something that would be more feasible in this situation?

@fakhar-ud-din, if we change the format of the filename from `<yt-id>_captions` to `<yt-id><random>_captions` would that resolve our integrity issue (the reason we wanted to delete these files in the first place)? Will there be any negative effects of doing so (apart from the duplicates)?

gumaerc:
@fakhar-ud-din IMO this PR is not necessary. https://github.com/mitodl/ocw-studio/pull/1792/ hasn't been reverted, the RC with it in it was simply closed. We should just start another RC with #1792 in it and release that. The other changes in this PR aren't really that necessary, but if there's a reason to add that video type content filter we should just make another PR addressing that specifically. The data issue caused by not catching the pathing issue in RC before running it for 3.091-fall-2018 in production should simply be addressed using the strategy in the last bullet point, but I believe we should just close this PR.

fakhar-ud-din:
I have been testing https://github.com/mitodl/ocw-studio/pull/1797 in the RC. Although this does fix the path issue, but when we try to go to a video, the captions request yield a 404 response. 
The captions URL formed is relative to the video page, which should not be the case. Check [this video example](https://live-qa.ocw.mit.edu/courses/7-01sc-fundamentals-of-biology-fall-2011/resources/genetics-3/).

We'd have to add a `/` when updating the metadata, so it would form an absolute URL instead of a relative one.
Moreover, the newly created markdown files are being created on the root of the course. See [this](https://github.mit.edu/ocw-content-rc/7.01sc-fall-2011/tree/main/courses/7-01sc-fundamentals-of-biology-fall-2011) example for the above mentioned video.

fakhar-ud-din:
Closing this PR since all the required changes have been made in the following
https://github.com/mitodl/ocw-studio/pull/1809
https://github.com/mitodl/ocw-studio/pull/1797

PR #1725, opened 2023-03-27
Added OCW_HUGO_THEMES_SENTRY_ENV to ocw-studio

fakhar-ud-din:
#### Pre-Flight checklist
- [x] Settings
  - [ ] New settings are documented and present in `app.json`
  - [x] New settings have reasonable development defaults, if applicable
  - [ ] Opened issue for DevOps regarding necessary configuration changes to deployed environments

#### What are the relevant tickets?
#1686

#### What's this PR do?
This PR adds a new env variable in ocw-studio for sentry
`OCW_HUGO_THEMES_SENTRY_ENV`

#### How should this be manually tested?
(Required)

PR #1811, opened 2023-05-18
SENTRY_ENV added to the themes-pipelines

### LINKED ISSUE #1686: Pass SENTRY_ENV to pipelines
*This is configuration/pipeline work for https://github.com/mitodl/ocw-hugo-themes/issues/1022*

### Acceptance Criteria
- [ ] Pipelines that use `OCW_HUGO_THEMES_SENTRY_DSN` should also set `OCW_HUGO_THEMES_SENTRY_ENV`
- [ ] ~Appropriate values should be set for rc and prod. *Probably "producation" and "qa"*. See also https://github.com/mitodl/hq/discussions/365~ The value can probably be based on `OCW_STUDIO_ENVIRONMENT`

fakhar-ud-din:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #1686

#### What's this PR do?
This PR adds the `SENTRY_ENV` in the theme-assests-pipeline in the concourse

1. Checkout to this branch
2. Make sure docker containers are up, `docker-compose up`
3. Run `docker-compose exec web ./manage.py upsert_theme_assets_pipeline`
4. Go to `concourse:8080` and trigger a `ocw-theme-assests` build
5. The build should complete successfully

### DIFF
--- a/app.json
+++ b/app.json
@@ -286,10 +286,6 @@
       "description": "The sentry DSN that will be used in ocw-hugo-themes",
       "required": false
     },
-    "OCW_HUGO_THEMES_SENTRY_ENV": {
-      "description": "The sentry ENV that will be used in ocw-studio",
-      "required": false
-    },
     "OCW_IMPORT_STARTER_SLUG": {
       "description": "The slug of the WebsiteStarter to assign to courses imported from ocw-to-hugo",
       "required": false
--- a/content_sync/pipelines/concourse.py
+++ b/content_sync/pipelines/concourse.py
@@ -590,7 +590,7 @@ class ThemeAssetsPipeline(GeneralPipeline, BaseThemeAssetsPipeline):
             )
             .replace(
                 "((ocw-hugo-themes-sentry-env))",
-                settings.OCW_HUGO_THEMES_SENTRY_ENV or "",
+                settings.ENVIRONMENT or "",
             )
             .replace("((atc-search-params))", self.instance_vars)
         )
--- a/content_sync/pipelines/concourse.py
+++ b/content_sync/pipelines/concourse.py
@@ -588,6 +588,10 @@ class ThemeAssetsPipeline(GeneralPipeline, BaseThemeAssetsPipeline):
                 "((ocw-hugo-themes-sentry-dsn))",
                 settings.OCW_HUGO_THEMES_SENTRY_DSN or "",
             )
+            .replace(
+                "((ocw-hugo-themes-sentry-env))",
+                settings.OCW_HUGO_THEMES_SENTRY_ENV or "",
+            )
             .replace("((atc-search-params))", self.instance_vars)
         )
         self.upsert_config(config_str, self.PIPELINE_NAME)
--- a/content_sync/pipelines/definitions/concourse/theme-assets-pipeline.yml
+++ b/content_sync/pipelines/definitions/concourse/theme-assets-pipeline.yml
@@ -42,6 +42,7 @@ jobs:
       params:
         SEARCH_API_URL: ((search-api-url))
         SENTRY_DSN: ((ocw-hugo-themes-sentry-dsn))
+        SENTRY_ENV: ((ocw-hugo-themes-sentry-env))
       run:
         path: sh
         args:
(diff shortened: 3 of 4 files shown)


HussainTaj-arbisoft (commented):
I might be missing some context, do let me know if I'm wrong about any of the following.

This change only adds the SENTRY_ENV for the themes pipeline. I don't think it closes https://github.com/mitodl/ocw-studio/issues/1686. From what I understand, for https://github.com/mitodl/ocw-studio/issues/1686, we also need to add that env var for course build, and mass build pipelines.

Additionally, I'm not sure where `OCW_HUGO_THEMES_SENTRY_ENV` is being set. If I had to guess, for local development we'll add it in the studio .env, and for prod and rc we need to add an env var in this [file in salt ops](https://github.com/mitodl/salt-ops/blob/a363eb9011391a123d8202467e599dbb33a2c85f/pillar/heroku/ocw-studio.sls#L72)

PS: Do you think we can improve the testing instructions? A build's success does not always signify a successful env var change.

fakhar-ud-din:
So another PR was merged prior to this which actually sets the env variable and also sets a default [here](https://github.com/mitodl/ocw-studio/pull/1725)

Seems like that variable needed to be passed in the themes assets pipeline as mentioned by @ChristopherChudzicki in [this comment](https://github.com/mitodl/ocw-studio/issues/1686#issuecomment-1551911863)

Regarding the testing instructions, I am not quite sure what else would be the best way to test this because as per my intuition, the test would only pass if newly added env variable is being correctly used/ updated.

@ChristopherChudzicki could you please provide more context on this variable?

HussainTaj-arbisoft:
@fakhar-ud-din @ChristopherChudzicki thank you for the answers. I was able to verify the changes by exploring the generated source for a course (`/static_shared/js/course*.js`) like Chris showed. The source has the following lines

Screenshot 2023-05-22 at 3 34 23 PM

Raising an error didn't report anything to Sentry, (likely because of some misconfiguration in my local setup). But I believe this confirmation in the source is sufficient for local testing.

@fakhar-ud-din this change works, I agree with @ChristopherChudzicki about using `settings.ENVIRONMENT` (or some other existing env var) for this configuration. Alternatively, we'll have to add the env var `OCW_HUGO_THEMES_SENTRY_ENV` in our deployments (which might not be part of the current ticket's scope so we can do it in a new one). Let me know what direction you want to choose, I'll proceed with my review accordingly.

fakhar-ud-din:
@HussainTaj-arbisoft I did check the code as well, and yes the changes are being replicated in the source.
Moreover, I think sticking with the `settings.ENVIRONMENT` would be a good idea. We shouldn't be needlessly adding new env variables so I will proceed with making variable name changes accordingly.

@ChristopherChudzicki Should we remove the addition of the variable names from the `app.json` as well as from the `settings.py` too? I believe those wouldn't be required once we use `settings.ENVIRONMENT`.

ChristopherChudzicki:
> @ChristopherChudzicki Should we remove the addition of the variable names from the app.json as well as from the settings.py too? I believe those wouldn't be required once we use settings.ENVIRONMENT.

@fakhar-ud-din  Yes, I think that would be good. 

@HussainTaj-arbisoft re
> Raising an error didn't report anything to Sentry, (likely because of some misconfiguration in my local setup). But I believe this confirmation in the source is sufficient for local testing.

Looking at the screenshot, you're missing the `OCW_HUGO_THEMES_SENTRY_DSN` for sentry. You can get it from RC (Sentry DSNs aren't really secret, so we *could* put them in our `env_sample` files. But in general, it's nice to NOT send sentry errors to sentry during local development—unless you're testing sentry—so we do not put them in env_sample.)

ChristopherChudzicki (changes requested):
Let's make the change that @HussainTaj-arbisoft suggested:

> Should we remove the addition of the variable names from the app.json as well as from the settings.py too? I believe those wouldn't be required once we use settings.ENVIRONMENT.

HussainTaj-arbisoft (approved):
👍 LGTM

ChristopherChudzicki (approved):
Looks good—I see the `env` value coming through in builds.

Issue #1814, opened 2023-05-19
Resources referenced inside ocw-www are deleted upon gdrive sync


HussainTaj-arbisoft:
### Steps to Reproduce

1. Create/Select a course site.
2. Open the GDrive folder for this site, and upload a test file (say `example.txt`).
3. Navigate to the course's resources tab and hit `Sync w/ Google Drive`.
4. Once the sync is successful, publish the course to live.
5. Open ocw-www website. Create a Resource Collection, and add `example.txt` to it.
6. Delete `example.txt` from GDrive folder.
7. Navigate back to the course's resources tab and hit `Sync w/ Google Drive`.

### Expected Behavior

Drive sync for the course is unsuccessful. Deleting `example.txt` failed. The error states that `example.txt` is being used.

### Actual Behavior

The resource is deleted successfully.

### Related Issues (Additional Information)

In https://github.com/mitodl/ocw-studio/issues/1754 a fix was published. The fix works for active starters only. ocw-www starter is kept inactive in prod and rc.

pdpinch:
I don't think we should limit this check to active starters. Can that be changed?

HussainTaj-arbisoft:
> I don't think we should limit this check to active starters. Can that be changed?

Yes, it can.

PR #1818, opened 2023-05-23
fix: use all starters to discover references

HussainTaj-arbisoft:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-studio/issues/1814

#### What's this PR do?
Uses all starters instead of only active starters in `iter_all_configs`.

1. Navigate to your local setup of ocw-studio and checkout `hussaintaj/enable-inactive-starters`.
2. Start/restart studio.
3. Open ocw-studio admin site and ensure that the website starter `www` or others of its kind are all `inactive`.
4. Create/Select a course site.
5. Open the GDrive folder for this site, and upload a test file (say example.txt).
6. Navigate to the course's resources tab and hit Sync w/ Google Drive.
7. Once the sync is successful, publish the course to live.
8. Open ocw-www website. Create a Resource Collection, and add example.txt to it.
9. Delete example.txt from GDrive folder.
10. Navigate back to the course's resources tab and hit Sync w/ Google Drive.
11. **Expected Behavior**: You should see an error stating that the resource could not be deleted because it is being used.
12. Delete the file from GDrive that you create in step 5.
13. Delete the resource reference from the collection you created in step 8.
14. Navigate back to the course's resources tab and hit Sync w/ Google Drive.
15. **Expected Behavior**: this time the sync is successful and the resource example.txt is deleted.

### DIFF
--- a/websites/models.py
+++ b/websites/models.py
@@ -476,7 +476,7 @@ class WebsiteStarter(TimestampedModel):
     @staticmethod
     def iter_all_config_items(website: Website) -> Iterator[Tuple[bool, ConfigItem]]:
         """
-        Yields ConfigItem for all active starters.
+        Yields ConfigItem for all starters.
 
         Args:
             website (Website): The website being scanned.
@@ -490,9 +490,7 @@ class WebsiteStarter(TimestampedModel):
                 f"Website {website} does not have a starter. Cannot iterate config."
             )
 
-        all_starters = WebsiteStarter.objects.filter(
-            status__in=WebsiteStarterStatus.ALLOWED_STATUSES
-        )
+        all_starters = WebsiteStarter.objects.all()
 
         for starter in all_starters:
             for item in SiteConfig(starter.config).iter_items():


ChristopherChudzicki (approved):
👍 

We could do this in a separate PR, but we might want better help text for the error message:

Screenshot 2023-05-24 at 2 07 32 PM

I believe the WebsiteContent object in that screenshot is the resource list within ocw-www... Would be nice to show the content object's parent site.

HussainTaj-arbisoft:
Thank you for the review @ChristopherChudzicki 

> I believe the WebsiteContent object in that screenshot is the resource list within ocw-www... Would be nice to show the content object's parent site.

Yes, you're right. It would be very convenient to have the website name.

Right now we don't have much on the specifications of the error messages. I think we should wait for some user feedback so that we can better engineer the requirement. Once we know what a normal message should look like, it would be easier to incorporate cross-site references.

Issue #1821, opened 2023-05-24
Upgrade ocw-course-publisher to tag 0.5


HussainTaj-arbisoft:
### Background / Purpose

Now that we have [ocw-course-publisher 0.5](https://hub.docker.com/r/mitodl/ocw-course-publisher) published (refer to https://github.com/mitodl/ol-infrastructure/issues/1513), it's time to configure it inside OCW Studio.

### Description

In our concourse pipeline definitions in OCW-Studio, we have pinned the image version to `0.4`. That needs to be updated to `0.5`.

### Acceptance Criteria

The following pipelines use the new ocw-course-publisher image:
- [ ] site build pipeline
- [ ] mass build pipeline
- [ ] theme asset build pipeline

HussainTaj-arbisoft:
This upgrade has been successfully completed.

PR #1826, opened 2023-05-31
enhancement: upgrade publisher

HussainTaj-arbisoft:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/ocw-studio/issues/1821

#### What's this PR do?
Upgrades `ocw-hugo-publisher` to `0.5` (Hugo 0.112.0 and Go 1.20.4). Hugo changed `--baseUrl` to `--baseURL`, so that has been updated too.

1. Navigate to your local ocw-studio project directory.
2. Checkout branch `hussaintaj/1821-upgrade-publisher`
3. Start studio with `docker-compose up`
4. Choose a course you'd like to test this change on.
5. Run
    ```bash
    docker-compose exec web ./manage.py backpopulate_pipelines --filter <course-id>
    docker-compose exec web ./manage.py upsert_theme_assets_pipeline
    docker-compose exec web ./manage.py upsert_mass_build_pipeline
    docker-compose exec web ./manage.py upsert_mass_build_pipeline -o
    ```
6. Open your local concourse http://concourse:8080.
7. Trigger `ocw-theme-assets` build and wait for success.
    **Expected Behavior**: Successful build and image used has the correct tag/version.
    
    Screenshot 2023-05-31 at 6 29 17 PM
8. Go to your course site on studio UI http://localhost:8043/sites/name
9. Publish the course to live and draft both and wait for success.
    **Expected Behavior**: Successful build and image used has the correct tag/version.
    Screenshot 2023-05-31 at 6 53 12 PM
    Screenshot 2023-05-31 at 6 53 36 PM
10. Return to the Concourse UI and trigger a mass build. You might want to lower the number of courses that run in the mass build. We can achieve that by setting `publish_date` and `draft_publish_date` to null on all of the courses except the ones you want to test with.
    1. ```bash
        docker-compose exec web ./manage.py shell
        from websites.models import Website
        Website.objects.all().update(publish_date=None, draft_publish_date=None)
        ```
    2. Open https://localhost:8043/admin and set a publish_date so that it appears in the mass build.
    
    **Expected Behavior**: Successful build and image used has the correct tag/version.
    
    <img width="1369" alt="Screenshot 2023-05-31 at 6 55 31 PM" src="https://github.com/mitodl/ocw-studio/assets/71316217/7944f71f-5265-4fb2-ba0c-46d12ff8af81">


gumaerc:
I tested this thoroughly today, and there is a major issue with the `--baseURL` setting.  I made a post on the Hugo support forums about it here: https://discourse.gohugo.io/t/issue-with-baseurl-in-0-112-0-0-112-6/44656

Basically, in versions 0.112.0 and up, the setting seems to not work with the way we have implemented nav URLs. Menus in Hugo can be implemented in a [number of ways](https://gohugo.io/content-management/menus/), but the way we have done it is to create a file in the content repos at `config/_default/menus.yaml`. It was done this way to facilitate the requirements of:

 - Must support internal and external links
 - Must support a custom defined order that includes the order of external vs internal links

There are two other ways menus can be implemented:

 - Defining the menu entry of a given page in front matter of the page to be included in the menu
 - Defining the menu in the `config.yaml` file

If we went with the first option, we would have to restructure the way `ocw-studio` writes menus to instead:

 - Write only external links to the `config/_default/menus.yaml` file, preserving their weight relative to all other menu items including internal ones
 - Write internal links to the front matter of the appropriate page, omitting the `url` property because it would be derived at build time from the position of the content, also preserving the weight relative to all other menu items

The second option is not viable, as we define a universal `config.yaml` file for all course sites and Hugo does not support merging config files as far as I know.

However, something did come to mind as I was researching it and that was a reply that Joe Mooring of the Hugo project had to an unrelated thread I posted in the Hugo support forums. Apparently there is an undocumented property on menu items that can be specified in a menu config called `pageRef`: https://discourse.gohugo.io/t/interaction-between-the-baseurl-and-relativeurls-settings/42770/2?u=gumaerc

As I was writing this post, he replied to the thread for this issue that I created above and confirmed that this is indeed an unintended behavioral change with Hugo 0.112.0. Either way, I tested using `pageRef` instead of `url` in the menu config, and I think we should switch to using it. I think we should wait a bit though and see if the Hugo team addresses this so the `url` property continues to function as expected. If we change our menu configuration to use `pageRef`, we will still need to mass-publish every site to make sure the configuration is updated in every Github repo before the changes in this PR go live. If we simply change the key from `url` to `pageRef` in this PR and release it, rebuilds of already published sites will result in broken nav links.

So, in summary, I believe this PR should be set on hold and another should be written up to move from `url` to `pageRef` in our menu definitions, then once that change is released and all Github repos are updated with it, we can release this PR.

gumaerc:
I created an issue for what I describe needs to be done above here: https://github.com/mitodl/ocw-studio/issues/1833

HussainTaj-arbisoft:
@gumaerc thank you for the review and for looking into the issue. Thanks to your forum post, [this fix](https://github.com/bep/hugo/commit/c80425c1e99140e11c800f527b2478dbe443d783) should be available in [v0.112.7](https://github.com/gohugoio/hugo/milestone/248). If I understand everything correctly, the fix should allow us to upgrade Hugo separately from https://github.com/mitodl/ocw-studio/issues/1833. Let's have a discussion on how we should proceed.

gumaerc (approved):
:+1:

PR #1828, opened 2023-06-01
fix @types/react versions

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

To help renovate with #1789 

#### What's this PR do?
This PR:
- has no runtime affects; it only affects our type declarations for Typescript typechecking 
- makes us use a consistent version of `@types/react` and `@types/react-dom`

See "Background" for more.

Because this only affects typechecking, I'm not sure any manual testing is required.

If you look at [`yarn.lock` diff](https://github.com/mitodl/ocw-studio/pull/1828/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deL2769-R2760) you'll see that we were previously installing multiple versions of `@types/react`:
```
# one place
  resolution: "@types/react@npm:17.0.43"
# elsewhere
  resolution: "@types/react@npm:17.0.44"
```
These two versions seem compatible enough—typechecking passes on master. But this problem was about to get worse in #1789 (where it is causing failures). And since we only use one version of react, we should only be using one version of the types.

**Why did we have multiple versions of `types@/react`?** 

1. Some NPM packages do not provide their own type declarations (e.g., because they are not written in Typescript). When that happens, we install type declarations from [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped), a repo maintained by the Typescript people where the community adds type declarations. (Similar to [typeshed](https://github.com/python/typeshed) in Python).
2. When specifying dependencies, DefinitelyTyped uses (a)`dependencies` not `peerDependencies`, and (b) specifies the version as `*`, meaning "any version".  _This is a choice made by the maintainers of DefinitelyTyped, not by the community members who contribute types; the package.json files are constructed by the DefinitelyTyped publishing process._ 
3. Yarn uses the latest possible version when installing `*`-dependencies.

So we manually specify `@types/react@16` version 16 because that what we use, but `@types/react-dom` specifies `@types/react@*`, and we end up with multiple versions.

For more, see:
- https://github.com/yarnpkg/yarn/issues/4489#issuecomment-1308694300 from the creator of Yarn
- https://github.com/facebook/react/issues/24304#issuecomment-1111559894, from react maintainers
- https://github.com/microsoft/DefinitelyTyped-tools/issues/433, where a proper fix is proposed and discussed

In the meantime, we can tell `yarn` to resolve `@types/react` and `@types/react-dom` to specific versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,8 +55,8 @@
     "@types/node": "^16.4.11",
     "@types/pluralize": "^0.0.29",
     "@types/ramda": "^0.27.44",
-    "@types/react": "^17.0.15",
-    "@types/react-dom": "^17.0.9",
+    "@types/react": "^16.9.46",
+    "@types/react-dom": "^16.9.8",
     "@types/react-outside-click-handler": "^1.3.0",
     "@types/react-redux": "^7.1.18",
     "@types/react-router": "^5.1.16",
@@ -167,7 +167,9 @@
     "> 1%"
   ],
   "resolutions": {
-    "turndown": "7.0.0"
+    "turndown": "7.0.0",
+    "@types/react": "^16.9.46",
+    "@types/react-dom": "^16.9.8"
   },
   "packageManager": "yarn@3.1.0"
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -2674,21 +2674,12 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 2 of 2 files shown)


HussainTaj-arbisoft (approved):
👍 LGTM

Thank you for the detailed explanation.

For testing, I primarily made sure the webpack build was working. And even though it was unnecessary, I ran a sort-of random manual regression test. Things are working as expected.

PR #1830, opened 2023-06-01
update typescript to 4.9.5

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

This should help resolve issues with #1789 

#### What's this PR do?
This PR updates our Typescript from 4.5.x to 4.9.5.

Check that Studio still runs / functions normally. 

There is no change to source code beside a type annotation change. So the only change at runtime will be changes due to the TS upgrade.

### DIFF
--- a/static/js/lib/ckeditor/plugins/util.ts
+++ b/static/js/lib/ckeditor/plugins/util.ts
@@ -48,7 +48,7 @@ const ensureEncasedInQuotes = (text: string, singleQuotes = false) => {
   return q + text + q
 }
 
-export function buildAttrsString(attrs: RegExpMatchArray | null): string {
+export function buildAttrsString(attrs: string[] | null): string {
   return attrs ?
     attrs
       .map(attr =>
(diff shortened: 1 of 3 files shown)


static/js/lib/ckeditor/plugins/util.ts
```diff
@@ -48,7 +48,7 @@ const ensureEncasedInQuotes = (text: string, singleQuotes = false) => {
   return q + text + q
 }
 
-export function buildAttrsString(attrs: RegExpMatchArray | null): string {
+export function buildAttrsString(attrs: string[] | null): string {
```
ChristopherChudzicki, line 51: Typescript 4.9 introduced a small breaking change: Previously `string[]` was assignable to `RegExpMatchArray`; it no longer is.
- We call `buildAttrsString` with both `string[]` and `RegExpMatchArray` argument types
- `string[]` is the looser type (`RegExpMatchArray` satisfies the `string[]` interface (and has some extra properties like `.groups`, too). We only use the `string[]`) and is sufficient for our purposes, so let's just use that.

See also https://github.com/microsoft/TypeScript/pull/49682#discussion_r915601446

HussainTaj-arbisoft (approved):
👍 LGTM

Issue #1829, opened 2023-06-01
Update Prettier, Eslint


ChristopherChudzicki:
We should update Prettier and Eslint:
- We currently use prettier v1, which does not support some recent typescript features (e.g., the `satisfies` keyword).
- renovate already wants to update Typescript in #1789 

Renovate cannot do this upgrade for us, because Prettier changed some of its preferences between v1 and v2, which means our code formatting is going to change.

PR #1831, opened 2023-06-01
Update linting, formatting

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#1829 , #1832 

#### What's this PR do?
Currently, if we update Typescript, we will get warnings like
>```
>WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript->eslint/typescript-estree.
>
>You may find that it works just fine, or you may not.
>
>SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.9.0
>
>YOUR TYPESCRIPT VERSION: 4.9.5
>
>Please only submit bug reports when using the officially supported version.
>```

To resolve this warning (and to use new TS features like [`satisfies`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator)) we need to update Prettier from v1 to v2.

Unfortunately... Prettier changed how it handles chained method calls from v1 to v2. See https://github.com/prettier/prettier/issues/7884 for details. _IMO the v1 way looked better for fluent APIs like enzyme has, but...whatever. We should stop using Enzyme anyway._

The only changes here are autofixes by Prettier v2. There should be no behavior change.

So... run Studio and make sure it works. Be sure to either `docker compose up` or manually `docker exec watch yarn install`


HussainTaj-arbisoft (approved):
👍 LGTM

I tried every UI element I could see.
- Edited all the content of an existing course and www site
- Created a new course from scratch, populated and published it

Everything is working.

PR #1843, opened 2023-06-13
chore: upgrade actions' versions

HussainTaj-arbisoft:
Closes https://github.com/mitodl/ocw-hugo-themes/issues/922

This PR:
- Updates versions of actions `checkout`, `setup-node`, and `setup-python`
- Removes the cache step because `actions/cache@v3` takes care of it

The original ticket was meant for ocw-hugo-theme repo, but since that was already done, I took the opportunity to update this repo's workflows.

# How can this be tested?

I believe that this can be tested online, there is no need to test it locally.

The tester only needs to verify four things:

- [x] The code is okay
- [x] actions/checkout@v3 works
- [x] actions/setup-note@v3 and actions/setup-python@v4 work.
- [ ] caching works

For these verifications, you can review the runner logs of the two workflow runs here:
- (First run) https://github.com/mitodl/ocw-studio/actions/runs/5254507265/jobs/9493198468
- (Second run) https://github.com/mitodl/ocw-studio/actions/runs/5254800744/jobs/9493867704

> One may choose to use [act](https://github.com/nektos/act). Though, I find it to be very slow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
           - 9200:9200
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: update
         run: sudo apt-get update -y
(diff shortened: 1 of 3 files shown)


ibrahimjaved12:
> Removes the cache step because actions/cache@v3 takes care of it

I think you meant checkout@v3 perhaps? There's no cache@v3.

Anyway, looks good to me, cache is working. And so is everything else.

ibrahimjaved12 (approved):

HussainTaj-arbisoft:
> > Removes the cache step because actions/cache@v3 takes care of it
> 
> I think you meant checkout@v3 perhaps? There's no cache@v3.

Sorry, I meant to say that the `setup-*` actions take care of the caching now.

Issue #1833, opened 2023-06-01
move to using `pageRef` instead of `url` for internal menu items


gumaerc:
#### Summary

While testing our `course-v2` theme in `ocw-hugo-themes` with Hugo 0.112.x, I ran into the issue I describe in this Hugo support thread: https://discourse.gohugo.io/t/issue-with-baseurl-in-0-112-0-0-112-6/44656/4. Basically, Hugo 0.112.x breaks `baseURL` support on menu items that specify the `url` property. The `pageRef` property was previously undocumented as noted in this thread: https://discourse.gohugo.io/t/interaction-between-the-baseurl-and-relativeurls-settings/42770/2?u=gumaerc, but after the discussion in this thread they added it to the menu documentation: https://gohugo.io/content-management/menus/

I imagine that the issue with `baseURL` will eventually be fixed, but we should still switch to using `pageRef` for our internal links, as it is a direct way to reference a page being rendered by Hugo.

#### Requirements

 - [ ] For internal links, use the `pageRef` key with the same value (i.e. `/pages/syllabus`) instead of `url`
 - [ ] It needs to be verified that these links reference a real page being rendered by Hugo because if it references a page that does not exist an error will be thrown, so we should thoroughly unit test this functionality
 - [ ] External links should still use the `url` property

PR #1845, opened 2023-06-14
Use pageRef with new internal navigation menu items

pt2302:
Closes https://github.com/mitodl/ocw-studio/issues/1833.

For any new internal navigation menu items, use `pageRef` instead of `url` as the property to be passed to Hugo. The reasons for this change are documented in the issue that this closes; essentially, while `url` works for internal menu items, using `pageRef` is more robust to potential changes and is the preferred approach in Hugo.

Note: there needs to be a follow-up PR to update the existing menu items for internal items to use `pageRef` instead of `url`. 

# How can this be tested?

On a local instance of OCW Studio, choose any course and click on `Menu` under `Settings`. Add some internal menu items and external menu items. Publish the course, and verify that the generated `config/_default/menus.yaml` in the GitHub repository has `pageRef` for the internal menu items and has `url` for the external menu items. Also, verify that the links are working properly.

### DIFF
--- a/static/js/components/widgets/MenuField.tsx
+++ b/static/js/components/widgets/MenuField.tsx
@@ -28,7 +28,6 @@ interface MenuFieldProps {
 export type HugoItem = {
   identifier: string
   name: string
-  pageRef?: string
   url?: string
   weight: number
   parent?: string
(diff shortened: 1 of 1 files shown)


gumaerc (changes requested):
In testing this, I wasn't able to get it to push `pageRef` up properly to Github. After modifying [this line](https://github.com/mitodl/ocw-studio/blob/4746aca059ceac1a24753c864dadd8ea9f2a4ac2/content_sync/serializers.py#L230) to set `pageRef` as opposed to `url`, the key was properly set in Github. We should also set up some testing surrounding this, both in the serializer and in the frontend.

gumaerc (approved):
:+1:

PR #1849, opened 2023-06-15
fix a useless test

ChristopherChudzicki:
None

This PR fixes a useless test. See comments.

# How can this be tested?
This PR does not change application code, only changes a test that was previously doing nothing.

### DIFF
--- a/static/js/components/AuthenticationAlert.test.tsx
+++ b/static/js/components/AuthenticationAlert.test.tsx
@@ -54,28 +54,27 @@ describe("Prompting for authentication", () => {
   ])(
     "prompts for authentication when APIs reject with auth errors",
     async () => {
-      const { result, setMockWebsiteResponse, website, user } = await setup()
-
-      setMockWebsiteResponse(authRejectionBody, 403)
-
-      const siteLink = await waitFor(() => result.getByText(website.title))
-      await act(() => user.click(siteLink))
-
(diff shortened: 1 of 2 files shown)


static/js/test_util.ts
```diff
@@ -178,9 +178,11 @@ export const withFakeLocation = async (
   cb: () => Promise<void> | void
```
ChristopherChudzicki, line 178: This function runs a callback `cb` using a fake `window.location` property. See comment above function. Mocking window.location is a pain with JSDom.

But... When I first wrote it... I did it wrong. The callback never ever ran. Now it does.

There should have been no `catch`. Only `try { ... } finally { ... }`.

static/js/components/AuthenticationAlert.test.tsx
```diff
@@ -54,28 +54,27 @@ describe("Prompting for authentication", () => {
...
-        await waitFor(() => {
-          expect(window.location.href).toBe("/login/saml/?idp=default")
-        })
+        expect(window.location.href).toBe("/login/saml/?idp=default")
+        result.unmount()
       })
 
-      result.unmount()
+      // Test is inside a callback. Let's make sure it actually ran.
+      expect.assertions(3)
```
ChristopherChudzicki, line 77: Previously, the callback inside `withFakeLocation` was never actually being run.

Now it is being run, and we are sure of that because of `expect.assertions(3)`.

Issue #1856, opened 2023-06-27
Menu YAML for some legacy courses has both url and pageRef


pt2302:
In https://github.com/mitodl/ocw-studio/pull/1845, the `url` property was replaced with `pageRef` for internal menu items. However, for legacy data that stores both `url` and `identifier` fields, updating the menu will now result in the YAML having both `url` and `pageRef`, as below:

```
leftnav:
- identifier: c9b592be-0215-c31f-6f65-02243f2987bb
  name: Syllabus
  pageRef: /pages/syllabus
  url: /pages/syllabus
  weight: 10
- identifier: 66c5c691-0366-d451-4ec6-011d0e9ba305
  name: Calendar
  pageRef: /pages/calendar
  url: /pages/calendar
  weight: 20
- identifier: 12a1aaf9-be7a-ca1d-756d-c7f978d7513c
  name: Lecture Notes
  pageRef: /pages/lecture-notes
  url: /pages/lecture-notes
  weight: 30
- identifier: 59d10fc0-a31a-74c3-07d0-42b79008c871
  name: Assignments
  pageRef: /pages/assignments
  url: /pages/assignments
  weight: 40
```

(from https://github.mit.edu/ocw-content-rc/16.333-fall-2004/blob/main/config/_default/menus.yaml)

PR #1857, opened 2023-06-30
Remove url property from menu items if pageRef is set

pt2302:
Closes https://github.com/mitodl/ocw-studio/issues/1856.

In https://github.com/mitodl/ocw-studio/pull/1845, the `url` property was replaced with `pageRef` for internal menu items. However, for legacy data that stores both `url` and `identifier` fields, updating the menu resulted in the YAML having both `url` and `pageRef` properties. This PR fixes that issue by removing the `url` property if `pageRef` is set (for internal links with a valid `identifier`).

# How can this be tested?
1. Find a course that has a navigation menu with internal links that uses both `url` and `identifier`. Some examples: `16.333-fall-2004`, `11.945-spring-2006`.
2. Run `docker compose exec web ./manage.py reset_sync_states --type navmenu --filter <course_id>`
3. Run `docker compose exec web ./manage.py sync_website_to_backend --filter <course_id>`
4. Check the markdown for the git repository on the main branch for the course, and verify that only `pageRef` (and not `url`) is set for internal menu items.
5. Test publishing the course and verify that all of the internal and external links work properly.

### DIFF
--- a/content_sync/serializers_test.py
+++ b/content_sync/serializers_test.py
@@ -77,6 +77,11 @@ EXAMPLE_MENU_FILE_YAML = f"""mainmenu:
     name: Ext Link
     weight: 10
     url: http://example.com
+  - identifier: {EXAMPLE_UUIDS[0]}
+    name: Page 1 url and pageRef
+    weight: 20
+    pageRef: content/page-1.md
+    url: content/page-1.md
 """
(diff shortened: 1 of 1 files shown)


gumaerc (approved):
:+1:

PR #1861, opened 2023-07-06
chore: create populate_file_sizes command

HussainTaj-arbisoft:
Closes https://github.com/mitodl/hq/issues/1683

This PR adds a populate_file_sizes command that populates the file_size metadata field of resource-type content for existing courses.

We fetch the file size of resource content in the following way:
1. If the resource content object has `file`, `metadata.file`, or `metadata.file_location` attribute then fetch the file size from S3.
2. Otherwise, if the content has a `metadata.video_files.archive_url` then fetch the size from the URL.

In both cases, we use a HEAD request and use the Content-Length header. Since the data is not encrypted, this approach works.

For DriveFiles, we simply fetch the size of `s3_key` field.

# How can this be tested?

## Prerequisites

* You should have the latest version of ocw-hugo-projects. And your ocw-studio starters should have the update configuration. Refer to https://github.com/mitodl/ocw-hugo-projects/pull/253 for more details.
* It would be good to have an RC clone. Refer to https://github.com/mitodl/hq/discussions/1189 in case you haven't done that before.
* You may need one local clone of a legacy course with its corresponding data in your local `AWS_STORAGE_BUCKET`. In case you need more details refer to https://github.com/mitodl/ocw-studio/pull/1763.

## Steps

1. Open a terminal inside your local ocw-studio setup.
2. Check out the branch `hussaintaj/1683-file-size`.
3. Run
     ```
     docker-compose run --rm web bash
     ```
4. Run
     ```
     ./manage.py populate_file_sizes --override-existing --filter course-id
     ```
     > --override-existing instructs the command to re-fetch sizes for objects that already have a size.
5. Inspect the logs. **Expected Behavior**:
   - This command finds and populates resource file sizes. You should see an output similar to:
      ```
      Scheduling populate_file_sizes_bulk...
      Waiting on task...
      Done.
      ```
      The rest of the logs will be inside the celery worker.
   - You should only see the error message `Could not read size for ...` if the URL or s3 key is indeed invalid. Some of our data is like that.
   - You should only see the message `... has no file associated to it.` if the resource or drive_file has no file linked to it. Some of our data is like that.
6. Open the admin site and explore your course's content objects. **Expected behavior**: You should see the `file_size` value in the metadata field.
7. (Optional) Publish the course and verify that you can see the `file_size` attribute in the frontmatter of resources' markdown.

# Additional Context

See also https://github.com/mitodl/hq/issues/1678


gumaerc (changes requested):
This works as advertised and I have no problems with the code in the management command itself, however I do have one concern. This has not been set up in a Celery task, and I think that could potentially be problematic when we go to run this unfiltered in RC / production to populate file sizes on every resource. Without being in a task, this will be run directly on the Django web process itself and could potentially cause a deadlock in the app with all the reaching out to GDrive / AWS it does. I think we should move this code to a Celery task. If you need an example there are plenty of other management commands that do this, such as the tasks that backpopulate pipelines.

gumaerc (approved):
:+1: LGTM!

HussainTaj-arbisoft:
I'll merge and test this tomorrow to avoid blocking any releases today.

HussainTaj-arbisoft:
The combination `chort(chain())` behaves unintuitively. I was suspicious of this so I ran a test to ensure that subtasks ran sequentially. They did not, so I removed the `chord`. It runs sequentially now.

PR #1879, opened 2023-07-14
fix: strip '/' from keys in populate_file_sizes

HussainTaj-arbisoft:
https://github.com/mitodl/hq/issues/1683

https://github.com/mitodl/ocw-studio/pull/1861

Normally course content would have a `file.name` as `course/abc/resource` but some of our legacy course content has `file.name` that starts with `/`, which makes an s3 key `/course/abc/resource`. This key is invalid (_when used directly to query size from s3_.)

This PR simply strips any `/` on the sides before querying s3 for file sizes.

Example: https://ocw-studio-rc.odl.mit.edu/admin/websites/websitecontent/489580/change/

# How can this be tested?
The issue does not appear locally. Both `/courses/abc/resource` and `courses/abc/resource` work the same in our local MinIO setup.

This cannot be tested locally. One could repeat the testing steps in https://github.com/mitodl/ocw-studio/pull/1861 but we won't know if this change worked until we move to RC.

### DIFF
--- a/gdrive_sync/utils.py
+++ b/gdrive_sync/utils.py
@@ -22,6 +22,7 @@ def fetch_content_file_size(
     )
 
     if file_key:
+        file_key = file_key.strip("/")
         size = bucket.Object(file_key).content_length
     elif content.metadata.get("video_files", {}).get("archive_url"):
         # Some of our video resources are directly linked to YT videos, and their


pdpinch:
How did you discover this issue?  Can you give an example of a course and
resource with a file.name starting with a slash?

HussainTaj-arbisoft:
> How did you discover this issue? Can you give an example of a course and resource with a file.name starting with a slash?

I ran the command for the course https://draft-qa.ocw.mit.edu/courses/5-36-biochemistry-laboratory-spring-2009, when the resource size information was not filled out correctly I saw the logs to figure out what happened.

There isn't any problem with the resources in the published course though. The course appears to be working correctly. I suppose we handle this case everywhere we use the keys, or maybe it's limited to the way I used it.

An example where there is a leading slash `/` https://ocw-studio-rc.odl.mit.edu/admin/websites/websitecontent/435992/change

New resources don't have a leading slash `/` for example https://ocw-studio-rc.odl.mit.edu/admin/websites/websitecontent/521369/change

gumaerc (approved):
I wasn't able to find much on why this is an issue, but I have a hypothesis surrounding what is being talked about in this old post: https://groups.google.com/g/boto-users/c/mv--NMPUXoU

Basically, in AWS S3 the key isn't a file path like we traditionally think about them. It's more like a unique ID string. Say you tell `boto3` to upload a file with code like this:

```
import boto3

client = boto3.client("s3")
client.upload_file("path/to/file.txt", "your-bucket", "/path/to/key.txt")
```

The resulting S3 path to the file will be `s3://your-bucket//path/to/key.txt` because `boto3` inserts the slash after the bucket name. In a traditional filesystem, this would be auto-corrected. S3 doesn't actually know what a directory is, it just knows there's a file with a key matching that string. You can see this on the web if you try and request a file directly from S3. An additional slash in your path will not be corrected, and you will simply get a 404 back. For example:

https://ocw-content-live-qa.s3.amazonaws.com/courses/21l-011-the-film-experience-fall-2013/c7c1a8f078c559256fbe9ed93ba9e5b8_MIT21L_011F13_2006_Notes.pdf

vs.

https://ocw-content-live-qa.s3.amazonaws.com/courses/21l-011-the-film-experience-fall-2013//c7c1a8f078c559256fbe9ed93ba9e5b8_MIT21L_011F13_2006_Notes.pdf

My hypothesis is that Minio is smarter about this and actually strips out consecutive slashes when requesting or uploading files. I'd like to better test scenarios like this and make sure that all of our code that uploads or grabs data from S3 makes sure the keys are properly formatted aka:

 - No leading slashes
 - No consecutive slashes

...but that's beyond the scope of simply fixing the issue here so the command for populating file sizes works. I'll work on writing up an issue for this in a way that makes sense.

gumaerc:
I found more about this issue here: https://github.com/minio/minio/issues/5958

Basically, Minio has to store your file on an actual local filesystem so it has to have a real file path, whereas Amazon S3 is not stored on a traditional file system so behavior like what I describe above is possible.

HussainTaj-arbisoft:
> Basically, Minio has to store your file on an actual local filesystem so it has to have a real file path, whereas Amazon S3 is not stored on a traditional file system so behavior like what I describe above is possible.

This makes a lot of sense. Thanks.

Issue #1852, opened 2023-06-20
Tune uWSGI settings based on application profile


rhysyngsun:
See my writeup on how to approach this: https://github.com/mitodl/hq/discussions/393

The current settings on xPro (workers = 3, threads = 25) is probably a good starting point for values.

PR #1886, opened 2023-07-17
Tune uWSGI settings

ibrahimjaved12:
Closes https://github.com/mitodl/ocw-studio/issues/1852

This PR aims to tune the uWSGI settings in order to optimize application's performance and resource utilization. The tuning process involves profiling the application's performance, conducting load tests, and monitoring memory usage. Key settings to be adjusted are reload-on-rss, max-requests, workers, and threads. By finding the right balance between these settings, we can enhance application stability, and improve overall efficiency.

# How can this be tested?
We have to perform load testing as mentioned in the [discussion post](https://github.com/mitodl/hq/discussions/393). One of the ways (and the way I did) was using using [Locust](https://docs.locust.io/en/1.0.2/what-is-locust.html).
It provides a UI interface at localhost:8089 which lets you easily perform load testing by providing number of users, swarm rate (number of users to increase per second), and host. You can use Locust, or any other tool. Here are steps with Locust:

We will want to make these changes to install and use Locust:
First checkout this branch. Then

docker-compose.yml (under services):
```
  locust:
      image: locustio/locust
      ports:
        - "8089:8089"
      volumes:
        - ./:/src
      command: >
        -f /src/locustfile.py
      networks:
        - default-network

```
Modify docker-compose.yml (web container):
```
  web:
    build:
      context: .
      dockerfile: Dockerfile
    deploy:
      resources:
        limits:
          cpus: "2"
          memory: "1g"
```
locustfile.py (this will be on project's root level, i.e next to `.env` file for instance):
```
from locust import HttpUser, task, between

class DjangoAdminUser(HttpUser):
endpoints = [
"/sites/<site>/type/",
]

def on_start(self):
    self.client.get("/admin/login/")
    self.login()

def login(self):
    csrf_token = self.client.cookies['csrftoken']
    response = self.client.post(
        "/admin/login/",
        {
            "username": "<your local django admin username>",
            "password": "<your local django admin password>",
            "csrfmiddlewaretoken": csrf_token,
        },
        headers={"Referer": "/admin/login/"},
    )
    if response.status_code == 200:
        self.navigate_to_endpoints()
@task
def navigate_to_endpoints(self):
    for endpoint in self.endpoints:
        self.client.get(endpoint)
```

Add uwsgitop to requirements.txt and configure the [uWSGI stats server](https://uwsgi-docs.readthedocs.io/en/latest/StatsServer.html) (e.g. stats = /tmp/uwsgi-stats.sock in uwsgi.ini) (As mentioned in the discussion post)

Then you will run:
`docker-compose build`
`docker-compose up`

Wait for it to finish loading up. Sometimes your requests can fail if `nginx` has not properly loaded. In that case you can go to `nginx` docker container from your docker application, restart it, and wait for it to finish restarting. (When this happens, all of your requests should fail)

Next you will go to `localhost:8089` this is where Locust's UI interface lies. And then you will try experimenting different values for users/swarm rate, but you will use `http://nginx:8043` for host. And you'll follow the steps as mentioned in the discussion post.

You'll be able to monitor data using Locust UI and uWSGI stats server using: `docker-compose exec web uwsgitop /tmp/uwsgi-stats.sock`


ChristopherChudzicki (commented):
Try as I might, I couldn't get the RSS per worker over 110. 

**Question:** When I started sending lots of requests with locust, I would get errors  (in the `locust` container) along the lines of:

```
  File "src/gevent/greenlet.py", line 908, in gevent._gevent_cgreenlet.Greenlet.run
  File "/opt/venv/lib/python3.10/site-packages/locust/user/users.py", line 176, in run_user
    user.run()
  File "/opt/venv/lib/python3.10/site-packages/locust/user/users.py", line 142, in run
    self.on_start()
  File "/src/locustfile.py", line 10, in on_start
    self.login()
  File "/src/locustfile.py", line 13, in login
    csrf_token = self.client.cookies['csrftoken']
  File "/opt/venv/lib/python3.10/site-packages/requests/cookies.py", line 334, in __getitem__
    return self._find_no_duplicates(name)
  File "/opt/venv/lib/python3.10/site-packages/requests/cookies.py", line 413, in _find_no_duplicates
    raise KeyError(f"name={name!r}, domain={domain!r}, path={path!r}")
KeyError: "name='csrftoken', domain=None, path=None"
2023-07-26T17:58:26Z <Greenlet at 0xffff82b80360: run_user(<locustfile.DjangoAdminUser object at 0xffff82b881)> failed with KeyError
```
Did you ever run into errors like that?

**Question:** If our `reload_on_rss` is only 135, couldn't we have more workers? I think we have 1GB memory on heroku.

ibrahimjaved12:
> Try as I might, I couldn't get the RSS per worker over 110.

Really? How many workers did you have + how many users did you use to send request (and spawn rate)?

>
> **Question:** When I started sending lots of requests with locust, I would get errors (in the `locust` container) along the lines of:
> 
> ```
> Traceback (most recent call last):
>   File "src/gevent/greenlet.py", line 908, in gevent._gevent_cgreenlet.Greenlet.run
>   File "/opt/venv/lib/python3.10/site-packages/locust/user/users.py", line 176, in run_user
>     user.run()
>   File "/opt/venv/lib/python3.10/site-packages/locust/user/users.py", line 142, in run
>     self.on_start()
>   File "/src/locustfile.py", line 10, in on_start
>     self.login()
>   File "/src/locustfile.py", line 13, in login
>     csrf_token = self.client.cookies['csrftoken']
>   File "/opt/venv/lib/python3.10/site-packages/requests/cookies.py", line 334, in __getitem__
>     return self._find_no_duplicates(name)
>   File "/opt/venv/lib/python3.10/site-packages/requests/cookies.py", line 413, in _find_no_duplicates
>     raise KeyError(f"name={name!r}, domain={domain!r}, path={path!r}")
> KeyError: "name='csrftoken', domain=None, path=None"
> 2023-07-26T17:58:26Z <Greenlet at 0xffff82b80360: run_user(<locustfile.DjangoAdminUser object at 0xffff82b881)> failed with KeyError
> ```
> 
> Did you ever run into errors like that?
> 

I don't think I have seen this before. When I would bombard it with requests, I generally saw the web docker container for error messages. And once in a while I would get 50X error message server related. And other times:
`2023-07-19 20:15:47 psycopg2.OperationalError: connection to server at "db" (10.1.0.103), port 5432 failed: FATAL:  sorry, too many clients already`
But Nathan told me we use pgbouncer in RC/Production. Therefore, this error message might not be super relevant.
I will look into Locust container tomorrow to see if I get any error messages there.

> **Question:** If our `reload_on_rss` is only 135, couldn't we have more workers? I think we have 1GB memory on heroku.

Yes. that's correct. 135*4 = 540mb
Theoretically, we could probably use something like 150MB * 6 containers = 900mb

But the thing is, when I started using more than 4 workers, the ratio of failed requests substantially increased for even less than 100 users per second. Where previously, 3-4 workers did just fine for up to 250 users per second.

Now as I understand, I believe local development, and RC might also function differently to a good extent. So I think we would probably have to go with one set of values on a good judgment, and we might want to revise them after testing those values in RC.

Also, when I was using 4 workers, and suppose, each worker had consumed 120mb. So 120mb*4= 480 mb. But in docker it would actually show around ~565mb if you go in stats. Is there any other memory also utilized? For this reason, we might also want to leave some memory to ensure our workers also do not go over the limit.

ChristopherChudzicki:
> But the thing is, when I started using more than 4 workers, the ratio of failed requests substantially increased for even less than 100 users per second. Where previously, 3-4 workers did just fine for up to 250 users per second.

👍 

> Now as I understand, I believe local development, and RC might also function differently to a good extent. So I think we would probably have to go with one set of values on a good judgment, and we might want to revise them after testing those values in RC.

I agree. I think the real test will be on rc/prod, and this looks reasonable, so I am going to approve.

ChristopherChudzicki (approved):

Issue #1863, opened 2023-07-07
Relation Widget Improvements (Resource Lists, Video Galleries, ...)


ChristopherChudzicki:
As a course author, I want to more easily create resource lists and video galleries.

Resource lists and video galleries are powered by the Relation widget with `sortable: true`. The sortable relation widget includes a selection field with a typeahead for filtering the search space. This select field has a few deficiences:
- it shows a limited number of items
- some of the displayed items are disabled (the ones already added)
- To add an item, users must (a) enter search text, (b) click the option, (c) verify the selection by clicking click "Add". Step (c) is redundant **and** it clears the filter text, which can be annoying.

#### Designs and Mockups

<img width="750px" alt="resource_list" src="https://github.com/mitodl/ocw-studio/assets/9010790/ad7fe13a-d10e-425c-aa9c-3e85fbb57a74">

#### Acceptance Criteria:

- [ ] 1. Clicking an option should immediately add its value to the sortable list. The "Add" button should be removed.
- [ ] 2. Adding an item to the list should preserve the filter text.
- [ ] 3. Already-selected items should be filtered from the options listing rather than disabled. [^1]
- [ ] 4. The options-listing should include more items. Ideally this would be an asynchronously loaded infinite scroll. The library we use for the select field is `react-select`, which does not support asynchronous infinite scroll. But there is a wrapper around it that does: https://www.npmjs.com/package/react-select-async-paginate

[^1]: The Relation widget already behaves this way when `multiple: true`,  `sortable: false`.

pdpinch:
Related to https://github.com/mitodl/hq/issues/1700

I'd like to be careful with rolling this out. It will need some extra testing on RC. Is there a reasonable way to put it behind a feature flag?

ChristopherChudzicki:
> I'd like to be careful with rolling this out. It will need some extra testing on RC. Is there a reasonable way to put it behind a feature flag?

I think we should be able to do that. Would flagging by env be good enough, or do we want it by user?

[...]

PR #1888, opened 2023-07-18
feat: select widget improvements

HussainTaj-arbisoft:
Closes https://github.com/mitodl/ocw-studio/issues/1863

Related PR https://github.com/mitodl/salt-ops/pull/1573

This PR updates our select widgets for UX improvement.

This PR adds the following four features.

* **Quick Add**: In the sortable select UI, add an option directly to the list when it is selected
* **Preserve Search Text**: Search query to filter options should persist between menu transitions
* **Hide Selected Options**: Menu should not show selected options
* **Infinite Scroll**: Load more options when the user reaches the end of the menu

All of the features above are disabled by default and are enabled through the use of feature flags. The flags are `FEATURE_SELECT_FIELD_INFINITE_SCROLL`, `FEATURE_SORTABLE_SELECT_PRESERVE_SEARCH_TEXT`, `FEATURE_SORTABLE_SELECT_HIDE_SELECTED`, `FEATURE_SORTABLE_SELECT_QUICK_ADD`

## `loadOptions` prop changes

SelectField had a significant callback prop whose signature has been changed in this PR. This change was required to support `AsyncPaginate` component.

New signature: `(s: string, ops: Options[], a: Additional) => {hasMore?: bool, options: Options[], additional?: Additional}` where `Additional` is `{callback?: (options: Option[]) => void}`

Old signature: ` (s: string, cb: (options: Option[]) => void) => void`

The existing code has been updated to conform to this new signature. A shim (wrapper) function is created for AsyncSelect to act as an adapter. I chose this approach because the new signature is more flexible than the old one.

## Feature flags

MIT ol-django app has [env feature flag support.](https://github.com/mitodl/ol-django/blob/97c5fd981733562ff171fa9a98985e8c58988419/src/mitol/common/envs.py#L332). ocw-studio already uses one or two flags for studio backend. This PR uses that feature and forwards the flags to the frontend builds through the config view.

# Screenshots (if appropriate):

The new UX will be shown in the following recordings.

https://github.com/mitodl/ocw-studio/assets/71316217/1bca02a2-6c03-43ab-8d0c-4277aad4c06b

https://github.com/mitodl/ocw-studio/assets/71316217/a40d50ed-d46a-4078-b345-64a8e1917325

# How can this be tested?

## Prerequsites

* Download sample data for a course from [here](https://drive.google.com/drive/folders/1kFQ_HklR7RPa0vH3iUGanFBlSl1_5jF9?usp=sharing)
* It's good to have an RC db clone on your local machine

## What this change affects

This change affects the following UI elements. _Edit after the first review: now some of these don't exhibit new behavior. But I'm leaving the list here for completeness._ 

* Course
   * Resources
      * Resource Type
      * Learning Resource Type
      * License
   * Metadata
      * Level
      * Term
      * Topics
      * Course Image
      * Course Thumbnail Image
      * Instructors
      * Department Numbers
      * Learning Resource Types
   * Resource Lists
      * Resources
   * Video. Gallery
      * Videos
* www
   * Resource Collections
      * Resources
   * Course List
      * Courses
   * Course Collections
      * Cover Image
      * Featured Courses List
      * Course Lists
   * Promos
      * Image
   * Stories
      * Image
   * Resources
      * Resource Type
   * Homepage
      * Featured Promos

## Steps

1. Navigate to your ocw-studio local setup and checkout branch `hussaintaj/1863-relation-widget`.
2. Don't set any feature flags yet. First, we'll test if the existing UI and UX work as they used to.
3. Prepare a course.
   1. Open or create a course.
   2. Go to the GDrive folder of this course and add the files you downloaded from the prerequisite sections.
   3. Sync these files with the course in the studio UI.
   4. Wait for the sync to complete.
   5. Publish the course to production. This will enable us to cross-site reference resources.
4. Prepare www
   1. If you have an RC clone you should have a www site that should work. Otherwise, you'll have to create a new one and repeat the previous steps to populate data.
5. Ideally we should be testing all the UI elements mentioned above. But testing a small set should suffice too. This set of UI elements includes.
   * Course
      * Resource Lists
         * Resources
      * Metadata
         * Course Image
   * www
      * Resource Collections
         * Resources
      * Course Collections
         * Cover Image
         * Featured Courses List
      * Homepage
         * Featured Promos
   
      **Expected Behavior**: Make sure these UI elements work as expected, similar to what we have on production.
6. Now open `.env` and set `FEATURE_SORTABLE_SELECT_PRESERVE_SEARCH_TEXT=true`.
7. Restart Studio.
8. Test the SortableSelect UI elements (e.g. mentioned in step 5).
    * **Expected Behavior**:  The enable features should be working as expected. See screen recordings for reference.
9. Repeat steps 6 through 8, enabling each feature one by one, i.e.
    ```
    FEATURE_SELECT_FIELD_INFINITE_SCROLL=true
    FEATURE_SORTABLE_SELECT_HIDE_SELECTED=true
    FEATURE_SORTABLE_SELECT_QUICK_ADD=true
   ```

There is a lot going on in this PR so here's a checklist you can use to track your review progress.

- [ ] When all of the flags are off, the UI works similarly to what we have in production
- [ ] When `FEATURE_SORTABLE_SELECT_PRESERVE_SEARCH_TEXT` is enabled, the search query is preserved
- [ ] When `FEATURE_SELECT_FIELD_INFINITE_SCROLL` is enabled, more menu options are dynamically loaded
- [ ] When `FEATURE_SORTABLE_SELECT_HIDE_SELECTED` is enabled, selected options are hidden instead of being disabled
- [ ] When `FEATURE_SORTABLE_SELECT_QUICK_ADD` is enabled, clicking an option adds it to the list. There's no "Add" button


main/views.py
```diff
@@ -23,6 +23,8 @@ def _index(request):
         "release_version": settings.VERSION,
         "sentry_dsn": settings.SENTRY_DSN,
         "gdrive_enabled": is_gdrive_enabled(),
+        "features": settings.FEATURES,
```
HussainTaj-arbisoft, line 26: ocw-studio only has one feature flag at this time, namely `FEATURE_USE_LOCAL_STARTERS`. So, any kind of filter here right now is unnecessary IMO.

static/js/components/widgets/SelectField.tsx
```diff
@@ -1,36 +1,58 @@
...
+  classNamePrefix?: string
+  value?: null | undefined | string | string[]
   onChange: (event: ChangeEvent<HTMLSelectElement>) => void
   multiple?: boolean
   options: Array<string | Option>
-  loadOptions?: (s: string, cb: (options: Option[]) => void) => void
+  loadOptions?: LoadOptions<Option, Option[], Additional | undefined>
   placeholder?: string
   defaultOptions?: Option[]
+  cacheUniques?: ReadonlyArray<any>
```
HussainTaj-arbisoft, line 27: `AsyncPaginate` caches search results internally. This prop is a set of values used to identify a cache inside `AsyncPaginate` component. This prop is useful when two SelectFields are related. For example, resource collection selects a website in one select and the resource in the second select field. When the website select changes, the resource select needs to reset/change its cache.

static/js/components/widgets/SelectField.tsx
```diff
@@ -1,36 +1,58 @@
...
 }
 
+export interface Additional {
+  callback?: (options: Option[]) => void
+}
+
 interface Props {
   name: string
-  value: null | undefined | string | string[]
+  classNamePrefix?: string
```
HussainTaj-arbisoft, line 19: `classNamePrefix` for `Select` makes its internal class names predictable. Which makes it easier to target internal elements in unit tests.

static/js/components/widgets/SelectField.tsx
```diff
@@ -74,29 +96,77 @@ export default function SelectField(props: Props): JSX.Element {
...
+
+  const handleMenuClosed = useCallback(
+    () => setPlaceholder(searchText),
+    [setPlaceholder, searchText]
+  )
+
+  const handleMenuOpen = useCallback(
+    () => setSearchText(placeholder ?? ""),
+    [setSearchText, placeholder]
+  )
```
HussainTaj-arbisoft, line 117: `Select` hides the typed input element when the field loses focus.

This workaround replaces the placeholder with the typed search text so it is visible when the field loses focus. And populates the the input field when the field regains focus. When the field loses focus, the input, although hidden, is not persisted because of resulting glitches.

ibrahimjaved12:
@HussainTaj-arbisoft  I have a few concerns related to user experience:

I think the main crux behind wanting to have preserve option was so that one category of things could be distinct from the others, right?
Eg. When an author is adding files, he/she might write "file" to preserve it so he/she can look at all the files and keep on adding them. Or "lecture" for all lecture notes, etc etc. But when we have a category, where all enlisted items do not relate to each other by name, having the preserve text option adds more work for the author than not having the option imho

Course:
- Level
  - is preserve text option needed? 
  - There are 4 levels and the only way preserve option would make sense for the use case of "graduate" for both "undergraduate" and "graduate"
- Department Numbers 
  - is preserve text option needed?
  - It would be needed for the use case of "21" because there are like 4-5 numbers with prefix of 21. But all other numbers, they don't have anything common with them.
  - Imagine having to add numbers: 2, 10, 22, CC. Every time you add one, you have to remove it to add the other one, because they're all mostly unrelated. Seems a bit work. 
- Term
  - Since only one item can be selected, therefore it's not exactly a problem to have preserve text option. But if it's not much work, I think it may be better to not preserve text when we have only one option to select (in all cases). What do you think? 
- Topics:
  - Hide selected option:
    - I think because of the nested nature in this case we didn't go for hide selected option? I mean for the "last" nested category.
    - Would it be more useful to have preserve text option only in the first category? UX wise.

HussainTaj-arbisoft:
> * Term
>   
>   * Since only one item can be selected, therefore it's not exactly a problem to have preserve text option. But if it's not much work, I think it may be better to not preserve text when we have only one option to select (in all cases). What do you think?

You're right.

> * Topics:
>   
>   * Hide selected option:
>     
>     * I think because of the nested nature in this case we didn't go for hide selected option? I mean for the "last" nested category.

Yes.
>   * Would it be more useful to have preserve text option only in the first category? UX wise.

That sounds a bit more complicated to do. And doesn't seem to be related to the scope of our ticket. Perhaps we can do it in another ticket if the need arises.

@ibrahimjaved12 Thank you for the review. You're correct. This does sound like a UX issue.

Preserving text for those fields was not a part of the ticket. I thought it would be better to have consistent behavior. I have no objections to removing the behavior from these fields. I'll start working on it.

HussainTaj-arbisoft:
@ibrahimjaved12 I've removed `preserve text` from `SelectField`. Now search queries are only preserved from `SortableSelect` component.

ibrahimjaved12 (approved):
Looks good to me.

static/js/components/widgets/SelectField.tsx
```diff
@@ -1,36 +1,57 @@
...
-    isOptionDisabled
+    placeholder: initialPlaceholder,
+    isOptionDisabled,
+    isOptionSelected,
+    hideSelectedOptions,
+    preserveSearchText = false,
+    cacheUniques
   } = props
+  const [searchText, setSearchText] = useState("")
+  const [placeholder, setPlaceholder] = useState("")
```
ChristopherChudzicki, line 51: I noticed a small UI bug with `SortableSelect`. When `FEATURE_SORTABLE_SELECT_QUICK_ADD` is in use, 
1. Search for "3play" (or anything)
2. Close the menu, so `placeholder = "3play"`
3. Open the search menu, and clear it.

The `"3play"` placeholder is still shown:

Screenshot 2023-08-14 at 11 51 13 AM

**Comment / Question:**  Is the placeholder state value necessary? Is there ever a time placeholder and searchtext are intentionally different?

ChristopherChudzicki (commented):
@HussainTaj-arbisoft I just looked this over and it's working really well. Really big improvement on the UX, and not a simple change to implement. 

I left two comments/questions, one of which is about a small UI bug.

This PR introduces some code that is unnecessary once we fully adopt the new feature. For example:
1. (very obvious) The feature flags themselves
2. props like `preserveSearchText`
3. (not obvious, IMO): `Additional` interface. *Assuming I'm correct that it won't be used once new feature is enabled.*

Could you please open an issue to remove that code once the we've fully adopted the new behavior? *Or update the acceptance criteria on https://github.com/mitodl/ocw-studio/issues/1863* 

---

Next time we introduce a feature flag for a significant redesign, maybe we should consider just copying the relevant files so that we have a clean separation between the old feature and the new, rather than conditionals mixed in. That way we can avoid unnecessary props, and removing the old code will be easier. Plus, the old feature would be more less likely to break accidentally. *This is just a thought. Certainly in simple cases a conditional in an existing component seems reasonable.*

static/js/components/widgets/SelectField.tsx
```diff
@@ -1,36 +1,57 @@
...
+import { isFeatureEnabled } from "../../util/features"
 
 export interface Option {
   label: string
   value: string
 }
 
+export interface Additional {
+  callback?: (options: Option[]) => void
+}
```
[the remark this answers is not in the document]
HussainTaj-arbisoft: > Is that correct?

Yes, your observations are correct. Additionally, the `callback` also affects `RelationField` through the `useWebsiteSelectOptions` hook.

`react-select` can either update the options through a [callback or a promise](https://react-select.com/async#loading-asynchronously). I don't know why we have used both ways. The code in the [website hook](https://github.com/mitodl/ocw-studio/blob/8544fd21e5537df359527cc522a5d0a70657fb72/static/js/hooks/websites.ts#L112) makes a debounced fetch if we are using a callback. So I assume we use the callback way when we need debounced search. 


`AsyncPaginate` has its own internal debounce feature. I forgot to enable it before. But since it came up, I'll configure it now.

> whether we are actually using that data anywhere we plan to keep.

Nope. We only need `Additional` to pass around the `callback`. Once we stop using `AsyncSelect` we won't need it anymore. At that point, we might also want to update the `useWebsiteSelectOptions` hook.

That being said, there will always be a space for `Additional` in the signatures. For example, the type `LoadOptions<Option, Option, Additional>` would become `LoadOptions<Option, Option, undefined>`.

static/js/components/widgets/SelectField.tsx
HussainTaj-arbisoft: I think I understand what's happening here. I'll fix that.

> Is the placeholder state value necessary?

Yes. When the field is unfocused, `Select` hides the input element. Which hides the search text. I tried handling it through styling but that didn't work. Handling it through the styles also causes a problem with showing selected values.

Basically, this is what happens:

https://github.com/mitodl/ocw-studio/assets/71316217/1dbab922-20ac-488b-a639-e4ad037488a1

>  Is there ever a time placeholder and searchtext are intentionally different?

Yes, the placeholder and the search text are different when the field is unfocused. More specifically the `searchText` is `""` when the field is unfocused.

The `Select` field has only one area to show the following three types of values (in order of presendence)

1. Search text (`inputValue` prop)
2. Field value
3. Placeholder

If we don't set `searchText=""`, the field won't show the field value. So we use `placeholder` as a backup **and** as a way to show the search text when nothing is selected.

HussainTaj-arbisoft:
@ChristopherChudzicki thank you for the review.

I have fixed the UI bug you discovered and answered your questions. Let me know if something I said is still unclear.

I have created https://github.com/mitodl/ocw-studio/issues/1919 for the cleanup tasks.

static/js/components/widgets/SelectField.tsx
ChristopherChudzicki: > The Select field has only one area to show the following three types of values (in order of presendence)
> 
> 1. Search text (inputValue prop)
> 2. Field value
> 3. Placeholder
> If we don't set searchText="", the field won't show the field value. So we use placeholder as a backup and as a way to show the search text when nothing is selected.

Thanks, great explanation. I think I was confused because SortableSelects never show the field value in the select field (right?); other selects, do, though. 

Anyway, that makes sense 👍

static/js/components/widgets/SelectField.tsx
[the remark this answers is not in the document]
ChristopherChudzicki: > That being said, there will always be a space for Additional in the signatures. For example, the type LoadOptions<Option, Option, Additional> would become LoadOptions<Option, Option, undefined>.

In the the `AsyncPaginate` prop `loadOptions`, yes. But I think `SelectFieldProps["loadOptions"]` could drop the 3rd arg once it's unused. Would make it clearer that we don't use the data. (Once we aren't using it, anyway).

ChristopherChudzicki (approved):
👍 Thanks for the explanations! Looks good to me.

PR #1894, opened 2023-07-20
add ol-concourse common components

gumaerc:
Closes https://github.com/mitodl/ocw-studio/issues/1895

This PR adds the `ol-concourse` library as a dependency. Some basic building blocks for our Concourse pipelines have been implemented as classes in `content_sync.pipelines.definitions.concourse.common` in the various Python files there:

 - `identifiers.py`: [`Identifier`](https://concourse-ci.org/config-basics.html#schema.identifier) objects for referring to commonly used resources the same way across your pipeline definitions
 - `image_resources.py`: [`AnonymousResource`](https://concourse-ci.org/tasks.html#schema.anonymous_resource) objects that represent Docker registry images, intended to be assigned to `image_resource` in `TaskConfig` objects
 - `resource_types.py`: [`ResourceType`](https://concourse-ci.org/resource-types.html) objects meant to be assigned to `Resource` objects via a common `Identifier` 
 - `resources.py`: [`Resource`](https://concourse-ci.org/resources.html) objects for commonly used resources in our Concourse pipelines
 - `steps.py`: [`Step`](https://concourse-ci.org/steps.html) objects common to our Concourse pipelines, along with extensions of the base `Step` objects that use `StepModifierMixin` and add our standard error handling steps, such as setting the pipeline status in `ocw-studio` and sending a Slack alert

# How can this be tested?
This PR simply adds the library and classes, they are not used yet by the application. Therefore, not much testing is required but you could instantiate some of the classes in a Django shell:

 - Spin up `ocw-studio` and in a separate terminal, run `docker-compose exec web ./manage.py shell`
 - For example, we can import the Slack alert step with `from content_sync.pipelines.definitions.concourse.common.steps import SlackAlertStep`
 - Create an example alert step by running `slack_alert = SlackAlertStep("failed", "it failed")`
 - Run `slack_alert.json(indent=2)` to generate some JSON resembling the alert step and verify that the properties we assigned are in the right spot
 - Feel free to test the other classes by instantiating them

### DIFF
--- a/content_sync/pipelines/definitions/concourse/common/steps.py
+++ b/content_sync/pipelines/definitions/concourse/common/steps.py
@@ -1,0 +1,233 @@
+import json
+from urllib.parse import urlencode, urljoin
+
+from django.conf import settings
+from ol_concourse.lib.models.pipeline import (
+    Command,
+    DoStep,
+    GetStep,
+    Identifier,
+    PutStep,
+    Step,
+    TaskConfig,
+    TaskStep,
+    TryStep,
+)
+
+from content_sync.pipelines.definitions.concourse.common.identifiers import (
+    OCW_STUDIO_WEBHOOK_RESOURCE_TYPE_IDENTIFIER,
+    OPEN_DISCUSSIONS_RESOURCE_IDENTIFIER,
+    SLACK_ALERT_RESOURCE_IDENTIFIER,
+)
+from content_sync.pipelines.definitions.concourse.common.image_resources import (
+    CURL_REGISTRY_IMAGE,
+)
+from main.utils import is_dev
+
+
+def add_error_handling(
+    step: Step, step_description: str, pipeline_name: str, instance_vars: str
+):
+    """
+    Add error handling steps to any Step-like object
+
+    Args:
+        step(Step): The Step-like object to add the error handling steps to
+        step_description(str): A description of the step at which the failure occurred
+        instance_vars(str): A query string of the instance vars from the pipeline to build a URL with
+
+    Returns:
+        None
+    """
+    concourse_base_url = settings.CONCOURSE_URL
+    concourse_team = settings.CONCOURSE_TEAM
+    concourse_path = f"/teams/{concourse_team}/pipelines/{pipeline_name}{instance_vars}"
+    concourse_url = urljoin(concourse_base_url, concourse_path)
+    on_failure_steps = [
+        OcwStudioWebhookStep(pipeline_name=pipeline_name, status="failed")
+    ]
+    on_error_steps = [
+        OcwStudioWebhookStep(pipeline_name=pipeline_name, status="errored")
+    ]
+    on_abort_steps = [
+        OcwStudioWebhookStep(pipeline_name=pipeline_name, status="errored")
+    ]
+    if not is_dev():
+        on_failure_steps.append(
+            SlackAlertStep(
+                alert_type="failed",
+                text=f"Failed - {step_description} : {concourse_url}",
+            )
+        )
+        on_error_steps.append(
+            SlackAlertStep(
+                alert_type="errored",
+                text=f"Concourse system error - {step_description} : {concourse_url}",
+            )
+        )
+        on_abort_steps.append(
+            SlackAlertStep(
+                alert_type="errored",
+                text=f"Concourse system error - {step_description} : {concourse_url}",
+            )
+        )
+    step.on_failure = TryStep(try_=DoStep(do=on_failure_steps))
+    step.on_error = TryStep(try_=DoStep(do=on_error_steps))
+    step.on_abort = TryStep(try_=DoStep(do=on_abort_steps))
+
+
+class GetStepWithErrorHandling(GetStep):
+    """
+    Extends GetStep and adds error handling
+    """
+    def __init__(
+        self, step_description: str, pipeline_name: str, instance_vars: str, **kwargs
+    ):
+        super().__init__(**kwargs)
+        add_error_handling(
+            self,
+            step_description=step_description,
+            pipeline_name=pipeline_name,
+            instance_vars=instance_vars,
+        )
+
+
+class PutStepWithErrorHandling(PutStep):
+    """
+    Extends PutStep and adds error handling
+    """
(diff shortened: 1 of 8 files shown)


content_sync/pipelines/definitions/concourse/common/steps.py
```diff
@@ -0,0 +1,248 @@
...
+    OPEN_DISCUSSIONS_RESOURCE_IDENTIFIER,
+    SLACK_ALERT_RESOURCE_IDENTIFIER,
+)
+from content_sync.pipelines.definitions.concourse.common.image_resources import (
+    CURL_REGISTRY_IMAGE,
+)
+from main.utils import is_dev
+
+
+def add_error_handling(
```
ChristopherChudzicki, line 29: Is there a reason to do the the OCW Studio Webhook  / Slack alert at the step-level instead of job-level?

I guess doing it at the step-level is pretty easy now, though.

**Also:** Should the typehint on step arg be `step: StepModifierMixin`?

content_sync/pipelines/definitions/concourse/common/steps.py
```diff
@@ -0,0 +1,248 @@
...
+        )
+        on_abort_steps.append(
+            SlackAlertStep(
+                alert_type="errored",
+                text=f"Concourse system error - {step_description} : {concourse_url}",
+            )
+        )
+    step.on_failure = TryStep(try_=DoStep(do=on_failure_steps))
+    step.on_error = TryStep(try_=DoStep(do=on_error_steps))
+    step.on_abort = TryStep(try_=DoStep(do=on_abort_steps))
```
ChristopherChudzicki, line 84: **Suggestion:** I think this should throw an error if `step` already has a failure/error/abort hook. (Or it could merge them, but I dunno if we need that now.)

content_sync/pipelines/definitions/concourse/common/steps.py
```diff
@@ -0,0 +1,248 @@
...
+    on_failure_steps = [
+        OcwStudioWebhookStep(pipeline_name=pipeline_name, status="failed")
+    ]
+    on_error_steps = [
+        OcwStudioWebhookStep(pipeline_name=pipeline_name, status="errored")
+    ]
+    on_abort_steps = [
+        OcwStudioWebhookStep(pipeline_name=pipeline_name, status="errored")
+    ]
+    if not is_dev():
```
ChristopherChudzicki, line 63: **Suggestion:** Add the slack hook even in dev.
- There's no downside as far as I know. That step just errors.
- Makes code & tests simpler

content_sync/pipelines/definitions/concourse/common/steps.py
```diff
@@ -0,0 +1,248 @@
...
+                        timeout="1m",
+                        params={"alert_type": alert_type, "text": text},
+                    )
+                ]
+            ),
+            **kwargs,
+        )
+
+
+class ClearCdnCacheStep(TaskStep):
```
ChristopherChudzicki, line 162: Why not `TaskStepWithErrorHandling`?

content_sync/pipelines/definitions/concourse/common/steps_test.py
```diff
@@ -0,0 +1,50 @@
...
+        types = []
+        for step in steps:
+            types.append(type(step))
+        if is_dev:
+            assert OcwStudioWebhookStep in types
+            assert SlackAlertStep not in types
+        else:
+            assert OcwStudioWebhookStep in types
+            assert SlackAlertStep in types
+    print(mock_step.json(indent=2))
```
ChristopherChudzicki, line 39: Did you mean to leave this in?

ChristopherChudzicki (approved):

content_sync/pipelines/definitions/concourse/common/steps.py
gumaerc: So, re: your first point, having error handlers at the step level lets us indicate more specifically about what failed in the Slack message. I suppose you're going to find out what went wrong pretty quickly once you pull up the pipeline in Concourse, but the issue remains that the URLs that the Slack alert shows by default don't properly link to our pipelines with the instance var in the query string, i.e.

https://cicd.odl.mit.edu/teams/ocw/pipelines/live/jobs/build-online-ocw-site/builds/2

vs where the link should go:

https://cicd.odl.mit.edu/teams/ocw/pipelines/live/jobs/build-online-ocw-site/builds/2?vars.site=%22matrix-calculus-for-machine-learning-and-beyond%22

What I was able to accomplish here is a link to the pipeline, but not the specific build, as that information is [intentionally withheld](https://github.com/concourse/concourse/issues/790) because they "don't want developers relying on the build number as a version number." Really, this is probably an issue with `concourse-slack-alert` itself and we should file it in their Github. If we set the alerts at the job level that would probably greatly reduce the size of the resulting JSON pipeline definition that is pushed up to Concourse, but the whole point of using this library is to obfuscate that part. We can neatly manage it in Python code, not so much just writing raw pipeline definitions.

As to the second point, I thought `Step` was more appropriate as that is what's logically being extended, but `StepModifierMixin` makes more sense.

content_sync/pipelines/definitions/concourse/common/steps_test.py
gumaerc: Nope!

content_sync/pipelines/definitions/concourse/common/steps.py
gumaerc: This is a good point, and part of what these changes are all about. It's a lot easier to make this kind of decision when it doesn't involve stripping out a bunch of YAML.

content_sync/pipelines/definitions/concourse/common/steps.py
gumaerc: Yea, I think we could add `add_on_error_step` etc. functions later that would append a step to the `do` property, but the `add_error_handling` function was only meant to be called once so this makes sense.

gumaerc:
@ChristopherChudzicki Thanks for the suggestions, this should be ready for another look

ChristopherChudzicki (approved):

PR #1896, opened 2023-07-21
overhaul theme assets pipeline definition

gumaerc:
Closes https://github.com/mitodl/ocw-studio/issues/1899

This PR re-writes the `theme-assets-pipeline.yml` pipeline definition as a class called `ThemeAssetsPipelineDefinition` that extends the `ol-concourse` `Pipeline` class.

# How can this be tested?
 - Make sure the following variables are set in your env, and make sure that you have Minio configured:
```
OCW_STUDIO_ENVIRONMENT=dev
AWS_ARTIFACTS_BUCKET_NAME=ol-eng-artifacts
AWS_PREVIEW_BUCKET_NAME=ocw-content-draft
AWS_PUBLISH_BUCKET_NAME=ocw-content-live
```
- Open a Django shell with `docker-compose exec web ./manage.py shell` and run the following:
```
content_sync.pipelines.definitions.concourse.theme_assets_pipeline import ThemeAssetsPipelineDefinition
import quote from urllib
from urllib.parse import quote
theme_assets_pipeline_definition = ThemeAssetsPipelineDefinition(
    artifacts_bucket="ol-eng-artifacts",
    preview_bucket=settings.AWS_PREVIEW_BUCKET_NAME,
    publish_bucket=settings.AWS_PUBLISH_BUCKET_NAME,
    ocw_hugo_themes_branch="main",
    instance_vars=f"?vars={quote(json.dumps({'branch': 'main'}))}"
)
f = open("theme-assets-pipeline-test.yml", "w")
f.write(theme_assets_pipeline_definition.json(indent=2, by_alias=True))
f.close()
```
 - You should have a file in the root of your `ocw-studio` repo now called `theme-assets-pipeline-test.yml`
 - You will need the `fly` CLI set up for the next part with your locally running Concourse instance added as a target, see https://concourse-ci.org/fly.html#fly-login for more info about how to use `fly login` to do that
 - Assuming you now have a fly target called `local`, run `fly -t local set-pipeline -p ocw-theme-assets-test --instance-var branch="main" -c theme-assets-pipeline-test.yml` and hit y at the prompt to push up the pipeline
 - Visit the Concourse web UI at http://localhost:8080 and login with test / test
 - Find the pipeline we just pushed up, `ocw-theme-assets-test`, unpause the pipeline and run it
 - Verify that the pipeline succeeds, look at the logs and confirm that the build completed successfully and the artifacts were uploaded to all the proper places
 - Change your `OCW_STUDIO_ENVIRONMENT` to `qa` and repeat the above steps, but stop before you push the pipeline up
 - Compare the pipeline definition to the `dev` one, the main differences should be:
   - No AWS connection info
   - Open Discussions webhook present
   - Slack alerts present

### DIFF
--- a/content_sync/pipelines/definitions/concourse/theme_assets_pipeline_test.py
+++ b/content_sync/pipelines/definitions/concourse/theme_assets_pipeline_test.py
@@ -14,6 +14,9 @@ from content_sync.pipelines.definitions.concourse.theme_assets_pipeline import (
 
 @pytest.mark.parametrize("is_dev", [True, False])
 def test_generate_theme_assets_pipeline_definition(mocker, is_dev):
+    """
+    The theme assets pipeline definition should contain the expected properties
+    """
     mock_is_dev = mocker.patch(
         "content_sync.pipelines.definitions.concourse.theme_assets_pipeline.is_dev"
     )
--- a/content_sync/pipelines/definitions/concourse/theme_assets_pipeline_test.py
+++ b/content_sync/pipelines/definitions/concourse/theme_assets_pipeline_test.py
@@ -1,0 +1,77 @@
+import json
+from urllib.parse import quote
+
+import pytest
+
+from content_sync.constants import DEV_ENDPOINT_URL
+from content_sync.pipelines.definitions.concourse.common.identifiers import (
+    OCW_HUGO_THEMES_GIT_IDENTIFIER,
+)
+from content_sync.pipelines.definitions.concourse.theme_assets_pipeline import (
+    ThemeAssetsPipelineDefinition,
+)
+
+
+@pytest.mark.parametrize("is_dev", [True, False])
+def test_generate_theme_assets_pipeline_definition(mocker, is_dev):
+    mock_is_dev = mocker.patch(
+        "content_sync.pipelines.definitions.concourse.theme_assets_pipeline.is_dev"
+    )
+    mock_is_dev.return_value = is_dev
+    artifacts_bucket = "ol-eng-artifacts"
+    preview_bucket = "ocw-content-preview"
+    publish_bucket = "ocw-content-publish"
+    ocw_hugo_themes_branch = "main"
+    instance_vars = f"?vars={quote(json.dumps({'branch': ocw_hugo_themes_branch}))}"
+    pipeline_definition = ThemeAssetsPipelineDefinition(
+        artifacts_bucket=artifacts_bucket,
+        preview_bucket=preview_bucket,
+        publish_bucket=publish_bucket,
+        ocw_hugo_themes_branch=ocw_hugo_themes_branch,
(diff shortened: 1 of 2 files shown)


content_sync/pipelines/definitions/concourse/theme_assets_pipeline_test.py
```diff
@@ -0,0 +1,80 @@
...
+    ocw_hugo_themes_branch = "main"
+    instance_vars = f"?vars={quote(json.dumps({'branch': ocw_hugo_themes_branch}))}"
+    pipeline_definition = ThemeAssetsPipelineDefinition(
+        artifacts_bucket=artifacts_bucket,
+        preview_bucket=preview_bucket,
+        publish_bucket=publish_bucket,
+        ocw_hugo_themes_branch=ocw_hugo_themes_branch,
+        instance_vars=instance_vars,
+    )
+    rendered_definition = json.loads(pipeline_definition.json(indent=2))
```
Ardiea, line 36: I know you don't have any `TryStep`s in this pipeline but w.r.t. https://github.com/mitodl/ol-infrastructure/issues/1637 

I know you don't have any `TryStep`s in this pipeline but

If you add the argument `by_alias=True` to the call to `json()` it will render the `try_` as it should be: `try`.

gumaerc: There actually are `TryStep`s in the `ClearCdnCacheStep` and `SlackAlertStep`, so this is useful. In testing I've just been calling `.replace("try_", "try")` but this is much better. Still, it would be good for the alias to be rendered by default in this case as you would never want `try_` to show up in the resulting pipeline definition.

mbertrand (approved):
LGTM,  the upsert-themes pipeline still works too.  The qa environment version is much more complex and I didn't do a deep dive comparison yet between that and what's produced by the master branch.  I assume RC is the best place to test that anyway, so approving now since I have to head out for a couple hours but I can do a more in-depth comparison when I get back if you'd like.

gumaerc:
> LGTM, the upsert-themes pipeline still works too. The qa environment version is much more complex and I didn't do a deep dive comparison yet between that and what's produced by the master branch. I assume RC is the best place to test that anyway, so approving now since I have to head out for a couple hours but I can do a more in-depth comparison when I get back if you'd like.

@mbertrand Yea, this PR doesn't modify `concourse.py`, so `upsert_theme_assets_pipeline` is unchanged as of now. This PR just implements the `ol-concourse` based pipeline definition class. I plan on opening a PR later once we have replacements for all the existing pipelines to transition to actually using them. The "qa" scenario should be testable from the perspective of simply rendering the JSON while `OCW_STUDIO_ENVIRONMENT` is not set to `dev`, but RC is a fine place to test that since this doesn't actually replace the pipeline definition that is used with the management command.

Issue #1897, opened 2023-07-24
Configure uWSGI stats server


ibrahimjaved12:
This ticket is part of https://github.com/mitodl/ocw-studio/issues/1852

### Explanation
We need to perform load testing on RC before and after [this PR](https://github.com/mitodl/ocw-studio/pull/1886) gets merged, to compare its performance.
For this we need to have `uwsgitop` present in ocw-studio-rc as well in order to monitor how its workers are operating.

### Solution
We need to add uwsgitop to requirements.txt and configure the [uWSGI stats server](https://uwsgi-docs.readthedocs.io/en/latest/StatsServer.html) (e.g. stats = /tmp/uwsgi-stats.sock in uwsgi.ini)

PR #1898, opened 2023-07-24
Configure uWSGI stats server

ibrahimjaved12:
Closes https://github.com/mitodl/ocw-studio/issues/1897

This PR adds `uwsgitop` in project and configures uWSGI stats server. This will be used to monitor how workers are operating, their response time, etc.

# How can this be tested?
This can be locally tested by checkout out to this branch, running `docker-compose build`, `docker-compose up` and then `docker-compose exec web uwsgitop /tmp/uwsgi-stats.sock` in another terminal tab. Make sure that it runs.

Although we really want to test its working in RC though. But for now this will do until this PR merges.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -3364,6 +3364,16 @@ files = [
 ]
 
 [[package]]
+name = "uwsgitop"
+version = "0.11"
+description = "uWSGI top-like interface"
+optional = false
+python-versions = "*"
+files = [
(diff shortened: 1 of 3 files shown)


ChristopherChudzicki (commented):
Now that #1887 is merged, could you resolve the conflict with
```
docker compose exec web poetry add uwsgitop
```

ChristopherChudzicki (approved):

PR #1900, opened 2023-07-25
overhaul site pipeline definition

gumaerc:
Closes https://github.com/mitodl/ocw-studio/issues/1905

This PR re-writes `site-pipeline.yml` as a class called `SitePipelineDefinition` that extends the `ol-concourse` `Pipeline` class.

# How can this be tested?
 - Spin up `ocw-studio` and make sure you have the following configured:
   - Concourse
   - Theme assets pipeline upserted & run
   - Minio
   - Github
   - Concourse `fly` CLI
 - Enter a Django shell using `docker-compose exec web ./manage.py shell`
 - Paste in the following code, replacing `test-site` with the `name` property of the site you wish to test:
```python
from websites.models import *
from content_sync.pipelines.definitions.concourse.site_pipeline import (
    SitePipelineDefinitionConfig,
    SitePipelineDefinition,
)

site_name = "6-041sc-probabilistic-systems-analysis-and-applied-probability-fall-2013"
site = Website.objects.get(name=site_name)
pipeline_config = SitePipelineDefinitionConfig(
    site=site,
    pipeline_name="draft",
    instance_vars=f"?vars=%7B%22site%22%3A%20%22{site_name}%22%7D",
    site_content_branch="preview",
    static_api_url="https://draft.ocw.mit.edu",
    storage_bucket_name="ol-ocw-studio-app",
    artifacts_bucket="ol-eng-artifacts",
    web_bucket="ocw-content-draft",
    offline_bucket="ocw-content-offline-draft",
    resource_base_url="",
    ocw_studio_url="http://10.1.0.102:8043",
    ocw_hugo_themes_branch="main",
    ocw_hugo_projects_branch="main",
)
pipeline_definition = SitePipelineDefinition(config=pipeline_config)
f = open("site-pipeline-test.yml", "w")
f.write(pipeline_definition.json(indent=2, by_alias=True))
f.close()
```
 - After running the above, you should have a file called `site-pipeline-test.yml` at the root of your repo
 - Use the `fly` CLI to push up the pipeline definition to your locally running Concourse, again replacing `test-site` with the `name` of your site:
```bash
fly -t local set-pipeline -p draft --instance-var site=test-site -c site-pipeline-test.yml
```
 - Open the Concourse web UI at http://localhost:8080/ and log in
 - Search for the pipeline we just pushed up by typing the `name` of your site in the search bar and click on the pipeline in the `draft` group
 - Open the `ocw-studio` UI at http://localhost:8043/ in a new tab, pull up your site and publish the draft version there
 - Switch back over to the Concourse UI
 - The pipeline run should have started, watch it and ensure that it completes
 - Manipulate the arguments to the `SitePipelineDefinition` object in the shell code above to push up different versions of the pipeline, for example:
   - `live` version of the above pipeline:
     - `pipeline_name="live"`
     - `site_content_branch="release"`
     - Remove the `--buildDrafts` argument from the Hugo args
     - In the `fly` command, `-p live` instead of `-p draft`
   - `draft` version of `ocw-www`:
     - `site_name = "ocw-www"`
     - `is_root_website=True`
     - `base_url="/"`
     - `static_resources_subdirectory="/ocw-www/`
     - remove `--baseUrl` arguments from Hugo commands
     - `delete_flag=""`

# Additional Context
This process should test much of the functionality, but some of it should be tested in RC. The branch that this PR merges, `cg/overhaul-site-pipeline`, does not include the changes to `concourse.py` that would actually switch over to using `SitePipelineDefinition` with the `backpopulate_pipelines` management command. So, this PR simply adds the class and tests for it.

# Screenshots
image


content_sync/pipelines/definitions/concourse/site_pipeline.py
```diff
@@ -0,0 +1,656 @@
...
+                {
+                    "endpoint": DEV_ENDPOINT_URL,
+                    "access_key_id": (settings.AWS_ACCESS_KEY_ID or ""),
+                    "secret_access_key": (settings.AWS_SECRET_ACCESS_KEY or ""),
+                }
+            )
+        offline_build_gate_resource = Resource(
+            name=self._offline_build_gate_identifier,
+            type=self._keyval_resource_type.name,
+            check_every="never",
```
Ardiea, line 197: My only question here is: "Are the `check_every` settings defined via this code consistent with the current `check_every` definitions across the existing jobs? This was a _major_ source of performance degradation. We just don't want to inadvertently revert that.

Ardiea: Specifically w.r.t. the ocw_hugo_themes_resource, ocw_hugo_projects_resource and the other resources defined below.

mbertrand (changes requested):
The PR description mentions running `backpopulate_pipelines` but I don't think that actually tests the new code?  Should I test in a similar way to https://github.com/mitodl/ocw-studio/pull/1896 or just leave it for RC?

Code looks great but could use some inline documentation, had a suggestion about all the inputs required for `SitePipelineDefinition`, take or leave.

gumaerc:
@mbertrand Sorry for the confusion. The branch this PR merges is `cg/overhaul-site-pipeline`, and the branch I suggested be checked out for testing is `cg/overhaul-site-pipeline-testing`, which contains all the changes in this PR and also changes to `concourse.py` to actually use the definition with `backpopulate_pipelines`. I didn't want that change to be in this PR, but it's easiest for testing this because the constructor is so large and requires so much contextual information that it would be annoying to reproduce in a Django shell. You could of course instantiate an object in a shell, but I figured it would be easier to test this way.

gumaerc:
@mbertrand After mocking up a new staggered mass build pipeline locally and trying to use the jobs defined here, I want to make some changes before merging this. In the new mass build pipeline, multiple build jobs for multiple sites will be added to the same pipeline definition. In order to accomplish this, the identifiers for certain things such as the site content repo need to be truly unique. Your other suggestions make sense and I'll be addressing those as well.

content_sync/pipelines/definitions/concourse/site_pipeline.py
```diff
@@ -0,0 +1,656 @@
...
+    _slack_resource = SlackAlertResource()
+
+    _config = SitePipelineDefinitionConfig()
+
+    class Config:
+        arbitrary_types_allowed = True
+
+    def __init__(
+        self,
+        site: Website,
```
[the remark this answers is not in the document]
blarghmatey: Pydantic models automatically generate the `__init__` signature based on the class attributes, so you should be able to avoid that entirely.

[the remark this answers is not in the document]
gumaerc: @blarghmatey The `SitePipelineDefinitionConfig` class I created is not a Pydantic class, just a normal Python class that I built to more conveniently pass arguments around. Is it worth extending a Pydantic class to make property assignment more automatic?

gumaerc:
@mbertrand I made changes to this based on everyone's suggestions, as well as some changes to make the different parts of the pipeline more modular. I found that this was going to be necessary while experimenting with building a new mass build pipeline using the code I've written here. The resource types , resources and tasks are all now their own classes that can be instantiated individually. This way, when I go to build the mass build pipeline I can dynamically create jobs with the tasks and resources necessary, then add them to a different pipeline object. The resources that would need to be unique when combined into the same job have also had their identifiers modified to be unique:

 - Site content git repo (`short_id` appended to the end)
 - Webpack manifest (`ocw_hugo_themes_branch` appended at the end)
 - OCW Studio webhook (`short_id` appended at the end)
 
 This should help when going to build the mass build pipeline, as when these resources are added to the pipeline for multiple sites they will not have the same identifier and overwrite each other, but sites that use the same resource (the `ocw-hugo-themes` resource for instance) will not have that resource duplicated, only the ones that need to be unique.
 
 I also updated the testing instructions to be able to test this in a Django shell without checking out another branch. Let me know if you have questions.

blarghmatey:
@gumaerc I just published v0.5.0 of ol_concourse. This updated Pydantic to V2 which should be largely invisible to your usage, as well as adding the config to properly serialize the `try` step. Let me know if you run into any issues.

content_sync/pipelines/definitions/concourse/site_pipeline.py
```diff
@@ -0,0 +1,820 @@
...
+        self.pipeline_name = pipeline_name
+        self.static_api_url = static_api_url
+        self.storage_bucket_name = storage_bucket_name
+        self.artifacts_bucket = artifacts_bucket
+        self.web_bucket = web_bucket
+        self.offline_bucket = offline_bucket
+        self.resource_base_url = resource_base_url
+        if (
+            self.site_content_branch == "preview"
+            or settings.OCW_STUDIO_ENVIRONMENT not in PRODUCTION_NAMES
```
mbertrand, line 131: Should be `settings.ENVIRONMENT` instead.  This will error if `site_content_branch == "release"`

From settings.py:
```python
ENVIRONMENT = get_string(
    name="OCW_STUDIO_ENVIRONMENT",
    default="dev",
    description="The execution environment that the app is in (e.g. dev, staging, prod)",
    required=True,
)
```

mbertrand (changes requested):
Still looking over, but here's one thing I noticed in the meantime

content_sync/pipelines/definitions/concourse/site_pipeline.py
gumaerc: Good catch, although this should be `settings.ENV_NAME` as seen here: https://github.com/mitodl/ocw-studio/blob/9ecb6e0e65db3c8ead578519bead8af0986a4fb2/content_sync/pipelines/concourse.py#L437

content_sync/pipelines/definitions/concourse/site_pipeline_test.py
```diff
@@ -0,0 +1,562 @@
...
+):
+    """
+    The site pipeline definition should contain the expected properties
+    """
+    settings.AWS_ACCESS_KEY_ID = "test_access_key_id"
+    settings.AWS_SECRET_ACCESS_KEY = "test_secret_access_key"
+    settings.CONCOURSE_IS_PRIVATE_REPO = concourse_is_private_repo
+    settings.OCW_HUGO_THEMES_SENTRY_DSN = "test_sentry_dsn"
+    settings.ROOT_WEBSITE_NAME = "root-website"
+    settings.OCW_STUDIO_ENVIRONMENT = env_name
```
mbertrand, line 92: I think this should be:
```python
settings.ENV_NAME = env_name
```

mbertrand (approved):
👍 assuming tests pass

PR #1903, opened 2023-07-25
remove requirements.txt

ChristopherChudzicki:
#1887 

This removes requirements.txt and requirements.in. They should have been removed in #1887 , but I made an error when resolving a merge conflict.

### How can this be tested?
1. Tests should pass.
2. If you like, you can check that `pyproject.toml` is consistent with the removed requirements.txt by:
```
git checkout HEAD~1
cat requirements.txt | xargs poetry add
```
`poetry.lock` and `pyproject.toml` should be unchanged. *Well, the lockfile might change if new versions of dependencies are published before this gets reviewed.*


gumaerc (approved):
:+1:

Issue #1938, opened 2023-09-06
OcwStudioWebhookStep and other PutStep steps unnecessarily pulling in all outputs as inputs


gumaerc:
### Steps to Reproduce

Create and push up any site using `SitePipelineDefinition`, for example:

```
from websites.models import *
from content_sync.pipelines.definitions.concourse.site_pipeline import (
    SitePipelineDefinitionConfig,
    SitePipelineDefinition,
)

site_name = "6-041sc-probabilistic-systems-analysis-and-applied-probability-fall-2013"
site = Website.objects.get(name=site_name)
pipeline_config = SitePipelineDefinitionConfig(
    site=site,
    pipeline_name="draft",
    instance_vars=f"?vars=%7B%22site%22%3A%20%22{site_name}%22%7D",
    site_content_branch="preview",
    static_api_url="https://draft.ocw.mit.edu",
    storage_bucket="ol-ocw-studio-app",
    artifacts_bucket="ol-eng-artifacts",
    web_bucket="ocw-content-draft",
    offline_bucket="ocw-content-offline-draft",
    resource_base_url="",
    ocw_studio_url="http://10.1.0.102:8043",
    ocw_hugo_themes_branch="main",
    ocw_hugo_projects_branch="main",
)
pipeline_definition = SitePipelineDefinition(config=pipeline_config)
f = open("site-pipeline-test.json", "w")
f.write(pipeline_definition.json(indent=2, by_alias=True))
f.close()
```

### Expected Behavior

The `ocw-studio-webhook`, `open-discussions-webhook` and `slack-alert` steps should not need to pull in any volumes in order to run

### Actual Behavior

These steps streams all volumes before they run:

```
selected worker: ip-10-1-1-129
streaming volume ocw-hugo-themes-git from ip-10-1-0-84
streaming volume site-content-git from ip-10-1-0-84
streaming volume static-resources-s3 from ip-10-1-3-59
streaming volume ocw-hugo-projects-git from ip-10-1-0-94
```

This is extra apparent if the site being built has a significant amount of data being pulled in during the `static-resources-s3` step

PR #1939, opened 2023-09-06
explicitly set inputs to a blank list on put steps that don't need them

gumaerc:
Closes https://github.com/mitodl/ocw-studio/issues/1938

This PR fixes an issue caused by not explicitly setting `inputs` to a blank list on objects of type `PutStep`. By default, the `put` step pulls in *all* outputs as inputs which causes major issues if any steps in the pipeline are outputting large amounts of data.

# How can this be tested?
This cannot actually be tested locally, because volume streaming is not a thing when running Concourse through Docker.

# Additional Context
This issue was found while testing the new `MassBuildSitesPipelineDefinition` in RC, where volume streaming is enabled and sites with large amounts of archive videos exist.


gumaerc:
@pt2302 I added the test

Issue #1943, opened 2023-09-06
ClearCdnCacheStep does not function properly


gumaerc:
### Steps to Reproduce

 - Push up a pipeline that uses `ClearCdnCacheStep` to Concourse, the easiest of which to test is the theme assets pipeline with `docker-compose exec web ./manage.py upsert_theme_assets_pipeline`
 - Run the pipeline

### Expected Behavior

The `ClearCdnCacheStep` runs properly

### Actual Behavior

It errors out

PR #1944, opened 2023-09-06
fix ClearCdnCacheStep

gumaerc:
Closes https://github.com/mitodl/ocw-studio/issues/1943

This PR fixes a number of issues with `ClearCdnCacheStep`, including:

 - The `path` property on the `Command` in the `run` property is just plain wrong; it's `sh` when it should be `curl`
 - When pushing up JSON pipeline definitions to Concourse, strings in the `args` field that have a space in them are automatically wrapped in single quotes. If you add manual single quotes, you will get doubles. This isn't the case for YAML, where if you wrap a string in single quotes you only end up with that set of single quotes.
 - The `settings.CONCOURSE_HARD_PURGE` boolean was being evaluated incorrectly
 - The use of `ClearCdnCacheStep` in the online and offline steps of `site_pipeline.py` were not properly setting `fastly_var`

A new test was added to make sure `ClearCdnCacheStep` renders properly.

# How can this be tested?
This needs to be tested in RC, as the fastly service ID and API token are stored in Vault. Nonetheless, you should still be able to run `docker-compose exec web ./manage.py upsert_theme_assets_pipeline` with your `OCW_STUDIO_ENVIRONMENT` set to values other than `dev` and not get an error.  If you try and run the pipeline, you will get an error about the `fastly_draft` / `fastly_live` vars missing.

### DIFF
--- a/content_sync/conftest.py
+++ b/content_sync/conftest.py
@@ -12,6 +12,12 @@ def mock_environments(settings, request):
     settings.ENV_NAME = request.param
 
 
+@pytest.fixture(params=[True, False])
+def mock_concourse_hard_purge(settings, request):
+    """Fixture that tests with True and False for settings.CONCOURSE_HARD_PURGE"""
+    settings.CONCOURSE_HARD_PURGE = request.param
+
+
 @pytest.fixture
 def mock_branches(settings, mocker):
     """Return mock github branches with names"""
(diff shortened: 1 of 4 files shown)


mbertrand (approved):
Code looks good, pipeline was generated successfully, 👍 assuming tests pass.

PR #1930, opened 2023-08-30
config,refactor: Add more extensive pre-commit config and lint rules

blarghmatey:
# What are the relevant tickets?
N/A

# Description (What does it do?)
This does a few things:

- Replaces usage of pylint with ruff for linting Python code
- Applies all automated Ruff fixes to the Python code
- Adds noqa comments to all existing linter violations found by Ruff
- Adds Prettier v3.x
- Applies prettier and eslint fixes to all relevant files
- Updates eslint config to be compatible with prettier
- Adds pre-commit rules for running all linters and fixes automatically
- Removes issue and PR templates from the repo so that it will fall back to the ones in mitodl/.github

# How can this be tested?
This was tested by running `pre-commit run --all-files` and ensuring that the output all passes.

# Additional Context
This is related to the proposals outlined in https://github.com/mitodl/hq/discussions/1422

PR #1946, opened 2023-09-08
update poetry lock file

gumaerc:
N/A

Recently https://github.com/mitodl/ocw-studio/pull/1930 was merged to update the pre-commit and linting configuration in `ocw-studio`. It caused some issues with poetry as well as running pre-commit locally. This PR implements some fixes:

 - `poetry lock && poetry install` was run to update the `poetry.lock` file
 - `shfmt` was removed from the `pre-commit` configuration for now as it throws an error when you run it locally

# How can this be tested?
If you rebase a PR off of this branch, Github Actions should run successfully. You should also be able to make a commit to your branch locally without pre-commit throwing an error.


ibrahimjaved12 (approved):
LGTM

Issue #1972, opened 2023-09-14
incorrect ocw-studio webhook url in MassBuildSitesPipelineDefinition


gumaerc:
# Description/Context
When generating a pipeline definition with `MassBuildSitesPipelineDefinition`, the `ocw-studio-webhook` resource ends up having a `url` property of `https://ocw-studio-rc.odl.mit.edu/api/websites/mass-build-sites/pipeline_status/`. When a call to this is made, `ocw-studio` returns a 404 because `mass-build-sites` is not a site that exists to set the status on.

# Plan/Design
 - Instead of hard coding the site name in the URL, the var should be used instead, like `https://ocw-studio-rc.odl.mit.edu/api/websites/((.:site.site_name))/pipeline_status/`

PR #1973, opened 2023-09-14
fix ocw-studio-webhook URL in MassBuildSitesPipelineDefinition

gumaerc:
Closes https://github.com/mitodl/ocw-studio/issues/1972

This PR makes a change to `mass_build_sites.py` so that the `site_name` Concourse var is used in the `ocw-studio-webhook` resource's `url` property, rather than hard coding the value. This allows the mass build to update the status on individual sites it is building while maintaining the functionality on the individual site pipeline.

# How can this be tested?
 - Follow the instructions in https://github.com/mitodl/ocw-studio/pull/1923 to smoke test the mass build pipeline
 - When running the first batch of the pipeline, instead of just letting it run, abort the build when any site gets to the `static-resources-s3` step
 - The `ocw-studio-webhook` step should fire to set the status on each site that is aborted and the URL should reflect the site being aborted, i.e. something like `http://10.1.0.102:8043/api/websites/10-01-ethics-for-engineers-artificial-intelligence-spring-2020/pipeline_status/`
 - If you visit the `ocw-studio` UI for any given site (i.e. http://localhost:8043/sites/10-01-ethics-for-engineers-artificial-intelligence-spring-2020/ in the example above) the status should show "aborted"

# Additional Context
This was discovered while testing the pipeline in RC at the CDN cache clearing step, which does not get run in dev. Therefore, we are testing it here with the abort hook. This issue was not discovered previously because while testing in RC previous to https://github.com/mitodl/ocw-studio/pull/1960 being merged, the `ocw_studio_url` property had been edited from the dev version which uses `http` instead of `https`. Calling the RC or production servers with `http` redirects to `https` and returns a 300, which does not cause the step to fail.

### DIFF
--- a/content_sync/pipelines/definitions/concourse/mass_build_sites.py
+++ b/content_sync/pipelines/definitions/concourse/mass_build_sites.py
@@ -22,7 +22,6 @@ from content_sync.pipelines.definitions.concourse.common.identifiers import (
     KEYVAL_RESOURCE_TYPE_IDENTIFIER,
     MASS_BUILD_SITES_BATCH_GATE_IDENTIFIER,
     MASS_BUILD_SITES_JOB_IDENTIFIER,
-    MASS_BULID_SITES_PIPELINE_IDENTIFIER,
     OCW_HUGO_PROJECTS_GIT_IDENTIFIER,
     OCW_HUGO_THEMES_GIT_IDENTIFIER,
     WEBPACK_MANIFEST_S3_IDENTIFIER,
@@ -131,6 +130,7 @@ class MassBuildSitesResources(list[Resource]):
     """
 
     def __init__(self, config: MassBuildSitesPipelineDefinitionConfig):
+        site_pipeline_vars = get_site_pipeline_definition_vars("site")
         webpack_manifest_resource = WebpackManifestResource(
             name=WEBPACK_MANIFEST_S3_IDENTIFIER,
             bucket=config.artifacts_bucket,
@@ -149,7 +149,7 @@ class MassBuildSitesResources(list[Resource]):
         self.append(ocw_hugo_projects_resource)
         self.append(
             OcwStudioWebhookResource(
-                site_name=MASS_BULID_SITES_PIPELINE_IDENTIFIER,
+                site_name=site_pipeline_vars["site_name"],
                 api_token=settings.API_BEARER_TOKEN or "",
             )
         )
--- a/content_sync/pipelines/definitions/concourse/mass_build_sites.py
+++ b/content_sync/pipelines/definitions/concourse/mass_build_sites.py
@@ -130,7 +130,7 @@ class MassBuildSitesResources(list[Resource]):
     """
 
     def __init__(self, config: MassBuildSitesPipelineDefinitionConfig):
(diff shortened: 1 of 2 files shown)


mbertrand (commented):
When I run the pipeline, there seem to be batches within batches. For instance, I started "batch-1", and there are x sites in the batch, and it starts to run 10 sites at a time.  When I abort, those 10 sites get a webhook sent updating the publish status to "Aborted", but the other sites in "batch-1" that had not been started don't.  Is that the desired behavior?

Screenshot 2023-09-20 at 9 38 34 AM


--------------

Something else I noticed, that may not be related to this PR: if I let the pipeline go ahead and run for awhile, sites that fail on the "build-online-site" step get a webhook sent,  but the published status is set to null:


Screenshot 2023-09-20 at 9 49 54 AM

Screenshot 2023-09-20 at 9 51 45 AM

mbertrand (commented):
One other question: this pipeline definition should be updated before every run to ensure it has all the latest sites, right?

gumaerc:
@mbertrand These are all interesting questions about how the new pipeline works. This PR specifically is just about correcting the base URL used when constructing the webhook URL, but I think all of these are worth thinking about.

Firstly, the way this works is that it runs "batches of batches." Basically, when constructing the pipeline `ocw-studio` divides all publishable courses into the jobs you see in the pipeline, controlled by `settings.OCW_MASS_BUILD_BATCH_SIZE`. This is mostly done to keep the Concourse UI manageable so it doesn't freeze trying to load a gigantic list of `across` steps. Then, each `across` step has `max_in_flight` set, controlled by `settings.OCW_MASS_BUILD_MAX_IN_FLIGHT`, which controls how many sites are built at any given time. I think it does make sense to only show aborted for sites that were actually in progress, as the other site builds would technically have not even started yet. Although, this does make a lot of noise with the Slack alerts. I would prefer to find a way to send only one alert to Slack to indicate that the mass build was aborted / failed so maybe we need to tune that. However, I do think it makes sense to set an aborted status on any site that was actually in progress. What if they were aborted on the upload step, and artifacts were actually uploaded to the bucket but the CDN cache was not cleared? I think we would want to see that site's build status as aborted.

I noticed that the `STATIC_API_BASE_URL` setting your builds are using is pointing to your locally hosted Minio sites. You may want to apply the following settings:

```
OCW_STUDIO_DRAFT_URL=http://localhost:8044/
OCW_STUDIO_LIVE_URL=http://localhost:8045/
STATIC_API_BASE_URL_DRAFT=https://draft.ocw.mit.edu
STATIC_API_BASE_URL_LIVE=https://ocw.mit.edu
```

This will make it so that the links in the publish drawer properly point to Minio, allowing you to see the output of your builds. However, the `STATIC_API_BASE_URL` setting that each build uses will pull from prod draft / live. I will need to look into the null status, that doesn't seem right.

Finally, per your last comment, yes this pipeline definition will need to be updated each time a site becomes eligible for publishing or a site becomes "unpublished." The plan is to do this automatically whenever any of those events occurs, and that code will be implemented in a future PR that actually sets the pipeline to be used in `concourse.py` Currently this pipeline is still in the testing phase.

mbertrand (approved):
👍

Issue #1978, opened 2023-09-19
Download course archives are missing for freshly (single site) built courses


HussainTaj-arbisoft:
Go to any course and view the download page.

e.g. https://ocw.mit.edu/courses/15-014-applied-macro-and-international-economics-ii-spring-2016/download/ 

## Expected Behavior

Clicking on the button should download the offline course archive.

## Current Behavior

We're redirected to a 404.

## Steps to Reproduce

1. Open any course in RC.
2. Publish the course.
3. Open the download page, and click download.

## Possible Solution

@gumaerc debugged this to be an issue with a bash variable `IS_ROOT_WEBSITE`.

PR #1931, opened 2023-08-31
use new site pipeline definition

gumaerc:
# What are the relevant tickets?
https://github.com/mitodl/ocw-studio/issues/1928

# Description (What does it do?)
This PR replaces the use of the static `site-pipeline.yml` definition with the new `SitePipelineDefinition` object based pipeline.

# How can this be tested?
 - Spin up `ocw-studio`, making sure you have Minio and a custom Github org configured and have run the theme assets pipeline at least once (`docker-compose exec web ./manage.py upsert_theme_assets_pipeline`)
 - Pick a course to test and note its `short_id` property
 - Run `docker-compose exec web ./manage.py backpopulate_pipelines --filter short_id`, replacing `short_id` with the ID of your site
 - In the `ocw-studio` UI, publish any site
 - Open the Concourse UI at http://localhost:8080/ and find the pipeline for your test site
 - Watch the pipeline run, inspecting each step as it runs and make sure that everything looks correct:
   - Steps being run in the proper order
   - Files being synced appropriately to the right places
   - No errors
 - Back in the `ocw-studio` UI, click the link in the publish drawer to browse the deployed site and verify that everything looks good

PR #1979, opened 2023-09-19
explicitly check value of IS_ROOT_WEBSITE as an integer

gumaerc:
Closes https://github.com/mitodl/ocw-studio/issues/1978

This PR corrects the logic surrounding how the `IS_ROOT_WEBSITE` environment variable is evaluated in pipelines. In https://github.com/mitodl/ocw-studio/pull/1900 we replaced the static `site-pipeline.yml` with the pydantic model `SitePipelineDefinition`. In https://github.com/mitodl/ocw-studio/pull/1931, we moved from injecting values into the pipeline directly to using the Concourse vars system, so that the steps can be reused in other pipelines and used with the `across` step. As part of this, `IS_ROOT_WEBSITE` was injected as an environment variable to be evaluated at runtime. This was necessary to maintain compatibility with the way the `across` step works. If we were to evaluate `is_root_website` in Python and then not include blocks of code based on that, then the pipeline definition would be run that way for each of the sites in the `across`. The main issue here is that the `IS_ROOT_WEBSITE` env variable had to be added as an integer, rather than a boolean. Concourse threw errors setting a boolean value in the var source. The logic performed a falsy check on these values, which did not function correctly. Instead, these checks have been set up to explicitly check for the numerical 0 or 1 value. Tests have been added to make sure the proper checks show up at the right place in the pipeline, but there's no good way to truly unit test this functionality, as it would require running the pipeline.

# How can this be tested?
 - Make sure you have an up to date database restore and are set up for publishing courses locally
 - Run the following for a test site (noted here by `test-site` that is *not* the root website and also the root website, noted by `ocw-www`: 
```
docker compose exec web ./manage.py backpopulate_pipelines --filter test-site
docker compose exec web ./manage.py backpopulate_pipelines --filter ocw-www
```
 - Publish both the sites and inspect the output in Concourse
 - For `test-site`, in `upload-online-build` under `online-site-job`, you should see `+ '[' 0 = 1 ']'`. The `aws s3 sync` command should have a `--delete` flag and `--exclude` flags for the zip archives, which prevent them from being removed in the destructive sync.
 - For `test-site`, in `build-offline-site` under `offline-site-job`, you should see `[ 0 = 0 ]` which indicates that this is not the root website and a zip archive should be built. You should see the archive being created. In the `upload-offline-build` step, you should see `'[' 0 = 0 ']'`, indicating that this is not the root website and zip archives should be uploaded.
 - For `ocw-www`, in `upload-online-build` under `online-site-job`, you should see `+ '[' 1 = 1 ']'`. The `aws s3 sync` command should have no `--exclude` flags and no `--delete` flag.
 - For `ocw-www`, in `build-offline-site` under `offline-site-job`, you should see `[ 1 = 0 ]` which indicates that this is the root website and a zip archive should not be built. You should see no output below this line. In the `upload-offline-build` step, you should see `'[' 1 = 0 ']'` with nothing below it, indicating that this is the root website and zip archives should not be uploaded.

### DIFF
--- a/content_sync/pipelines/definitions/concourse/site_pipeline.py
+++ b/content_sync/pipelines/definitions/concourse/site_pipeline.py
@@ -495,7 +495,7 @@ class SitePipelineOnlineTasks(list[StepModifierMixin]):
             ] = settings.AWS_SECRET_ACCESS_KEY
 
         online_sync_command = f"""
-        if [ $IS_ROOT_WEBSITE ] ; then
+        if [ $IS_ROOT_WEBSITE = 1 ] ; then
             aws s3{get_cli_endpoint_url()} sync {SITE_CONTENT_GIT_IDENTIFIER}/output-online s3://{pipeline_vars['web_bucket']}/{pipeline_vars['base_url']} --metadata site-id={pipeline_vars['site_name']}{pipeline_vars['delete_flag']}
         else
             aws s3{get_cli_endpoint_url()} sync {SITE_CONTENT_GIT_IDENTIFIER}/output-online s3://{pipeline_vars['web_bucket']}/{pipeline_vars['base_url']} --exclude='{pipeline_vars['short_id']}.zip' --exclude='{pipeline_vars['short_id']}-video.zip' --metadata site-id={pipeline_vars['site_name']}{pipeline_vars['delete_flag']}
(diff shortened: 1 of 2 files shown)


HussainTaj-arbisoft (approved):
Working as expected. I can see the correct files in MinIO as well.

👍 LGTM

PR #1984, opened 2023-09-22
Fix italics font size issue in CKEditor

pt2302:
Closes https://github.com/mitodl/hq/issues/2215.

Previously, there was an issue where making text italic in the `Caption` or `Credit` fields in `Image Metadata` for an Image Resource would increase the font size in CKEditor. While this did not affect published output, it was not intended. This PR resolves that issue. 

# How can this be tested?

For an Image resource in OCW Studio, verify that text in the `Caption` and `Credit` fields under `Image Metadata` can now be made italic using the Markdown editor without changing the font size.

Verify that the font sizes in the full Markdown editor (such as in `Pages`) are not affected when changing to and from italics for any of the Heading types. 

# Additional Context
This PR does something similar for block quotes: https://github.com/mitodl/ocw-studio/pull/974.

### DIFF
--- a/static/js/components/widgets/ObjectField.tsx
+++ b/static/js/components/widgets/ObjectField.tsx
@@ -45,9 +45,9 @@ export default function ObjectField(props: Props): JSX.Element {
         >
           {field.label}
         </label>
-        <i className="material-icons">
+        <span className="material-icons">
           {collapsed ? "expand_more" : "expand_less"}
-        </i>
+        </span>
       </div>
       {collapsed ? null : (
         <div className="object-sub-fields">
--- a/static/scss/ckeditor.scss
+++ b/static/scss/ckeditor.scss
@@ -9,12 +9,6 @@
   }
 
   .ck-content {
-    p {
-      i {
-        font-size: initial;
-      }
-    }
-
     blockquote {
       font-style: initial;
     }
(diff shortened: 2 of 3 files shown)


HussainTaj-arbisoft (approved):
LGTM 👍

ChristopherChudzicki:
Took a look at the root cause here since @pt2302 asked me about it.

@pt2302 The root cause of this is shown in screenshot below:
- metadata fields are an "object-widget", which causes them to be styled a little differently **and** have the collapse/expand button
- objet-widget uses CSS selector `.object-widget i` to control font size. 
    - I **strongly** suspect this is only intended to select the expand more / expand less icon
- CKEditor also uses `<i>` tag for italic fonts. 

Screenshot 2023-09-25 at 9 31 30 AM


**Icons in HTML:** There is some debate around whether using `<i>` for icons is appropriate, see e.g., https://stackoverflow.com/questions/11135261/what-are-the-advantages-disadvantages-of-using-the-i-tag-for-icons-instead-of. IMO, (*the second answer is the best: it's reasonable and common, so fine.*). But:
    - bootstrap 5 uses `<i>` for icons https://icons.getbootstrap.com/icons/2-circle/
    - Google's material-icons recommends using `<span>`s, apparently https://developers.google.com/fonts/docs/material_icons#using_the_icons_in_html

### What to do

IMO, the best thing to do is probably to change the font-size selector from `.object-widget i` to `.object widget .material-icons` (as opposed to applying extra CSS to ckeditor). The problem isn't CKEditor styling things wrong, the problem is our other CSS selecting too broadly.

PR #1986, opened 2023-09-23
Make return type of is_gdrive_enabled Boolean

pt2302:
Closes https://github.com/mitodl/hq/issues/2459.

Changes the return type of the `is_gdrive_enabled()` function into a Boolean value, which is what it was intended to be.

# How can this be tested?
Start up OCW Studio on this branch using `docker compose up` and verify that the page source reads `"gdrive_enabled": true`, assuming that Google Drive integration is set up. Verify that Google Drive works as expected, including syncing.


HussainTaj-arbisoft (approved):
LGTM 👍 

Python is too quirky at times.

Issue #1998, opened 2023-10-13
root website sync takes a long time


gumaerc:
# Description/Context
The site pipeline, defined in `content_sync.pipelines.definitions.concourse.site_pipeline.py`, takes a much longer time to run than other sites. More specifically, When the root website is using the [`ocw-www` configuration from `ocw-hugo-projects`](https://github.com/mitodl/ocw-hugo-projects/blob/main/ocw-www/ocw-studio.yaml) and a large number of "instructor" content is added, the s3 sync time at the end of the pipeline is heavily inflated. This happens because of the large amount of small files that need to be synced, and the fact that the sync is happening against the root of the bucket. This could also be a problem for any future content added to a root website configuration where thousands of items will be added as well. This happens because:

 - Since the root website exists at the root of the S3 bucket, if a `sync` command is used to push this content up then a comparison must be made by the AWS CLI against every single object in the bucket, including all other sites already deployed
 - Every file transferred with the AWS CLI uses its own connection from a pool of connections that is created in the background, the default of which is 10 connections max as defined by `max_concurrent_connections`

# Plan/Design
 - For the root website, use `cp` instead of `sync` so that comparisons don't need to be run
 - In sync steps in the pipeline, set `max_concurrent_connections` to a settings value (so it can be easily changed) that defaults to 10

PR #1999, opened 2023-10-13
root website pipeline improvements

gumaerc:
Closes https://github.com/mitodl/ocw-studio/issues/1998

This PR makes some optimizations to the AWS CLI commands used when syncing build output to S3:

 - The `aws s3 sync` commands for the root website have been changed to `aws s3 cp --recursive`. This reduces the amount of time it takes to copy large amounts of small files into S3 because a comparison doesn't need to be made with all of those files against everything that already exists in the bucket.
 - The `AWS_MAX_CONCURRENT_CONNECTIONS` setting was added with a default of 10 (the same as the CLI default) and this value is set on `max_concurrent_connections` in the AWS CLI configuration before sync steps are run. This allows us to experiment with setting the value higher to see how it improves performance, or if it causes issues with things like the mass build.

# How can this be tested?
 - Check out the `master` branch and make sure you have a recent database restore as well as all the prerequisites outlined in the README for publishing courses locally set up. In your `.env` file, also set `AWS_MAX_CONCURRENT_CONNECTIONS=100`, then spin up `ocw-studio` with `docker-compose up`
 - Upsert the `ocw-www` pipeline to your local Concourse with `docker-compose exec web ./manage.py backpopulate_pipelines --filter ocw-www`
 - Open the `ocw-studio` UI (http://localhost:8043/sites/) as well as the Concourse UI (http://localhost:8080) in two tabs
 - Publish `ocw-www` from the `ocw-studio` UI, then find the running pipeline in the Concourse UI by typing `status: running` into the search bar
 - Run the build 2 more times by clicking the plus sign in the upper right of the Concourse UI in `online-site-job`
 - Take the average runtime of the 3 runs
 - Check out this PR branch and repeat the above steps from the backpopulate step and get an average runtime against this branch
 - The average runtime on this branch should be a bit shorter
 - If you're comparing against QA / production, keep in mind that your local build time should be much shorter because copying to Minio locally takes a lot less time than syncing to S3 proper

### DIFF
--- a/app.json
+++ b/app.json
@@ -38,6 +38,10 @@
       "description": "S3 artifacts bucket name.",
       "required": false
     },
+    "AWS_MAX_CONCURRENT_CONNECTIONS": {
+      "description": "The max concurrent connections used by cp and sync AWS CLI commands",
+      "required": false
+    },
     "AWS_OFFLINE_PREVIEW_BUCKET_NAME": {
       "description": "S3 offline preview bucket name.",
       "required": false
--- a/content_sync/pipelines/definitions/concourse/site_pipeline.py
+++ b/content_sync/pipelines/definitions/concourse/site_pipeline.py
@@ -452,6 +452,9 @@ class SitePipelineOnlineTasks(list[StepModifierMixin]):
                 timeout="20m",
                 attempts=3,
                 params={
+                    "AWS_MAX_CONCURRENT_CONNECTIONS": str(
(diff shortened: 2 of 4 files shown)


gumaerc:
@pt2302 That's not as dramatic as my results, but at least you were able to see some improvement. My average was around 2:30 for `master` and 1:45 with the changes. Either way, this is safe enough to merge and test in RC so we'll see what happens there.

Issue #2001, opened 2023-10-18
Add videos to the online site content bucket


HussainTaj-arbisoft:
Video download links in our online sites do not work for sites recently published with the single site definition. They appear to be missing from the content bucket.

## Expected Behavior
A video file is downloaded.

## Current Behavior
We get a 404.

## Steps to Reproduce
1. Visit https://live-qa.ocw.mit.edu/courses/1-ht-test-site-spring-2023/pages/video/
2. Open the download menu
3. Select "Download video"

PR #1963, opened 2023-09-12
filter out videos during online mass build

gumaerc:
# What are the relevant tickets?
Closes https://github.com/mitodl/ocw-studio/issues/1965

# Description (What does it do?)
This PR adds two new optional arguments to the constructor of `SitePipelineOnlineTasks`, `destructive_sync` with a default of `True` and `filter_videos` with a default of `False`. This maintains the existing functionality of `SitePipelineDefinition` by default for single site pipelines, but allows this functionality to be overridden in the case of the online version of `MassBuildSitesPipelineDefinition`. Not handling videos during the online mass build keeps the build time down. Since adding videos to every single site is not a thing that ever really happens, this works out well. The single site pipeline will still pull in the videos for a site and publish them, so when a new video is added to a course and it is published they will go live as usual. 

# How can this be tested?
 - Follow the instructions in https://github.com/mitodl/ocw-studio/pull/1931 to smoke test the individual site pipeline
 - Follow the instructions in https://github.com/mitodl/ocw-studio/pull/1923 to smoke test the mass build pipeline
 - While testing the mass build, inspect the `static-resources-s3` step in any given site in the online mass build and ensure that the sync is run with `--exclude *.mp4`
 - Also ensure that during the `upload-online-build` step, the sync is *not* run with `--delete`

PR #2003, opened 2023-10-18
don't remove videos during single site online pipeline

gumaerc:
Closes https://github.com/mitodl/ocw-studio/issues/2001

In https://github.com/mitodl/ocw-studio/pull/1963, we added the `filter_videos` and `destructive_sync` optional boolean arguments to `SitePipelineOnlineTasks`. This allows the online mass build to get tasks which filter out mp4 files when pulling in static resources as well as running the final sync without the `--delete` flag. The single site pipelines do not filter out videos by default, because their final sync to S3 is run with `--delete`, as long as they are not the root website. A line at the end of the online site build that removes videos after the build was missed, a hangover from the old pipeline definition that did this. This causes a destructive sync to be run without the videos being present, removing them from the destination S3 bucket. This PR removes that line, so that during single site builds videos are not removed, but the mass build still avoids handling them.

# How can this be tested?
 - Make sure you are set up for publishing courses locally
 - Create a site or pull up an existing site in the `ocw-studio` UI with the minimum metadata needed to publish a site and publish it at least once
 - Log into the Minio console at http://localhost:9001
 - Browse to the `ol-ocw-studio-app` bucket, then under the `courses` folder find the folder with the name matching your site
 - Drop in an example mp4 file into the folder
 - From the `ocw-studio` UI, publish the course again
 - Back in Minio, browse to the draft / live bucket (whichever you published to)
 - Browse to the folder matching your course and ensure that your example video is in the folder

# Additional Context
We are not testing this locally using the traditional video workflow because spoofing YouTube / AWS MediaConvert is difficult and the test above verifies that videos present in the `ocw-studio` bucket are not removed from the publish bucket. The full video flow will be tested in RC.

### DIFF
--- a/content_sync/pipelines/definitions/concourse/site_pipeline.py
+++ b/content_sync/pipelines/definitions/concourse/site_pipeline.py
@@ -485,7 +485,6 @@ class SitePipelineOnlineTasks(list[StepModifierMixin]):
                             cp ../{WEBPACK_MANIFEST_S3_IDENTIFIER}/webpack.json ../{OCW_HUGO_THEMES_GIT_IDENTIFIER}/base-theme/data
                             hugo {pipeline_vars['hugo_args_online']}
                             cp -r -n ../{STATIC_RESOURCES_S3_IDENTIFIER}/. ./output-online{pipeline_vars['static_resources_subdirectory']}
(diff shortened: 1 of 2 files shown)


HussainTaj-arbisoft (approved):
LGTM 👍

Issue #2057, opened 2024-01-04
test site fixture exports contain fully qualified urls in the file field


gumaerc:
## Expected Behavior
All paths to files should be relative in fixture exports

## Current Behavior
The `file` property of exported `WebsiteContent` objects contains a fully qualified URL pointing to the S3 bucket where the file was stored on the exporting system

## Steps to Reproduce
1. Run `./manage.py export_test_sites`
2. Inspect the output at `test_site_fixtures/test_website_content.json`

PR #2058, opened 2024-01-04
fix file path formatting issue in test course data export

gumaerc:
Closes https://github.com/mitodl/ocw-studio/issues/2057

This PR fixes an issue with `ExportWebsiteContentSerializer`. Currently when you run content through it, the `file` property on the objects contains a fully qualified URL to the content in S3. File paths in exports should be relative. The serializer was altered to keep only the relative path to the file, and the checked in fixtures were updated.

# How can this be tested?
 - Run `docker compose exec web ./manage.py export_test_sites`
 - Verify that the `file` paths in `test_website_content.json` do not contain fully qualified URLs. Check the changed files in this PR to see an example of what they used to be in `test_site_fixtures/test_website_content.json`


HussainTaj-arbisoft (approved):
LGTM 👍

Issue #2060, opened 2024-01-04
end to end test pipeline doesn't use proper SITEMAP_DOMAIN when building sites


gumaerc:
## Expected Behavior
The `SITEMAP_DOMAIN` setting used when building sites should be the domain from `STATIC_API_BASE_URL_TEST` so the tests pass

## Current Behavior
The `SITEMAP_DOMAIN` setting is set to the "live" server in a given environment, then the tests fail because the domains do not match what is expected

## Steps to Reproduce
1. Upsert the test pipeline with `./manage.py upsert_e2e_test_pipeline`
2. Run the pipeline and inspect the results

PR #2061, opened 2024-01-04
allow sitemap domain override

gumaerc:
Closes https://github.com/mitodl/ocw-studio/issues/2060

This PR alters `SitePipelineDefinitionConfig`, adding the optional `sitemap_domain` argument. If this is set, this value is used as the `SITEMAP_DOMAIN` env variable when running Hugo builds. Otherwise, the default of `settings.SITEMAP_DOMAIN` is used from `ocw-studio`. `EndToEndTestPipelineDefinition` was then modified to pass the domain of the test site here so that it is properly used in the Hugo builds for the test sites before running Playwright.

# How can this be tested?
 - In your `.env` file, set `SITEMAP_DOMAIN=live-qa.ocw.mit.edu`
 - Run `docker compose exec web ./manage.py upsert_e2e_test_pipeline` to update your test pipeline definition
 - Use the `fly` CLI to inspect the pipeline definition:
   - If you don't have `fly` installed, you will need to do so. Follow the steps in the docs to install `fly` and add a target for your locally running Concourse called `local`
   - Run `fly -t local get-pipeline -p e2e-test-pipeline | grep SITEMAP_DOMAIN`
   - Ensure that `SITEMAP_DOMAIN` is set to `10.1.0.102:8046`
 - Run the pipeline and it should pass

### DIFF
--- a/content_sync/pipelines/definitions/concourse/e2e_test_site_pipeline.py
+++ b/content_sync/pipelines/definitions/concourse/e2e_test_site_pipeline.py
@@ -207,6 +207,7 @@ class EndToEndTestPipelineDefinition(Pipeline):
             resource_base_url=static_api_base_url,
             ocw_hugo_themes_branch=themes_branch,
             ocw_hugo_projects_branch=projects_branch,
+            sitemap_domain=site_pipeline_vars["sitemap_domain"],
             namespace=namespace,
         )
         www_config.is_root_website = True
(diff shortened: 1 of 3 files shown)


HussainTaj-arbisoft (approved):
The changes look good.

The diff still shows changes from https://github.com/mitodl/ocw-studio/pull/2058. Maybe a fresh rebase is needed?

Issue #2065, opened 2024-01-05
mocking ocw-studio api response in e2e test pipeline does not work on real S3


gumaerc:
### Expected Behavior
When running the end to end test pipeline against a site hosted in AWS S3, requests to the `ocw-studio` API should be mocked by static JSON

### Current Behavior
Requests to the API result in a 404

### Steps to Reproduce
1. Run the end to end test pipeline in an environment that is connected to AWS S3 proper rather than Minio

PR #2018, opened 2023-11-06
end to end testing pipeline

gumaerc:
# What are the relevant tickets?
Closes https://github.com/mitodl/ocw-studio/issues/2033

# Description (What does it do?)
This PR adds a new pipeline definition, `TestPipelineDefiniton`. This pipeline is designed for running the end to end Playwright tests defined in `ocw-hugo-themes` against the output of the steps defined in `SitePipelineOnlineTasks`. The resulting pipeline will be automatically triggered on pushes to `ocw-hugo-themes` or `ocw-hugo-projects` based on configured branches and will alert us of any problems running the end to end tests against content that was built and deployed using the same steps we use to build and deploy any other site using pipelines generated by `ocw-studio`.

# How can this be tested?
 - Make sure your instance of `ocw-studio` is set up to publish websites according to the readme
 - Set the following environment variables:
```
OCW_WWW_TEST_SLUG=ocw-ci-test-www
OCW_COURSE_TEST_SLUG=ocw-ci-test-course
AWS_TEST_BUCKET_NAME=ocw-content-test
AWS_OFFLINE_TEST_BUCKET_NAME=ocw-content-offline-test
```
 - Download the JSON fixtures from this Google Drive folder: https://drive.google.com/drive/folders/1iqLw2C2Atk1_O94z7PpUC9X7JQe_IxSn
 - In your local Django admin, make sure you don't have existing sites that would conflict with the test courses. You can do this by going to http://localhost:8043/admin/websites/website/ and looking for `ocw-ci-test-course` and `ocw-ci-test-www`. If these websites already exist, delete them.
 - Place `test_websites.json` and `test_website_content.json` at the root of your `ocw-studio` folder
 - Run `docker compose exec web ./manage.py loaddata test_websites.json` and you should see `Installed 2 object(s) from 1 fixture(s)`
 - Run `docker compose exec web ./manage.py loaddata test_website_content.json` and you should see `Installed 43 object(s) from 1 fixture(s)`
 - Run `docker compose exec web ./manage.py reset_sync_states --filter "ocw-ci-test-course, ocw-ci-test-www" --skip_sync`
 - Run `docker compose exec web ./manage.py sync_website_to_backend --filter "ocw-ci-test-course, ocw-ci-test-www" --delete`
[...]

PR #2066, opened 2024-01-05
fix api mocking in e2e test pipeline

gumaerc:
Closes https://github.com/mitodl/ocw-studio/issues/2065

This PR alters `upload-fixtures-step` in the end to end test pipeline definition to upload the API fixtures as HTML files in the correct place. Since AWS S3 treats every URL path as a string rather than an actual path, a request to something like `https://live-qa.ocw.mit.edu/api/websites/` will result in a 404. In this PR, we upload `__fixtures__/api/websites.json` and `publish__fixtures__/api/publish.json` as `/api/websites/index.html` and `/api/publish/index.html` respectively into the target bucket. This means that web CDN's fronting an S3 bucket that are set up to automatically deliver `index.html` for URL paths that do not include an extension will deliver the JSON content at the expected URLs.

### How can this be tested?
 - If you don't have the end to end testing sites in your local database, follow the steps in this PR to get those: https://github.com/mitodl/ocw-studio/pull/2018
 - Upsert the test pipeline definition with `docker compose exec web ./manage.py upsert_e2e_test_pipeline`
 - Go to http://localhost:8080/, log in and find `e2e-test-pipeline` and run it
 - Ensure that the pipeline passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,6 +363,7 @@ OCW_WWW_TEST_SLUG=ocw-ci-test-www
 OCW_COURSE_TEST_SLUG=ocw-ci-test-course
 AWS_TEST_BUCKET_NAME=ocw-content-test
 AWS_OFFLINE_TEST_BUCKET_NAME=ocw-content-offline-test
+STATIC_API_BASE_URL_TEST=http://10.1.0.102:8046
 ```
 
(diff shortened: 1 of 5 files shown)


ChristopherChudzicki (approved):
After we added `corepack enable`, this ran successfully on my M1 on OS Sonoma 14.2.1 (23C71)

README.md
```diff
@@ -363,6 +363,7 @@ OCW_WWW_TEST_SLUG=ocw-ci-test-www
 OCW_COURSE_TEST_SLUG=ocw-ci-test-course
 AWS_TEST_BUCKET_NAME=ocw-content-test
 AWS_OFFLINE_TEST_BUCKET_NAME=ocw-content-offline-test
+STATIC_API_BASE_URL_TEST=http://10.1.0.102:8046
```
ChristopherChudzicki, line 366: You could put this in `.env.example` (or whatever it is called)

PR #2071, opened 2024-01-16
Multiple open catalog webhook endpoints

### LINKED ISSUE #1994: Send webhook requests to mit-open instead of/in addition to open-discussions (RC only for now)
# Description/Context
Now that the OCW ETL code for mit-open is on RC, it would be nice for the RC instance of ocw-studio to send webhook requests there instead of or in addition to open-discussions RC.

Eventually, when mit-open is live with a production instance, the same should be done on prod ocw-studio.

# Plan/Design
If switching from open-discussions to mit-open, just update the `OPEN_DISCUSSIONS_URL` setting and then update all the concourse pipelines.

If we decide to send webhook requests to both, the concourse pipeline generation code will need to be updated to send the requests to multiple endpoints instead of just one.

The variable `OPEN_DISCUSSIONS_URL` should also be renamed, maybe something like `OPEN_WEBHOOK_URL` (make plural if we decide to support multiple endpoints).

mbertrand:
Closes #1994

Related salt-ops PR: https://github.com/mitodl/salt-ops/pull/1609

Modifies the pipeline generation code to allow for sending webhook requests to multiple urls.
Renames the following env vars:
  - `OCW_NEXT_SEARCH_WEBHOOK_KEY` to `OPEN_CATALOG_WEBHOOK_KEY`
  -  `OPEN_DISCUSSIONS_URL` to `OPEN_CATALOG_URLS`

### How can this be tested?
- Change the value of `OCW_STUDIO_ENVIRONMENT` to anything except `dev`
- Add this to your .env: 
  ```
  OPEN_CATALOG_URLS=http://discussions-rc.odl.mit.edu,https://mitopen-rc.odl.mit.edu
  OPEN_CATALOG_WEBHOOK_KEY=<former value of OCW_NEXT_SEARCH_WEBHOOK_KEY>
  ```
- Run the following commands:
  ```
  ./manage.py upsert_theme_assets_pipeline --delete
  ./manage.py upsert_mass_build_pipeline --delete
  ./manage.py upsert_site_removal_pipeline --delete
  ./manage.py backpopulate_pipelines --filter <site-name>
  ```
- Using fly, save the pipeline configurations locally and make sure there are sections present to send webhooks to both open catalog urls:
  ```
   fly -t local get-pipeline -p 'mass-build-sites/offline:false,prefix:"",projects_branch:main,starter:"",themes_branch:main,version:live ' > mass_publish.yml
   fly -t local get-pipeline -p remove-unpublished-sites > remove_unpublished.yml
   fly -t local get-pipeline -p live/site:site_name > single_site.yml
  ```

### Additional Context
This assumes that the webhook key is always the same for all endpoints.  In salt-ops they all use the [same source](https://github.com/mitodl/salt-ops/blob/main/pillar/heroku/mitopen.sls#L127).

When deployed to RC/production, all the existing pipelines will need to be updated

### DIFF
--- a/content_sync/pipelines/definitions/concourse/theme_assets_pipeline.py
+++ b/content_sync/pipelines/definitions/concourse/theme_assets_pipeline.py
@@ -22,7 +22,6 @@ from content_sync.pipelines.definitions.concourse.common.image_resources import 
 )
 from content_sync.pipelines.definitions.concourse.common.resources import (
     GitResource,
-    OpenCatalogResource,
     SlackAlertResource,
 )
 from content_sync.pipelines.definitions.concourse.common.steps import (
@@ -60,10 +59,6 @@ class ThemeAssetsPipelineDefinition(Pipeline):
         "clear-draft-cdn-cache-task"
     ).root
     _clear_live_cdn_cache_identifier = Identifier("clear-live-cdn-cache-task").root
-
-    _open_catalog_resources = [
-        OpenCatalogResource(catalog_url) for catalog_url in settings.OPEN_CATALOG_URLS
-    ]
     _slack_resource = SlackAlertResource()
 
     def __init__(  # noqa: PLR0913
--- a/content_sync/pipelines/definitions/concourse/theme_assets_pipeline.py
+++ b/content_sync/pipelines/definitions/concourse/theme_assets_pipeline.py
@@ -22,7 +22,7 @@ from content_sync.pipelines.definitions.concourse.common.image_resources import 
 )
 from content_sync.pipelines.definitions.concourse.common.resources import (
     GitResource,
-    OpenDiscussionsResource,
+    OpenCatalogResource,
     SlackAlertResource,
 )
 from content_sync.pipelines.definitions.concourse.common.steps import (
@@ -61,7 +61,9 @@ class ThemeAssetsPipelineDefinition(Pipeline):
     ).root
     _clear_live_cdn_cache_identifier = Identifier("clear-live-cdn-cache-task").root
 
-    _open_discussions_resource = OpenDiscussionsResource()
+    _open_catalog_resources = [
+        OpenCatalogResource(catalog_url) for catalog_url in settings.OPEN_CATALOG_URLS
+    ]
     _slack_resource = SlackAlertResource()
 
     def __init__(  # noqa: PLR0913
--- a/main/settings.py
+++ b/main/settings.py
@@ -499,8 +499,8 @@ DRIVE_UPLOADS_PARENT_FOLDER_ID = get_string(
     description="Gdrive folder for video uploads",
     required=False,
 )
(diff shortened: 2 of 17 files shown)


content_sync/pipelines/definitions/concourse/theme_assets_pipeline.py
```diff
@@ -61,7 +61,9 @@ class ThemeAssetsPipelineDefinition(Pipeline):
     ).root
     _clear_live_cdn_cache_identifier = Identifier("clear-live-cdn-cache-task").root
 
-    _open_discussions_resource = OpenDiscussionsResource()
+    _open_catalog_resources = [
+        OpenCatalogResource(catalog_url) for catalog_url in settings.OPEN_CATALOG_URLS
+    ]
```
mbertrand, line 66: @gumaerc can this variable be removed? It doesn't seem to be used anywhere in the pipeline.

gumaerc (changes requested):
This looks good aside from a couple unnecessary resource definitions, one of which you already noted. Another thing I wanted to mention is that when I test non-dev things like this, instead of changing my environment in my `.env` file, I usually go in and temporarily remove the if statements limiting the steps in question to dev only. This means we can test pushing the pipeline up to our local Concourse instance and run it, and the pipeline will mostly work as expected. This doesn't work for things like testing Fastly cache clearing (although it might with the proper vars in place) but for something like this Open Discussions / MIT Open will accept the webhook request as long as you have the proper key. I tested all of the pipelines in question here that way and all looks good. I did get a 503 when it went to hit MIT Open but that looks to be because as of the time of writing this, the database is down or improperly configured. I will test again once it's back up.

content_sync/pipelines/definitions/concourse/theme_assets_pipeline.py
mbertrand: Removed

gumaerc:
Testing this again, the remove-unpublished-sites pipeline fails trying to hit https://mit-open-rc.odl.mit.edu/api/v0/ocw_next_webhook/ with a 404

main/settings.py
```diff
@@ -499,16 +499,16 @@
...
-    name="OCW_NEXT_SEARCH_WEBHOOK_KEY",
+OPEN_CATALOG_WEBHOOK_KEY = get_string(
+    name="OPEN_CATALOG_WEBHOOK_KEY",
     default="",
     description="Open discussions webhook key",
     required=False,
 )
-OPEN_DISCUSSIONS_URL = get_string(
-    name="OPEN_DISCUSSIONS_URL",
+OPEN_CATALOG_URLS = get_delimited_list(
```
gumaerc, line 508: Can we add example values for these new settings in `.env.example`, and document their usage in `README.md`?

gumaerc (approved):
:+1: 

Looks good to me now, although wait to merge this until the new env settings are fully deployed all the way to production.

Issue #2076, opened 2024-01-18
end to end test pipeline needs cache clearing step


gumaerc:
### Description/Context
The newly added end to end testing pipeline (`content_sync/pipelines/definitions/concourse/e2e_test_site_pipeline.py`) does not currently perform cache clearing outside of dev. In order to assure that tests are run against the latest built content, cache should be cleared during the builds of the test sites.

### Plan/Design
 - In `EndToEndTestPipelineDefinition`, when initializing `SitePipelineOnlineTasks`, the `skip_cache_clear` argument should not be passed if the current environment is not `dev` (it defaults to `False`)

PR #2078, opened 2024-01-18
e2e test pipeline cache clearing

gumaerc:
Closes https://github.com/mitodl/ocw-studio/issues/2076

This PR adds a cache clearing step to the end to end testing pipeline that is added outside of dev. An optional argument was also added to `SitePipelineOnlineTasks`: `skip_search_index_update`. This boolean defaults to false. The end to end test pipeline will set it to true, which will prevent the test pipeline from calling the catalog sites to update the search index, as the test sites should definitely not show up in the search index.

### How can this be tested?
 - On line 587 of `content_sync/pipelines/definitions/site_pipeline.py`, you will see:
```python
        if not is_dev() and not skip_cache_clear:
            self.append(clear_cdn_cache_online_step)
```
 - Change this to:
```python
        # if not is_dev() and not skip_cache_clear:
        self.append(clear_cdn_cache_online_step)
```
 - This will ensure that even though we are running the site in local dev mode, the cache clearing step will be inserted into the pipeline definition
 - Spin up `ocw-studio` and make sure that you are all set up for publishing sites
 - Run `docker compose exec web ./manage.py upsert_e2e_test_pipeline`
 - Browse to http://localhost:8080 and find the test pipeline, then trigger a build
 - Eventually the pipeline should error out
 - In the site across step, you should see a `clear-cdn-cache` step with the error `failed to interpolate task config: undefined vars: fastly_test.api_token, fastly_test.service_id`
 - Further testing will need to be done on RC, but this indicates that the proper values were inserted into the pipeline definition and it was pushed up to Concourse without error
 - Revert the changes we made above and repeat this process and the end to end pipeline should skip the cache clear and run normally

### DIFF
--- a/content_sync/pipelines/definitions/concourse/e2e_test_site_pipeline.py
+++ b/content_sync/pipelines/definitions/concourse/e2e_test_site_pipeline.py
@@ -37,7 +37,6 @@ from content_sync.pipelines.definitions.concourse.common.resources import (
     OcwHugoProjectsGitResource,
     OcwHugoThemesGitResource,
     OcwStudioWebhookResource,
-    OpenCatalogResource,
     SlackAlertResource,
     WebpackManifestResource,
 )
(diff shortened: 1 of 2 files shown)


gumaerc:
@pt2302 This should be all set now

Issue #2082, opened 2024-01-29
Can't use import_website_starters with non-main branch


pt2302:
### Expected Behavior
Setting the `GITHUB_WEBHOOK_BRANCH=<some non-main branch>` variable in the `.env` file and running `docker compose exec web ./manage.py import_website_starters https://github.com/mitodl/ocw-hugo-projects` should pull in the starters from that branch.

### Current Behavior
The `GITHUB_WEBHOOK_BRANCH=<some non-main branch>` variable in the `.env` file and running `docker compose exec web ./manage.py import_website_starters https://github.com/mitodl/ocw-hugo-projects` also requires explicitly passing a commit hash in order to work.

### Steps to Reproduce
1. Set `GITHUB_WEBHOOK_BRANCH=<some non-main branch>` variable in the `.env` file.
2. Run `docker compose exec web ./manage.py import_website_starters https://github.com/mitodl/ocw-hugo-projects` and observe that the starters that get pulled in are from the `main` branch of `ocw-hugo-projects`, rather than from the branch specified in the `.env` file.

PR #2083, opened 2024-01-30
Correctly set branch when commit hash is not given

pt2302:
Closes https://github.com/mitodl/ocw-studio/issues/2082.

This PR adds a reference to the branch set in the `.env` variable `GITHUB_WEBHOOK_BRANCH` when getting the contents; previously, this was not being set properly.

### How can this be tested?
1. Set `GITHUB_WEBHOOK_BRANCH=<some non-main branch>` variable in the `.env` file.
2. Run `docker compose exec web ./manage.py import_website_starters https://github.com/mitodl/ocw-hugo-projects` and observe that the starters that get pulled in are from the `<some non-main branch>` branch of `ocw-hugo-projects`, rather than from `main`.


gumaerc (approved):
:+1:

PR #2087, opened 2024-02-06
set allow_unicode to True where we use slugify

gumaerc:

- [ ] Desktop screenshots

### DIFF
--- a/websites/models.py
+++ b/websites/models.py
@@ -58,7 +58,7 @@ def validate_yaml(value):
 
 def validate_slug(value):
     """Validator function to ensure that the value is a properly-formatted slug"""  # noqa: D401, E501
-    slugified = slugify(value)
+    slugified = slugify(value, allow_unicode=True)
     if slugified != value:
         msg = f"Value '{value}' is not a proper slug (slugified version: {slugified})"
         raise ValidationError(msg)
@@ -256,11 +256,11 @@ class Website(TimestampedModel):
                     content = self.websitecontent_set.filter(type=section_type).first()
                     if content:
                         value = get_dict_field(content.metadata, section_field)
-                if not value:  # noqa: SIM108
+                if not value:
                     # Incomplete metadata required for url
                     value = section
                 else:
-                    value = slugify(value.replace(".", "-"))
+                    value = slugify(value.replace(".", "-"), allow_unicode=True)
                 url_format = url_format.replace(section, value)
         return url_format
 
--- a/websites/views.py
+++ b/websites/views.py
@@ -190,7 +190,7 @@ class WebsiteViewSet(
             and self.request.data.get("title")
         ):
             request_data = self.request.data.copy()
-            request_data["name"] = slugify(request_data["title"])
+            request_data["name"] = slugify(request_data["title"], allow_unicode=True)
             kwargs["data"] = request_data
 
         return serializer_class(*args, **kwargs)
@@ -539,7 +539,9 @@ def _get_derived_website_content_data(
         added_data["filename"] = get_valid_new_filename(
             website_pk=website_pk,
             dirpath=dirpath,
-            filename_base=slugify(get_valid_base_filename(slug, content_type)),
+            filename_base=slugify(
+                get_valid_base_filename(slug, content_type), allow_unicode=True
+            ),
         )
     return added_data
 
--- a/websites/views_test.py
+++ b/websites/views_test.py
@@ -577,7 +577,7 @@ def test_websites_autogenerate_name(drf_client):
     )
     website_title = "My Title"
     website_short_id = "my-title"
-    slugified_title = slugify(website_title)
+    slugified_title = slugify(website_title, allow_unicode=True)
     resp = drf_client.post(
         reverse("websites_api-list"),
         {"title": website_title, "short_id": website_short_id, "starter": starter.id},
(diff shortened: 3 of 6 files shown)


pdpinch:
What slug/URL do we get when the title is set to テスト ページ ?

websites/models.py
```diff
@@ -256,11 +256,11 @@ def url_path_from_metadata(self, metadata: dict | None = None):
                     content = self.websitecontent_set.filter(type=section_type).first()
                     if content:
                         value = get_dict_field(content.metadata, section_field)
-                if not value:  # noqa: SIM108
+                if not value:
                     # Incomplete metadata required for url
                     value = section
                 else:
-                    value = slugify(value.replace(".", "-"))
+                    value = slugify(value.replace(".", "-"), allow_unicode=True)
```
HussainTaj-arbisoft, line 263: The Studio UI doesn't allow Unicode characters in the url_path. I wonder if we should keep `allow_unicode=False` here since Unicode characters are invalid for this field.

Screenshot 2024-02-07 at 5 13 45 PM

websites/models_test.py
```diff
@@ -283,6 +283,28 @@ def test_website_url_path_from_metadata(missing_keys):
     assert website.url_path_from_metadata(metadata) == expected_url_path
 
 
+def test_website_url_path_from_metadata_unicode():
+    """None should be returned for a site without a starter"""
```
HussainTaj-arbisoft, line 287:
```suggestion
    """url_path_from_metadata should correctly generate a url/format for unicode input."""
```

websites/models_test.py
```diff
@@ -283,6 +283,28 @@ def test_website_url_path_from_metadata(missing_keys):
...
+    starter = WebsiteStarterFactory.create(
+        config={
+            "site-url-format": "[meta:course_nr]-[meta:title]-[meta:term]-[meta:year]"
+        }
+    )
+    website = WebsiteFactory.create(starter=starter, not_published=True)
+    course_nr = "1-1" if metadata.get("course_nr") is not None else "[meta:course_nr]"
+    title = "テスト-ウェブサイト" if metadata.get("title") is not None else "[meta:title]"
+    term = "fall" if metadata.get("term") is not None else "[meta:term]"
+    year = metadata.get("year", "[meta:year]")
```
HussainTaj-arbisoft, line 303: Unless you meant to parameterize `metadata`, this can be simplified.

```suggestion
    course_nr = "1-1"
    title = "テスト-ウェブサイト"
    term = "fall"
    year = "2025"
```

websites/views.py
```diff
@@ -190,7 +190,7 @@ def get_serializer(self, *args, **kwargs):
             and self.request.data.get("title")
         ):
             request_data = self.request.data.copy()
-            request_data["name"] = slugify(request_data["title"])
+            request_data["name"] = slugify(request_data["title"], allow_unicode=True)
```
HussainTaj-arbisoft, line 193: Unicode `name`s don't seem to work for me in Studio. Seems like a frontend issue, because I see no request being made to fetch data from the backend.

Screenshot 2024-02-07 at 5 26 49 PM

websites/views_test.py
```diff
@@ -577,7 +577,27 @@ def test_websites_autogenerate_name(drf_client):
...
+def test_websites_autogenerate_name_unicode(drf_client):
+    """The autogenerated name should be compatible with unicode characters"""
+    superuser = UserFactory.create(is_superuser=True)
+    drf_client.force_login(superuser)
+    starter = WebsiteStarterFactory.create(
+        source=constants.STARTER_SOURCE_GITHUB,
+        status=constants.WebsiteStarterStatus.ACTIVE,
+    )
+    website_title = "テスト ページ"
+    website_short_id = "テスト_ページ"
```
HussainTaj-arbisoft, line 599: It seems like GitHub doesn't allow unicode and our Studio UI forbits it as well for `short-id`. We should probably not test this here, and maybe add a test to ensure short-id is not unicode compatible.

Screenshot 2024-02-07 at 5 31 12 PM

HussainTaj-arbisoft (commented):
The changes seem good. It solves the `filename` issue we had. I do have a few suggestions and comments. 

I think we can merge this change; however, supporting Unicode characters throughout Studio might require more thought.

---

@gumaerc I was unable to test the following functions locally with unicode names. We'll probably need to test them on RC:

- Video/caption workflow (a video with a special name). To test YouTube, MediaConvert, and 3Play integrations with unicode names.
- Search Index

---

This trivial, not urgent, issue with the `file` field showing encoded names.

Screenshot 2024-02-07 at 4 46 42 PM

HussainTaj-arbisoft:
Perhaps we should make only the `filename`s Unicode compatible and leave all other fields incompatible for now.

gumaerc:
> What slug/URL do we get when the title is set to テスト ページ ?

We would get a slug of `テスト-ページ`, which modern browsers do support but older operating systems / browsers may have issues with this. @HussainTaj-arbisoft has highlighted a number of areas where Unicode characters will be problematic, and I'm going to step back the scope of this PR to take that into account, although I didn't get a chance to address this today.

gumaerc:
@HussainTaj-arbisoft This is ready for another look

HussainTaj-arbisoft (approved):
Looks good 👍 

A reminder that I'm unable to test the video workflow locally, so we should probably do that on RC. I don't expect any issues with it; it's just to be safe.

Issue #2088, opened 2024-02-07
Google Drive Backfill Misses Some Files in S3


pt2302:
### Expected Behavior
Running `./manage.py backfill_gdrive_folder --filter <course name>` for a legacy course should create Google Drive files for the all the non-video resources in the course.

### Current Behavior
The backfill fails, with the error message `botocore.exceptions.ClientError: An error occurred (404) when calling the HeadObject operation: Not Found`.

### Steps to Reproduce
Locally, this can be tested by setting the `ENVIRONMENT` variable to anything but `dev`, entering appropriate credentials for AWS (instead of MinIO), and then running the `backfill_gdrive_folder` management command for a legacy course.

### Possible Solution
This is caused by a difference in the way that MinIO and AWS S3 handle paths. A related prior discrepancy for caption file paths was resolved here: https://github.com/mitodl/ocw-studio/pull/1797.

PR #2089, opened 2024-02-07
Fix S3 path for Google Drive backfill

pt2302:
Closes https://github.com/mitodl/ocw-studio/issues/2088.

This PR updates the call to download a file from S3 to strip a leading forward slash, if it exists. While MinIO properly handles a path with an extra leading forward slash, S3 does not.

### How can this be tested?
Note: Testing is similar to testing for https://github.com/mitodl/ocw-studio/pull/2081, except that the AWS storage is used instead of MinIO. This can be done by setting the following variables in the OCW Studio `.env` file:
```
ENVIRONMENT=<anything except `dev`>
AWS_ACCESS_KEY_ID=<use value from RC>
AWS_SECRET_ACCESS_KEY=<use value from RC>
AWS_STORAGE_BUCKET_NAME=<use value from RC>
```
1. Pick a legacy course that has no Google Drive content but does contain non-video resources for testing. An example is `24-915-linguistic-phonetics-fall-2015`. 
2. Spin up OCW Studio with `docker compose up`. 
3. Run `docker compose exec web ./manage.py backfill_gdrive_folder --filter <course name or short-id>`. 
4.  Check the Google Drive folder for the website to ensure that the resources have been uploaded correctly. Also, check the Django admin to ensure that DriveFile objects have been created for each resource.

### DIFF
--- a/gdrive_sync/management/commands/backfill_gdrive_folder.py
+++ b/gdrive_sync/management/commands/backfill_gdrive_folder.py
@@ -155,8 +155,11 @@ class Command(WebsiteFilterCommand):
         )
         try:
             self.s3.download_fileobj(
-                settings.AWS_STORAGE_BUCKET_NAME, str(resource.file), file_obj
+                settings.AWS_STORAGE_BUCKET_NAME,
+                f"{resource.file}".lstrip("/"),
+                file_obj,
             )
+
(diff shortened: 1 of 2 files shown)


gumaerc (approved):
👍 LGTM

Issue #2095, opened 2024-02-09
add COURSE_SEARCH_API_URL and CONTENT_FILE_SEARCH_API_URL


gumaerc:
### Description/Context
In order to support the work being done in https://github.com/mitodl/ocw-hugo-themes/pull/1313, two new settings / environment variables need to be added: `COURSE_SEARCH_API_URL` and `CONTENT_FILE_SEARCH_API_URL`. These values need to be set in `ThemeAssetsPipelineDefinition` so that the new search API URL's are baked into the Javascript bundles generated for the site.

### Plan/Design
 - [x] Add `COURSE_SEARCH_API_URL` as a Django setting pulled from the environment with a sane default
 - [x] Add `CONTENT_FILE_SEARCH_API_URL` as a Django setting pulled from the environment with a sane default
 - [x] In `ThemeAssetsPipelineDefinition`, inject the values of these settings into the `params` section of `build-ocw-hugo-themes-task`

PR #2109, opened 2024-02-13
new params

abeglova:
closes https://github.com/mitodl/ocw-studio/issues/2095

This pr adds two new environmental variables to the theme to support using the new search apis

### How can this be tested?
I'm not sure if this can be tested locally. I think just the tests passing is good enough

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -35,6 +35,8 @@ CONTENT_SYNC_THEME_PIPELINE=content_sync.pipelines.concourse.ThemeAssetsPipeline
 # Values used in pipeline definitions
 OCW_STUDIO_BASE_URL=http://MYNAME.ocw-studio.odl.local:8043/
 SEARCH_API_URL=https://discussions-rc.odl.mit.edu/api/v0/search/
+COURSE_SEARCH_API_URL=https://mit-open-rc.odl.mit.edu/api/v1/learning_resources_search/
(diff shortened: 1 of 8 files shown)


gumaerc (changes requested):
All looks good, but we don't need to modify the YML pipeline definitions as they aren't used anymore:

.env.example
```diff
@@ -35,6 +35,8 @@ CONTENT_SYNC_THEME_PIPELINE=content_sync.pipelines.concourse.ThemeAssetsPipeline
 # Values used in pipeline definitions
 OCW_STUDIO_BASE_URL=http://MYNAME.ocw-studio.odl.local:8043/
 SEARCH_API_URL=https://discussions-rc.odl.mit.edu/api/v0/search/
+COURSE_SEARCH_API_URL=https://mit-open-rc.odl.mit.edu//api/v1/learning_resources_search/
```
gumaerc, line 38: This should be `COURSE_SEARCH_API_URL=https://mit-open-rc.odl.mit.edu/api/v1/learning_resources_search/` without the double slash

gumaerc (changes requested):
Just noticed a small typo:

.env.example
abeglova: oops

gumaerc (approved):
👍

Issue #2111, opened 2024-02-16
Create management command to copy videos from one course to another


pt2302:
### Description/Context
As indicated in https://github.com/mitodl/hq/issues/3556, sometimes videos that have already been uploaded to YouTube need to be copied over from one course to another. This should bypass the YouTube submission and transcoding processes, and just link the newly-created video resources to the existing YouTube videos. It should also copy over the transcript/captions, if any, to the new course.

### Plan/Design
The arguments to the management command should be `source_course`, `destination_course`, and (optionally) a list of video resource UUIDS. If provided, only those videos are copied over; otherwise all videos are copied over. `Video` objects and `DriveFile` objects, if available for the source course, would also need to be copied over.

Issue #2117, opened 2024-02-26
resources published locally include storage bucket in URL


gumaerc:
### Expected Behavior
The `file` property in the front matter of markdown files published by `ocw-studio` locally are relative paths pointing to the file from the root of any given host

### Current Behavior
The `file` property is prefixed with the value of `AWS_STORAGE_BUCKET_NAME`, i.e. `/ol-ocw-studio-app/courses/ocw-ci-test-course/example_pdf.pdf`

### Steps to Reproduce
1. Import or create a course in `ocw-studio` using the local Docker setup
2. Publish said course
3. Browse to the Github repo for the course and find a resource

### Possible Solution
When using the setting `OCW_STUDIO_USE_S3=true`, the `django-storages` backend automatically generates the `url` property of `FileField`s using the configured AWS settings. In local development, AWS settings are overridden to use a locally running instance of Minio. An automatically generated URL to a resource in AWS proper looks something like: `'https://ol-ocw-studio-app-qa.s3.amazonaws.com/courses/6-450-principles-of-digital-communications-i-fall-2006/c59828b4e470fd50daf31f0f7251456c_quiz1_04_soln.pdf'`. Locally, you will see URLs like `'http://10.1.0.100:9000/ol-ocw-studio-app/courses/ocw-ci-test-course/example_pdf.pdf'`. This happens because `endpoint_url` is overridden to hit Minio, and a different URL strategy is used.

In the `full_metadata` property of the `WebsiteContent` model, the file URL is set using `full_metadata[file_field["name"]] = urlparse(file_url).path`. If you look at the above URLs, you will see why the bucket name is being included locally. Since the bucket name is part of the path and not the domain, it gets included when you call the `path` property of the parsed URL. The bucket name should be filtered out of the path if it is being included.

gumaerc:
Closing this issue, as it was found to be an issue with my local configuration as noted in https://github.com/mitodl/ocw-studio/pull/2124

PR #2120, opened 2024-02-29
Copy videos from one course to another

pt2302:
Closes https://github.com/mitodl/ocw-studio/issues/2111.

This PR adds a management command to copy videos and associated transcripts from one course to another. The command can be run with an optional `video_uuids` argument to copy only specific videos; otherwise all videos from the course are copied.

While this command will copy Google Drive files that already exist from one course's folder to another course's folder, this PR does not currently create Google Drive files if those are missing; that can be handled separately by running the management command added here: https://github.com/mitodl/ocw-studio/pull/2081. 

This PR also includes some refactoring to move functions that are used in multiple management command to utility libraries.

### How can this be tested?
Pick a course that has videos and copy them to another course, by running 
```
docker compose exec web ./manage.py copy_videos --source_course <source_course> --destination_course <destination course> --video_uuids <optional; comma-separated list of video-uuids>
```

Publish the course and verify that the videos, transcripts, and captions all look as expected. Also verify that the paths for the files are now pointing to the destination course.

### DIFF
--- a/videos/utils.py
+++ b/videos/utils.py
@@ -69,7 +69,7 @@ def create_new_content(source_obj, to_course):
     new_text_id = uuid_string()
     if source_obj.file:
         new_s3_loc = copy_obj_s3(source_obj, to_course)
-        new_dirpath = get_dirpath_and_filename(new_s3_loc)[0]
+        new_dirpath = "content/resources"
         new_filename = get_dirpath_and_filename(new_s3_loc)[1]
     else:
         new_s3_loc = source_obj.file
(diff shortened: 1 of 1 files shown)


videos/utils.py
```diff
@@ -46,3 +50,63 @@ def get_content_dirpath(slug, collection_type):
...
+
+
+def create_new_content(source_obj, to_course):
+    """Create new WebsiteContent object from source_obj in to_course,
+    or update existing object if it exists.
+    """
+    new_text_id = uuid_string()
+    if source_obj.file:
+        new_s3_loc = copy_obj_s3(source_obj, to_course)
+        new_dirpath = get_dirpath_and_filename(new_s3_loc)[0]
```
gumaerc, line 72: There seems to be an issue here with the `dirpath` attribute. Take a look at my test course that I've been working with:

https://github.com/gumaerc-testorg/video-copy-test-site/tree/main

As you can see, there is a `courses/video-copy-test-site` folder in the Github repo that shouldn't be there. I tried deleting my website and recreating it, and the problem persisted. The `dirpath` property controls where the markdown representing a piece of content is stored. In this case, these files should be under `content/resources`.

gumaerc (changes requested):
The idempotency and filtering issues are fixed, but I ran into a new problem:

videos/utils.py
gumaerc: As a separate note, if you look at the content in that folder you'll see `file` paths like `/ol-ocw-studio-app/courses/video-copy-test-site/-EgKluVR2Ug_captions.vtt`. This may be related to https://github.com/mitodl/ocw-studio/issues/2117 but we should check and make sure the correct `file` attribute is being written here.

gumaerc (approved):
👍 LGTM!

Issue #2113, opened 2024-02-16
The mass build sites pipeline's prefix feature doesn't properly set Hugo baseURL


gumaerc:
### Expected Behavior
If I specify `--prefix` to `upsert_mass_build_pipeline`, that prefix should be applied to the `baseURL` setting 

### Current Behavior
The prefix is not included in the `baseURL` setting, and links are all relative to the root of the site using `site.get_url_path()`

### Steps to Reproduce
1. Upsert a mass build pipeline using the `upsert_mass_build_pipeline` management command and specify `--prefix` with a custom prefix
2. Run the pipeline and let it deploy a few sites
3. Once one of the sites has deployed, browse to it and try clicking any of the internal links

### Possible Solution
Take the prefix into account when setting `baseURL` in Hugo

PR #2121, opened 2024-02-29
fix links in prefixed sites built with the mass-build-sites pipeline

gumaerc:
Closes https://github.com/mitodl/ocw-studio/issues/2113

This PR takes the `prefix` argument being sent to `SitePipelineDefinition` and in addition to prefixing the S3 path with it, the Hugo `baseURL` argument is now also prefixed. This means that links rendered inside of sites deployed at a prefix URL will correctly link you to the prefixed page, at least for content generated by Hugo.

### How can this be tested?
 - Make sure you have run through the basic requirements for setting up `ocw-studio` and can publish sites locally
 - Run `docker compose exec web ./manage.py upsert_mass_build_pipeline --prefix test-prefix`
 - Browse to http://localhost:8080 and log into the Concourse web UI
 - Find the prefixed mass build pipeline, unpause it and trigger the first batch. You can pause it after, as you only need to run through one batch to test this.
 - In the batch's `across` step, find of of the sites it published and copy the `url_path`. Browse to the site by appending the `url_path` and the prefix to http://localhost:8045/. It should be something like `http://localhost:8045/test-prefix/courses/site-name`
 - Click the links in the sidebar to browse the different parts of the site and verify that you stay under the `test-prefix`

### DIFF
--- a/content_sync/pipelines/definitions/concourse/site_pipeline.py
+++ b/content_sync/pipelines/definitions/concourse/site_pipeline.py
@@ -161,6 +161,7 @@ class SitePipelineDefinitionConfig:
         self.offline_bucket = offline_bucket
         self.static_api_url = static_api_url
         self.sitemap_domain = sitemap_domain
+        self.prefix = f"{prefix.strip('/')}/" if prefix != "" else prefix
         self.url_path = site.get_url_path()
         self.resource_base_url = resource_base_url
         self.ocw_studio_url = get_ocw_studio_api_url()
@@ -198,7 +199,7 @@ class SitePipelineDefinitionConfig:
         base_online_args.update(
             {
                 "--config": f"../{OCW_HUGO_PROJECTS_GIT_IDENTIFIER}/{starter_slug}/config.yaml",  # noqa: E501
-                "--baseURL": f"/{self.base_url}",
+                "--baseURL": f"/{self.prefix}{self.base_url}",
                 "--destination": "output-online",
             }
         )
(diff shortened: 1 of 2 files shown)


ibrahimjaved12 (commented):
Tested on both `localhost:8044`(draft) and `localhost:8045`(live). Tried multiple links there on few courses, works as expected.

However I noticed that it does not:
- Work for `ocw-www` (http://localhost:8045/ or http://localhost:8044/)
  - Course links directing to `ocw-www` eg. When you're on a course's homepage, and you click to its relevant department. It redirects to the search page with department as query, but the URL is without the given prefix.

I think prefixing  URLs on `ocw-www` would also be good to have?

gumaerc:
@ibrahimjaved12 Thanks for your review. I looked into home page links, and I believe it's a bit beyond the scope of what this PR is doing. This PR simply sets the Hugo `--baseUrl` parameter to include the prefix. This means that links generated by Hugo (`.Permalink`, `.RelPermalink`, `pageRef`) will contain the prefix and work as expected. However, The links to the search page are mostly hard coded in `ocw-hugo-themes` to be relative to the root of the site. I think that in order to support this, we would need a PR in `ocw-hugo-themes` to be able to pull the prefix from an env variable and set it where necessary. This especially applies to the links within the search page itself, as the search page is a separate React app not generated by Hugo.

ibrahimjaved12 (approved):
Looks good to me then 👍

gumaerc:
I created an issue here for adding support to the `www` theme for prefixed deployments: https://github.com/mitodl/ocw-hugo-themes/issues/1365

Issue #2126, opened 2024-03-07
Add units tests for video copying management command


pt2302:
### Description/Context
In https://github.com/mitodl/ocw-studio/pull/2120, a management command for copying videos from one course to another was introduced. However, this PR did not include unit tests for the command. Since this command is quite complex, there should be unit tests added for testing the relevant functionality.

### Plan/Design
The unit tests should verify that everything works in the various possible situations in which the management command could be called, including for courses that have Google Drive files associated with the videos and captions/transcripts.

PR #2131, opened 2024-03-13
Fix Google Drive copy and adding test

pt2302:
Closes https://github.com/mitodl/ocw-studio/issues/2126.

This PR fixes the Google Drive copy functionality of the `copy_videos` management command to properly copy Google Drives files as well, if present. It also adds a test that the copied resource is created with the correct attributes.

### How can this be tested?
Testing is similar to https://github.com/mitodl/ocw-studio/pull/2120, but first make sure that (at least) the transcripts/captions in the course have associated files in Google Drive. This would be either a newer course authored through OCW Studio or a legacy course with Google Drive backfilled. To backfill Google Drive for a course, first run through the testing instructions for https://github.com/mitodl/ocw-studio/pull/2081 and then run `docker compose exec web ./manage.py backfill_gdrive_folder --filter <course name or short-id>`.  Note that Google Drive backfill only backfills non-video resources (in this case, captions and transcripts).

At this point, run the management command `docker compose exec web ./manage.py copy_videos --source_course <source_course> --destination_course <destination course> --video_uuids <optional; comma-separated list of video-uuids>`, and now verify that any of the Google Drive files have also been copied from the source course to the destination course.

### DIFF
--- a/videos/tasks_test.py
+++ b/videos/tasks_test.py
@@ -898,7 +898,7 @@ def test_create_drivefile(mocker):
     mock_update_or_create = mocker.patch.object(DriveFile.objects, "update_or_create")
 
     create_drivefile(
-        mock_gdrive_file, new_resource, destination_course, files_or_videos
+        mock_gdrive_file.file_id, new_resource, destination_course, files_or_videos
     )
 
     mock_get_drive_service.assert_called_once()


gumaerc:
@pt2302 I was able to run this locally and confirm that the transcript files were copied across to the proper gdrive folder. Reading this over again though, I find myself wondering if the code in `move_videos.py` should be moved out into a Celery task. Copying videos around could take a decent amount of time depending on the amount of videos and where they are stored / where they are being copied to. S3 could be a lot slower in a real environment that Minio is locally. Running this code directly on the command as opposed to running it in celery could stall the web server in a scenario where the command was being run directly on the machine running the web server. In theory firing up a Heroku runner through the CLI to do this wouldn't have any adverse effects, but still I think it's probably better to run this code in a Celery task.

gumaerc (changes requested):
As noted above, this functions properly locally but I worry a bit about this being used in production without the main code in a Celery task.

gumaerc (approved):
👍 

Looks good, but I do have a suggestion:

videos/management/commands/copy_videos.py
```diff
@@ -101,141 +86,7 @@ def handle(self, **options):
...
-            )
-            new_transcript_file = new_transcript_resource.file
-
-            video_captions_resource = WebsiteContent.objects.filter(
-                file=video_captions_file
-            ).first()
-            new_captions_resource = create_new_content(
-                video_captions_resource, destination_course
+            copy_video_resource.delay(
+                source_course.uuid, destination_course.uuid, video.text_id
```
gumaerc, line 91: Maybe we can add some logging here, so it's more apparent that progress is being made? Currently running the command returns nothing. For example, `backpopulate_pipelines` does something like this:

```python
        self.stdout.write(
            f"Started celery task {task} to upsert pipelines for {len(website_names)} sites"  # noqa: E501
        )
```

It gives a little bit more assurance that the command is doing what it's supposed to do, and also gives you the UUID of the task in case you need to jump in a shell and troubleshoot. The `backpopulate_pipelines` command will also print errors from the task if there are any and print a message when the task is complete.

gumaerc (approved):
👍 

Looks good, but I do have a suggestion:

Issue #2140, opened 2024-04-10
Django 3.2 end of life


gumaerc:
### Description/Context
Django 3.2 is end-of-life as of April 1st, 2024: https://endoflife.date/django. We should upgrade to Django 4.2 to gain LTS support until April 2026.

### Plan/Design
 - [ ] Upgrade any dependencies of Django necessary to upgrade to 4.2.11
 - [ ] Upgrade Django itself to 4.2.11

PR #2141, opened 2024-04-10
Django 4.2.11 / Python 3.12.2 upgrade

gumaerc:
Closes https://github.com/mitodl/ocw-studio/issues/2140

This PR upgrades Django to 4.2.11, along with some other dependencies that need to be upgraded to support it. Python was also upgraded to the latest supported version, 3.12.2.

### How can this be tested?
 - Run `docker compose up --build` on this branch to rebuild your docker containers with the latest versions of everything
 - Visit http://localhost:8043/ and make sure the site is reachable
 - If you are set up for publishing sites locally, test doing so


gumaerc:
I dealt with all the deprecation warnings that could be solved and posted issues in libraries we use that have unfixed deprecations:

https://github.com/jaraco/cssutils/issues/49
https://github.com/makinacorpus/django-safedelete/issues/240

ibrahimjaved12:
There's one thing I wanted to mention. 
I was trying this tool [django-upgrade](https://github.com/adamchainz/django-upgrade) on this branch and it ended up changing/reformatting 13 files when I gave it target of `django 4.2`: `git ls-files -z -- '*.py' | xargs -0 django-upgrade --target-version 4.2`


I analyzed those changes individually, and they were:
- Removing `default_app_config`
- Using `@admin.register()`, `@admin.display()` (over `admin.site.register`)
- `django.urls` over `django.conf.urls`
- `path` over `re_path`

None of these earlier code chunks are depreciated or breaking, so it was safe to merge the PR. However, each of these is according to modern Django practices and yes we may want to reformat the code accordingly sometime soon or later.

PR #2146, opened 2024-04-18
Update actions: checkout, setup-python, setup-node, node-version

ibrahimjaved12:
None, but there are Renovate PRs:
https://github.com/mitodl/ocw-studio/pull/2143
https://github.com/mitodl/ocw-studio/pull/2144
https://github.com/mitodl/ocw-studio/pull/2145

This PR updates versions of actions `checkout`, `setup-node`, `setup-python` and also updates `node-version`.
I have added them all in one place for easier testing via runner logs of workflow runs.
I upgraded `node-version` from `16.15.0` to `20`. The reason for this is `yarn v4` uses `>=18.12.0`. So we would be able to update `yarn` later .

### How can this be tested?
Check out the runner logs of PR's workflow runs, and you can compare it with any other PR's workflow runs to make sure everything is good.

Here are two runs of this PR:
https://github.com/mitodl/ocw-studio/actions/runs/8750466425
https://github.com/mitodl/ocw-studio/actions/runs/8751107510

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,8 @@ jobs:
         - 9200:9200
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
+
 
     - name: update
       run: sudo apt-get update -y
@@ -60,7 +61,7 @@ jobs:
         virtualenvs-in-project: true
 
     - name: Set up Python
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5
       with:
         python-version: "3.10.12"
         cache: "poetry"
@@ -120,10 +121,10 @@ jobs:
   frontend-tests:
     runs-on: ubuntu-20.04
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
 
     - name: Setup NodeJS
-      uses: actions/setup-node@v4
+      uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4
       with:
         node-version: 20
         cache: yarn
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -60,7 +61,7 @@ jobs:
         virtualenvs-in-project: true
 
     - name: Set up Python
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5
```
HussainTaj-arbisoft, line 64: Although this is the correct commit hash, what if we can use the version string instead?

```suggestion
      uses: actions/setup-python@v5.1.0
```

Reference: https://github.com/actions/setup-python/releases/tag/v5.1.0

.github/workflows/ci.yml
```diff
@@ -44,7 +44,8 @@ jobs:
         - 9200:9200
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
```
HussainTaj-arbisoft, line 47: Same here.

```suggestion
    - uses: actions/checkout@v4.1.1
```

Reference: https://github.com/actions/checkout/releases/tag/v4.1.1

.github/workflows/ci.yml
```diff
@@ -120,12 +121,12 @@ jobs:
   frontend-tests:
     runs-on: ubuntu-20.04
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
 
     - name: Setup NodeJS
-      uses: actions/setup-node@v3
+      uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4
```
HussainTaj-arbisoft, line 127:
```suggestion
      uses: actions/setup-node@v4.0.2
```

Reference: https://github.com/actions/setup-node/releases/tag/v4.0.2

HussainTaj-arbisoft (changes requested):
This looks good. I just think the version strings are more readable. Let me know if there is a specific reason to choose hashes.

HussainTaj-arbisoft (approved):
@ibrahimjaved12, thank you for the explanation. It makes sense. LGTM!

Renovate should be using version numbers; does it think we're bots too? 😆

PR #2184, opened 2024-05-22
upgraded PostgreSQL version from 12.8 to 16.3

umar8hassan:
This closes https://github.com/mitodl/hq/issues/4323

PostgreSQL version upgrade from 12.8 to 16.3. Follow instructions in https://github.com/mitodl/hq/issues/4323 to migrate data locally (if required)

### How can this be tested?
1. Restart Containers with the following commands, and check if ocw-studio is running:
    - `docker-compose down`
    - `docker-compose up -d`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ LABEL maintainer "ODL DevOps <mitx-devops@mit.edu>"
 # Add package files, install updated node and pip
 WORKDIR /tmp
 
-# Install packages and add repo needed for postgres 9.6
+# Install packages and add repo needed for postgres 16.3
 COPY apt.txt /tmp/apt.txt
 RUN apt-get update
 RUN apt-get install -y $(grep -vE "^\s*#" apt.txt  | tr "\n" " ")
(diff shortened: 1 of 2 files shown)


blarghmatey:
This will also need to include an update to the GH workflow [here](https://github.com/mitodl/ocw-studio/blob/master/.github/workflows/ci.yml#L16) to ensure that the tests are using the proper version of the database.

umar8hassan:
> This will also need to include an update to the GH workflow [here](https://github.com/mitodl/ocw-studio/blob/master/.github/workflows/ci.yml#L16) to ensure that the tests are using the proper version of the database.

Thanks @blarghmatey for taking time to review this. This would have been our next step as mentioned here https://github.com/mitodl/hq/issues/4242.

ibrahimjaved12 (approved):
Looks good to me. While we are at it, I think we should also use this opportunity to update this stale comment with current version of `posgres`: https://github.com/mitodl/ocw-studio/blob/master/Dockerfile#L7

PR #2188, opened 2024-05-27
Remove the internal external radio buttons

umar8hassan:
Closes https://github.com/mitodl/hq/issues/4101

It removes the Item type radio button in `Menu Item Form`.
Screenshot 2024-05-27 at 5 52 27 AM

### How can this be tested?
1. Login to `cow-studio`
2. Add a new site course or use and existing one
3. In the course setting, visit page for `Menu`
Screenshot 2024-05-27 at 5 55 10 AM

4. Add a new menu item or open an existing one
5. The item form should not have radio button as shown in the above picture in description

### DIFF
--- a/static/js/components/forms/MenuItemForm.test.tsx
+++ b/static/js/components/forms/MenuItemForm.test.tsx
@@ -53,8 +53,7 @@ describe("MenuItemForm", () => {
     })
     expect(wrapper.prop("initialValues")).toEqual({
       menuItemTitle: "",
-      menuItemType: "internal",
-      internalLink: "",
+      contentLink: "",
     })
   })
 
@@ -62,15 +61,13 @@ describe("MenuItemForm", () => {
     const activeItem = {
       text: "text",
       targetContentId: "content-id",
-      menuItemType: "internal",
     }
     const wrapper = renderForm({
       activeItem,
     })
     expect(wrapper.prop("initialValues")).toEqual({
       menuItemTitle: activeItem.text,
-      internalLink: activeItem.targetContentId,
-      menuItemType: "internal",
+      contentLink: activeItem.targetContentId,
     })
   })
 
@@ -80,11 +77,11 @@ describe("MenuItemForm", () => {
       {
         values: {
           menuItemTitle: "",
-          internalLink: "",
+          contentLink: "",
         },
       },
     )
-    const relationField = wrapper.find('RelationField[name="internalLink"]')
+    const relationField = wrapper.find('RelationField[name="contentLink"]')
     expect(relationField.exists()).toBe(true)
   })
 
@@ -101,12 +98,12 @@ describe("MenuItemForm", () => {
       {
         values: {
           menuItemTitle: "",
-          internalLink: value,
+          contentLink: value,
         },
         setFieldValue: setFieldValueStub,
       },
     )
-    const relationField = wrapper.find('RelationField[name="internalLink"]')
+    const relationField = wrapper.find('RelationField[name="contentLink"]')
     expect(relationField.exists()).toBe(true)
     expect(relationField.prop("value")).toEqual(value)
     expect(relationField.prop("valuesToOmit")).toEqual(existingMenuIds)
@@ -117,6 +114,6 @@ describe("MenuItemForm", () => {
     }
     // @ts-expect-error Not using a full event
     relationField.prop("onChange")(fakeEvent)
-    expect(setFieldValueStub).toHaveBeenCalledWith("internalLink", "abc")
+    expect(setFieldValueStub).toHaveBeenCalledWith("contentLink", "abc")
   })
 })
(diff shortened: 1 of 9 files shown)


arslanashraf7:
@ibrahimjaved12 Could you review this?

static/js/components/forms/MenuItemForm.tsx
```diff
@@ -21,28 +21,18 @@ interface Props {
...
-  menuItemType: yup.mixed().oneOf([LinkType.Internal, LinkType.External]),
-  externalLink: yup.string().label("External link").url().when("menuItemType", {
-    is: LinkType.External,
-    then: yup.string().required(),
-  }),
-  internalLink: yup.string().label("Internal link").when("menuItemType", {
-    is: LinkType.Internal,
-    then: yup.string().required(),
-  }),
+  internalLink: yup.string().required().label("Internal link"),
```
ibrahimjaved12, line 30: I think the label should be just `Link` now
image

Also, do you think perhaps we should use something else instead of `internalLink`?
Initially the variable name set out distinction for internal links, and external links, but now it is kind of ambiguous.

Even though the "Content" is internal, but content will also contain external resources.
Maybe it could be something like `link` or `contentLink`, or anything else ...

ibrahimjaved12 (changes requested):
The changes are working as expected. And I see that you also updated tests, and removed a lot outdated code.

However, I think we may be able to use some more cleanup.

When I search `external` within the codebase, there are some places where `external links` is still mentioned (I don't mean within context of external resources). 

We should use this opportunity to update it also. Also it appears a few management commands are currently also using `external URL`.

For example, these are a few files:
`ocw-studio/websites/management/commands/markdown_cleaning/rules/validate_urls.py `
`ocw-studio/websites/management/commands/markdown_cleaning/rules/broken_link_fix.py`
`ocw-studio/websites/constants.py `- `EXTERNAL_IDENTIFIER_PREFIX = "external-"` - is this used anywhere?

Also there is `isExternalLinkId` which seems to be used  a lot in the codebase.

We should remove `external link` related stuff wherever possible, make sure it does not break the management commands. And update `internal link` related variables' names for their better clarity now. 

Just search through `external` and `internal` in the codebase for cleanup. More thorough cleanup would depend on priority, but we should do at least what we can here.

umar8hassan:
> The changes are working as expected. And I see that you also updated tests, and removed a lot outdated code.
> 
> However, I think we may be able to use some more cleanup.
> 
> When I search `external` within the codebase, there are some places where `external links` is still mentioned (I don't mean within context of external resources).
> 
> We should use this opportunity to update it also. Also it appears a few management commands are currently also using `external URL`.
> 
> For example, these are a few files: `ocw-studio/websites/management/commands/markdown_cleaning/rules/validate_urls.py ` `ocw-studio/websites/management/commands/markdown_cleaning/rules/broken_link_fix.py` `ocw-studio/websites/constants.py `- `EXTERNAL_IDENTIFIER_PREFIX = "external-"` - is this used anywhere?
> 
> Also there is `isExternalLinkId` which seems to be used a lot in the codebase.
> 
> We should remove `external link` related stuff wherever possible, make sure it does not break the management commands. And update `internal link` related variables' names for their better clarity now.
> 
> Just search through `external` and `internal` in the codebase for cleanup. More thorough cleanup would depend on priority, but we should do at least what we can here.


@ibrahimjaved12 I have checked through a lot of code. `external` mostly shows up in management commands like `nav_item_to_external_resource` which is used to convert an obsolete type of `external link` to new type of `external resource`. This should be kept as we might need it in case some external links needs to be converted in production

There shouldn't be any breaking change as far as I can see. However, removing the suggested commands or constants will take a lot of effort alone with updating/removing the relevant tests. I shall track this in a separate issue and prioritize if need be.

umar8hassan:
@pdpinch what's your take on this?

static/js/components/forms/MenuItemForm.test.tsx
```diff
@@ -53,24 +53,21 @@ describe("MenuItemForm", () => {
     })
     expect(wrapper.prop("initialValues")).toEqual({
       menuItemTitle: "",
-      menuItemType: "internal",
-      internalLink: "",
+      contentLink: "",
```
ibrahimjaved12, line 56: 👍🏻

static/js/components/forms/MenuItemForm.test.tsx
```diff
@@ -53,106 +53,39 @@ describe("MenuItemForm", () => {
...
-          externalLink: "",
-          internalLink: "",
-        },
-      },
-    )
-    const extLinkField = wrapper.find('Field[name="externalLink"]')
-    expect(extLinkField.exists()).toBe(true)
-    expect(wrapper.find('RelationField[name="internalLink"]').exists()).toBe(
-      false,
-    )
```
ibrahimjaved12, line 152: Awesome. Good job getting rid of stale tests as well. This will simplify our codebase.

pdpinch:
> @pdpinch what's your take on this? ⬆️

Sorry I haven't had a chance to comment until now. I appreciate @ibrahimjaved12's zeal for removing code, but I also don't want to hold up this change too long. We should remove the UI as soon as reasonably possible in order to avoid course authors being tempted to use it and creating more data we need to migrate.

ibrahimjaved12 (approved):
Looks good to me 👍🏻 There's a minor suggestion though. It's up to you.

Issue #2198, opened 2024-06-03
Refactor HTTP code constants for external resources


pt2302:
### Description/Context
https://github.com/mitodl/ocw-studio/blob/d6d5cd100ae8c935aeb6cc2fb2faebb1e05988fe/external_resources/constants.py#L4-L10 defines constants for external resources that could instead be imported from `rest_framework.status`, a library that is used elsewhere in the `ocw-studio` codebase.

### Plan/Design
Use `rest_framework.status` imports rather than defined constants for HTTP status codes.

PR #2199, opened 2024-06-03
Refactor HTTP status codes for external resources

pt2302:
Closes https://github.com/mitodl/ocw-studio/issues/2198.

This PR updates the constants for external resources to use `rest_framework.status` imports rather than defining the constants directly, to be consistent with the usage elsewhere in the `ocw-studio` codebase.

### How can this be tested?
No functionality is affected by this PR. However, for completeness, test that the external resource link-checking functionality is not affected by following the same testing instructions as for https://github.com/mitodl/ocw-studio/pull/2171.

### DIFF
--- a/external_resources/api_test.py
+++ b/external_resources/api_test.py
@@ -1,6 +1,7 @@
 """Tests for External Resources API"""
 
 import pytest
+from rest_framework.status import HTTP_200_OK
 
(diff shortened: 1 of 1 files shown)


umar8hassan (approved):
Looks good to me.

PR #2231, opened 2024-07-02
added encoding/decoding for URL in markdown editor

umar8hassan:
closes https://github.com/mitodl/hq/issues/4590

This PR aims to encode/decode URLs in text editor in `ocw-studio` for correct browser readable URLs.
The opening and closing brackets will be saved as `%28` and `%29` respectively for being later used in course compilation.

### How can this be tested?
1. Switch to branch `umar/4590-links-to-urls-that-have-parentheses`
2. Create any resource and link a URL with parentheses, e.g `https://commons.wikimedia.org/wiki/File:Saint_Joseph_charpentier_(La_Tour).jpg` in the text editor.
Screenshot 2024-07-02 at 5 47 47 PM

3. Save and publish changes.
4. Visit the published site. and Click on the URL it follow proper URL.

### Additional Testing
1. The content data in markdown should be save in encoded format.
Screenshot 2024-07-02 at 5 54 42 PM

### DIFF
--- a/static/js/lib/ckeditor/plugins/Markdown.test.ts
+++ b/static/js/lib/ckeditor/plugins/Markdown.test.ts
@@ -183,11 +183,11 @@ describe("Handling of raw HTML", () => {
 
     const html = `
     <p>
-    <a href="http://host.com/with_(parenthese)/uri">Hello World</a>
+    <a href="http://host.com/with_(parentheses)/uri">Hello World</a>
     </p>
     `
     expect(encodeParentheses(html2md(html))).toBe(
-      `[Hello World](http://host.com/with_%28parenthese%29/uri)`,
+      `[Hello World](http://host.com/with_%28parentheses%29/uri)`,
     )
   })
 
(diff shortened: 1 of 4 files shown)


pdpinch:
Does this only address parentheses? Will there be other characters that hit this type of problem, or is there something special about parentheses?

umar8hassan:
> Does this only address parentheses? Will there be other characters that hit this type of problem, or is there something special about parentheses? 

@pdpinch this happens for `()` only. Markdown has `[title](url)` syntax to store hyperlinks. When a url with parentheses is inserted it becomes `[title](url_with\(parentheses\))`. The `()` inside url needs to be escaped to meet the markdown syntax. This escaped markdown string is saved in DB which is then used for site compilation.

The html url encoding for `()` is `%28%29`. However, for the escaped characters `\(\)` it is encoded to `%5c%28%5c%29` which makes the URL invalid on compiled site.

To handle this, we encode/fix the url before saving it to DB. Then the correct encoded url is used in compilation. On author side, we decode it to show the original utl with parentheses.

PR #2254, opened 2024-07-24
added change to restrict the parentheses encoding/decoding to anchor …

umar8hassan:
closes 

### How can this be tested?
1. switch to branch `umar/4950-mathjax-not-rendering` in `ocw-studio` and spin up containers.
2.  Create any type of resource e.g.`Page` and open editor
3. Add a text in markdown e.g `Einstein theorized \(E=mc^2\) and I say \(d > 1\)`
4. Link a url with parentheses to any plain text like `URL`, e.g
`https://commons.wikimedia.org/wiki/File:Saint_Joseph_charpentier_(La_Tour).jpg`
5. Save and publish changes.
6. Visit the published site. The markdown text should be converted to math symbols. 
7. Click on the URL, it should resolve to a proper URL in browser.

### Additional Testing
1. The content data in markdown should be save in encoded format.
Screenshot 2024-07-24 at 3 57 42 PM

### DIFF
--- a/static/js/lib/ckeditor/plugins/Markdown.test.ts
+++ b/static/js/lib/ckeditor/plugins/Markdown.test.ts
@@ -186,7 +186,7 @@ describe("Handling of raw HTML", () => {
     <a href="http://host.com/with_(parentheses)/uri">Hello World</a>
     </p>
     `
-    expect(encodeParentheses(html2md(html))).toBe(
+    expect(html2md(encodeParentheses(html))).toBe(
       `[Hello World](http://host.com/with_%28parentheses%29/uri)`,
     )
   })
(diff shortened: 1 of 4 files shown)


ibrahimjaved12 (approved):
This is with plain text:
image


This is how it appears in CKEditor, even though I had `linked` the whole sentences at once: 
image
The link does not apply to it, as its not allowed along with these other functions which are greyed:
image

But the fix is working.

Before:

image

After:
image

The link: `https://en.wikipedia.org/wiki/Unicorn_(disambiguation)` is redirected and works in browser as: `https://en.wikipedia.org/wiki/Unicorn_%28disambiguation%29`



Approving this as this was the expected functionality within this PR, and its working as intended.

PR #2258, opened 2024-07-30
Always enable publishing

pt2302:
Closes https://github.com/mitodl/hq/issues/4191.

This PR allows for `ocw-www` or any course to be published regardless of whether there have been changes to the content. Previously, there needed to be unpublished changes in order for a course to be republished. The reason for this change is that updating the New Courses rotator requires republishing `ocw-www`, but there may have been no changes to `ocw-www` itself.

### How can this be tested?
Using the OCW Studio UI, publish `ocw-www`. Verify that a job was started and successfully completed in the publishing pipeline on Concourse, at `http://concourse:8080`. Then, without making any changes to the content, publish again. Verify that another publishing job was started and successfully completed.

### DIFF
--- a/static/js/components/PublishDrawer.test.tsx
+++ b/static/js/components/PublishDrawer.test.tsx
@@ -186,12 +186,12 @@ describe("PublishDrawer", () => {
         expect(wrapper.find(".publish-option label").at(idx).text()).toBe(label)
       })
 
-      it("disables the button if there is no unpublished content", async () => {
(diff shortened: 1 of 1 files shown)


pdpinch:
Let me know when this gets onto RC so I can ask Jen to test it as well.

gumaerc (approved):
👍 Looks good to me!

It doesn't even make an empty commit to the git repo like I thought, you can just click the button to trigger a build.

PR #2271, opened 2024-08-07
Add SlackAlertStep in remove_unpublished_sites pipeline

ibrahimjaved12:
Closes https://github.com/mitodl/hq/issues/4857

Adds Slack notification for unpublished sites failures and pipeline abort operation.

This isn't exact screenshot, as the messages should be: `"Failed to remove unpublished site {short_id} (Site Name: {site_name}). Check the pipeline logs for more details."` OR `"User aborted the unpublish operation for site {short_id} (Site Name: {site_name}). Check the pipeline logs for more details."` But it shows the notifications working in action.

image

### Test Context:
1. You will need to set `OCW_STUDIO_ENVIRONMENT` to something other than `dev`. You can set it to `rc`. 
2. You can set `OCW_STUDIO_BASE_URL` to an incorrect value, such as `abc` to cause the unpublish pipeline to fail. Abort is simple, you unpublish the site, go to the pipeline, and abort it. But for abort, you will want to use correct values. You can set it to RC: `OCW_STUDIO_BASE_URL=http://ocw-studio-rc.odl.mit.edu` for abort. I think you may also need to have the database dump? you can try `ocw-www` site if you don't have dump.
3. The error messages will not be able to run in development. I added `site_name`, and `short_id`, which will be used via Vault in RC/Production, but they won't work locally. So in order to test, you should checkout to this branch, and modify the messages:
```
 text=f"""
                User aborted the unpublish operation for site {short_id} (Site Name: {site_name}). Check the pipeline logs for more details.
                """,  # noqa: E501
            )
```
to without the `short_id` and `site_name`. Same goes for fail message. (Example in screenshot). Only then the procedure will work.
4. You can create your own Slack webhook, but it should be easier if you use my personal Slack workspace webhook in `resources.py`, and `mass-build-sites.yml`. Under `((slack-url))`. Just DM me and I'll provide you with the URL.

### Steps:
5. Checkout to this branch.
6. Spin up OCW Studio. Modify the code from branch as shown above to test locally.
7. Upsert the `site_removal_pipeline`: `docker-compose exec web ./manage.py upsert_site_removal_pipeline`
  - Unpause the pipeline.
8. Go to any site, unpublish it. Immediately go to the `site_removal_pipeline` at `http://concourse:8080/` , and abort it.
9. For failure operation, use incorrect values for `OCW_STUDIO_BASE_URL`.

### DIFF
--- a/content_sync/pipelines/definitions/concourse/remove_unpublished_sites.py
+++ b/content_sync/pipelines/definitions/concourse/remove_unpublished_sites.py
@@ -74,9 +74,6 @@ class UnpublishedSiteRemovalPipelineDefinition(Pipeline):
     def __init__(self, **kwargs):
         base = super()
         base.__init__(**kwargs)
-        namespace = ".:site."
-        short_id = f"(({namespace}short_id))"
-        site_name = f"(({namespace}site_name))"
         common_pipeline_vars = get_common_pipeline_vars()
         web_bucket = common_pipeline_vars["publish_bucket_name"]
         offline_bucket = common_pipeline_vars["offline_publish_bucket_name"]
@@ -223,14 +220,14 @@ class UnpublishedSiteRemovalPipelineDefinition(Pipeline):
             resources.append(self._slack_resource)
             job.on_failure = SlackAlertStep(
                 alert_type="failed",
-                text=f"""
-                Failed to remove unpublished site {short_id} (Site Name: {site_name}). Check the pipeline logs for more details.
+                text="""
+                Failed to remove unpublished site. Check the pipeline logs for more details.
                 """,  # noqa: E501
             )
             job.on_abort = SlackAlertStep(
                 alert_type="aborted",
-                text=f"""
-                User aborted the unpublish operation for site {short_id} (Site Name: {site_name}). Check the pipeline logs for more details.
+                text="""
+                User aborted the unpublish operation for site. Check the pipeline logs for more details.
                 """,  # noqa: E501
             )
         job.plan = tasks
--- a/content_sync/pipelines/definitions/concourse/remove_unpublished_sites.py
+++ b/content_sync/pipelines/definitions/concourse/remove_unpublished_sites.py
@@ -74,6 +74,8 @@ class UnpublishedSiteRemovalPipelineDefinition(Pipeline):
     def __init__(self, **kwargs):
         base = super()
         base.__init__(**kwargs)
+        namespace = ".:site."
+        short_id = f"(({namespace}short_id))"
         common_pipeline_vars = get_common_pipeline_vars()
         web_bucket = common_pipeline_vars["publish_bucket_name"]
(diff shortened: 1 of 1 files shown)


content_sync/pipelines/definitions/concourse/remove_unpublished_sites.py
```diff
@@ -212,6 +218,22 @@ def __init__(self, **kwargs):
...
         ]
         job = Job(name=self._remove_unpublished_sites_job_identifier, serial=True)
-
+        if not is_dev():
+            resource_types.append(slack_notification_resource())
+            resources.append(self._slack_resource)
+            job.on_failure = SlackAlertStep(
+                alert_type="failed",
+                text=f"""
+                Failed to remove unpublished site {short_id} (Site Name: {site_name}). Check the pipeline logs for more details.
```
gumaerc, line 227: > The error messages will not be able to run in development. I added site_name, and short_id, which will be used via Vault in RC/Production, but they won't work locally.

I get what you're going for here, but I don't think that this works exactly like you're intending it to. `short_id` and `site_name` are not set via Vault in RC / production. The `get_unpublished_sites_task` fetches the list of sites to be unpublished from `ocw-studio` then iterates them in an `AcrossStep`:

```python
            DoStep(
                do=across_tasks,
                across=[
                    AcrossVar(
                        var="site",
                        values=f"((.:{self._unpublished_sites_var_identifier}.sites))",
                        max_in_flight=5,
                    )
                ],
            ),
```

This creates the `site` namespace, but it is only available from within tasks running in the `AcrossStep`. When the pipeline fails *outside* the across step, such as when hitting the `ocw-studio` API to get the list of sites, `site_name` and `short_id` are not available. If we want to include this information in the Slack alert, we would need to provide different `on_failure` / `on_abort` steps both inside and outside the context of the `AcrossStep`. The best way to do this would be to wrap the tasks in `across_tasks` in `add_error_handling` or manually setting the `on_failure` / `on_abort` of these steps. The global error handler on the job should not try to use these variables, as they are not available outside the context of the `AcrossStep`. It might be best just to remove this information from the error entirely. I've found that when debugging an issue it isn't super helpful to have this information in Slack, and I think this is especially true with the unpublish sites pipeline, since there is only one pipeline with no instance vars. The purpose of the Slack alert is to let you know that something went wrong. Generally you are just going to click the link to bring up the pipeline, and detailed error information will be available in the pipeline logs.

gumaerc (changes requested):
I think we should revise this to be a bit more simple:

content_sync/pipelines/definitions/concourse/remove_unpublished_sites.py
ibrahimjaved12: Thanks for the detailed explanation 👍🏻 I'll update the PR to remove the step. The error alert (with the link) itself should be sufficient.

gumaerc (approved):
👍 Looks good now!

PR #2279, opened 2024-08-22
Umar/5124 create external resource with add links

umar8hassan:
Closes https://github.com/mitodl/hq/issues/5124

This change is aimed to add External Resources on Link Add/Update. The syntax used for Link will be replaced with syntax for Resources in Markdown i.e. `{{ % resource_link "uuid" "title" % }}`

### How can this be tested?
1. Switch to `umar/5124-create-external-resource-with-add-links` branch in `ocw-studio`
2. Open text editor
3. Add a link 
Screenshot 2024-08-23 at 11 01 13 AM
4. The syntax of link should be replaced with resource syntax in editor
Screenshot 2024-08-23 at 10 58 04 AM
5. The resource should not be editable
Screenshot 2024-08-23 at 11 01 20 AM

6. Type a URL and upon space/enter, it should convert to a Resource Link
  - If the same URL as above, verify that the existing resource ID is being used.
  - If a different URL a new resource id will be used
7. verify the use case without selecting any text and adding a link with the AddLink button

### Additional Testing
Enable/disable `OCW_STUDIO_CUSTOM_LINKUI_ENABLE` feature flag on PostHog to enable/disable the feature for users.
Setup PostHog with instruction in the Readme.


static/js/lib/ckeditor/plugins/CustomLinkPlugin.ts
```diff
@@ -0,0 +1,93 @@
...
+export default class CustomLinkUI extends Plugin {
+  static get pluginName(): string {
+    return "CustomLinkUI"
+  }
+
+  static get requires() {
+    return [LinkUI]
+  }
+
+  private get syntax() {
```
ibrahimjaved12, line 19: I think this isn't being used anywhere?

static/js/lib/ckeditor/plugins/CustomLinkPlugin.ts
```diff
@@ -0,0 +1,93 @@
...
+      for (const range of ranges) {
+        // Iterate over all items in the range
+        for (const item of range.getItems()) {
+          // Check if the item is a text node or text proxy
+          if (item.is("text") || item.is("textProxy")) {
+            title += item.data
+          }
+        }
+      }
+      console.log(title)
```
ibrahimjaved12, line 45: You should remove this

static/js/lib/ckeditor/plugins/CustomLinkPlugin.ts
```diff
@@ -0,0 +1,93 @@
...
+
+      customLinkHook(this.editor, href, siteName, title)
+    })
+
+    console.log("CustomLinkUI Plugin is initialized")
```
ibrahimjaved12, line 50: And this

ibrahimjaved12 (approved):
Looks good to me

pdpinch:
Is this behind a (posthog) feature flag? I want to be able to control the roll-out of this feature, limiting it to certain users until it's been thoroughly tested.

pdpinch:
Is this blocked by #2291 ?

umar8hassan:
> Is this blocked by #2291 ?

Yes. It was. But I can rebase on the branch and work on it now, since PostHog has been integrated with core functionality. Additional changes for PostHog integration will be done later in separate tickets.

ibrahimjaved12 (approved):
👍🏻

pdpinch:
When fixing https://github.com/mitodl/ocw-studio/pull/2279#issuecomment-2373995475 be sure to add a test

Otherwise this should get merged Thursday 9/26

ibrahimjaved12 (changes requested):
I tried all these things: 
1. Pasting link and pressing enter
2. typing link and pressing enter
3. Text selection and adding link via button
4. No text but adding link via button
All of them work.

However, there is a corner case. If you try (4), the code uses the link as title, which has length violations. I faced some weird behaviors, but I think they were all caused because of ER title (filename).

When I paste this [314 characters URL](https://www.google.com/search?q=ocw&sca_esv=9177e7b6d39e290e&source=hp&ei=bXr-ZpuHOa-I7NYP65iO2QI&iflsig=AL9hbdgAAAAAZv6IfYUTsKdoeMHIaE14JwXybFOnf08d&ved=0ahUKEwjb8f-8iPKIAxUvBNsEHWuMIysQ4dUDCBY&uact=5&oq=ocw&gs_lp=Egdnd3Mtd2l6IgNvY3dI0gFQAFiWAXAAeACQAQCYAQCgAQCqAQC4AQPIAQD4AQGYAgCgAgCYAwCSBwCgBwA&sclient=cws-wiz) into the editor and press Enter, it creates an ER with the same link as title we get this error:

```
web-1               |     return self.cursor.execute(sql, params)
web-1               |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
web-1               | django.db.utils.DataError: value too long for type character varying(125)
web-1               | 
```

This is caused by: 
`CONTENT_FILENAME_MAX_LEN = 125
`
used in `websites/models.py`:
```
    filename = models.CharField(
        max_length=CONTENT_FILENAME_MAX_LEN,
        null=False,
        blank=True,
        default="",
        help_text="The filename of the file that will be created from this object WITHOUT the file extension.",  # noqa: E501
    )
```
This limit is too small now that links are mapped to titles. URLs are easily longer than 200 characters (in a recent PR I set character limit of 500 for `wayback_url` because there were 300+ characters ERs in production).

Also, I'm not sure about the implications of having super long titles either. Apart from OCW studio breaking, talking from data point of view.

The admin panel looks like this, all the fields are pushed to the right:
image

And OCW Studio breaks up like this:
image

Upon trying all these combinations, I somehow also created an empty ER:
image

But I wasn't able to replicate this.
The WebsiteContent in Django admin looks like this (top one -- no name)
image

This happened when I was trying long URLs repeatedly with some length variations, in combinations of 1-4 use cases shared above. Normally it doesn't even allow, but somehow it allowed once. I think the length was culprit as that is what I was targeting.

Also, I wanted to share another thing.

These are two URLs:
`https://www.google.com/search?q=ocwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww`

`https://www.google.com/search?q=ocw&sca_esv=9177e7b6d39e290e&source=hp&ei=bXr-ZpuHOa-I7NYP65iO2QI&iflsig=AL9hbdgAAAAAZv6IfYUTsKdoeMHIaE14JwXy`

The first one is 134 characters
The second one is 141 characters
Both are over 125 characters.

If you create ER for the first one (134 characters), it won't allow you. But the second one will allow you. This is because there is also some parsing involved, the symbols are stripped off, which makes the 2nd one shorter, and the 1st one longer (than 125 characters).

I think a solution here might be to simply strip off some last X characters and follow a cap for `filename` while passing the data. It can be same as 125 characters, or something like 200 characters.

ibrahimjaved12 (approved):
This fixes the errors. However, there is still some length related corner cases for sake of UI (Studio and admin panel). Created a ticket for this: https://github.com/mitodl/hq/issues/5716

Other than that, I also found one other thing. If we type text and create an empty ER on it, the editor allows it. Even though otherwise we don't allow this:

image
I think this is because we were able to create empty anchor links before in the same way But now it is empty ER, so in this context it is different.

I leave it to you to deal with this as you like. The PR functionalities work as expected.

Issue #2168, opened 2024-04-25
separately track publish_date and build_date


gumaerc:
### Description/Context
First, a bit of prior discussion on this topic for reference:

https://github.com/mitodl/hq/discussions/3169
https://github.com/mitodl/hq/issues/2862

Currently, the `publish_date` field on a `Website` object is updated at the end of a run of `SitePipeline` in Concourse. When the site has been successfully published, a webhook call is made back to `ocw-studio` that tells it to update the `publish_date` field for that site. The main issue with this is that the pipeline code is not exclusively triggered by user interaction. When there is a change to `ocw-hugo-themes` or `ocw-hugo-projects` a rebuild of the entire website is triggered, thus updating `publish_date` for every website. The issue with this is that courses that have not had any publish action by a content author have their `publish_date` updated, even though no actual content changes were made. This makes it impossible to determine if a course has actually had changes published to production, or if the site has simply been rebuilt using the existing content.

### Plan/Design
Currently, we have `draft_publish_date` and `live_publish_date` fields on the `Website` model. We should:

 - Add `draft_build_date` and `live_build_date` as `DateTimeField` fields to the `Website` model
 - The `update_website_status` function, which is called by the `pipeline_status` endpoint, should update these new build date fields instead of the publish date fields
 - The `publish_website` function should be changed to update the publish date fields (`draft_publish_date` or `live_publish_date`) after Github operations have completed, setting the appropriate status based on whether it succeeds or fails

PR #2284, opened 2024-08-30
Umar/2168 separately track publish date and build date

umar8hassan:
Fixes https://github.com/mitodl/ocw-studio/issues/2168

This change separately tracks the build and publish dates for live/draft versions. The publish dates are updated on `mass_build` which indicates no content change. Build dates and publish dates are both updated on `publish from studio`.

### How can this be tested?
1. Switch to `umar/2168-separately-track-publish_date-and-build_date` branch `ocw-studio`
2. Check migrations applied. `docker-compose exec web ./manage.py showmigrations websites` 
    - Last applied migration should be `0057_website_draft_build_date_website_live_build_date`
    - If migration is not applied, apply migrations: `docker-compose exec web ./manage.py migrate websites`
3. Create a site in `ocw-studio` locally at `localhost:8043/sites` or use an existing one.
4. Make any changes and publish site in `ocw-studio` and Wait for the publish operation to complete. 
5. Go to Django admin at `localhost:8043/admin` and search for the site in `Websites`.
6. In Django admin verify that only `publish date` will be updated for the site.
7. Run this command to trigger a mass build locally. Replace the variables with appropriate values: 
    `docker-compose exec web ./manage.py mass_publish <live or draft> --filter <site-name>`
8. In Django admin verify that `Build date` and `Publish date` have been updated.


pdpinch:
> I think that we want the opposite changes. Specifically, we want build date updates to be triggered both by user-initiated publish actions and by the mass build pipeline (i.e., the last time that the course site was built) and publish date updates to be triggered only by user-initiated publish actions. So the publish date should stay the same when the mass build pipeline runs.

This is correct, but I'd like to clarify a bit more. Am I right that: 

build: current release branch of theme and content are checked out and hugo builds the HTML. 

publish: main branch of content repo is merged into release branch and then build process is followed.

PR #2290, opened 2024-09-02
added unpublish status in site listing

umar8hassan:

- [ ] Desktop screenshots

### DIFF
--- a/static/js/pages/SitesDashboard.tsx
+++ b/static/js/pages/SitesDashboard.tsx
@@ -20,7 +20,7 @@ import Dropdown from "../components/Dropdown"
 import UnpublishDialog from "../components/UnpublishDialog"
 import { useURLParamFilter, usePagination } from "../hooks/search"
 import { usePermission } from "../hooks/permissions"
-import { Permission } from "../constants"
+import { Permission, PublishStatus } from "../constants"
 
 function getListingParams(search: string): WebsiteListingParams {
   const qsParams = new URLSearchParams(search)
@@ -95,8 +95,13 @@ export default function SitesDashboard(): JSX.Element {
               key={site.uuid}
             >
               <div className="d-flex flex-row">
-                {site.publish_date && !site.unpublished ? (
+                {site.live_publish_status === PublishStatus.Pending ||
+                site.live_publish_status === PublishStatus.Started ? (
+                  <div className="text-warning">In Progress...</div>
+                ) : site.publish_date && !site.unpublished ? (
                   <div className="text-success">Published</div>
+                ) : site.unpublished ? (
+                  <div className="text-danger">Unpublished</div>
                 ) : (
                   <div className="text-dark">Draft</div>
                 )}
--- a/static/js/types/websites.ts
+++ b/static/js/types/websites.ts
@@ -258,6 +258,7 @@ export type Website = WebsiteStatus & {
   url_path: string | null
   url_suggestion: string
   s3_path: string | null
+  live_publish_status: PublishStatus | null
 }
 
 type WebsiteRoleEditable = typeof ROLE_ADMIN | typeof ROLE_EDITOR


pdpinch:
Sorry, I'm still thinking about this and I'd like to bring in some UX advice. 

To help out the UX team, can you list out all the possible status for:
1. the site listing page ("Draft," "Published," etc.) 
2. the publish status indicator ("Succeeded","Failed", anything else?) 
3. the publish drawer

gumaerc:
> Sorry, I'm still thinking about this and I'd like to bring in some UX advice.
> 
> To help out the UX team, can you list out all the possible status for:
> 
>     1. the site listing page ("Draft," "Published," etc.)
> 
>     2. the publish status indicator ("Succeeded","Failed", anything else?)
> 
>     3. the publish drawer

1. With Umar's change here, the possible values on the listing page (called `SitesDashboard`) are "Published," "Draft" or "Unpublished." In the site right now outside of this PR, it's only "Published" or "Draft" (sites go back to "Draft" when unpublished currently)
2. The publish status indicator has the following statuses it can display, based on the build status in Concourse:
```ts
const publishStatusIndicatorClass = (status: PublishStatus): string => {
  switch (status) {
    case PublishStatus.NotStarted:
      return "bg-secondary"
    case PublishStatus.Pending:
    case PublishStatus.Started:
      return "bg-warning"
    case PublishStatus.Aborted:
    case PublishStatus.Errored:
      return "bg-danger"
    case PublishStatus.Success:
      return "bg-success"
    default:
      return ""
  }
}
```
3. The drawer reuses the `PublishStatusIndicator` component so it displays the same status above

mbilalmughal:
After discussing the issue with Umer, I believe that not having labels for both statuses may still confuse users. For best practices and good UX, it’s important to guide users, and I think labels will help with that. I've proposed a solution to include separate labels for each. Please see the attached image and let me know if this resolves the problem.

In the first screenshot, I proposed the label 'Deployment,' while in the second screenshot, I used 'Site Status.' Umer preferred 'Site Status,' but I would rather avoid using 'Status' twice.

- [Figma Link](https://www.figma.com/proto/7icc6eD58sipYwydutYE4u/OCW-Logo-Update?page-id=168%3A11&node-id=168-14&node-type=frame&viewport=1467%2C983%2C1.17&t=njl38CNr900g4ntS-1&scaling=min-zoom&content-scaling=fixed)

- Reading from the beginning, it would say,
'Publish Status - Succeeded'
'Deployment - Failed'

{D1956456-5BEA-418A-BA64-0A6EB8A92B07}

- 2nd Screenshot
'Publish Status - Succeeded'
'Site Status - Failed'

{97D3C1BC-484B-4C76-AC24-6FC7025E33E4}

pdpinch:
Looks a bit like a Christmas Tree, or a traffic signal. 🙂

I appreciate the desire for clarity, but sometimes more information leads to increased cognitive load and confusion.

- The values for "publish status" seem transient to me, and not of much use on the list view. They are more valuable in the course publisher drawer, where the user is wondering what's happening right now, and is tempted to press the publish button over and over. 
- Looking at the values in the screenshots, they seem very redundant. Is that just because of the sample you chose?

mbilalmughal:
After discussing it with umer & reading the [thread](https://mitodl.slack.com/archives/GQ89D0Z5M/p1725381582568899)
Carey Comment:

> I think we should label and show both, but we should see what Peter thinks

> "publish status" vs "deployment status"
> if not messaged correctly it could cause more confusion though
> Consistent labels on both pages

Peter Comment:
> They are more valuable in the course publisher drawer

I think on the listing page, we need to either show both statuses for consistency or not show them at all. This means we should either display both or none. Currently, we are only showing one status, which is the deployment status. Also, the use of multiple colors is usually unavoidable when showing different statuses, but I’m open to any suggestions.

Drawer
Drawer-1
Drawer-2
Listing

pdpinch:
Sorry guys, this still isn't working for me. 

First, I think there are two different bits of information to communicate. One is the current state of the course -- is it never published, published or unpublished (to production). Most of the time, this is all the user wants to know. Then there's the state of change of the publication. that could be requested by "not started", "in progress", "failed" or "succeeded". Most of the time this is static. 

Second, I think we should be judicious about what information to show. For the list view, the most common use case is to lok through the list and see if a course has been published or not. The state of change is usually irrelevant and only of interest to someone who cares about that specific course. Can we give that less prominence? Maybe put it in a hover state, or a drop-down, or behind some of kind of "updating..." indicator? 

I'll discuss this a bit with Jen today to validate my assumptions about values and priority.

pdpinch:
@umar8hassan can you pick this up again and use it to close https://github.com/mitodl/hq/issues/6017 ?

Let's narrow the scope and remove any changes to the publish drawer. We can save those for a future issue.

umar8hassan:
@pdpinch, would you agree to closing this pull request as the linked [issue](https://github.com/mitodl/hq/issues/4858) has already been closed? And I have already put up a PR for https://github.com/mitodl/hq/issues/6017 as mentioned in https://github.com/mitodl/ocw-studio/pull/2290#issuecomment-2595665730

We can reopen this if needed in the future.

Issue #2275, opened 2024-08-08
Add PostHog integration to OCW Studio


pt2302:
### User Story
After merging https://github.com/mitodl/ocw-studio/pull/2255, external resources can now be deleted in OCW Studio. A related feature is to allow instructors to be deletable as well. These features should be able to be turned on/off using feature flags, and using PostHog to manage these feature flags is consistent with other Open Learning products.

### Description/Context
OCW Studio should be integrated with PostHog, and the deletion functionality should be wrapped in a PostHog feature flag as the first use case of this integration.

### Acceptance Criteria
- [ ] Set up the PostHog dashboard for the relevant features.
- [ ] Add the relevant dependencies to OCW Studio.
- [ ] Set up a feature flag for deleting content (currently, external resources and instructors) in OCW Studio.

PR #2291, opened 2024-09-04
Add PostHog Integration and Make Instructors Deletable

pt2302:
Closes https://github.com/mitodl/ocw-studio/issues/2275 and closes https://github.com/mitodl/hq/issues/5053.

This PR introduces the following changes:
1.  Sets up PostHog integration, including relevant documentation updates.
2.  Makes instructors in `ocw-www` deletable.
3. Wraps the functionality for deleting both instructors and external resources in the PostHog feature flag `OCW_STUDIO_CONTENT_DELETABLE`. The functionality for making external resources deletable was introduced in https://github.com/mitodl/ocw-studio/pull/2255.

### How can this be tested?
First, set up PostHog integration as described in the updated README for the repo (including setting the required environment variables) and restart the OCW Studio containers. Navigate to the OCW project page on PostHog and toggle the `OCW_STUDIO_CONTENT_DELETABLE` feature flag. When this is set to on, both instructors and external resources should be deletable in your local OCW Studio. When it is set to off, neither should be deletable. 

Also, experiment with rolling out the feature flag only to individual users. Add the login email address for your local instance of OCW Studio to `distinct_id equals <email address>` and verify that the feature is enabled only for the user with that email address.

### DIFF
--- a/README.md
+++ b/README.md
@@ -509,7 +509,7 @@ This feature sets up a cron job to validate external resource urls. The workflow
 
 PostHog is used for dynamically testing and rolling out features that may not be ready for permanent deployment as part of OCW Studio. For example, we use the feature flag `OCW_STUDIO_CONTENT_DELETABLE` to control whether content can be deleted.
 
-These variables should be set in the `.env` file for PostHog integration:
+The following variables should be set in the `.env` file for PostHog integration:
 
 ```
 POSTHOG_ENABLED=True
@@ -517,7 +517,7 @@ POSTHOG_API_HOST=https://app.posthog.com
 POSTHOG_PROJECT_API_KEY=<obtain from the PostHog dashboard>
 ```
 
-These variables can be optionally set:
+The following variables can be optionally set to configure PostHog requests:
 
 ```
 POSTHOG_FEATURE_FLAG_REQUEST_TIMEOUT_MS=<3000 by default>


static/js/components/RepeatableContentListing.tsx
```diff
@@ -37,16 +38,32 @@ import SiteContentEditorDrawer from "./SiteContentEditorDrawer"
...
-  const isExternalResource = configItem.name === "external-resource"
+
   const website = useWebsite()
 
+  const [isContentDeletable, setIsContentDeletable] = useState(false)
+
+  useEffect(() => {
+    const checkFeatureFlag = async () => {
+      const flagEnabled =
+        posthog.isFeatureEnabled("OCW_STUDIO_CONTENT_DELETABLE") ?? false
```
umar8hassan, line 57: This checks for feature flag without any user id verification. It'd be a good idea to send user information as well, so that we may control the rollout of feature to specific users. keeping that in mind, I'd suggest the following change.

```suggestion 
        posthog.isFeatureEnabled("OCW_STUDIO_CONTENT_DELETABLE", SETTINGS.user.email ) ?? false
```

umar8hassan:
While testing, it was seen that even if the Feature flag has been set to `false` the deletable option persisted in menu. I tried changing to other pages then back, but the option did not disappeared unless page was refreshed.
This even allowed me to create a new resource and delete it with Delteable flag set to `False`.

static/js/lib/util.ts
```diff
@@ -1,6 +1,27 @@
...
+    capture_pageview: false,
+    capture_pageleave: false,
+    cross_subdomain_cookie: false,
+    persistence: "localStorage+cookie",
+    loaded: function (posthog) {
+      posthog.setPersonPropertiesForFlags({
+        environment: environment,
+      })
+    },
+  })
```
umar8hassan, line 23: Another way to identify users could be with
```suggestion
  })
  posthog.identify(SETTINGS.user.email, { key: value,  <any optional properties> })
```

static/js/components/RepeatableContentListing.tsx
umar8hassan: Let's do this instead https://github.com/mitodl/ocw-studio/pull/2291#discussion_r1747231345

umar8hassan (approved):
Looks good.
I have tested the basic functionality of PostHog Feature flags. Currently, the condition is set to allow only the `<REDACTED>` email_id. I impersonated to test the feature flag working.
Screenshot 2024-09-11 at 5 54 57 PM
Screenshot 2024-09-11 at 6 04 47 PM
Screenshot 2024-09-11 at 5 53 57 PM
Screenshot 2024-09-11 at 5 54 06 PM

Issue #2305, opened 2024-09-12
Bug: PostHog Requires User to be Logged In


pt2302:
### Expected Behavior
If a user is not logged in, the `posthog.identify` function gracefully handles this.

### Current Behavior
This is causing OCW Studio to crash with the error: `TypeError: null is not an object (evaluating 'SETTINGS.user.email')`.

### Steps to Reproduce
Spin up OCW Studio and log out. The main page will not load.

### Possible Solution
Handle connecting to PostHog in the case when the user is not logged in.

PR #2306, opened 2024-09-12
Fix PostHog initialization if user is not logged in

pt2302:
Closes https://github.com/mitodl/ocw-studio/issues/2305.

This PR handles the case when the user is not logged in for initializing PostHog.

### How can this be tested?
Testing instructions are analogous to https://github.com/mitodl/ocw-studio/pull/2291 but verify that there are no errors after logging out and logging back into OCW Studio.


umar8hassan (approved):
It looks good 👍 
The error was reproduced in the local environment after removing cookies and user sessions. The PR fixed the broken login page in local environment

PR #2308, opened 2024-09-16
Add Wayback Machine integration: Update external_resources model and tasks

ibrahimjaved12:
Part of https://github.com/mitodl/hq/issues/5355
Specifically, https://github.com/mitodl/hq/issues/5516 & https://github.com/mitodl/hq/issues/5517

### Related PRs:
This should be reviewed together: https://github.com/mitodl/ocw-hugo-projects/pull/307
We also have to update `ocw-hugo-themes` to not use `is_broken` and `backup_url` stuff, this will be done in next PR. For now this can be reviewed.

Adds Wayback Machine integration:

1. We have a migration file that removes `backup_url`,  `is_broken`, and `backup_url_status_code` from WebsiteContent.metadata for type `external-resource`. It adds `status` (from `ExternalResourceState`) and `wayback_url`.
2. It re-defines `Status` from `ExternalResourceState`: To be only about External Resource (not `backup_url` -- now `wayback_url`). The statuses remain same as `unchecked`, `valid`, `broken`, and `check_failed`.
3. Wayback ER submission is only done in case of ER creation and while link checking functionality (which occurs once a week). It is not done in case of course publication to respect the API rate limits. But we will create a management command as next steps so we can have more control over the functionality.
4. Wayback Status Checking is implemented for batches, not for singular jobs. This is currently set for every 6 hours.
5. PostHog integration is not set up yet. This is for next steps.
6. Tests are not added yet. This is for next steps too.
7. Documentation / readme for the functionality will be in follow up tasks as well.

The following fields were added: `wayback_job_id`, `wayback_status`, `wayback_status_ext`, `wayback_url`, `wayback_http_status`, `wayback_last_successful_submission`. Each one serves a distinct purpose.

`wayback_job_id` is used for status retrieval of a specific job (not a specific URL).
`wayback_status` is `pending`, `error`, or `success` based on the response. Wayback Status Checking Task will look for `pending` jobs to update their status.
An `error` is not necessarily a bad thing in this context. `wayback_status_ext` gives more information about it. Here are some examples for `wayback_status_ext`:  `error:too-many-daily-captures` which occurs if we have sent the URL too many times for submission. (I can't recall how exactly is too many times-- I got this error when I sent the URL 5 times (on a day basis I think), but I also got this error with the error message that you can send URL again next hour).

Another example is `error:no-access` eg. for reuter.com

Lastly, `wayback_last_successful_submission` is used as a check -- to only submit URLs if they were already **successfully** submitted more than a month ago.

### How can this be tested?
First, DM me for `access_key` and `secret_key`.

Checkout to the `ocw-hugo-projects` branch [in the PR](https://github.com/mitodl/ocw-hugo-projects/pull/307) and copy the code for `ocw-www/ocw-studio.yaml` and `ocw-course-v2/ocw-studio.yaml` and update your Starters configuration for both via django admin.

Run migration: `docker-compose exec web ./manage.py migrate`
There are different ways to run `submit_url_to_wayback_task`. When you create an External Resource, `submit_url_to_wayback_task` is designed to run after it in celery queue.

Another way is, you can tie it to publishing operation of a site:

Eg. in content_sync/tasks.py:

```
from external_resources.tasks import submit_website_resources_to_wayback_task
@app.task(acks_late=True, autoretry_for=(BlockingIOError,), retry_backoff=True)
@single_task(10)
def publish_website_backend_draft(website_name: str):
    """
    Create a new backend preview for the website.
    """
    try:
        submit_website_resources_to_wayback_task.delay(website_name)
        api.publish_website(website_name, VERSION_DRAFT)
        return True  # noqa: TRY300
    except:  # pylint:disable=bare-except  # noqa: E722
        log.exception("Error publishing draft site %s", website_name)
        return website_name
```

Or you can change frequency in settings.py:
```
# Check External Resources settings
CHECK_EXTERNAL_RESOURCE_STATUS_FREQUENCY = get_int(
    name="CHECK_EXTERNAL_RESOURCE_STATUS_FREQUENCY",
    default=604800,
    description="Frequency (in seconds) to check potentially broken external urls",
    required=False,
)

```
To something like 120 (2 minutes), it will `call submit_url_to_wayback_task` for the specific ERs. (Make sure your user account has this functionality enabled in PostHog - I don't know if `check_external_resources_status` function is integrated with PostHog or not. So make sure you got that working.)

You can also run it in Django shell, like:
```
from external_resources.tasks import submit_website_resources_to_wayback_task
submit_website_resources_to_wayback_task.delay("website_name")
```
and then monitor celery logs.

Similarly, for status checking, you can change its frequency to something like 120 seconds:
```
# Wayback Machine Settings
UPDATE_WAYBACK_JOBS_STATUS_FREQUENCY = get_int(
    name="UPDATE_WAYBACK_JOBS_STATUS_FREQUENCY",
    default=21600,  # 6 hours in seconds
    description="Frequency (in seconds) to check the status of Wayback Machine jobs",
    required=False,
)
```
And then monitor celery logs.

Or, Execute the following in Django shell:

```
from external_resources.tasks import update_wayback_jobs_status_batch
update_wayback_jobs_status_batch()
```

Sample output:
```
In [2]: update_wayback_jobs_status_batch()
[2024-10-02 10:57:02] INFO 236 [root] tasks.py:292 - [cb9c8fa62f30] - Wayback Machine job spn2-ba94eaa471c8a8d7cd50a6c9d32b875a2f4c9fbd is still pending for resource ER2 (ID: 524780)
[2024-10-02 10:57:02] INFO 236 [root] tasks.py:292 - [cb9c8fa62f30] - Wayback Machine job spn2-7cee1ed123f872db426caf0117c2e11a6451107f is still pending for resource A new-style External Resource link (ID: 524805)
[2024-10-02 10:57:02] INFO 236 [root] tasks.py:292 - [cb9c8fa62f30] - Wayback Machine job spn2-2cc2a2661679e84570fea922810b4e04fdf7ce2e is still pending for resource Google Scholar (ID: 524810)
[2024-10-02 10:57:02] INFO 236 [root] tasks.py:292 - [cb9c8fa62f30] - Wayback Machine job spn2-0acc66b592e76df43eb62dc8f55507803bf811d2 is still pending for resource Egg (ID: 524827)
[2024-10-02 10:57:02] INFO 236 [root] tasks.py:274 - [cb9c8fa62f30] - Wayback Machine job spn2-f0a7ed7698d4fee327e2a67648f19e150f3c2950 succeeded for resource 25 (ID: 524853)
[2024-10-02 10:57:02] INFO 236 [root] tasks.py:292 - [cb9c8fa62f30] - Wayback Machine job spn2-6ac52eb35b1436618af76b3a612fcba55624b0c0 is still pending for resource OCW (backed by old site) (ID: 522845)
[2024-10-02 10:57:02] INFO 236 [root] tasks.py:292 - [cb9c8fa62f30] - Wayback Machine job spn2-b4d54115279c8359fb7cace1b4033b98b5e821e4 is still pending for resource 9 (ID: 524837)
[2024-10-02 10:57:02] INFO 236 [root] tasks.py:292 - [cb9c8fa62f30] - Wayback Machine job spn2-92c6ac4981c5e4ee0b80c721ea920f81dc64a534 is still pending for resource Part 1: Differentiation (ID: 524930)
[2024-10-02 10:57:02] INFO 236 [root] tasks.py:292 - [cb9c8fa62f30] - Wayback Machine job spn2-692982656e06b4a4a6a96977839a19327e11ea8a is still pending for resource Introductory Programming Courses (ID: 524871)
[2024-10-02 10:57:02] INFO 236 [root] tasks.py:292 - [cb9c8fa62f30] - Wayback Machine job spn2-e3d9f30a9a28b8f9f98f3154937403d03d3645f5 is still pending for resource Part 2: Integration (ID: 524931)
[2024-10-02 10:57:02] INFO 236 [root] tasks.py:292 - [cb9c8fa62f30] - Wayback Machine job spn2-d9ca3f4157ed21c48683b5487392e8f4118fa060 is still pending for resource Part 3: Coordinate Systems & Infinite Series ...
```


pdpinch:
We should probably put this behind a feature flag. It doesn't need to be a posthog flag necessarily, but I'd like to be able to turn it on and off without having to do a code deploy.

pdpinch:
Planning to have this ready for review on Monday 9/30

pdpinch:
@ibrahimjaved12 can you say more about this change?

> Wayback ER submission is only done in case of ER creation and **while link checking functionality (which occurs once a week).** 

Does this mean that we re-submit external resource URLs to the WayBack machine every time we check the link? That seems more often than is necessary. And we're tying it to an unrelated activity that we might want to change the frequency of.

pdpinch:
> PostHog integration is not set up yet. This is for next steps.

What's the plan for PostHog? I think all we need is an environment-level feature flag that we can turn on for RC and leave off on production until we are ready to turn it on.

ibrahimjaved12:
@pdpinch 

> Does this mean that we re-submit external resource URLs to the WayBack machine every time we check the link? That seems more often than is necessary. And we're tying it to an unrelated activity that we might want to change the frequency of.

I added a variable for this: `wayback_last_successful_submission`  which basically tracks time upon a successful submission.
In any case, next time a link is submitted only if there has been more than one month upon the last successful submission.

> What's the plan for PostHog? I think all we need is an environment-level feature flag that we can turn on for RC and leave off on production until we are ready to turn it on.

PostHog was for next subtask, but I did add feature flag here in this PR: 
```
ENABLE_WAYBACK_TASKS = get_bool(
    name="ENABLE_WAYBACK_TASKS",
    default=True,
    description="Enables tasks related to Wayback Machine submissions and status checks",  # noqa: E501
    required=False,
)
```

external_resources/tasks.py
```diff
@@ -108,3 +113,202 @@ def check_external_resources_for_breakages(self):
...
+@app.task(
+    bind=True,
+    acks_late=True,
+    rate_limit=WAYBACK_MACHINE_TASK_RATE_LIMIT,
+    autoretry_for=(BlockingIOError, ConnectionError, Timeout),
+    retry_backoff=True,
+    retry_backoff_max=128,
+    max_retries=7,
+)
+@single_task(7)
```
umar8hassan, line 144: With our current settings of a single worker for this task and a rate_limit of `0.11/s` set with this change, is it required to have an additional `redis lock`?

external_resources/tasks.py
```diff
@@ -108,3 +113,202 @@ def check_external_resources_for_breakages(self):
         return self.replace(celery.group(tasks))
 
     return None
+
+
+@app.task(bind=True)
+def submit_website_resources_to_wayback_task(self, website_name):
```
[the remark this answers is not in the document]
ibrahimjaved12: This will be used for the management command

external_resources/tasks.py
```diff
@@ -108,3 +113,202 @@ def check_external_resources_for_breakages(self):
...
+    max_retries=7,
+)
+@single_task(7)
+def submit_url_to_wayback_task(self, resource_id):
+    """Submit an External Resource URL to the Wayback Machine."""
+
+    try:
+        state = ExternalResourceState.objects.get(content_id=resource_id)
+        if state.wayback_last_successful_submission:
+            one_month_ago = timezone.now() - timedelta(days=30)
```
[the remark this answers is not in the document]
ibrahimjaved12: Good call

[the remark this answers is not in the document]
ibrahimjaved12: But would this be better in constants? @umar8hassan

umar8hassan:
There is another thing to consider, as the Wayback machine takes snapshots of site a page, we should have a resubmission button in the studio to force update the snapshot of site on user discretion

external_resources/tasks.py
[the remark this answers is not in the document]
ibrahimjaved12: Done

external_resources/migrations/0003_remove_externalresourcestate_backup_url_response_code_and_more.py
```diff
@@ -0,0 +1,162 @@
...
+            content.save(update_fields=["metadata"])
+
+
+def migrate_metadata_backward(apps, schema_editor):
+    """
+    Backward migration to revert changes to the metadata fields of WebsiteContent.
+    Note: This migration only restores the field names without
+    preserving their original data values.
+    """
+    WebsiteContent = apps.get_model("websites", "WebsiteContent")
```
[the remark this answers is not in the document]
ibrahimjaved12: Done

main/settings.py
```diff
@@ -732,6 +762,12 @@
     },
 }
 
+if ENABLE_WAYBACK_TASKS:
+    CELERY_BEAT_SCHEDULE["check-wayback-jobs-status-every-12-hours"] = {
```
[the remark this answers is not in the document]
ibrahimjaved12: Done

external_resources/tasks.py
ibrahimjaved12: The `rate_limit` is sufficient but it does not fully serve our purpose. Multiple workers will be able to pick up the task independently and that's where the `@single_task` comes in handy. It ensures only one worker processes the task at a time.

Another question could arise from this:
Could `single_task()` only suffice?

`rate_limit` divides the tasks based on the time given. And `single_task()` locks the task to a single worker for the duration specified, but it doesn't manage time-based distribution like `rate_limit` does. 

Using both together ensures efficient division of tasks across workers while making sure only one worker works on the task at the given time.

external_resources/migrations/0003_remove_externalresourcestate_backup_url_response_code_and_more.py
```diff
@@ -0,0 +1,162 @@
...
+        ),
+        migrations.AddField(
+            model_name="externalresourcestate",
+            name="wayback_status",
+            field=models.CharField(
+                blank=True,
+                choices=[
+                    ("pending", "Pending"),
+                    ("success", "Success"),
+                    ("error", "Error"),
```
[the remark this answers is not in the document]
ibrahimjaved12: As discussed, let's not use constants in this migration scope

ibrahimjaved12:
The requested functionalities have been added, but the PR is on hold because the Internet Archive is currently down. Until it gets back up, we won't be able to test the API.

external_resources/signals.py
```diff
@@ -0,0 +1,58 @@
...
+                "wayback_url": "",
+                "wayback_status": "",
+                "wayback_status_ext": "",
+                "wayback_http_status": None,
+                "wayback_last_successful_submission": None,
+            }
+            ExternalResourceState.objects.update_or_create(
+                content=instance, defaults=defaults
+            )
+            if settings.ENABLE_WAYBACK_TASKS:
```
umar8hassan, line 44: I introduced `posthog` feature flag-checking [here](https://github.com/mitodl/ocw-studio/pull/2316/files#diff-e48ade8b030312c01525fc67ffbe535d8e900ddf191885be7fcc8d084cb6a999). Which should be easy to use if needed.
Could you check with @pdpinch if we need posthog feature flag for this rather than having an environment variable.

external_resources/signals.py
```diff
@@ -0,0 +1,58 @@
...
+                submit_url_to_wayback_task.delay(instance.id)
+                log.info(
+                    "New external resource created: %s. Submitting to Wayback Machine.",
+                    instance.title,
+                )
+                log.debug(
+                    "Created ExternalResourceState for WebsiteContent id=%s",
+                    instance.id,
+                )
+            else:
```
umar8hassan, line 54: We should remove this `else` statement. we don't need to display any log if the feature is disabled.
PS. The log text is wrong. maybe a copy-paste issue?

external_resources/api.py
```diff
@@ -55,9 +58,52 @@ def is_external_url_broken(
...
-    url = external_resource.metadata.get("backup_url", "")
-    return is_url_broken(url)
+def submit_url_to_wayback(
+    url: str,
+) -> Optional[str]:
+    """
+    Submit the external resource URL to the Wayback Machine and
+    return the response
+    """
+    if not url:
```
umar8hassan, line 68: This check id a duplicate.
in `tasks.py` on line 173, just before calling this function, we have the same check for `Null` values.

external_resources/api.py
```diff
@@ -55,9 +58,52 @@ def is_external_url_broken(
...
+        return None
```
umar8hassan, line 69: If you do wanna keep this check, we should return the same  data type (i.e., `{}`) from this method

external_resources/api.py
```diff
@@ -55,9 +58,52 @@ def is_external_url_broken(
...
+
+
+def check_wayback_jobs_status_batch(job_ids: list[str]) -> list[dict]:
+    """
+    Check the status of multiple Wayback Machine jobs in batch.
+    """
+    if not job_ids:
+        return []
+
+    headers = {
```
umar8hassan, line 95: Since the headers are fixed and used multiple times, it would have been better to have them in a constants file.

external_resources/tasks.py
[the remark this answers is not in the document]
umar8hassan: We have a separate PR for management command. could this be moved there so that it may be tracked along with its usage and other changes?

main/settings.py
```diff
@@ -650,6 +650,27 @@
     required=False,
 )
 
+# Wayback Machine Settings
+ENABLE_WAYBACK_TASKS = get_bool(
```
umar8hassan, line 654: We should mention the Wayback related config in the readme

umar8hassan (changes requested):

main/settings.py
ibrahimjaved12: Mentioned here: https://github.com/mitodl/ocw-studio/pull/2326

main/settings.py
```diff
@@ -732,6 +768,12 @@
     },
 }
 
+if ENABLE_WAYBACK_TASKS:
```
umar8hassan, line 771: @ibrahimjaved12, as discussed, this feature flag in the environment requires restarting web workers. This kind of arrangement is not suitable for RC or Production. Please remove this check and always register the tasks.
You can control the logic of tasks to be empty with Posthog feature flag.

main/settings.py
```diff
@@ -732,6 +768,12 @@
...
+    CELERY_BEAT_SCHEDULE["update-wayback-jobs-status"] = {
+        "task": "external_resources.tasks.update_wayback_jobs_status_batch",
+        "schedule": UPDATE_WAYBACK_JOBS_STATUS_FREQUENCY,
+    }
+
 if CHECK_EXTERNAL_RESOURCE_TASK_ENABLE:
```
umar8hassan, line 777: This needs to be dealt with as well, but probably i a separate task along with changes required to convert it to a posthog feature flag

umar8hassan (changes requested):

external_resources/api.py
ibrahimjaved12: Removing this check.

ibrahimjaved12:
This PR is ready again! @umar8hassan 
I merged the branches https://github.com/mitodl/ocw-studio/pull/2323 https://github.com/mitodl/ocw-studio/pull/2320 into this branch to add PostHog checks to all relevant places at once.

umar8hassan (approved):
Looks Good

PR #2310, opened 2024-09-18
Save status codes for external resource link checking

pt2302:
Closes https://github.com/mitodl/hq/issues/5515.

This PR adds the new fields `url_status_code` and `backup_url_status_code` to the metadata for external resources, which are updated when the automated link checking task is done. This allows for more nuanced information about broken links status, and whether the link is actually broken.

### How can this be tested?
Spin up Django shell with `docker compose exec web ./manage.py shell`.

Then, run the following commands:

```
from external_resources.tasks import check_external_resources
from websites.models import WebsiteContent
from websites.constants import CONTENT_TYPE_EXTERNAL_RESOURCE
external_resources = WebsiteContent.objects.filter(type=CONTENT_TYPE_EXTERNAL_RESOURCE)
external_resource_ids = list(external_resources.values_list("id", flat=True))
check_external_resources(external_resource_ids) #this step will take a long time
```
After the link-checking completes, run

```
for resource in external_resources:
    print(resource.metadata)
```
and confirm that the new fields are in the metadata and look correct.

### DIFF
--- a/external_resources/constants.py
+++ b/external_resources/constants.py
@@ -26,6 +26,7 @@ RESOURCE_UNCHECKED_STATUSES = [
 EXTERNAL_RESOURCE_TASK_RATE_LIMIT = "100/s"
 EXTERNAL_RESOURCE_TASK_PRIORITY = 4  # Lowest priority from range (0 - 4)
 
+# constants for user agent
 USER_AGENT_STRING = (
     "Mozilla/5.0 (Windows NT 10.0; Win64; x64) "
     "AppleWebKit/537.36 (KHTML, like Gecko) "
@@ -34,3 +35,8 @@ USER_AGENT_STRING = (
 )
 
 USER_AGENT_TIMEOUT = 30
+
+# metadata fields
+METADATA_IS_BROKEN = "is_broken"
+METADATA_URL_STATUS_CODE = "url_status_code"
+METADATA_BACKUP_URL_STATUS_CODE = "backup_url_status_code"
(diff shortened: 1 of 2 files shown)


external_resources/tasks.py
```diff
@@ -58,6 +61,10 @@ def check_external_resources(resources: list[int]):
             log.debug(ex)
             state.status = ExternalResourceState.Status.CHECK_FAILED
         else:
+            # Update the metadata of the resource with the status codes
+            resource.metadata[METADATA_URL_STATUS_CODE] = url_status
+            resource.metadata[METADATA_BACKUP_URL_STATUS_CODE] = backup_url_status
```
umar8hassan, line 66: @pt2302 We have these status codes already stored in `WebsiteContent.external_resource_state.external_url_response_code` and `WebsiteContent.external_resource_state.backup_url_response_code`. Do we have a use case planned to duplicate these to `WebsiteContent.metadata`?

external_resources/constants.py
```diff
@@ -25,3 +25,18 @@
...
+
+# constants for user agent
+USER_AGENT_STRING = (
+    "Mozilla/5.0 (Windows NT 10.0; Win64; x64) "
+    "AppleWebKit/537.36 (KHTML, like Gecko) "
+    "Chrome/114.0.0.0 "
+    "Safari/537.36"
+)
+
+USER_AGENT_TIMEOUT = 30
```
umar8hassan, line 37: @pt2302 We have observed this timeout is not enough for some URLs. This should be ~60-90 sec. However, a longer timeout may have a direct impact on link checking. We should consider adding this timeout as an environment variable.

pdpinch: Let's address this in a follow-up PR so we can get this merged. 

I'm not sure I understand what you mean by "direct impact on link checking" but we could discuss that in a separate issue.

umar8hassan (approved):
I have tested the link-checking functionality in the local setup, and it looks good to me.

PR #2316, opened 2024-09-24
added page type in deletable items

umar8hassan:
Closes https://github.com/mitodl/hq/issues/5358

This PR allows users to delete a WebsiteContent only if it is not referenced in other content.

### How can this be tested?
1. Make sure you have permission to delete content (Your email is added to the allowed list on PostHog)
2. Go to any resource and check if the Delete menu is there.
3. Refer to the instructions [here](https://github.com/mitodl/ocw-studio/pull/2322) for populating referencing content fields.
4. When the content is being referenced elsewhere, deletion is prevented and it shows a warning message like in the screenshot below

Screenshot 2024-10-28 at 5 54 17 PM

### Additional Testing
Feature Flag testing with PostHog
1. Ensure you set `POSTHOG_ENABLED=True`, `POSTHOG_API_HOST`, and `POSTHOG_PROJECT_API_KEY` values in `.env` in your local `ocw-studio` setup.
2. Rebuild containers
3. The flag should already enabled on the `dev` environment. If needed, visit PostHog to configure the `OCW_STUDIO_CONTENT_DELETABLE_REFERENCES` flag.
4. With the flag disabled, the content is deleted even if it has references

### DIFF
--- a/main/posthog.py
+++ b/main/posthog.py
@@ -1,4 +1,4 @@
-from posthog import Posthog
+from posthog import Posthog, feature_enabled
 
 from main.settings import (
     ENVIRONMENT,
@@ -12,3 +12,8 @@ if POSTHOG_ENABLED:
 
     if ENVIRONMENT == "dev":
         posthog.debug = True
+
+
+def is_feature_enable(feature_key, user_id):
+    """Check for Feature flag"""
+    return POSTHOG_ENABLED and feature_enabled(feature_key, user_id)
--- a/main/posthog.py
+++ b/main/posthog.py
@@ -1,4 +1,4 @@
-from posthog import Posthog, feature_enabled
+from posthog import Posthog
 
 from main.settings import (
     ENVIRONMENT,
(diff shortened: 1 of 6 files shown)


ibrahimjaved12 (changes requested):
The functionality works as expected. 👍🏻

But can you please add the error message prior to clicking Delete button on Dialog.

umar8hassan:
> The functionality works as expected. 👍🏻
> 
> But can you please add the error message prior to clicking Delete button on Dialog.

@ibrahimjaved12. To show the error message before the delete dialogue, I would need to check if the content has any reference. I'd have to create another API that would be called by the `Delete` button from the vertical menu. That's an API per deletable content type. With this PR, we have three deletable content types.

I would suggest we log this change request in a separate ticket. @pdpinch can triage it.

pdpinch:
I agree with Umar. But is this behind a flag so we can merge but control the roll out?

umar8hassan:
> I agree with Umar. But is this behind a flag so we can merge but control the roll out?

This is an addition to the existing delete feature, which is already behind a feature flag. Should we add a separate flag to check the references before delete?

ibrahimjaved12 (approved):
Looks good to me

PR #2322, opened 2024-10-11
Add referencing_pages to external resources

pt2302:
Part of https://github.com/mitodl/hq/issues/5736. Adding `referencing_pages` for newly-created or newly-linked external resources will be handled in a follow-up PR.

This PR adds the `referencing_pages` field to the WebsiteContent Django model, and also updates the parsing rules for legacy links.

### How can this be tested?

Run 

`docker compose exec web ./manage.py markdown_cleanup link_to_external_resource --skip-sync --commit --filter <course ID>`

and 

`docker compose exec web ./manage.py markdown_cleanup nav_item_to_external_resource --skip-sync --commit --filter <course ID>`

and verify that the `referencing_pages` in Django admin looks as expected.


ibrahimjaved12 (approved):
I tested the functionality in different contexts, and it works as expected:
image
image
image
image
image

The `referencing_pages` field was auto-populated for these WC along with the creation of ERs.

However, I do have a suggestion regarding the field name `referencing_pages`:

While I understand that the original use case focused on pages, and pages are essentially a type of WebsiteContent. In the current implementation, we also handle `navmenu` entries. The `referencing_pages` field applies to both new and existing content.

Given this broader context, naming the field `referencing_pages` feels somewhat restrictive, especially since ERs can be referenced by Video Galleries, Resource Links, and other types of content. A more general-purpose name like `referencing_contents` (with updated help text) would reflect its usage and make it clearer that it is not limited to pages alone.

What do you think about this adjustment?

pdpinch:
> What do you think about this adjustment?

That would be great. 

@ibrahimjaved12 or @umar8hassan would you be able to take over the PR since Paul is on vacation now?

ibrahimjaved12:
@umar8hassan could you take over this task? It will also require a migration, and since you're already handling migrations for the Wayback-related work, you can streamline both as you see fit.

umar8hassan:
Tested PR after updating field name:
Screenshot 2024-10-18 at 5 51 58 PM
Screenshot 2024-10-18 at 5 52 18 PM
Screenshot 2024-10-18 at 5 39 03 PM
Screenshot 2024-10-18 at 6 00 44 PM

PR #2326, opened 2024-10-15
Update Documentation for External Resources Model with Wayback Machine and API Changes

ibrahimjaved12:
Closes https://github.com/mitodl/hq/issues/5835

This PR updates the README for the External Resources model to reflect the newly added functionalities, including:

- Integration of Wayback Machine API for URL submissions.
- Task scheduling and submission frequency controls.
- Updates to the metadata migration process for ExternalResourceState, including the removal of `backup_url` and related fields, and the addition of `wayback_url`, `wayback_status`, and other Wayback Machine-related fields.
- Management commands for:
  - Submitting external resources to the Wayback Machine.
  - Updating the status of Wayback Machine jobs.
- Detailed workflow of checking external resource URLs, and incorporating the Wayback Machine integration.

### Reviewing
Click on '[View File](https://github.com/mitodl/ocw-studio/blob/90d4c7dd425c9520bda6dc0c17656f698bd5ae04/external_resources/README.md)' in Files changed to see the markdown rendered in its formatted view. [Or click on "Display the rich diff"](https://github.com/mitodl/ocw-studio/pull/2326/files?short_path=89699bf#diff-89699bfca8c2e9c50b01a40afdd5e7d2e8b84867d7497e3138e09027658f2a4e).

### DIFF
--- a/external_resources/README.md
+++ b/external_resources/README.md
@@ -1,48 +1,173 @@
 # External Resource Availability Workflow
 
-This document describes the workflow for the external resources validation tasks.
+This document describes the workflow for validating external resources (link checking) and integrating with the Internet Archive's Wayback Machine.
 
 **SECTIONS**
 
 1. [Overview](#overview)
-1. [Enabling Task](#enabling-task)
+1. [Enabling Tasks](#enabling-tasks)
+1. [Wayback Machine Integration](#wayback-machine-integration-1)
+1. [Wayback Machine Removal Requests](#wayback-machine-removal-requests)
 1. [Frequency Control](#frequency-control)
 1. [Rate Limiting](#rate-limiting)
 1. [Task Priority](#task-priority)
-
+1. [Management Commands](#management-commands)
+1. [Code References](#code-references)
 
 # Overview
 
-This assumes that celery beat scheduler is installed and enabled, which is required for the task scheduling.
+This assumes that **Celery beat scheduler** is installed and enabled, which is required for the task scheduling.
 
-Frequency for the task is set to `1/week`. After each week, all external resources, new or existing, will be validated regardless of their last status.
+The frequency for **external resource checking** is set to once per week (`1/week`), as defined by the `CHECK_EXTERNAL_RESOURCE_STATUS_FREQUENCY` variable. All external resources, both new and existing, are validated weekly, regardless of their last status.
 
-The high-level description of the process is below, and each subsequent section contains additional details, including links to the relevant code.
+The **Wayback Machine integration** is enabled by default. Valid external resources are submitted for archiving as part of the external resource validation task. The system monitors the submission status and avoids resubmitting resources within a specified interval of `30` days, as defined by the `WAYBACK_SUBMISSION_INTERVAL_DAYS` variable.
 
-* Task is automatically added in scheduler on system start.
-* On execution, all available external resources are retrieved from DB.
-* Gathered data is divided into preconfigured batch sizes.
-* All batches are grouped into a single celery task and executed.
-* Each batch-task iterates over batch to validate availability of each resource and its backup resource if available.
-* The status of resource is then added to DB.
-* Batch tasks have a preconfigured rate-limiter and lower priority by default.
+The status of Wayback Machine archiving jobs is updated every `6` hours to track the progress of pending jobs, as defined by the `UPDATE_WAYBACK_JOBS_STATUS_FREQUENCY` variable.
 
+These variables can be found in [settings.py](/main/settings.py).
 
-## Enabling Task
-The task for external resource checking can be enabled/disabled using the `CHECK_EXTERNAL_RESOURCE_TASK_ENABLE` defined in [here](/main/settings.py). However, once scheduled, the task can be removed only if `Celery` is restarted along with toggling `CHECK_EXTERNAL_RESOURCE_TASK_ENABLE` to `False`.
+If a resource is valid, the `check_external_resources` task triggers the `submit_url_to_wayback_task` to archive it.
 
-## Frequency Control
+Here’s a high-level description of the process:
 
-The task frequency (in seconds) is set using the `CHECK_EXTERNAL_RESOURCE_STATUS_FREQUENCY` in [here](/main/settings.py). Default value for the frequency is set to `604800 seconds -> 1 week`.
+### External Resource Validation:
 
+- Task is automatically added in scheduler on system start.
+- On execution, all available external resources are retrieved from DB.
+- Gathered data is divided into preconfigured batch sizes, defined by `BATCH_SIZE_EXTERNAL_RESOURCE_STATUS_CHECK` in [websites/constants.py](/websites/constants.py), which is currently set to `100`.
+- All batches are grouped into a single Celery task and executed.
+- Each batch-task iterates over the batch to validate availability of each resource and its backup resource if available.
+- The status of resource is then added to DB.
+- Batch tasks have a preconfigured rate limiter and lower priority by default.
 
-## Rate Limiting
+### Wayback Machine Integration:
 
-The rate-limit for the external resource batch-tasks is set using `EXTERNAL_RESOURCE_TASK_RATE_LIMIT` in [here](constants.py). The assigned value for the rate-limiter is set to `100/s`.
+- When external resource validation occurs, valid external resources are submitted to the Wayback Machine for archiving.
+- The status of Wayback Machine archiving jobs is tracked and updated periodically (currently, every 6 hours).
+- New external resources are automatically submitted to the Wayback Machine upon creation.
 
+# Enabling Tasks
 
-## Task Priority
+### External Resource Checking:
 
-Batch-task priority is set using the `EXTERNAL_RESOURCE_TASK_PRIORITY` in [here](constants.py). The default priority for each celery task has been preconfigured to `2` out of range `0(lowest) - 4(highest)`. External resource tasks have lowest (`0`) priority by default.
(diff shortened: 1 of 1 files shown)


external_resources/README.md
```diff
@@ -1,48 +1,162 @@
 # External Resource Availability Workflow
 
-This document describes the workflow for the external resources validation tasks.
+This document describes the workflow for validating external resources and integrating with the Wayback Machine.
```
pdpinch, line 3:
```suggestion
This document describes the workflow for validating external resources (link checking) and integrating with the Internet Archive's Wayback Machine.
```

external_resources/README.md
```diff
@@ -1,48 +1,162 @@
...
+1. [Code References](#code-references)
 
 # Overview
 
-This assumes that celery beat scheduler is installed and enabled, which is required for the task scheduling.
+This assumes that **Celery beat scheduler** is installed and enabled, which is required for the task scheduling.
+
+The frequency for **external resource checking** is set to once per week (`1/week`). All external resources, both new and existing, are validated weekly, regardless of their last status.
+
+The **Wayback Machine integration** is enabled by default. Valid external resources are submitted for archiving as part of the external resource validation task. The system monitors the submission status and avoids resubmitting resources within a specified interval. The status of Wayback Machine archiving jobs is updated every 6 hours to track the progress of pending jobs.
```
pdpinch, line 23: > within a specified interval.

Where is this specified?

external_resources/README.md
```diff
@@ -1,48 +1,162 @@
...
+
+Here’s a high-level description of the process:
+
+### External Resource Validation:
+
+- Task is automatically added in scheduler on system start.
+- On execution, all available external resources are retrieved from DB.
+- Gathered data is divided into preconfigured batch sizes, defined by `BATCH_SIZE_EXTERNAL_RESOURCE_STATUS_CHECK` in [websites/constants.py](/websites/constants.py), which is currently set to `100`.
+- All batches are grouped into a single Celery task and executed.
+- Each batch-task iterates over batch to validate availability of each resource and its backup resource if available.
```
pdpinch, line 35:
```suggestion
- Each batch-task iterates over the batch to validate availability of each resource and its backup resource if available.
```

external_resources/README.md
```diff
@@ -1,48 +1,162 @@
...
+- The status of resource is then added to DB.
+- Batch tasks have a preconfigured rate limiter and lower priority by default.
+
+### Wayback Machine Integration:
+
+- When external resource validation occurs, valid external resources are submitted to the Wayback Machine for archiving.
```
pdpinch, line 41: Does this mean that we could possibly submit _all_ the external resources to the Wayback Machine _every_ week?

external_resources/README.md
```diff
@@ -1,48 +1,162 @@
...
+
+# Enabling Tasks
+
+### External Resource Checking:
+
+The task for external resource checking can be enabled/disabled using the `CHECK_EXTERNAL_RESOURCE_TASK_ENABLE` defined in [settings.py](/main/settings.py). By default, this task is enabled.
+
+### Wayback Machine Tasks:
+
+Wayback Machine tasks can be enabled or disabled using the `ENABLE_WAYBACK_TASKS` defined in [settings.py](/main/settings.py). These tasks are also enabled by default.
```
pdpinch, line 53: Maybe too late for this comment, but I wish the naming scheme for `CHECK_EXTERNAL_RESOURCE_TASK_ENABLE` and `ENABLE_WAYBACK_TASKS` were more similar. 

i.e. 
`CHECK_EXTERNAL_RESOURCE_TASK_ENABLE` and `WAYBACK_TASKS_ENABLE`

or 
`ENABLE_CHECK_EXTERNAL_RESOURCE_TASK` and `ENABLE_WAYBACK_TASKS`

external_resources/README.md
```diff
@@ -1,48 +1,162 @@
...
+
+Note: Changes to these settings require restarting Celery workers to take effect.
```
pdpinch, line 55: will this still be true after we move to manage the settings in Posthog? (this is a question, not a request)

pdpinch:
I know we think of checking external resources and submitting external resources to the Wayback machine as related tasks, but I'm wondering if it might be clearer to separate the documentation. I'm not sure. It's just a question.

external_resources/README.md
```diff
@@ -1,48 +1,162 @@
...
+- **Submit Valid URLs:**
+  - After an external resource URL is validated and found to be valid, it is submitted to the Wayback Machine for archiving.
+  - The submission is handled by the `submit_url_to_wayback_task` task.
+- **Automatically Submit on Creation:**
+  - When a new external resource is created, it is automatically submitted to the Wayback Machine via the Django signal in [signals.py](/external_resources/signals.py).
+- **Track Job Statuses:**
+  - The status of Wayback Machine archiving jobs is tracked using the `wayback_status` field in the `ExternalResourceState` model.
+  - The system periodically updates the status of pending jobs using the `update_wayback_jobs_status_batch` task at an interval set by `UPDATE_WAYBACK_JOBS_STATUS_FREQUENCY` in [settings.py](/main/settings.py). Currently, it is set to `6` hours (21600 seconds).
+- **Control Resubmissions:**
+  - The system avoids resubmitting the same URL to the Wayback Machine within a specified interval, defined by `WAYBACK_SUBMISSION_INTERVAL_DAYS` in [settings.py](/main/settings.py).
```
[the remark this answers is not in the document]
ibrahimjaved12: It's currently set to `30` days

external_resources/README.md
ibrahimjaved12: I'm going to change `CHECK_EXTERNAL_RESOURCE_TASK_ENABLE` into `ENABLE_CHECK_EXTERNAL_RESOURCE_TASK`

external_resources/README.md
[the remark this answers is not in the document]
pdpinch: Sounds like a good default.

external_resources/README.md
ibrahimjaved12: We could technically submit all the external resources every week, but we are not doing this right now, as the `30` days `WAYBACK_SUBMISSION_INTERVAL_DAYS` variable prevents this.

Also, even though the link checking task is set to occur every week, its priority is lowest (`4`). And the `submit_url_to_wayback_task` function it calls, its priority is also low (`3`). So they may take much longer before the ERs actually finish execution.

external_resources/README.md
ibrahimjaved12: This is now only partly true, so I'm updating the documentation to reflect the changes. Wayback tasks are now governed by PostHog feature flags. However, the `ENABLE_WAYBACK_TASKS` setting in `settings.py` still plays a role:

```
ENABLE_WAYBACK_TASKS = get_bool(
    name="ENABLE_WAYBACK_TASKS",
    default=True,
    description="Enables tasks related to Wayback Machine submissions and status checks",  # noqa: E501
    required=False,
)
if ENABLE_WAYBACK_TASKS:
    CELERY_BEAT_SCHEDULE["update-wayback-jobs-status"] = {
        "task": "external_resources.tasks.update_wayback_jobs_status_batch",
        "schedule": UPDATE_WAYBACK_JOBS_STATUS_FREQUENCY,
    }
```

This variable is responsible for adding the `update-wayback-jobs-status` task to the Celery beat schedule. The reason for this is, we're using PostHog feature flags for Runtime modifications. But this is the kind of modification that would require a restart. So PostHog feature flags would not help with that.

However, from a single-source-of-truth perspective, we could raise the question of why we are not still using the PostHog feature flag for this.  Using a PostHog feature flag directly in `settings.py` would result in circular import issues under the current `posthog.py` structure, making `ENABLE_WAYBACK_TASKS` a separate, necessary source of truth -- for now.

ibrahimjaved12:
> I know we think of checking external resources and submitting external resources to the Wayback machine as related tasks, but I'm wondering if it might be clearer to separate the documentation. I'm not sure. It's just a question.

That is a valid point. Right now, the documentation is kind of written from a perspective of Celery tasks for ERs. Eg. we have heading sections such as: Frequency Control, Rate Limiting, Task Priority etc.

This approach groups the different functionalities (external resource checking and Wayback Machine submissions) together because they share similar Celery mechanics. While this makes it easier to understand the overall workflow and see how the components interact, it may slightly hinder the comprehension of individual functionalities.

Pro:
 - Overall flow easier to understand, get a bigger picture of how things are working + interacting, plus Code References at once which I think also helps

Con:
 - Someone wanting to learn about only one functionality, additional details about the other may be distracting or set someone a bit off-track

pdpinch:
> While this makes it easier to understand the overall workflow and see how the components interact, it may slightly hinder the comprehension of individual functionalities.

Ok, I'm convinced. Let's keep it as is.

external_resources/README.md
```diff
@@ -1,48 +1,168 @@
...
+
+### External Resource Checking:
+
+The task for external resource checking can be enabled/disabled using the `ENABLE_CHECK_EXTERNAL_RESOURCE_TASK` defined in [settings.py](/main/settings.py). By default, this task is enabled.
+
+### Wayback Machine Tasks:
+
+Wayback Machine tasks are governed by PostHog feature flags, with the feature flag key `OCW_STUDIO_WAYBACK_MACHINE_TASKS` defined in `ENABLE_WAYBACK_TASKS` in [constants.py](/external_resources/constants.py). In addition to this, the `ENABLE_WAYBACK_TASKS` setting in [settings.py](/main/settings.py) is responsible for adding the `update_wayback_jobs_status_batch` task to the Celery beat schedule.
+
+Note: Changes to PostHog feature flags take effect immediately, but modifications to Celery beat schedules still require a restart.
```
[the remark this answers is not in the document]
ibrahimjaved12: Done

pdpinch (approved):
LGTM!

PR #2332, opened 2024-10-25
populated referencing content field for the website content

umar8hassan:
closes https://github.com/mitodl/hq/issues/5851

This adds the references of the website contents that are using this WebsiteContent

### How can this be tested?
1. Open `ocw-studio`
2. Add a new page or edit an existing one
3. From resource picker, add a link for an external resource or page.
Screenshot 2024-10-28 at 2 13 50 PM

4. Go to the `Admin` panel and check the `referencing_content` field for the selected resource. it should have the reference of the page where the link was added.
5. Type links in editor which would create External Resources with page reference (refer [here](https://github.com/mitodl/ocw-studio/pull/2279) for details on testing this feature)

### DIFF
--- a/static/js/components/widgets/MarkdownEditor.tsx
+++ b/static/js/components/widgets/MarkdownEditor.tsx
@@ -31,7 +31,6 @@ import { useWebsite } from "../../context/Website"
 import { siteContentRerouteUrl } from "../../lib/urls"
 import { checkFeatureFlag } from "../../lib/util"
 import CustomLink from "../../lib/ckeditor/plugins/CustomLink"
-import { useParams } from "react-router"
 import { useReferences } from "../../context/References"
 
 export interface Props {
@@ -56,7 +55,6 @@ export default function MarkdownEditor(props: Props): JSX.Element {
   const { link, embed, value, name, onChange, minimal, allowedHtml } = props
   const throwSynchronously = useThrowSynchronously()
   const website = useWebsite()
-  const { uuid } = useParams<{ uuid: string }>()
   const [isCustomLinkUIEnabled, setIsCustomLinkUIEnabled] = useState(false)
   const { addReferences, removeReferences } = useReferences()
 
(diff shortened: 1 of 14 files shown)


static/js/components/widgets/MarkdownEditor.tsx
```diff
@@ -53,6 +55,7 @@ export default function MarkdownEditor(props: Props): JSX.Element {
   const { link, embed, value, name, onChange, minimal, allowedHtml } = props
   const throwSynchronously = useThrowSynchronously()
   const website = useWebsite()
+  const { uuid } = useParams<{ uuid: string }>()
```
ibrahimjaved12, line 58: The `uuid` isn't available for `useParams` in the context of a new `WebsiteContent` creation (eg. page). This means, if any ER is created or added to a page at the time of its creation, it won't be referenced currently.

As we address this, we should also ensure that it's a general purpose solution applicable for other `WebsiteContent` types as well where we may be able to use ERs.

ibrahimjaved12 (changes requested):
We also need to look into dereferencing. Under current implementation, it's only possible via Django admin.

When we remove the ER from the specific `WebsiteContent`, it should be removed from `Referencing content`.

ibrahimjaved12 (changes requested):
The functionality is working well. 👍🏻 However, could you please address two things:
1. `Console.log()` statements present in the code
2. Github Action warnings

ibrahimjaved12 (approved):
👍🏻

umar8hassan:
This would require a follow-up issue to handle the case of multiple references of a single content on a page.

#### Cases of Multiple References of Content on a Page
| Case | Occurrence | Implication | Included in this PR |
|--------|--------|--------|--------|
| 1. | The page added two references of the same content. | The reference will be added only once as duplicates are not allowed in the DB relation | Yes |
| 2. | The page removed two references of the same content. | It will remove the reference only once since a duplicate cannot be added following `Case 1` | Yes |
| 3. | The page added and removed a reference to the same content in a single update. | It will not change anything since changes will cancel out each other in the `Changeset Stack`. | Yes |
| 4. | The page removed one of the two references to content. | Since a duplicate following `Case 1` cannot be added, it will remove the only existing reference. The other reference won't have an entry in the DB. Thus, the content cannot prevent the `deletion` or any other operation. | No |
| 5. | The page removed the second reference from `Case 4` in the next update. | The attempt may fail since the only entry was removed following `Case 4`. | No |

PR #2334, opened 2024-10-28
Fix file size metadata mismatch when replacing Google Drive image in Studio

ibrahimjaved12:
Closes https://github.com/mitodl/hq/issues/5602

This PR addresses the issue where the `file_size` metadata of a `WebsiteContent` resource in Studio does not update correctly when an image is replaced on Google Drive and resynced.

### How can this be tested?
1. Checkout to this branch
2. Spin up OCW Studio locally
3. Upload an image to a course's Gdrive and sync
4. Go to admin panel, open the relevant WebsiteContent object, and check `file_size` from metadata.
5. Go to course's Gdrive again, delete the picture.
8. Upload any other picture there but with same name.
9. Sync with Gdrive again.
10. Go to admin panel, open the relevant WebsiteContent object, and check that `file_size` in metadata has successfully updated, while the other metadata fields stay same.

### DIFF
--- a/gdrive_sync/api.py
+++ b/gdrive_sync/api.py
@@ -487,7 +487,7 @@ def create_gdrive_resource_content(drive_file: DriveFile):
             )
         else:
             resource.file = drive_file.s3_key
-            if resource.metadata["file_size"] != drive_file.size:
+            if resource.metadata.get("file_size") != drive_file.size:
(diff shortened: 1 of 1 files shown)


umar8hassan (approved):
Looks good 👍

PR #2335, opened 2024-10-30
Add environment to PostHog feature flags

pt2302:
Closes https://github.com/mitodl/hq/issues/5910.

This PR adds the `environment` variable to users for PostHog, allowing feature flags to be rolled out for certain values of the environment variable, such as only on RC.

### How can this be tested?

First, set the variable `OCW_STUDIO_ENVIRONMENT` to `rc` in your `.env` file. Then, restart the Docker OCW Studio containers. Verify that the behavior described in the testing instructions for https://github.com/mitodl/ocw-studio/pull/2279 is valid. Now, change the value of `OCW_STUDIO_ENVIRONMENT` to any other value (such as `prod`). Restart the Studio containers. Verify that links are now editable as before.


umar8hassan (approved):
Looks Good

PR #2313, opened 2024-09-19
Fix TypeError for external resources link checking task

pt2302:
### What are the relevant tickets?
Closes https://github.com/mitodl/ocw-studio/issues/2312.

### Description (What does it do?)
This PR fixes a TypeError in the external resources link checking Celery task.

### How can this be tested?
Execute the following in Django shell:

```
from external_resources.tasks import check_external_resources_for_breakages
results = check_external_resources_for_breakages.apply()
```

Verify that the TypeError described in the issue does not appear.

PR #2350, opened 2024-11-20
Use return Task.replace instead of raise Task.replace for Celery tasks

pt2302:
Closes https://github.com/mitodl/hq/issues/5789.

This PR updates all of the OCW Studio Celery tasks to use `return Task.replace` instead of `raise Task.replace`, consistent with the current Celery API. Related previously-merged PR: https://github.com/mitodl/ocw-studio/pull/2313.

### How can this be tested?
Smoke-test the tasks that have been updated, especially Google Drive sync and video import. Ensure that the tasks are working properly.


main/settings.py
```diff
@@ -26,7 +26,7 @@
 
 # pylint: disable=too-many-lines
 
-VERSION = "0.132.1"
+VERSION = "0.133.0"
```
mbertrand, line 29: A rebase should clear out this difference

mbertrand (approved):
👍

PR #2353, opened 2024-12-07
Add function for listing YouTube caption tracks

pt2302:
Part of https://github.com/mitodl/hq/issues/1770.

This PR adds a function to the `YouTubeApi` class for querying YouTube and getting a list of all of the captions tracks for all of the videos in the Django database. The purpose of this is to inspect the captions that are currently on YouTube and check for duplicate caption tracks.

### How can this be tested?
1. Make sure that YouTube credentials for RC are set up in the `.env` file. 
2. [Optional] Add a line similar to `video_files = video_files[:20]` after the definition of `video_files` in the `get_all_video_captions` function to restrict this to some subset of video files in order to speed up testing.
3. Run `docker compose up` to spin up OCW Studio containers. 
4. Run `docker compose exec web python ./manage.py shell` to open the Django shell.
5. Execute the following in the Django shell:
```
from videos.youtube import YouTubeApi
captions = YouTubeApi.get_all_video_captions()
```
6. Inspect `captions` and verify that there is information about the captions tracks available.

### DIFF
--- a/videos/youtube.py
+++ b/videos/youtube.py
@@ -4,6 +4,7 @@ import http
 import logging
 import re
 import time
+from collections import Counter
 from io import BytesIO
 from urllib.parse import urljoin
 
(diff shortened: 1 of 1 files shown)


mbertrand (approved):
Works as intended.  👍  assuming this will be run from a shell and doesn't need a management command.

PR #2370, opened 2025-01-01
Add Video Resource Functionality

ibrahimjaved12:
Closes https://github.com/mitodl/hq/issues/4709

This PR adds functionality to add Video Resources using YouTube ID.

Functional changes introduced:
- Video thumbnail field is auto-populated using YouTube ID (if empty) on Save.
- In case of adding (as in the case of Add Video Resource Button), "File" and "Resource Type" will be hidden, defaulting to "Video" for Resource Type.
- Existing editing behavior for resources will be same, that is, File and Resource Type will not be hidden or affected. 
- These changes are managed through the PostHog Feature Flag `OCW_STUDIO_ADD_VIDEO_RESOURCE`.

image
image

### How can this be tested?
1. Checkout to this branch.
2. Spin up OCW Studio locally.
3. Make sure the `OCW_STUDIO_ADD_VIDEO_RESOURCE` feature flag is enabled in PostHog's web interface.
4. Go to `ocw-www` or any course site locally and go to resources.
5. Click on the new "Add Video Resource" button.
6. Add a title and YouTube ID, and any other fields you like. Save.
7. Embed the video in any page and make sure it works fine -- like a regular video resource.

### DIFF
--- a/static/js/components/RepeatableContentListing.test.tsx
+++ b/static/js/components/RepeatableContentListing.test.tsx
@@ -164,7 +164,7 @@ describe("RepeatableContentListing", () => {
       const addLink = wrapper.find("a.add")
       expect(driveLink.exists()).toBe(isGdriveEnabled && isResource)
       expect(syncLink.exists()).toBe(isGdriveEnabled && isResource)
-      expect(addLink.exists()).toBe(!isGdriveEnabled || !isResource)
+      expect(addLink.exists()).toBe(true)
     },
   )
 
@@ -489,8 +489,12 @@ describe("RepeatableContentListing", () => {
       }
       const { wrapper } = await render()
       expect(wrapper.find("h2").text()).toBe(configItem.label)
-      const link = wrapper.find(".cyan-button .add").at(1)
-      expect(link.text()).toBe(`Add ${expectedLabel}`)
+      const link = wrapper.find(".cyan-button .add").at(0)
+      if (configItem.name === "resource") {
+        expect(link.text()).toBe("Add Video Resource")
+      } else {
+        expect(link.text()).toBe(`Add ${expectedLabel}`)
+      }
       expect(link.prop("href")).toBe(
         siteContentNewUrl
           .param({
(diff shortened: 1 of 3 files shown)


umar8hassan:
@ibrahimjaved12 the flow works as expected. Can you please elaborate on how do we plan to add Video captions and sync in this workflow??

ibrahimjaved12:
@pdpinch 

> Can you please elaborate on how do we plan to add Video captions and sync in this workflow??

Essentially all we need are links to populate the captions/transcripts in metadata. Using Gdrive flow, the link takes the form: `/courses/<course_name>/<filename>.extension`

Under the Gdrive video workflow, we have the option of uploading captions/transcripts using filename such as `<video_name_in_drive>_captions.vtt` which are then linked with the video upon creation.

I was thinking we could do something similar here but using `youtube_id` instead (because title could have spaces or something else problematic). So it would become: `youtubeID_captions.vtt` etc. 

Workflow:
They would be uploaded to Gdrive. When we are creating/editing a Video Resource, on clicking Save -- it would check if the video resource has blank captions/transcripts then proceed to query Gdrive if something is present there for the YouTube ID. If found, populate the field, like we populate the thumbnails.

That's the simpler way we can go about this. Or should I research if 3play could be made a part of the workflow?
What do you think of this.

pdpinch:
I talked to Jen about this a bit today. She's going to follow up with Jason, but I think this is the way it will play out:

There are two different cases for adding existing videos. 

1. Sometimes we are adding videos that are already on the OCW YouTube channel. A good example of this is the podcast "videos" that are now posted on YouTube. In these cases, there will most likely be 3Play captions. It would be desirable to get the caption files from 3Play. 

2. Sometimes we are adding videos that are on someone else's channel. Most likely these will not have 3Play captions (although I suppose it is possible). In these cases, if we have captions to add, they will have to be added via gdrive, similar to how we handle caption edits now. 

@ibrahimjaved12 (and @pt2302), can you think of a workflow that might work with both of these options?

In any case, I think this can be a follow up PR. Even if the OCW editorial policy is "don't publish without captions" we can add the functionality and count on them not to use it until the captions are worked out.

ibrahimjaved12:
I have rebased the PR to simplify `SiteContentForm.tsx`. Could you take another look at it? @umar8hassan

umar8hassan (changes requested):
The PR works as expected but the form shows the extra Label for `File`
Screenshot 2025-01-14 at 6 40 48 PM

umar8hassan:
> The PR works as expected but the form shows the extra Label for `File` Screenshot 2025-01-14 at 6 40 48 PM

The behaviour pre-exists in the edit form, which makes it outside the scope of this ticket. I have reported this in a separate issue https://github.com/mitodl/hq/issues/6465

umar8hassan (approved):
Looks Good 👍

PR #2373, opened 2025-01-07
Remove "name" field from the metadata of stories in WebsiteContent

ibrahimjaved12:
Part of https://github.com/mitodl/hq/issues/6320

This PR introduces a manual migration to remove the name field from the metadata of stories in the WebsiteContent model. 

### How can this be tested?
Checkout to this branch, spin up OCW Studio locally, jump into the Web container in Docker and run: `python manage.py migrate`. Make sure `name` is no longer present in metadata.


umar8hassan (changes requested):
@ibrahimjaved12, the migration works but we should include an empty rollback with a warning message just in case.

umar8hassan (approved):
Looks great 👍

PR #2378, opened 2025-01-16
added management command to detect unrelated files in a website

umar8hassan:
Closes https://github.com/mitodl/hq/issues/6457

A management to check for unrelated files in a website.

### How can this be tested?
1. Switch to this branch and spin up `ocw-studio` in your local environment.
2. Create a `test` course in your local ocw-studio or select an existing one.
3. Got to your local instance of s3 (MinIO) which should at `localhost:9001` and log in.
5. Upload a file at `{AWS_STORAGE_BUCKET_NAME}/courses/{selected_course_name}/`
6. run `docker-compose exec -it web ./manage.py detect_unrelated_content` in terminal. This command should return the uploaded file name against the course.

### DIFF
--- a/websites/management/commands/detect_unrelated_content.py
+++ b/websites/management/commands/detect_unrelated_content.py
@@ -63,21 +63,21 @@ class Command(WebsiteFilterCommand):
                 unrelated_files[website.name] = unrelated_website_files
 
         if unrelated_files:
+            self.style.SUCCESS("Unrelated content found in the bucket!")
             content = json.dumps(unrelated_files, indent=4)
 
-            with NamedTemporaryFile(delete=False, suffix=".json") as tmp_file:
-                tmp_file_path = tmp_file.name
             if total_files > self.UNRELATED_FILES_THRESHOLD:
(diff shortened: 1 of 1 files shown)


ibrahimjaved12 (changes requested):
Currently we're getting a bunch of false-positives because of path mismatch. The S3 key does not have a leading `/`, while the WebsiteContent.file field does. This causes a mismatch during the comparison.

As a result, for the courses I tested, one of which is `14.770-fall-2017`, if you download it, there are 412 items in `static_resources`, and in the report, I was getting 410 files, whereas we see from Django admin there are many `WebsiteContent` objects present for resources.

I think we have to normalize the S3 keys by adding a leading `/`:
```
            files_in_s3 = (
                {f"/{content['Key']}" for content in files.get("Contents", [])}
                if files.get("KeyCount")
                else set()
            )
```

ibrahimjaved12 (approved):
Now I am getting 368 files for the `14.770-fall-2017` course, which is consistent with the expected output.

Looks good to me!

PR #2320, opened 2024-10-09
Add management commands for submitting and checking Wayback Machine resources

ibrahimjaved12:
### What are the relevant tickets?
Closes https://github.com/mitodl/hq/issues/5779

### Description (What does it do?)
This PR introduces management commands for handling Wayback Machine submissions and status updates for external resources.

* `submit_sites_to_wayback.py`: to submit external resources of websites to the Wayback Machine, with an option to force submission (`--force`) (skipping `WAYBACK_SUBMISSION_INTERVAL_DAYS` defined in `settings.py`)
* `update_wayback_status.py`: to update the status of pending Wayback Machine jobs, with an option to run synchronously (`--sync`)

The `--sync` argument, however, will only work if the filter for sites is given. By default, the behavior is async.
Note: without `--filter`, the task will run for all sites.

### How can this be tested?
First, you will need to follow these testing steps coming from the branch PR `ibrahim/5355-wayback-integration-models-tasks` https://github.com/mitodl/ocw-studio/pull/2308. The steps are:
1. DM me for `access_key` and `secret_key`.
2. Checkout to the `ocw-hugo-projects` branch [in the PR](https://github.com/mitodl/ocw-hugo-projects/pull/307) and copy the code for `ocw-www/ocw-studio.yaml` and `ocw-course-v2/ocw-studio.yaml` and update your Starters configuration for both via Django admin.
3. Run the migration: `docker-compose exec web ./manage.py migrate`

Run in web container and monitor `celery` logs: 
```
1. ./manage.py submit_sites_to_wayback --filter "course1" --force
2. ./manage.py update_wayback_status --filter "course1"
3. ./manage.py update_wayback_status --filter "course1" --sync
4. ./manage.py update_wayback_status --filter "course1, course2"
5. ./manage.py submit_sites_to_wayback --filter "course1"

```

PR #2379, opened 2025-01-16
Introduce Environment Variable to Control Wayback Machine Tasks

ibrahimjaved12:
Closes https://github.com/mitodl/hq/issues/6412

This PR ensures that Wayback Machine submissions are controlled by both `settings.ENABLE_WAYBACK_TASKS` and PostHog Feature Flag. Also updates documentation pertaining to this.

### How can this be tested?
The wayback functionalities will only work if both of these above flags are enabled.

1. Make sure PostHog Feature Flag is enabled for your user or `dev` while testing.
2. Checkout to this branch.
3. Spin up OCW Studio locally.
4. Create a new ER and immediately check Celery logs. You will see that there is no log that says "`Submitting URL to Wayback Machine for resource...`"
5. Run a management command from https://github.com/mitodl/ocw-studio/pull/2320. You will get the error: `Wayback Machine tasks are disabled via environment settings or PostHog feature flag`.
6. Now add `ENABLE_WAYBACK_TASKS=True` in your `.env`.
7. Restart OCW Studio.
8. Run steps 4-5 again, the tasks will execute this time.


umar8hassan (approved):
Looks Good 👍

PR #2407, opened 2025-01-30
added hover message and new statuses for the site listing page

umar8hassan:
closes https://github.com/mitodl/hq/issues/6017

Add the statuses as mentioned on the [ticket](https://github.com/mitodl/hq/issues/6017) to the Sites Listing Page and incorporate feedback from [here](https://github.com/mitodl/ocw-studio/pull/2407#issuecomment-2647798005)

### How can this be tested?
1. Switch to this branch
2. Go to site listings page on `http://localhost:8043/sites/`
3. Create a new site or use an existing one to publish on live.
4. Check that existing sites have the appropriate statuses shown as below
Screenshot 2025-01-30 at 10 56 41 AM
5. Also check hover is diplayed for statuses in `PublishStatus` enum

Note: Status is updated according to Live Publish Status

### DIFF
--- a/static/js/pages/SitesDashboard.test.tsx
+++ b/static/js/pages/SitesDashboard.test.tsx
@@ -196,7 +196,7 @@ describe("StatusWithDateHover component", () => {
       />,
     )
 
-    const element = container.firstChild
+    const element = container.firstChild as HTMLElement
     fireEvent.mouseEnter(element)
 
     expect(getByText(/Published at \(Jan 15, 2023/)).toBeInTheDocument()
@@ -211,7 +211,7 @@ describe("StatusWithDateHover component", () => {
       />,
     )
 
-    const element = container.firstChild
+    const element = container.firstChild as HTMLElement
 
     // Hover
     fireEvent.mouseEnter(element)
@@ -236,7 +236,7 @@ describe("StatusWithDateHover component", () => {
 })
 
 describe("formatDateTime function", () => {
-  it("formats date strings correctly", () => {
+  it("formats date strings", () => {
     const result = formatDateTime("2023-01-15T12:30:45Z")
 
     expect(result).toContain("2023")
@@ -246,7 +246,7 @@ describe("formatDateTime function", () => {
 })
 
 describe("Site status indicators", () => {
-  it("shows correct status for different site states", async () => {
+  it("shows status for different site states", async () => {
     const testHelper = new IntegrationTestHelper()
     const testWebsites = makeWebsites(5)
 
@@ -269,18 +269,20 @@ describe("Site status indicators", () => {
         unpublished: true,
         updated_on: "2023-01-15T12:30:45Z",
       },
-      // Draft site - has only draft_publish_date and draft_publish_status
+      // Draft site - has draft_publish_status and draft_publish_date
+      // but no publish_date and unpublished is false
       draftSite = {
         ...testWebsites[2],
         name: "draft-site",
         uuid: "test-uuid-3",
         draft_publish_date: "2023-01-15T12:30:45Z",
-        draft_publish_status: "draft",
+        draft_publish_status: PublishStatus.Success,
         publish_date: null,
         live_publish_status: null,
         unpublished: false,
         updated_on: "2023-01-15T12:30:45Z",
       },
+      // Published site - has publish_date and live_publish_status
       publishedSite = {
         ...testWebsites[3],
         name: "published-site",
@@ -290,6 +292,7 @@ describe("Site status indicators", () => {
         unpublished: false,
         updated_on: "2023-01-15T12:30:45Z",
       },
+      // Failed site - has live_publish_status marked as errored
       failedSite = {
         ...testWebsites[4],
         name: "failed-site",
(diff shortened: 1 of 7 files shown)


ibrahimjaved12:
@umar8hassan A few things:

1. Unpublished from Production: 
    - Does not show `datetime` on hover.
    - The status still immediately changes to `Unpublished from Production` even if the unpublish pipeline fails. 
2. `Datetime`:
  - It seems very difficult to read with all the attributes present. Could we perhaps use the same timezone that we use in Concourse so that we don't have to utilize `tzinfo` like this? (and get rid of `microsecond` as well ?) 
Current behavior: image
    
  - Also, should we consider changing the rendered text itself on hover? Otherwise it seems a bit difficult to get the datetime on hover, and there's also no way to copy it.
Like this instead on hover:
image

3. `Draft` status:
    - I'm currently unable to get this working.
    - It seems to briefly show for a second when I publish to `production` instead.
    - When I create a new course and publish to staging, it still shows `Never Published`. I think it is supposed to show `Draft` ?
4. The statuses in listing page do not change in real time, but instead are reflected on page refresh. This is because the listing is fetched once via a Redux Query, so any subsequent status updates aren't automatically reflected in the UI. Should we create a ticket to triage this behavior and determine what changes would be needed to change this?

@pdpinch Could you please look at (2) and (4).

pdpinch:
I'm OK with your plan for 2). 

Regarding 4), I think we can live with requiring a refresh to see updated state.

umar8hassan:
@ibrahimjaved12, this is ready for re-review (commenting because the re-review button was already clicked)

static/js/pages/SitesDashboard.tsx
```diff
@@ -115,28 +153,38 @@ export default function SitesDashboard(): JSX.Element {
...
-                {!site.publish_date && !site.live_publish_status ? (
+                {!site.publish_date &&
+                !site.live_publish_status &&
+                !site.draft_publish_date ? (
                   <div className="text-danger">Never Published</div>
                 ) : site.unpublished ? (
-                  <div className="text-dark">Unpublished from Production</div>
+                  <StatusWithDateHover
+                    statusText="Unpublished from Production"
+                    dateTime={site.updated_on}
```
ibrahimjaved12, line 163: Should `site.updated_on` be used here?

What does `updated_on` represent?

static/js/pages/SitesDashboard.test.tsx
```diff
@@ -267,6 +267,8 @@ describe("Site status indicators", () => {
         uuid: "test-uuid-2",
         publish_date: "2023-01-01T12:00:00Z",
         unpublished: true,
+        unpublish_status: PublishStatus.Success,
```
ibrahimjaved12, line 270: I know this is in a test, but this appears confusing. Do we want to make a new constant to use here instead for `unpublish`?

static/js/pages/SitesDashboard.tsx
```diff
@@ -81,13 +81,91 @@ export const StatusWithDateHover = ({
...
+      type: "published",
+      active: PublishStatus.Success === site.live_publish_status,
+      statusText: "Published",
+      className: "text-success",
+      dateTime: site.publish_date,
+    },
+    {
+      type: "unpublished",
+      active:
+        site.unpublished && site.unpublish_status === PublishStatus.Success,
```
ibrahimjaved12, line 119: `site.unpublish_status === PublishStatus.Success
`
And here as well. It would be good to use a different enum for Unpublishing statuses?

ibrahimjaved12 (changes requested):
Could you also please address the warnings. I'm still testing the functionality. I'll let you know if I come up with anything else.

ibrahimjaved12 (changes requested):
I don't think the changes are working as expected.

Two screenshots, first from last commit, and the second from commit `0c54c8b`.
Both represent same Website states.

image

image

The second one is more accurate. It also captures `Aborted` (and other states).

There's kind of a bug.. for a fraction second, the Website shows its true state but then it changes to an inaccurate state. I believe this was supposed to be the intended change, but it's not working correctly.

I have slowed down the video to 0.25x. If you slowly go between 22 seconds to 24 seconds, you will notice it in 23 seconds of the video.

https://github.com/user-attachments/assets/6274a10b-eac3-44f8-be5e-4d5ebe7b9175

It says `publish: aborted` and quickly changes to `Never Published`

Secondly, for me, as my local unpublish pipelines are not fully working (gets stuck on `not started`) -- which is actually a good way to test the changes, the status never changes to `Unpublish` despite I can see `UNPUBLISHED = TRUE` in django admin.

Also, to give information on `ibrahim-test-publish2`:
It was published to production first
Then it was unpublished.

First screenshot shows `Published`, which was the last **completed** status. (Last actual status was `Unpublish - not started`)
Second screenshot shows `Unpublished from Production`, which is again close but not precisely accurate because the job never started.

I think we probably meant to show here `Unpublish - not started`.

ibrahimjaved12:
Also, I didn't test `Failed` state. Do you have an example how I could try to trigger that?

umar8hassan:
> Also, I didn't test `Failed` state. Do you have an example how I could try to trigger that?

You can tweak the `date` and `status` of publish settings in django admin. `Failed` state is related to error in concourse of pipeline which may and may not occur depending on the local setup.

umar8hassan:
> I was able to test `Failed` by manually changing the status to `Failed` from Django admin.
> 
> However, now `Draft` isn't working at all: it just says `Never Published`

I had not tested it yet due to some issues with my local setup. But it's ready for review now.
Screenshot 2025-06-04 at 10 41 11 AM

PR #2426, opened 2025-02-27
Use latest tag for ocw-course-publisher

ibrahimjaved12:
Uses `latest` tag for `ocw-course-publisher`, which uses the upgraded versions for Go, Hugo, and Node as specified here: https://github.com/mitodl/ol-infrastructure/pull/2993

Also adds `export COREPACK_ENABLE_DOWNLOAD_PROMPT=0` before `yarn install` in `theme_assets_pipeline.py`, `e2e_test_site_pipeline.py`, and `mass-build-sites.yml` to fix the interactive prompt:

image

### How can this be tested?
I tested the `theme_assets_pipeline.py` pipeline locally and it worked fine. Rest of the stuff will be tested in RC.
Just make sure the code looks fine!

### DIFF
--- a/content_sync/pipelines/definitions/concourse/common/image_resources.py
+++ b/content_sync/pipelines/definitions/concourse/common/image_resources.py
@@ -8,7 +8,7 @@ https://github.com/mitodl/ol-infrastructure/tree/main/dockerfiles/ocw/node-hugo
 """
 OCW_COURSE_PUBLISHER_REGISTRY_IMAGE = AnonymousResource(
     type=REGISTRY_IMAGE,
-    source=RegistryImage(repository="mitodl/ocw-course-publisher", tag="latest"),
+    source=RegistryImage(repository="mitodl/ocw-course-publisher", tag="0.8"),
 )
 
 AWS_CLI_REGISTRY_IMAGE = AnonymousResource(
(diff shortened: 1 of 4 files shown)


blarghmatey (approved):

content_sync/pipelines/definitions/concourse/common/image_resources.py
```diff
@@ -8,7 +8,7 @@
 """
 OCW_COURSE_PUBLISHER_REGISTRY_IMAGE = AnonymousResource(
     type=REGISTRY_IMAGE,
-    source=RegistryImage(repository="mitodl/ocw-course-publisher", tag="0.6"),
+    source=RegistryImage(repository="mitodl/ocw-course-publisher", tag="latest"),
```
gumaerc, line 11: Can we specify 0.8 here instead of using `latest`? We have made breaking changes to the ocw-course-publisher image in the past that required additional changes here or in `ocw-hugo-themes`. Pinning the version allows us to gate those changes to coincide with using new versions of the Docker image.

gumaerc (changes requested):

content_sync/pipelines/definitions/concourse/common/image_resources.py
ibrahimjaved12: Sure, done!

PR #2438, opened 2025-03-05
Revert ocw course publisher tag temporarily

ibrahimjaved12:
Reverts https://github.com/mitodl/ocw-studio/pull/2426 because we just had a pre-mature release

PR #2440, opened 2025-03-05
Revert "Revert ocw course publisher tag temporarily"

ibrahimjaved12:
Reverts mitodl/ocw-studio#2438

This is the actual PR that needs to go through -- but after some time -- when ocw-hugo-themes RC is fully tested with Hugo's 0.142.0 version.


gumaerc (approved):
👍

PR #2443, opened 2025-03-10
Cleanup headless related code (Hugo build options)

ibrahimjaved12:
Part of https://github.com/mitodl/hq/issues/6895

Removes `headless: true` from tests and other places to cleanup redundant code and update it with new configuration.


umar8hassan (approved):
Looks Good 👍

PR #2448, opened 2025-03-14
skip implicit get

gumaerc:
Closes https://github.com/mitodl/hq/issues/6944

This PR sets `no_get` on `PutStep` objects created during the various OCW Concourse pipeline definitions. It is hoped that this will solve a problem seen in our RC and production Concourse environments where the "implicit GET" of the `PutStep` indefinitely hangs.

### How can this be tested?
It is my hypothesis that this is an issue with the Concourse `across` step, possibly related to running it on multiple workers. It is a difficult issue to reproduce on-demand, but happens frequently in said environments. In order to test this, we would need to deploy the code to our test environment and try running the mass build pipeline with the changes in this PR. Nonetheless, I think it's worth trying running this branch locally and running the following to test that upserting the pipelines is not broken locally:

- Make sure you have `ocw-studio` running locally (containers built on this branch) and configured to run pipelines in Concourse
- Run the following:
```
docker compose exec web ./manage.py upsert_theme_assets_pipeline
docker compose exec web ./manage.py upsert_mass_build_pipeline
docker compose exec web ./manage.py backpopulate_pipelines
```
- The last one will take a while to push up all the pipelines, so you can use `--filter` to filter on a single site if you wish.
- Run the theme assets pipeline
- Try publishing a course
- Run the mass build pipeline
- You will likely not see the exact same problem described as above, but as long as these pipelines run then we have not broken anything


gumaerc:
Ignore the stuff Gemini is saying about Posthog. I accidentally based this off another branch instead of main and have since rebased it back to main.

content_sync/pipelines/definitions/concourse/site_pipeline_test.py
```diff
@@ -427,10 +427,12 @@ def assert_base_build_tasks(tasks: list[dict], offline: bool):  # noqa: FBT001
                 open_discussions_webhook_step_online_params["version"]
                 == config.vars["pipeline_name"]
             )
+    offline_build_gate_put_task = online_site_tasks[-1]
     assert (
-        online_site_tasks[-1]["try"]["put"]
+        offline_build_gate_put_task["try"]["put"]
         == pipeline_definition._offline_build_gate_identifier  # noqa: SLF001
     )
+    assert online_site_tasks[-1]["try"]["no_get"] is True
```
umar8hassan, line 435: a nitpick, but maybe you wanted this here:
```suggestion
    assert offline_build_gate_put_task["try"]["no_get"] is True
```

umar8hassan (approved):
Looks Good 👍 
i have tested the single course (draft/live, w/wo offline) and mass publish (draft only) locally.
Screenshot 2025-03-18 at 6 32 58 AM
Screenshot 2025-03-18 at 6 23 23 AM

Screenshot 2025-03-18 at 6 21 46 AM

PR #2470, opened 2025-03-31
separate Posthog env vars between studio and published OCW

gumaerc:
Closes https://github.com/mitodl/hq/issues/7017

This PR sets up a separate set of env vars for Posthog to be used with the published OCW site on the frontend. The theme assets pipeline definition was configured to use these vars instead of the ones used by `ocw-studio` itself.

### How can this be tested?
- Make sure you have a test Posthog project to work with
- In your `.env`, set the following:
```
PUBLISH_POSTHOG_ENABLED=true
PUBLISH_POSTHOG_PROJECT_API_KEY=<your API key here>
```
- Spin up `ocw-studio`
- Run `docker compose exec web ./manage.py upsert_theme_assets_pipeline`
- Run the theme assets pipeline in Concourse
- Publish a site
- Open the site in a browser, look at the development console and ensure that you see "Posthog is enabled"
- Switch to `PUBLISH_POSTHOG_ENABLED=false` and `docker compose restart web celery`
- Re-run the steps above, ensuring that in the browser you now see "Posthog is disabled"

### Checklist:
- [x] https://github.com/mitodl/ol-infrastructure/pull/3079 should be merged and deployed

### DIFF
--- a/app.json
+++ b/app.json
@@ -487,6 +487,26 @@
       "description": "Frequency (in seconds) to run a check on potentially stalled publish builds",
       "required": false
     },
+    "PUBLISH_POSTHOG_API_HOST": {
+      "description": "API host for PostHog, published to pipelines",
+      "required": false
+    },
+    "PUBLISH_POSTHOG_ENABLED": {
+      "description": "Whether PostHog is enabled, published to pipelines",
+      "required": false
+    },
+    "PUBLISH_POSTHOG_FEATURE_FLAG_REQUEST_TIMEOUT_MS": {
+      "description": "Timeout (ms) for PostHog feature flag requests, published to pipelines",
+      "required": false
+    },
+    "PUBLISH_POSTHOG_MAX_RETRIES": {
+      "description": "Number of times requests to PostHog are retried if failed, published to pipelines",
+      "required": false
+    },
+    "PUBLISH_POSTHOG_PROJECT_API_KEY": {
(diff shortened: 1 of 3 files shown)


pdpinch:
After we make this change, and if we leave the values blank, will this turn off or turn on the login button in production?

gumaerc:
> After we make this change, and if we leave the values blank, will this turn off or turn on the login button in production?

Yes, that is correct. I have another PR up [here](https://github.com/mitodl/ol-infrastructure/pull/3079) to set the values only for RC in the Pulumi config, but @blarghmatey had some suggestions on that. I was holding off on that until we knew exactly what we wanted to do about production.

- Shut it off entirely?
- Opt out of data collection? https://github.com/mitodl/ocw-hugo-themes/pull/1573
- Turn off `autocapture` and specify which events to capture manually?
- Spin up our own Posthog server so we can firehose events to it from every service we run?

I put up PR's for the "shut it off entirely" approach to stem the bleeding as it were, but if we want to take a different approach that's possible. Maybe for now we should just shut it off, because we're not even really ready to use the data? This would also mean we can't check feature flags in production, but the only thing using that on the OCW frontend right now is the login functionality. It doesn't seem like we're going to prioritize releasing that anytime soon, so maybe shutting off Posthog in production is fine?

pdpinch:
I want to shut it off entirely until we get a chance to get our ducks in a row and also find out what enterprise billing looks like for posthog.

PR #2562, opened 2025-05-09
Upgrade mitol-django-transcoding and boto3 libraries

pt2302:
N/A; follow-up to https://github.com/mitodl/ol-django/pull/243.

This PR upgrades the `mitol-django-transcoding` and `boto3` libraries to the latest versions. Automatic updates of `boto3` had not been possible before because of the strict `boto3` version dependency in `mitol-django-transcoding`. Now that this has been resolved, this PR updates `ocw-studio` to use the latest versions.

### How can this be tested?
Smoke-test the functionality associated with S3, including syncing with Google Drive.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -148,34 +148,34 @@ files = [
 
 [[package]]
 name = "boto3"
-version = "1.37.11"
+version = "1.38.13"
 description = "The AWS SDK for Python"
 optional = false
-python-versions = ">=3.8"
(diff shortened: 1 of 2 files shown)


umar8hassan (approved):
Looks Good 👍 
I have tested the flow which includes GDrive sync, pushing files to S3, submitting job to mediaconvert. Verified that the job was valid and completed without error.

PR #2578, opened 2025-05-26
Refactor referenced items Delete Dialog

ibrahimjaved12:
https://github.com/mitodl/hq/issues/5686#issuecomment-2839573054

1. Simplify the text to just "This item is referenced by other items and cannot be deleted."
2. Check whether the item can be deleted when the ~modal~ page is loaded, and display the error message right away when the delete modal is opened.
3. When the item cannot be deleted, hide the delete button.

image
image
image

### How can this be tested?
1. Checkout to this branch
2. Spin up ocw-studio locally
3. Create two external resources and link one in a page.
4. Now try deleting the external resources. The one linked in the page should error out immediately when you open the delete modal, with only `Cancel` button visible. The other external resource should be deletable via `Delete` button.
5. You can also extend the testing by testing deletion of a page. First link a course page in the course's menu. Afterwards, try deleting the page and confirm its erroring same as above.

### DIFF
--- a/static/js/components/RepeatableContentListing.tsx
+++ b/static/js/components/RepeatableContentListing.tsx
@@ -167,7 +167,7 @@ export default function RepeatableContentListing(props: {
       setSelectedContent(content)
       if (content.is_deletable === false) {
         setDeleteError(
-          "This item is hehe referenced by other items and cannot be deleted.",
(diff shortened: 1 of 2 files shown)


gumaerc (approved):
👍 LGTM

PR #2584, opened 2025-06-04
fix: added check for resource links with highlighted text

umar8hassan:
closes https://github.com/mitodl/hq/issues/6953

This aims to fix the ceration of external resource on adding link to highlighted text in editor. Follow the steps in this [video](https://www.youtube.com/watch?v=rZ4lxn0FLHY) to replicate the issue locally 

### How can this be tested?
1. swicth to master branch and replicat the issue as described [here](https://www.youtube.com/watch?v=rZ4lxn0FLHY)
2. switch to this branch and verify that it should not create an external resource
3. verify that the added link leads to resource on published site and not to ocw studio internal link

### DIFF
--- a/static/js/lib/ckeditor/plugins/CustomLink.ts
+++ b/static/js/lib/ckeditor/plugins/CustomLink.ts
@@ -73,7 +73,7 @@ class CustomLinkCommand extends LinkCommand {
  * CustomLink Plugin for CKEditor 5
  *
  * PURPOSE:
- * This plugin automatically converts external URLs into tracked "external resources"
+ * This plugin automatically converts URLs into tracked "external resources"
  * in the OCW Studio content management system. It exists to solve several problems:
  *
  * 1. LICENSE TRACKING: External links need license warnings and rights management


pt2302 (changes requested):
Could you please add some unit tests for `CustomLink.ts`, specifically testing the `getExternalResource` and `updateHref` functions?

static/js/lib/ckeditor/plugins/CustomLink.ts
```diff
@@ -140,19 +147,32 @@ async function getExternalResource(
...
 
-function updateHref(
+export function updateHref(
   externalResource: { title?: string; textId?: string },
   editor: Editor,
   superExecute: { (customHref: string): void },
 ) {
+  // Validate inputs
+  if (!externalResource.title || !externalResource.textId) {
+    console.warn("Missing title or textId, skipping updateHref.")
```
pt2302, line 167: Would the user be expected to react to this warning? When would the `externalResource` be missing a `title` or a `textID`, and shouldn't that be handled when creating the external resource rather than when this function is called?

static/js/lib/ckeditor/plugins/CustomLink.test.ts
```diff
@@ -0,0 +1,954 @@
...
+    })
+
+    it("inserts text with linkHref if selection is collapsed", () => {
+      updateHref({ title: "T", textId: "id" }, editor, superExecute)
+      expect(editor.model.change).toHaveBeenCalled()
+      const writer = { insertText: jest.fn() }
+      editor.model.change.mock.calls[0][0](writer)
+      expect(writer.insertText).toHaveBeenCalledWith(
+        "T",
+        { linkHref: "resource://id" },
```
pt2302, line 683: This is not what resource links actually look like: see https://github.com/mitodl/ocw-studio/blob/master/static/js/lib/ckeditor/plugins/ResourceLink.ts

umar8hassan: updated in https://github.com/mitodl/ocw-studio/pull/2584/commits/24e799538e1afc652e34db290bb2acb4e0e644c1

static/js/lib/ckeditor/plugins/CustomLink.ts
umar8hassan: This was an extra guard and warning. removed in https://github.com/mitodl/ocw-studio/pull/2584/commits/14849b233fc54bd70b38a5089644982d0ec0d464 and updated tests

static/js/lib/ckeditor/plugins/CustomLink.ts
```diff
@@ -13,6 +13,7 @@ import { DiffItem } from "@ckeditor/ckeditor5-engine/src/model/differ"
 import Writer from "@ckeditor/ckeditor5-engine/src/model/writer"
 class CustomLinkCommand extends LinkCommand {
   execute(href: string, _options = {}) {
+    const syntax = this.editor.plugins.get(ResourceLinkMarkdownSyntax)
```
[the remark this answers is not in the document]
umar8hassan: The plugin already has a guard in place, as it is required in `CustomLink`.

static/js/lib/ckeditor/plugins/CustomLink.test.ts
```diff
@@ -0,0 +1,970 @@
...
+
+      const editor = await getEditor("")
+      const linkCommand = editor.commands.get("link")
+      expect(linkCommand).toBeDefined()
+
+      if (!linkCommand) return
+
+      // Create some text in the editor first
+      const root = editor.model.document.getRoot()
+      if (!root) return
```
ChristopherChudzicki, line 742: I **think** you have these early exists (`if (!root) return`) to avoid Typescript errors. Is that correcct?

A better approach would be to make a type-narrowing assertion like:
```ts
import invariant from "tiny-invariant"

invariant(root, "Optional error message")

// The code above code is roughly equivalent to
if (!root) {
  throw new Error("Optional Error Message")
}
// but IMO invariant is much more convenient.
```
That will narrow the type so that TS knows it is not undefined.

**The danger of early exist** is that maybe the early exit triggers when you didn't expect, and then test passes errorneously because fewer assertions are made.

You seem to have been careful most places to check—e.g., above `expect(linkCommand).toBeDefined()`. But (A) jest expectations don't narrow the type and (B) I think you missed it here for `root`? Generally safer to use a type-narrowing assertion for that reason.

static/js/lib/ckeditor/plugins/CustomLink.ts
```diff
@@ -23,15 +24,21 @@ class CustomLinkCommand extends LinkCommand {
       }
     }
 
-    getExternalResource(this.editor.config.get(WEBSITE_NAME), href, title).then(
-      (externalResource) => {
+    if (syntax.isResourceLinkHref(href)) {
+      super.execute(href)
+    } else {
+      getExternalResource(
```
ChristopherChudzicki, line 30: This seems good to me 👍 

However, general comments on `CustomLink` file:

1. I think it would be REALLY helpful to have docstrings explaining how this feature is supposed to work. The test file is also very helpful for that, but I think prose explaining what CustomLink is responsible for / why it exists would be helpful.
2. Based on the if/else above this comment, it seems to me that **every link is either a resource link or an external link**. Is that correct? Is that the terminology we use? E.g., is a link to ocw.mit.edu/courses/other-course an external link?  *This is a question...maybe that is the terminology we use! It is external to the specific course.*
3. These lines:
    ```ts
    this.editor.commands.add("link", new CustomLinkCommand(this.editor))
    console.log("CustomLink Plugin is initialized")
    ```ts

    Make me a little nervous. Looking at `MarkdownEditor.ts`, it seems to me we are adding `Link` (via default conrigs) and `CustomLink` both. I know that `Link` adds `LinkCommand`, which adds a `"link"` command. That is to say: I'm concerned that the command named "link" is a being added twice. Is it?
4. Does the ` this.editor.model.document.on("change:data", () => {...})` callback really need to run on every change? I would have thought that logic would run as part of `CustomLinkCommand` but I mgiht be wrong.

I realize that most of these things aren't changed in this PR, but at least (1) might be worth addressing here.

ChristopherChudzicki (commented):
Thanks for adding the test file! 

I left a couple comments. I lack some context here and haven't worked directly with CKEditor in a while, but they are things that jumped out to me as potentially odd. Let me know what you think.

Disclaimer: I haven't tried running this locally yet, but might be able to tomorrow.

static/js/lib/ckeditor/plugins/CustomLink.ts
```diff
@@ -23,18 +51,56 @@ class CustomLinkCommand extends LinkCommand {
...
+      })
+    }
   }
 }
 
+/**
+ * CustomLink Plugin for CKEditor 5
+ *
+ * PURPOSE:
+ * This plugin automatically converts external URLs into tracked "external resources"
```
ChristopherChudzicki, line 76:
```suggestion
 * This plugin automatically converts URLs into tracked "external resources"
```

My understanding is...

All URLs are "external resources", where "external" means eternal to the current site. (A URL within the current site would technically be handled as an external resource, too, but the expectation is that site authors use resource links directly for that sort of thing.)

static/js/lib/ckeditor/plugins/CustomLink.ts
```diff
@@ -95,14 +190,47 @@ export default class CustomLink extends Plugin {
...
 }
 
-async function getExternalResource(
+/**
+ * Create an external resource content item via the OCW Studio API.
+ *
+ * This function handles the core logic of external resource creation:
+ *
+ * 1. DOMAIN ANALYSIS: Determines if the URL is external by comparing its hostname
+ *    to SETTINGS.sitemapDomain. Same-domain URLs get no license warning.
```
ChristopherChudzicki, line 215: Again, it seems to me that all URLs are external resources. E.g., https://ocw-studio-rc.odl.mit.edu/sites/resstr-001-geographic-information-system-gis-tutorial/type/external-resource/edit/134ee4ce-749a-46e3-9baa-f1f062a1bce6/ 

So we don't determine external vs non-external via domain, we determine "needs content warning" vs "does not need content warning".

pt2302 (approved):
Looks good to me!

PR #2592, opened 2025-06-19
fix: stripped query parameters before downloading

umar8hassan:
https://github.com/mitodl/hq/issues/7526

The filter-webpack-artifacts pipeline step was failing with 404 errors because webpack.json contained font file paths with query parameters (e.g., "font.woff2?64889571"), but the actual S3 objects are stored without these query parameters.

In this PR, we have modified the pipeline command to strip query parameters using `jq` before downloading from S3, ensuring the correct object keys are used.

### How can this be tested?
1. Switch to this branch and test the offline build. it should complete without errors


pt2302:
@umar8hassan Can you please update the PR description, since `sed` is no longer being used?

PR #2593, opened 2025-06-19
chore(fix): updated color scheme and text for course states

umar8hassan:
https://github.com/mitodl/hq/issues/6017

This PR is a follow-up to https://github.com/mitodl/ocw-studio/pull/2407. In this PR we have incorporated the followinh changes:
- `Never Publish` to use an orange color.
- `Draft` to use a blue color
- Remove "from Production" from `Unpublished from Production`

### How can this be tested?
1. Switch to this branch
2. Go to the site listings page on `http://localhost:8043/sites/`
3. Create a new site or use an existing one to stage/publish.
4. Check that existing sites have the appropriate statuses


pdpinch (approved):
I left a couple of questions, but this is fine to merge otherwise.

PR #2602, opened 2025-06-24
Ability to delete video resources

ibrahimjaved12:
Closes https://github.com/mitodl/hq/issues/2991

Adds the ability to delete all kinds of video resources according to the plan specified in: https://github.com/mitodl/hq/issues/2991#issuecomment-2651158964

### How can this be tested?
Follow the video workflow. Make sure you got the env vars etc as defined here: https://github.com/mitodl/ocw-studio?tab=readme-ov-file#video-workflow

Note: You will need to use `DRIVE_SERVICE_ACCOUNT_CREDS` of ocw-studio-rc to actually enable files deletion in gdrive. Because I got the permissions set up there. I used RC environment for testing (so no minio).

Checkout to this branch. Add resource in your env var `OCW_STUDIO_DELETABLE_CONTENT_TYPES`:
`OCW_STUDIO_DELETABLE_CONTENT_TYPES=instructor,page,course-collection,resource`

To test pre-existing videos' deletion:
1. Click on Add New Resource and create a new video resource using `youtube_id`: `E8uZtq_vOYM`
  i. This will create the video resource with linked transcript and caption files.
2. Click on the 3 dots that appear on the right side of resource, for deletion. Confirm deletion.
3. Make sure all the 3 resources are fully deleted.

To test other videos (legacy or gdrive workflow):
1.  Add a video in videos_final and a `videoname_captions.vtt` file and a `videoname_transcript.pdf` file in `files_final` directory in gdrive.
2. Sync resources.
3. Under the normal flow, both the related files automatically get linked with video.. but this might not happen locally. So you can go to Django admin, the relevant WebsiteContent files, and from `File` field copy the relevant path of caption and transcript. And link them with the Video resource from OCW studio interface. Make sure there are no spaces in the start or end of the path!
4. Delete the video resource. Make sure all the 3 resources are fully deleted.

This can take a couple of seconds to delete. I'm going to create follow-up tickets for:
1. Video image_thumbnail process updating (right now it force-overwrites the thumbnail even when provided)
2. Closing off deletion dialog without waiting for deletion to complete
3. Updating READMEs

### DIFF
--- a/websites/deletion_utils_test.py
+++ b/websites/deletion_utils_test.py
@@ -1,0 +1,61 @@
+"""Tests for deletion utilities related to WebsiteContent."""
+
+import pytest
+
+from websites.constants import CONTENT_TYPE_RESOURCE, RESOURCE_TYPE_VIDEO
+from websites.deletion_utils import (
+    delete_related_captions_and_transcript,
+    delete_resource,
+)
+from websites.factories import WebsiteContentFactory, WebsiteFactory
+from websites.models import WebsiteContent
+
+
+@pytest.mark.django_db()
+def test_delete_video_deletes_captions_and_transcripts(mocker):
+    """
+    Deleting a video should also delete its associated caption and transcript WebsiteContent
+    objects.
+    """
+    website = WebsiteFactory.create()
+
+    mocker.patch("websites.deletion_utils.delete_drive_file")
+    mocker.patch("websites.deletion_utils.get_drive_service")
+    mocker.patch("websites.deletion_utils.delete_s3_objects.delay")
+
+    base_name = "E8uZtq_vOYM"
+    course_path = f"/courses/{website.name}/"
+    caption_path = f"{course_path}{base_name}_captions.vtt"
+    transcript_path = f"{course_path}{base_name}_transcript.pdf"
+
+    video_metadata = {
+        "resourcetype": RESOURCE_TYPE_VIDEO,
+        "video_files": {
+            "video_captions_file": caption_path,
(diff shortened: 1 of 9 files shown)


umar8hassan (changes requested):
delete is an irreversible operations, we should increase test coverage. and add tests for `deletion_utils` specifically.

websites/deletion_utils_test.py
```diff
@@ -0,0 +1,61 @@
...
+    transcript = WebsiteContentFactory.create(
+        website=website, file=transcript_path, filename=f"{base_name}_transcript"
+    )
+
+    assert WebsiteContent.objects.filter(pk=video.pk).exists()
+    assert WebsiteContent.objects.filter(pk=caption.pk).exists()
+    assert WebsiteContent.objects.filter(pk=transcript.pk).exists()
+
+    delete_related_captions_and_transcript(video)
+    delete_resource(video)
```
umar8hassan, line 57: Looking at these, it made me realize that it should have been the other way around. But the refactor might be too much effort now, is it?

We should have had to call a single method `delete_resource(video)`, and this method would have called `delete_related_captions_and_transcript` inherently, and also deleted all related objects like drivefile and content sync states if any.

umar8hassan (changes requested):
Code works but we need to fix the tests.

umar8hassan (approved):
Looks Good 👍 

As discussed, even though we had removed the deletion API for YouTube in the past to prevent any programmatic deletion, we should re-discuss the possibility now that we are working on deleting videos.
cc @pdpinch @pt2302

PR #2610, opened 2025-07-05
Display read-only page URL in Edit Page drawer

pt2302:
Part of https://github.com/mitodl/hq/issues/5357.

This PR adds a new read-only field to the `Edit Page` drawer where the page URL is displayed. This is part of making page URL editing easy-to-use and transparent. 

### How can this be tested?
In OCW Studio, verify that the Page URL field is displayed and looks correct in the `Edit Page` drawer.

Also, after following the testing instructions in https://github.com/mitodl/ocw-studio/pull/2601, verify that changing the Page Title also changes the displayed page URL after saving and re-opening the drawer. Note that this is currently behind a PostHog feature flag.

### DIFF
--- a/static/js/components/forms/SiteContentForm.test.tsx
+++ b/static/js/components/forms/SiteContentForm.test.tsx
@@ -247,7 +247,7 @@ test.each`
 
 test.each`
   isGdriveEnabled | isResource | willRender | filename
-  ${true}         | ${true}    | ${true}    | ${"courses/file.pdf"}
+  ${true}         | ${true}    | ${true}    | ${"file.pdf"}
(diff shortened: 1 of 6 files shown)


gumaerc (approved):
👍 

I'm going to approve this PR because I was able to get everything running and the functionality works as described. However, something did catch my eye. I'm not sure if there's a separate issue for this. After saving the new page title, I saw the updated URL. When I published the site however, the page had disappeared from the nav. I believe this is because the nav configuration is not updated as part of the URL change, and is still referencing the page by the old slug. After going into the Menu section and re-saving the menu, then republishing, I was able to see the nav item again.

PR #2638, opened 2025-07-24
chore: set number of attempts for the offline build gate step to three

umar8hassan:
https://github.com/mitodl/hq/issues/7898

This branch sets the number of attempts for the offline build gate step in the site pipeline, from one to three.

### How can this be tested?
1. Switch to the branch and spin up containers
2. update pipeline definition with the following command and verify that the offline build gate step has three attempts:
`docker compose exec web ./manage.py backpopulate_pipelines`


ibrahimjaved12 (approved):
👍

image

PR #2646, opened 2025-07-28
Skip recursion by using skip_sync flag variable for Add Video Resource

ibrahimjaved12:
None

Fixes a bug in RC introduced as side effect of https://github.com/mitodl/ocw-studio/pull/2618. If you create a video resource using Add Video Resource functionality without `archive_url`, it goes into an infinite loop.

### How can this be tested?
Checkout to this branch.
Use the Add Video Resource functionality using youtube_id `E8uZtq_vOYM`. In one test, don't use an archive_url. In another test, use this archive_url: `https://dn720405.ca.archive.org/0/items/mit-2020-vision/MIT_2020_Vision_Part_6_300k.mp4`. Make sure it's properly created with transcript and captions. (For single youtube_id in a course, transcript/captions will only be created once -- but they will be linked regardless).

### DIFF
--- a/videos/signals.py
+++ b/videos/signals.py
@@ -76,5 +76,6 @@ def sync_missing_caption(
             )
             instance.metadata = metadata
             metadata["file_size"] = None
+            instance.skip_sync = True
             instance.save(update_fields=["metadata"])
         sync_video_captions_and_transcripts(instance)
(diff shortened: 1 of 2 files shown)


umar8hassan (approved):
Looks Good 👍 
The code works as expected, although max recursion error didn't reproduce in my local setup.

PR #2654, opened 2025-07-30
Backfill file and file_size for videos and their captions/transcript

ibrahimjaved12:
https://github.com/mitodl/hq/issues/7807

This PR creates two management commands:
1.  backfill_video_file_size (backfills `metadata['file_size']` for video resources using archive_url if not already populated)
2. backfill_caption_transcript_file_size (Links s3 `file` and backfills `metadata['file_size']` for caption and transcript resources)

Even though they're one-time utilities, they are implemented as management commands to support Celery task queuing, as the operations involve external services (S3, archive URLs) and can be slow or error-prone.

### How can this be tested?
It can be a little tricky to test this, because when you make a video resource, using Django signals, all of the things that this management command backfills, will automatically be force-populated for new video resources.

So during my testing, I had disabled that by removing the relevant code. Which you can see on the last commit of this branch: https://github.com/mitodl/ocw-studio/pull/2654/commits/9e38e21b73a44a3beb45bf40ede2f1cbe5efc520

What you can do is, you can checkout to the 2nd last commit from this branch, that had included those reverted changes for testing.

And then you can create a New Video Resource using `youtube_id`: `E8uZtq_vOYM` and any working `archive_url` of a video: https://ia600708.us.archive.org/30/items/chalk-radio/S05E01_Anne_White_360p.mp4. Check from Django admin that its `metadata["file_size"]` is not populated.
1. Run management command: `backfill_video_file_size` using `--filter` with your course. It will populate its metadata["file_size"].
Check from Django admin for its caption and transcript WebsiteContent objects. They should not have file_size either, and should have empty `file` fields.
3. Run management command `backfill_caption_transcript_file_size` using `--filter` to backfill those.

### DIFF
--- a/videos/tasks.py
+++ b/videos/tasks.py
@@ -659,3 +659,46 @@ def populate_video_file_size(self, video_content_id: int):
             video.website.short_id,
         )
         raise self.retry(exc=exc) from exc
+
+
+@app.task(
+    bind=True,
+    acks_late=True,
+    retry_backoff=True,
+    retry_backoff_max=128,
+    max_retries=3,
+)
+def backfill_caption_or_transcript_file_size(self, resource_id: int):
+    """
+    Populate file and file_size for a caption or transcript resource.
+    Uses metadata["file"] to get the S3 path and fetch its size.
+    """
+
+    def _raise_missing_file_error(filename: str) -> None:
+        error_msg = f"Missing metadata['file'] for {filename}"
+        raise ValueError(error_msg)
+
+    try:
+        resource = WebsiteContent.objects.get(id=resource_id)
+        s3_path = (resource.metadata or {}).get("file")
+        if not s3_path:
+            _raise_missing_file_error(resource.filename)
+
+        s3_key = s3_path.lstrip("/")
+        s3 = get_boto3_resource("s3")
+        s3_obj = s3.Object(settings.AWS_STORAGE_BUCKET_NAME, s3_key)
+        file_size = s3_obj.content_length
+
+        resource.file = s3_path
+        resource.metadata["file_size"] = file_size
+        resource.skip_sync = True
+        resource.save(update_fields=["file", "metadata"])
+
+    except Exception as exc:
+        log.exception(
+            "Error backfilling file size for %s (%s) in course %s",
+            getattr(resource, "filename", f"ID {resource_id}"),
+            getattr(resource, "id", resource_id),
+            getattr(resource.website, "short_id", "?"),
+        )
+        raise self.retry(exc=exc) from exc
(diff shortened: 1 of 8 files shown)


videos/tasks.py
```diff
@@ -566,21 +566,96 @@ def copy_video_resource(source_course_id, destination_course_id, source_resource
...
-        video.save(update_fields=["metadata"])
+
+    def _raise_bad_response_error(status_code: int, video_title: str):
+        error_msg = f"Bad response: {status_code}"
+        log.error(
+            "Failed to populate file size for video %s: %s",
+            video_title,
+            error_msg,
+        )
+        raise requests.HTTPError(error_msg)
```
umar8hassan, line 587: I assume this is to meet the linter on raising a direct exception inside the function. we should move this to `exceptions.py` for the module.

videos/tasks.py
```diff
@@ -566,21 +566,96 @@ def copy_video_resource(source_course_id, destination_course_id, source_resource
...
+)
+def backfill_caption_or_transcript_file_size(self, resource_id: int):
+    """
+    Populate file and file_size for a caption or transcript resource.
+    Uses metadata["file"] to get the S3 path and fetch its size.
+    """
+
+    def _raise_missing_file_error(filename: str) -> None:
+        error_msg = f"Missing metadata['file'] for {filename}"
+        raise ValueError(error_msg)
```
umar8hassan, line 636: I assume this is to meet the linter on raising a direct exception inside the function. We should move this to `exceptions.py` for the module.

websites/management/commands/backfill_video_file_size.py
```diff
@@ -0,0 +1,66 @@
...
+    def handle(self, *args, **options):
+        super().handle(*args, **options)
+
+        sync_execution = options.get("sync", False)
+
+        websites = Website.objects.all()
+        websites = self.filter_websites(websites)
+
+        videos = WebsiteContent.objects.filter(
+            type="resource", metadata__resourcetype="Video", website__in=websites
```
umar8hassan, line 32: nitpick - use constant from `websites.constants`
```suggestion
            type=CONTENT_TYPE_RESOURCE, metadata__resourcetype=RESOURCE_TYPE_VIDEO, website__in=websites
```

umar8hassan (commented):
I was able to test `backfill_video_file_size` on my local setup, it works fine. ~But `backfill_caption_transcript_file_size` hasn't worked for me yet.~ it worked. The issue was with my celery task not being setup properly on the container build and restart.

umar8hassan (approved):
Looks Good 👍

PR #2680, opened 2025-08-16
Remove ocw_to_hugo dependency and refactor

pt2302:
Part of https://github.com/mitodl/hq/issues/7095.

This PR removes the `ocw_import` module from the `ocw-studio` codebase, and refactors tests that used functions from it.

### How can this be tested?
There should be no impact to current functionality. Tests should pass as expected.


umar8hassan (approved):
Looks Good 👍 
Please also remove directory `test_ocw2hugo` before merge.

Issue #2650, opened 2025-07-29
Python Upgrade: 3.12.x -> 3.13.x


umar8hassan:
### Description/Context
We are currently running Python 3.12.x, which has `3.12.10` as the last stable release. We should upgrade to Python 3.13.x for the latest updates. This kind of upgrade has been carried out before the list below was used to smoke test Studio. We can merge the following renovate PR when the list has been checked: https://github.com/mitodl/ocw-studio/pull/2649

### Plan/Design

We need to check the following list of items to make sure the upgrade didn't break anything:

 - [ ] Django admin
 - [ ] pylint / pytest / black
 - [ ] Editing content and saving it
   - [ ] Pages
   - [ ] Resources
   - [ ] Metadata
   - [ ] Menus
 - [ ] Pushing up pipelines to Concourse
   - [ ] Creating a new site
   - [ ] `backpopulate_pipelines`
   - [ ] `upsert_theme_assets_pipline`
   - [ ] `upsert_mass_build_pipeline`
 - [ ] Google drive sync
 - [ ] Youtube
 - [ ] 3Play
 - [ ] AWS S3
 - [ ] AWS MediaConvert

PR #2689, opened 2025-08-20
Update Poetry to version 2.1.4

zawan-ila:
This is needed for https://github.com/mitodl/ocw-studio/issues/2650

Of late, poetry has not been using the expected python version for its virtual environments in the CI. The expected python version is the one set up in the setup-python [step](https://github.com/mitodl/ocw-studio/blob/master/.github/workflows/ci.yml#L62-L66). What's happening is that poetry has been using the system python version for its environments (which is 3.12.3 on our runners). This can be verified by looking at the output of any recent CI run ([example](https://github.com/mitodl/ocw-studio/actions/runs/17013451368/job/48232531560)) Notice the python version set up in `Set up python` step (3.12.11) and the one being used in the `Tests` step (3.12.3)

This was reported as a [bug](https://github.com/python-poetry/poetry/issues/10490) in poetry and fixed in 2.1.4.

### How can this be tested?
1. Check out this branch
2. Rebuild the containers
3. Verify that studio loads correctly

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,7 @@ jobs:
     - name: Install poetry
       uses: snok/install-poetry@76e04a911780d5b312d89783f7b1cd627778900a # v1
       with:
-        version: 2.1.1
+        version: 2.1.4
         virtualenvs-create: true
         virtualenvs-in-project: true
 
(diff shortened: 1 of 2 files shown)


umar8hassan (approved):
Looks Good 👍

PR #2702, opened 2025-09-02
Umar/8322 nested shortcode syntax in hugo

umar8hassan:
https://github.com/mitodl/hq/issues/8322

Adds guards to prevent nested shortcodes in Hugo builds by detecting when markdown links are inside existing Hugo shortcode attributes and skipping their conversion to `resource_link` shortcodes. This prevents invalid nested shortcode syntax like `{{< shortcode attr="{{< resource_link ... >}}" >}}` which would break Hugo builds.

The implementation adds a `_is_inside_shortcode_attribute()` method that uses regex patterns to detect if a markdown link is positioned within a Hugo shortcode's attribute value (between quotes), and if so, preserves the original markdown link instead of converting it.

This PR also includes fixes for markdown text cleaning that prevent over-escaping of backticks and brackets, ensuring clean Hugo shortcode parameter generation without unnecessary escape characters.

## Testing
1. Add the following test data (or use your own test data) to an existing (or new) page using Django Admin.
```markdown
# Test Page with Shortcodes and Escaping Issues

Regular link that should be converted:
[MIT Website](https://mit.edu)

Image gallery with links in attributes (should NOT be converted):
{{< image-gallery >}}
{{< image-gallery-item href="test1.png" text="[OCW](https://ocw.mit.edu) - Educational content" >}}
{{< image-gallery-item href="test2.png" data-desc="Link to [GitHub](https://github.com) repository" >}}
{{< /image-gallery >}}

Links with problematic characters that need proper escaping:
[Download \`solution.py\` from \[Resources\]](resource:abc-123)  
[Code example with \`backticks\` and \[brackets\]](https://example.com/code)
[Math notation: \[A\] + \[B\] = \[C\]](https://example.com/math)

Another regular link that should be converted:
[Google](https://google.com)
```

2. ### Manual Testing
```bash
# Test the shortcode detection prevents nested shortcodes and escaping is fixed
docker-compose exec web ./manage.py markdown_cleanup link_to_external_resource --filter my_site

# Expected Results:
# - Regular links converted to resource_link shortcodes
# - Links inside shortcode attributes remain as markdown links  
# - Generated Hugo shortcodes have clean parameters without over-escaped backticks/brackets
# - Example: {{< resource_link "uuid" "Download `solution.py` from [Resources]" >}}
```

### DIFF
--- a/websites/management/commands/markdown_cleaning/cleaner.py
+++ b/websites/management/commands/markdown_cleaning/cleaner.py
@@ -144,12 +144,18 @@ class WebsiteContentMarkdownCleaner:
     def write_matches_to_csv(self, path: str, only_changes):
         """Write matches and replacements to csv."""
 
-        with open(path, "w", newline="") as csvfile:  # noqa: PTH123
+        with open(path, "w", newline="", encoding="utf-8") as csvfile:  # noqa: PTH123
             fieldnames = [
                 *self.csv_metadata_fieldnames,
                 *(f.name for f in fields(self.rule.ReplacementNotes)),
             ]
-            writer = csv.DictWriter(csvfile, fieldnames, quoting=csv.QUOTE_MINIMAL)
+            writer = csv.DictWriter(
+                csvfile,
+                fieldnames,
+                quoting=csv.QUOTE_MINIMAL,
+                escapechar="\\",
+                doublequote=False,
+            )
             writer.writeheader()
             for change in self.replacement_matches:
                 has_changed = change.original_text != change.replacement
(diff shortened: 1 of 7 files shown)


umar8hassan:
I ran another pass on the full production data and analyzed the data over the weekend. It seems that it solved the overescaping issue, but didn't fully resolve the double quote issues. I would need to revisit it.

gumaerc (approved):
👍 

@umar8hassan I was able to fully get through the mass build locally and only a single course had a problem after running your markdown cleanup script in this branch, and it was `11-s945-urbanizing-china-a-reflective-dialogue-fall-2013`. I think we're good to merge this and we can fix the issue with that one course manually. Good job!

PR #2704, opened 2025-09-05
fix: do not add incorrect source for gdrive videos

zawan-ila:
https://github.com/mitodl/hq/issues/8354

For videos added through the `Add Video Resource` button, we set their `metadata.video_metadata.source` to "youtube" if they have a youtube Id. However, this code also runs if you try **editing** a "conventional" gdrive video resource that has been published to youtube. Hence, that gdrive resource is incorrectly marked as being of source "youtube", and in a [post save signal](https://github.com/mitodl/ocw-studio/blob/master/videos/signals.py#L77-L78), its file size is set to None.

### How can this be tested?

1. Create a gdrive video resource and push it through the [video workflows](https://github.com/mitodl/ocw-studio/blob/master/videos/README.md) to ensure it is uploaded to youtube. 
2. Back in studio, verify that it has the correct file size.
3. Open the resource in studio, and save it again. 
4. The file size has been incorrectly set to None
5. Checkout this branch and repeat the above steps. The file size should be unaffected by saving the resource in studio.

### DIFF
--- a/static/js/components/forms/SiteContentForm.test.tsx
+++ b/static/js/components/forms/SiteContentForm.test.tsx
@@ -156,7 +156,7 @@ test.each(EDITOR_STATES)(
   async (editorState) => {
     const data = setupData()
     data.content.type = "resource"
-    data.content.metadata.resourcetype = "Video"
+    data.content.metadata = { resourcetype: "Video" }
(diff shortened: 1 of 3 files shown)


umar8hassan (approved):
Looks Good 👍

PR #2723, opened 2025-09-16
differentiate build date and publish date

zawan-ila:
https://github.com/mitodl/hq/issues/7763

This PR changes how and when build_date and publish_date are recorded for a site. Currently, the build_date is being updated whenever a `publish` is carried out and the publish_date is updated whenever a concourse pipeline build succeeds.

This change makes it so that the build_date **only** updates when a concourse build is successful, and the publish_date is **only** changed when a concourse build **triggered by a publish action** is successful. It does not appear to be easy to track if a build was triggered by a publish action. To that end, the approach I have used is to use the build_id of the concourse pipeline to identify if a build was triggered through a publish action. When publishing a site, we store the `latest_build_id_draft` and `latest_build_id_live` fields on the website object. Upon receiving a notification from concourse, we extract the build_id from the notification and match it against the `latest_build_id_*` fields to determine if the build was triggered by a concourse action.

### How can this be tested?
1. Backpopulate Pipelines
2. Publish a site through Studio UI and wait for concourse build to finish
3. Verify that both its build_date and publish_date are updated
4. Rerun the site build through concourse UI
5. Verify that this only updates the build_date. The publish_date should not be affected
6. Run the mass build
7. Verify that this only updates the build_date. The publish_date should not be affected


umar8hassan:
@zawan-ila changes work as expected. However, `build_id` updates only when the publish action is performed. I'd expect the build_id to update on each run, as its name suggests, regardless of action type.

zawan-ila:
> can this be merged, or is there further work?

There are some small changes (updating the help text for the field) that need to be made.

umar8hassan (approved):
Looks Good 👍 
Just a few nitpicks suggested before the merge

PR #2726, opened 2025-09-19
fix: git author attribution for django admin edits and gdrive sync

zawan-ila:
https://github.com/mitodl/hq/issues/2753

Git author attribution is [determined](https://github.com/mitodl/ocw-studio/blob/968e8de22c61ba3e73551742be9a151a89227870/content_sync/apis/github.py#L337-L347) through the `updated_by` field on WebsiteContent objects. This field is set properly for edits made through studio, but is not set while making edits to a WebsiteContent object from admin, or when resources are synced through the gdrive workflow.

In this PR, we make changes to ensure that the `updated_by` field is properly set when making edits through django admin. Similarly, during gdrive sync, the `updated_by` field for any created/updated

### DIFF
--- a/gdrive_sync/api.py
+++ b/gdrive_sync/api.py
@@ -447,7 +447,7 @@ def get_pdf_title(drive_file: DriveFile) -> str:
 
 
 @transaction.atomic
-def create_gdrive_resource_content(drive_file: DriveFile):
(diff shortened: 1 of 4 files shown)


pdpinch:
Is this ready to merge?

PR #2749, opened 2025-10-03
fix: Add handling for escaped charaters in markdown link

umar8hassan:
https://github.com/mitodl/hq/issues/8322

This PR fixes incorrect escaping of backticks and square brackets in markdown link titles when converting them to Hugo shortcodes during the markdown cleaning process.

### How can this be tested?
1. Create test content with problematic markdown:
```
Test Content
[\`\[T\]his is a problematic title](https://example.com)
```

2. Run the markdown cleaner on content containing such links:
```
docker-compose exec web ./manage.py markdown_cleanup link_to_external_resource --f <website-name>
```

3. Verify the output shows properly unescaped shortcode:
```
{{% resource_link "uuid" "`[T]his is a problematic title" %}}
```

### DIFF
--- a/websites/management/commands/markdown_cleaning/rules/link_to_external_resource.py
+++ b/websites/management/commands/markdown_cleaning/rules/link_to_external_resource.py
@@ -29,11 +29,8 @@ def unescape_link_text(text: str) -> str:
     """
     Unescape markdown-escaped characters allowed in shortcode titles.
     """
-    if not text:
-        return text
     for escaped, literal in UNESCAPE_MAP.items():
-        if escaped in text:
-            text = text.replace(escaped, literal)
+        text = text.replace(escaped, literal)
     return text
 
 


websites/management/commands/markdown_cleaning/rules/link_to_external_resource.py
```diff
@@ -251,11 +251,18 @@ def replace_match(
         config = self.starter_lookup.get_config(starter_id)
         link_text = toks.link.text
 
+        # Unescape markdown characters in the link text for use in shortcode
+        # Markdown may escape backticks and square brackets that should not be
+        # escaped in the shortcode title parameter
+        unescaped_link_text = (
+            link_text.replace(r"\`", "`").replace(r"\[", "[").replace(r"\]", "]")
```
pt2302, line 258: It would be better if this were a utility function, defined elsewhere. Otherwise, there is code duplication here and below.

pt2302: Updated in 0999ad92f1b567e07955c4553f3c40e3255c5ec7.

PR #2753, opened 2025-10-14
Fix some metadata tracking for website builds

zawan-ila:
https://github.com/mitodl/hq/issues/7763

1. Prevent offline build webhooks from modifying the `build_date` field on Website objects. The `build_date` will only be updated for online builds
2. Currently, on a successful publish, the `last_published_by` fields are set to None. We change the behavior to preserve the `last_published_by` field. This makes it a bit difficult to prevent mails to publishers on mass builds, as the mail sending is gated by the truthiness of the `last_published_by` field. To get around this, we make use of the `latest_build_id` fields by only sending mails for build ids matching those. This presents another problem, which is that we receive 2 webhooks for successful online builds, once when clearing the cdn cache, and another when the hugo build is successful. Hence, mails are sent twice per build. To work around this, we also prevent cdn-cache-clear webhooks from triggering mails.
3.  Ensures that the `publish_date` fields are correctly updated on mass-publish.

### How can this be tested?
1. The cdn-cache-clear steps are not included in local concourse pipelines. Change the pipeline code so that they are. For that, you may apply this [patch](https://gist.github.com/zawan-ila/6abcc42b4a1415b63941c452dec994a9).
2. Backpopulate site pipelines and upsert mass build pipelines to ensure they have the correct pipeline definitions
3. Publish a website from studio
4. Verify that the publish_date and build_date are updated correctly. Verify that the `last_published_by` field is not reset to None. Also, verify that the email is sent on publish. You can do this by checking the container logs. It shows as an error there as we haven't configured email locally. Also verify that the email is only sent once per publish.
5. Verify that the offline build runs and does not update the build_date fields
6. Run an online mass build. Verify that the build_dates are updated but not the publish_dates
7. Run an offline mass build. Verify that neither build_dates not publish_dates are updated.
8. Run a mass publish. Verify that the publish_dates are updated. Verify also that the `build_dates` are updated.

### DIFF
--- a/content_sync/pipelines/definitions/concourse/common/steps.py
+++ b/content_sync/pipelines/definitions/concourse/common/steps.py
@@ -202,6 +202,8 @@ class OcwStudioWebhookStep(TryStep):
         pipeline_name(str): The name of the pipeline to set the status on
         status: (str): The status to set on the pipeline (failed, errored, succeeded)
         build_type: (str, optional): The type of build ('online' or 'offline')
+        is_cdn_cache_step(bool, optional): Whether this step is being called from
+                                           a cdn cache purge step
     """
 
     def __init__(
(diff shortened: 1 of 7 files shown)


umarhassan-arbisoft (dismissed):
Looks Good 👍

PR #2755, opened 2025-10-16
Associate captions/transcripts with existing videos

pt2302:
Part of https://github.com/mitodl/hq/issues/6483.

This PR adds `Video Captions Resource` and `Video Transcript Resource` fields to the `Edit Resource` menu, allowing for course authors to manually update captions and transcript resources for an already-uploaded video (in contrast to the workflow for videos and captions/transcript being uploaded at the same time, as in https://github.com/mitodl/ocw-studio/pull/1683).

### How can this be tested?
First, paste the config file from https://github.com/mitodl/ocw-hugo-projects/blob/pt/associate_existing_captions/ocw-course-v2/ocw-studio.yaml into `ocw-course` in `Website starters` in Django admin. Then, either upload a new video or select an existing video. Update the `Video Captions Resource` and `Video Transcript Resource` fields, and make sure that the URLs are properly updated once the video resource is saved. Then, publish the course, and verify that the captions and transcript have been properly updated.

### DIFF
--- a/websites/serializers.py
+++ b/websites/serializers.py
@@ -54,8 +54,8 @@ ROLE_ERROR_MESSAGES = {"invalid_choice": "Invalid role", "required": "Role is re
 
 
 RELATION_URL_FIELDS = (
-    ("video_files.video_captions_resource", settings.YT_FIELD_CAPTIONS),
-    ("video_files.video_transcript_resource", settings.YT_FIELD_TRANSCRIPT),
+    (settings.YT_FIELD_CAPTIONS_RESOURCE, settings.YT_FIELD_CAPTIONS),
+    (settings.YT_FIELD_TRANSCRIPT_RESOURCE, settings.YT_FIELD_TRANSCRIPT),
 )
 
 
--- a/main/settings.py
+++ b/main/settings.py
@@ -572,8 +572,14 @@ FIELD_METADATA_TITLE = get_string(
 YT_FIELD_CAPTIONS = get_string(
     name="YT_FIELD_CAPTIONS",
     default="video_files.video_captions_file",
-    description="The site config metadata field for the caption url",
+    description="The site config metadata field for the captions URL",
 )
+YT_FIELD_CAPTIONS_RESOURCE = get_string(
+    name="YT_FIELD_CAPTIONS_RESOURCE",
+    default="video_files.video_captions_resource",
+    description="The site config metadata field for the captions resource",
+)
+
 YT_FIELD_ID = get_string(
     name="YT_FIELD_ID",
     default="video_metadata.youtube_id",
@@ -597,12 +603,17 @@ YT_FIELD_TAGS = get_string(
 YT_FIELD_THUMBNAIL = get_string(
     name="YT_FIELD_THUMBNAIL",
     default="video_files.video_thumbnail_file",
-    description="The site config metadata field for YouTube thumbnail url",
+    description="The site config metadata field for YouTube thumbnail URL",
 )
 YT_FIELD_TRANSCRIPT = get_string(
     name="YT_FIELD_TRANSCRIPT",
     default="video_files.video_transcript_file",
-    description="The site config metadata field for the transcript url",
+    description="The site config metadata field for the transcript URL",
+)
+YT_FIELD_TRANSCRIPT_RESOURCE = get_string(
+    name="YT_FIELD_TRANSCRIPT_RESOURCE",
+    default="video_files.video_transcript_resource",
+    description="The site config metadata field for the transcript resource",
 )
 
 
--- a/websites/serializers_test.py
+++ b/websites/serializers_test.py
@@ -917,12 +917,12 @@ def test_website_content_detail_serializer_syncs_video_relation_files(
     metadata_patch = {"video_files": {}}
     set_dict_field(
         metadata_patch,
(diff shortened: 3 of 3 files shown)


static/js/components/widgets/RelationField.tsx
```diff
@@ -133,10 +133,29 @@ export default function RelationField(props: Props): JSX.Element {
...
+  type WrappedRelationValue = { website?: string; content: string | string[] }
+  const effectiveValue = useMemo(() => {
+    if (
+      !crossSite &&
+      value &&
+      typeof value === "object" &&
+      !Array.isArray(value) &&
+      "content" in (value as any)
+    ) {
+      return (value as WrappedRelationValue).content
```
ChristopherChudzicki: **TLDR:** I think we should do this differently. The value passed to RelationField should not be an object.

Two things tipped me off:
- the types here clearly indicate that this should be a string or array of strings (latter case only if config has `multiple: true`, I imagine)
- In slack you wrote
    > yeah what's new in this PR is to actually permit for a single option that's not cross-site... as far as I could tell
Looking at the configs, I noticed that we actually do use a single-option, non-cross-site relationfield in one other place, and it's working fine there. **The course image field in site metadata**.

OK, so why is this behaving differently than the course image field—you're getting an object when we expect a string. The difference is that here we're using it inside an ObjectField.

In general, for our form fields:
1. fields have a `name` (like `<input type="text" name="field-name" />` in html)
2. when we edit them, the WebsiteContent object is altered, roughly to `websiteContent[name] = value`
3. The form handlers we use come from Formik, which apparently handles updating nested objects properly as long as the field name is like `video_metadata.youtube_id` (or whatever... just, `.`-separated.).
    - This happens in `renameNestedFields` function.
4. ....But apparently RelationField also does something special with the name for validation....It appends `.content`. So on our forms, if a field is a RelationWidget, then its name isn't actually `field_name`, it's name is `field_name.content`. (Extra `.content`). 
    - I don't entirely understand this yet? And it's late. RelationWidget says it's something to do with validation?
    - This also happens in `renameNestedFields` function

But (3) and (4) seem not be working quite right together... In particular, nested relation fields don't get the `.content` suffix.

**Which is all to say...:** If we fix `renameNestedFields` so that nested relation fields DO get the content suffix, I think everything will work fine / as intended.

For example:  https://github.com/mitodl/ocw-studio/compare/cc/relationfield-inside-objectfield ... Possibly this is the only frontend change we need for this PR? And some tests to go with it.

websites/serializers.py
```diff
@@ -41,14 +41,66 @@
...
+    if isinstance(value, list):
+        first = value[0] if value else None
+        return first[0] if isinstance(first, list | tuple) else first
+    return value if isinstance(value, str) else None
+
+
+def resource_file_path(resource: WebsiteContent | None) -> str | None:
+    if not resource:
+        return None
+    if getattr(resource, "file", None) and getattr(resource.file, "name", None):
```
ChristopherChudzicki: could use hasattr i think

ChristopherChudzicki (changes requested):
@pt2302 I didn't really look at the backend much, and I haven't successfully published a course. (It failed... I didn't look into why. GDrive is working nicely, though).

I left a long comment about RelationWidget... I wish this had been more explicit.... 

**Request:** Take a look at the suggested alternative frontend change. From what I can tell, it's more how this was supposed to work. But, like I said, I haven't fully tested your PR, so there could be an issue with it.

websites/serializers.py
pt2302: Simplified, although still using `getattr`, in 922d4b04c94b5bcccbf3277358c5b4726e42dd2d.

static/js/components/widgets/RelationField.tsx
pt2302: Updated in 2b90de0ead75cdf1347294c04e0277bfc0499037 and test added in 663aaf7220cdec5c66c9a268fc97ff62a48e90dc.

websites/serializers.py
```diff
@@ -41,14 +41,65 @@
...
+
+def relation_value_to_id(value) -> str | None:
+    if not value:
+        return None
+    if isinstance(value, dict):
+        return value.get("content") or None
+    if isinstance(value, list):
+        first = value[0] if value else None
+        return first[0] if isinstance(first, list | tuple) else first
+    return value if isinstance(value, str) else None
```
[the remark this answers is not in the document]
ChristopherChudzicki: @pt2302 Actually, I think I misunderstood how relation widget values are being stored in the database.

It seems like we always store relation widget values as 
```
{
  content: string or string[] or CrossSitePair[]
  website: uuid of resource's website 
}
```
which I did not understand.

### New question
Does ☝️ sound right to you? It seems correct based on resource lists and site thumbnail data, as well as the new video resource /trnascript data.

In that case, I think only the `if isinstance(value, dict):` line is ever triggered?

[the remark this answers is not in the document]
pt2302: Updated in 9bf776913610569924463430494e227530a33da1.

ChristopherChudzicki (approved):
👍 Looks good.

As mentioned, I have not tested publishing, but feel comfortable with this: publishing was not changed and the database data updates as I would expect.

But definitely check publishing on RC!

PR #2766, opened 2025-10-22
chore: added tasks in offline gate step to remove site from offline search index

umar8hassan:
closes https://github.com/mitodl/hq/issues/4473

This is part of an ongoing effort to exclude courses with `hide_download` set to `True`. This pull request aims to remove individual courses from the offline search index, which requires the root site (default `ocw-www`) to be rebuilt.

### How can this be tested?
1. switch to this branch and upsert course pepline for a test course (rebuild if necessary)
2. set the `Hide Download` to False and publish site. verify in MinIO that test site has offline package
3. set the `Hide Download` to True and re-ublish site. 
4. Verify that the zip package has been removed, websiteCpntent object of `website` type has been deleted, and it triggered a publish for root site as well.

### DIFF
--- a/content_sync/pipelines/definitions/concourse/site_pipeline.py
+++ b/content_sync/pipelines/definitions/concourse/site_pipeline.py
@@ -885,27 +885,25 @@ class SitePipelineDefinition(Pipeline):
         resources = SitePipelineResources(config=config)
         online_job = self.get_online_build_job(config=config)
 
-        # Only add offline gate step for non-root websites
+        # Create the inner put step with error handlers
+        inner_put_step = PutStep(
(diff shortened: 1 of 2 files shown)


gumaerc (changes requested):
This successfully removed the `WebsiteContent` that it needed to. However, rebuilding the root website does not work. There is also a slight security issue:

gumaerc (approved):
Everything works well with the latest changes. I toggled the flag and published a couple times, and `ocw-www` was rebuilt each time. The `WebsiteContent` object was properly removed / added as necessary and I didn't see the bearer token logged this time.

PR #2993, opened 2026-04-14
fix: do not change url for section pages

zawan-ila:
https://github.com/mitodl/hq/issues/10832

See [comment](https://github.com/mitodl/hq/issues/10832#issuecomment-4243181190)

### How can this be tested?
1. Create a section page in a site. This is a page with a filename of _index.md (like [this](https://ocw-studio-rc.odl.mit.edu/sites/pe-710-tennis-spring-2007/type/page/edit/de168d39-f840-f977-8435-fb3ef4fd9530/)) I don't think the studio interface supports building such pages and these are for legacy usage only.
2. Make some small edit to the page and save
3. Verify that the page url does not change.
4. Build the course and verify that the section page renders correctly.

### DIFF
--- a/drf_lint_baseline.json
+++ b/drf_lint_baseline.json
@@ -1,7 +1,5 @@
 [
   "content_sync/serializers.py:133:21:ORM001",
   "content_sync/serializers.py:54:21:ORM001",
-  "websites/serializers.py:257:23:ORM001",
-  "websites/serializers.py:683:16:ORM001",
-  "websites/serializers.py:700:20:ORM002"
+  "websites/serializers.py:257:23:ORM001"
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -177,6 +177,7 @@ ignore = [
     "RUF012",
     "UP007"
 ]
+external = ["ORM001", "ORM002"]
 
 [tool.ruff.lint.pydocstyle]
 convention = "pep257"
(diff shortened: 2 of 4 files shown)


pt2302:
Two questions/comments:
- what are the `ORM001` and `ORM002` directives, and what's the purpose of adding `external = ["ORM001", "ORM002"]` to `pyproject.toml`?
- it would be cleaner to move "_index" to a constants file rather than having it inline in the function.

zawan-ila:
> what are the ORM001 and ORM002 directives, and what's the purpose of adding external = ["ORM001", "ORM002"] to pyproject.toml?

These directives are from drf-lint. This [linter](https://github.com/mitodl/ocw-studio/pull/2965) is responsible for detecting N+1 queries. ORM001 and ORM002 are rules in the linter. 

Since this PR affects serializers.py, the line numbers for some of the existing exceptions added in drf-lint-baseline.json(which contains exceptions for the linter) change, which cause the linter to fail. That is why we add noqas for these. I have now removed the releated lines from drf-lint-baseline.json as well.

Now ruff does not like noqa codes that it can not recognize, in this case ORM001 and ORM002. Thus, we need to tell rust that these are codes for external linting tools with the `external` [directive](https://docs.astral.sh/ruff/rules/invalid-rule-code/#options).


> it would be cleaner to move "_index" to a constants file rather than having it inline in the function

done

pt2302 (approved):
Looks good to me!

PR #2999, opened 2026-04-21
feat: migarate captions & transcript file fields to resource fields

umar8hassan:
https://github.com/mitodl/hq/issues/10958

Removes the legacy `video_captions_file` and `video_transcript_file` string fields from the `video_files` metadata block on video resources.

### How can this be tested?
1. Switch to `main` branch
2. Visit a course site in Studio that has an existing video resource with captions or transcripts already linked.
  - OR if you have working 3play locally, you can add a video resource with `youtube_id: E8uZtq_vOYM` and it'd pull the related transcript and captions.
3. Switch to the branch: `umar/10958-remove-captions-and-transcript-file-fields`
4. Add content to starter config in admin from `cow-hugo-projects` branch: `umar/10958-remove-captions-and-transcript-file-fields`
5. Open the video resource editor, confirm that `video_captions_file` and `video_transcript_file` text fields are no longer present in the form.
6. Rebuild containers or run `docker-compose exec web python manage.py migrate` and verify migration `0073` applies cleanly without errors.
7. After migration, open the same video resource and confirm the linked caption/transcript content still appears correctly under the relation fields.

### DIFF
--- a/websites/migration_0074_test.py
+++ b/websites/migration_0074_test.py
@@ -1,0 +1,245 @@
+"""Tests for website migration 0073 — back-populate _resource fields from file-path lookup"""
+
+import importlib
+import uuid
+
+import pytest
+from django.apps import apps
+
+from websites.factories import WebsiteContentFactory, WebsiteFactory
+
+pytestmark = pytest.mark.django_db
+
+
+@pytest.fixture
+def migration_module():
+    """Import the migration module (names with leading digits need importlib)."""
+    return importlib.import_module(
+        "websites.migrations.0074_video_caption_transcript_files_to_resources"
+    )
+
+
+def test_migration_0073_backpopulates_resource_fields_from_file_path(
+    migration_module,
+):
+    """_resource fields are back-populated by matching the legacy _file path to WebsiteContent.file."""
+    website = WebsiteFactory.create()
+
+    captions_path = f"courses/{website.name}/abc_captions.vtt"
+    transcript_path = f"courses/{website.name}/abc_transcript.pdf"
+
+    captions = WebsiteContentFactory.create(
+        website=website,
+        filename="abc_captions_vtt",
+        type="resource",
+        file=captions_path,
+    )
+    transcript = WebsiteContentFactory.create(
+        website=website,
+        filename="abc_transcript_pdf",
+        type="resource",
+        file=transcript_path,
+        metadata={"resourcetype": "Document"},
+    )
+
+    video = WebsiteContentFactory.create(
+        website=website,
+        type="resource",
+        metadata={
+            "resourcetype": "Video",
+            "video_files": {
+                "video_captions_file": f"/{captions_path}",
+                "video_transcript_file": f"/{transcript_path}",
+            },
+        },
+    )
+
+    migration_module._backpopulate_resource_fields(apps, None)  # noqa: SLF001
+
+    video.refresh_from_db()
+    vf = video.metadata["video_files"]
+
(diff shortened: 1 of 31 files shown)


zawan-ila (commented):
I feel that we need to think more about https://github.com/mitodl/hq/issues/10958#issuecomment-4388622424. There appear to be a large number (~1000) of captions files (and most likely transcripts too) that do not have associated resources.

The way this migration is currently implemented will end up removing that caption file data from the resource objects and not be able to attach it to the video_captions_resource field (as there are no associated resources), **causing us to lose captions information for a bunch of video resources**. 

[PS] Please also fix the CI

websites/migrations/0074_video_caption_transcript_files_to_resources.py
```diff
@@ -0,0 +1,191 @@
...
+reliable discriminator for transcript PDFs.
+
+For captions, the same path-based lookup is used. The .vtt/.webvtt extension in the
+path is a reliable signal; no resourcetype filter is needed.
+
+The _resource fields are written in single-select format for now:
+  video_captions_resource   = {"content": "<text_id>", "website": "<name>"}
+  video_transcript_resource = {"content": "<text_id>", "website": "<name>"}
+
+Migration 0074 converts these to the multi-select list format used by the relation widget.
```
zawan-ila: 0075?

websites/migrations/0074_video_caption_transcript_files_to_resources.py
```diff
@@ -0,0 +1,191 @@
...
+                WebsiteContent.objects.filter(
+                    website=content.website,
+                    metadata__resourcetype="Document",
+                    file=transcript_path.lstrip("/"),
+                ).first()
+                or WebsiteContent.objects.filter(
+                    website=content.website,
+                    metadata__resourcetype="Document",
+                    metadata__file=transcript_path,
+                ).first()
```
zawan-ila: Why not use _find_resource_by_path? Is it some kind of a perf optimization?

websites/migration_0074_test.py
```diff
@@ -0,0 +1,245 @@
+"""Tests for website migration 0073 — back-populate _resource fields from file-path lookup"""
```
zawan-ila, line 1: Can we move this file to the migrations folder?

zawan-ila (approved):
Some minor suggestions/questions.

websites/migrations/0074_video_caption_transcript_files_to_resources.py
umar8hassan: because it needed `metadata__resourcetype="Document"`, a filter `_find_resource_by_path` didn't support. Rather than keep the duplication, extend `_find_resource_by_path` with an optional `resourcetype` kwarg.

updated in https://github.com/mitodl/ocw-studio/pull/2999/commits/f2df25eafdf50e2eec69fc8dae3591583df242d8

websites/migration_0074_test.py
umar8hassan: No, placing it in `websites/migrations/` will cause Django's `MigrationLoader` to attempt to load it as a migration.

websites/migrations/0074_video_caption_transcript_files_to_resources.py
umar8hassan: updated in https://github.com/mitodl/ocw-studio/pull/2999/commits/f2df25eafdf50e2eec69fc8dae3591583df242d8

PR #3030, opened 2026-05-04
feat: ingest drive files for multi language

umar8hassan:
https://github.com/mitodl/hq/issues/10983

Adds language and locale parsing for caption and transcript filenames ingested from Google Drive, and auto-links caption/transcript resources to their video during GDrive ingestion.

### How can this be tested?
1. In Google Drive, upload a caption file named `<video_title>_captions_en_US.vtt` and a French one named `<video_title>_captions_fr.vtt` for the same video. Trigger a GDrive sync.
2. Open the corresponding video resource in Studio. Confirm both captions appear in `video_captions_resource` as separate linked resources, each with correct language.
3. Publish the site (draft) and inspect the generated YAML front-matter for the video, and verify the English entry has `locale: US` and the French entry has no `locale` key.
4. Open a video resource that already had a single English caption before this change, and confirm the existing entry still renders correctly with `language: en` and no `locale`.
5. Trigger a GDrive sync for a site that has video files, and confirm that after ingestion, the video resource's `video_captions_resource` and `video_transcript_resource` fields are populated without any manual editor action.

### DIFF
--- a/drf_lint_baseline.json
+++ b/drf_lint_baseline.json
@@ -1,5 +1,5 @@
 [
   "content_sync/serializers.py:139:21:ORM001",
   "content_sync/serializers.py:58:21:ORM001",
-  "websites/serializers.py:269:23:ORM001"
+  "websites/serializers.py:266:23:ORM001"
 ]
--- a/gdrive_sync/api.py
+++ b/gdrive_sync/api.py
@@ -42,7 +42,7 @@ from gdrive_sync.constants import (
 )
 from gdrive_sync.models import DriveFile
 from main.s3_utils import get_boto3_resource
-from main.utils import get_dirpath_and_filename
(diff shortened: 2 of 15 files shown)


zawan-ila (commented):
Following the testing instructions does not produce the desired results. I added the two captions files as directed and then ran a sync. Only one of the resources showed up in the `video_captions_resources` field, instead of the expected two.

This is because at a number of places, we are still assuming a single-language world. For example in `link_resource_to_video`, we do

```
    video_files[resource_field] = {
        "content": [str(resource.text_id)],
        "website": website.name,
    }
```

instead of preserving any exising text_ids that may already exist (i.e we should append to the content list)

zawan-ila (approved):

PR #3032, opened 2026-05-06
fix: use correct base url for learn deployed v3 theme

zawan-ila:
https://github.com/mitodl/hq/issues/11124

ocw-course-v3 sites are currently built with a baseUrl like `/ocw-course-v3/courses/12-108-fall-2004`. However, in Learn, these courses are served under a different path `/courses/o/12-108-fall-2004`. Because of this mismatch, Hugo generates incorrect internal links. For example, navigation links resolve to `/ocw-course-v3/courses/12-108-fall-2004/pages/...` instead of the expected `/courses/o/12-108-fall-2004/pages/...`

This results in broken or incorrect navigation when users browse course pages on Learn. You can reproduce the issue at https://rc.learn.mit.edu/courses/o/7-016-introductory-biology-fall-2018 by clicking on any navigation link and observing the resulting urls in the browser bar (they'll include `ocw-course-v3`)

### How can this be tested?
1. Set OCW_EXTRA_COURSE_THEMES=ocw-course-v3 in your env file
2. Set OCW_EXTRA_THEMES_BASE_URLS={"ocw-course-v3": "/courses/o"} in your env file
2. Backpopulate pipelines `./manage.py backpopulate_pipelines`
3. Trigger a publish from studio
4. Verify that both the v3 and v2 pipelines in concourse run
5. Verify from the concourse task logs that baseURL passed to hugo for the v3 pipeline follows the `courses/o/<course_name>` format.
6. Verify that the baseURL passed to hugo is unaffected by these changes for the v2 site pipeline.

### DIFF
--- a/content_sync/pipelines/definitions/concourse/site_pipeline.py
+++ b/content_sync/pipelines/definitions/concourse/site_pipeline.py
@@ -78,12 +78,6 @@ BUILD_OFFLINE_SITE_IDENTIFIER = Identifier("build-offline-site").root
 UPLOAD_OFFLINE_BUILD_IDENTIFIER = Identifier("upload-offline-build").root
 CLEAR_CDN_CACHE_IDENTIFIER = Identifier("clear-cdn-cache").root
 
-# Extra themes can be served under a different published base path than the
-# source site path used in Studio.
-EXTRA_THEME_BASE_URL_PREFIXES = {
-    "ocw-course-v3": "/courses/o",
-}
-
 
 def get_site_pipeline_definition_vars(namespace: str):
     return {
@@ -127,7 +121,7 @@ def get_base_url_path(
     is_root_website: bool,
 ) -> str:
     """Get the Hugo baseURL path for a site pipeline."""
-    extra_theme_base_url_prefix = EXTRA_THEME_BASE_URL_PREFIXES.get(prefix)
+    extra_theme_base_url_prefix = settings.OCW_EXTRA_THEMES_BASE_URLS.get(prefix)
     if is_extra_theme and not is_root_website and extra_theme_base_url_prefix:
         root_prefix = site_root_path.strip("/")
         course_path = base_url
--- a/content_sync/pipelines/definitions/concourse/site_pipeline.py
+++ b/content_sync/pipelines/definitions/concourse/site_pipeline.py
@@ -216,6 +216,18 @@ class SitePipelineDefinitionConfig:
         base_url_path = (
             f"/{self.prefix}/{self.base_url}" if self.prefix else f"/{self.base_url}"
         )
+
+        # ocw-course-v3 extra-theme course pages are served on Learn under
+        # /courses/o/<course-path>, so generated Hugo URLs must use that path.
+        if (
+            self.prefix == "ocw-course-v3"
+            and self.is_extra_theme
+            and not self.is_root_website
+        ):
+            base_url_path = f"/courses/o/{
+                self.base_url.removeprefix('/courses/').removeprefix('courses/')
+            }"
+
         base_online_args.update(
             {
                 "--config": f"../{OCW_HUGO_PROJECTS_GIT_IDENTIFIER}/{starter_slug}/config.yaml",  # noqa: E501
--- a/content_sync/pipelines/definitions/concourse/site_pipeline.py
+++ b/content_sync/pipelines/definitions/concourse/site_pipeline.py
@@ -213,20 +213,20 @@ class SitePipelineDefinitionConfig:
         starter_slug = theme_slug if theme_slug else site.starter.slug
         base_hugo_args = {"--themesDir": f"../{OCW_HUGO_THEMES_GIT_IDENTIFIER}/"}
         base_online_args = base_hugo_args.copy()
-        base_url_path = (
-            f"/{self.prefix}/{self.base_url}" if self.prefix else f"/{self.base_url}"
-        )
-
-        # ocw-course-v3 extra-theme course pages are served on Learn under
-        # /courses/o/<course-path>, so generated Hugo URLs must use that path.
         if (
             self.prefix == "ocw-course-v3"
             and self.is_extra_theme
             and not self.is_root_website
         ):
-            base_url_path = f"/courses/o/{
-                self.base_url.removeprefix('/courses/').removeprefix('courses/')
-            }"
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -170,7 +170,7 @@ services:
     - concourse-keys:/concourse-keys
 
   concourse:
-    image: concourse/concourse:7.14
+    image: concourse/concourse:8.0
```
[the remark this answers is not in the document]
zawan-ila: Added a comment in the description

content_sync/pipelines/definitions/concourse/site_pipeline.py
```diff
@@ -205,9 +205,21 @@ def __init__(  # noqa: PLR0913 PLR0915
...
+            # ocw-course-v3 extra-theme course pages are served on Learn under
+            # /courses/o/<course-path>, so generated Hugo URLs must use that path.
+            base_url_path = f"/courses/o/{self.base_url.removeprefix('courses/')}"
+        else:
+            base_url_path = (
+                f"/{self.prefix}/{self.base_url}"
+                if self.prefix
+                else f"/{self.base_url}"
+            )
+
```
pt2302: This logic should be generalized, perhaps using a helper function or environment variables to set the prefixes. Otherwise, it's putting too much deployment logic in the pipeline definition.

content_sync/pipelines/definitions/concourse/site_pipeline_test.py
```diff
@@ -1030,6 +1030,71 @@ def test_site_pipeline_definition_config_theme_slug(settings, mocker, theme_slug
...
+        theme_slug="ocw-course-v3",
+    )
+
+    assert expected_template.format(base_url=config.base_url) in config.hugo_args_online
+    assert (
+        unexpected_template.format(base_url=config.base_url)
+        not in config.hugo_args_online
+    )
+
+
```
pt2302, line 1195: It would be better to tighten this test to check that the exact URL that is necessary is being produced. Something like

```
starter.config["root-url-path"] = "courses"
website = WebsiteFactory.create(
    starter=starter,
    url_path="courses/my-course",
)

config = SitePipelineDefinitionConfig(...)

assert "--baseURL /courses/o/my-course" in config.hugo_args_online
assert "--baseURL /courses/o/courses/my-course" not in config.hugo_args_online
```

docker-compose.yml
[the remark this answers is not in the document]
pt2302: It's not ideal to include this in one PR. But it would also be helpful to document exactly what errors with v7 were happening, and whether this is related to changes in this PR.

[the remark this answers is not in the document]
zawan-ila: Removed in https://github.com/mitodl/ocw-studio/pull/3032/changes/4dba5c4fc56ea1297982968c03c7f5b5f9c580ec

It was unrelated to the changes in this PR. I don't recall there being any errors. Just that the concourse worker would exit abruptly. I had recently switched to [Colima](https://colima.run/) so thought that it may be related to that.

I'll open a separate PR if I still face issues with 7.14 locally.

content_sync/pipelines/definitions/concourse/site_pipeline.py
zawan-ila: Added a helper function in https://github.com/mitodl/ocw-studio/pull/3032/changes/4dba5c4fc56ea1297982968c03c7f5b5f9c580ec

content_sync/pipelines/definitions/concourse/site_pipeline_test.py
zawan-ila: Made the test more explicit in https://github.com/mitodl/ocw-studio/pull/3032/changes/4dba5c4fc56ea1297982968c03c7f5b5f9c580ec

content_sync/pipelines/definitions/concourse/site_pipeline.py
```diff
@@ -77,6 +77,12 @@
 UPLOAD_OFFLINE_BUILD_IDENTIFIER = Identifier("upload-offline-build").root
 CLEAR_CDN_CACHE_IDENTIFIER = Identifier("clear-cdn-cache").root
 
+# Extra themes can be served under a different published base path than the
+# source site path used in Studio.
+EXTRA_THEME_BASE_URL_PREFIXES = {
+    "ocw-course-v3": "/courses/o",
+}
+
```
pt2302: This mapping should be an environment variable that is set, rather than hardcoded directly in the pipeline. Otherwise, the original bug could occur for any other theme that would be added in the future. I would suggest using the function `get_dict_of_str` from https://github.com/mitodl/ocw-studio/pull/3043 once that is merged.

pt2302 (changes requested):
The mapping from `EXTRA_THEMES` to `BASE_URL_PREFIXES` should be an environment variable, rather than being hardcoded in code.

content_sync/pipelines/definitions/concourse/site_pipeline.py
zawan-ila: Switched to using an env var in https://github.com/mitodl/ocw-studio/pull/3032/commits/86e46df0c70ad2c124d74fc4136fdfe09626ff08. I have also put up https://github.com/mitodl/ol-infrastructure/pull/4670 to add the env var to the deployment.

umar8hassan (changes requested):
mostly looks good. However, i am concerned about the s3 paths for built files which are still pointing to `/ocw-course-v3/..` instead of `../o/..`.

I built sites locally and the content was still being served from `/ocw-corse-v3/..` while `/o/..` threw `404` error. unless i am missing something, this would require us to proxy paths in CDN or update S3 paths.

Image
Image

zawan-ila:
@umar8hassan You are right. This needs some proxying which learn.mit.edu has already implemented. However, in the current ocw-studio setup, we do not replicate that.

If you want to test the flow end to end, I would suggest taking the build output from the studio build (from minio) and then following the testing instructions from step 4 onwards at https://github.com/mitodl/ocw-hugo-themes/pull/1797. You may also be able to use https://github.com/mitodl/ocw-hugo-themes/blob/main/package_scripts/build-course-v3-prefixed.ts if you are fine with testing the hugo build separately.

umar8hassan (approved):
Looks Good 👍 

Image
ss2

PR #3041, opened 2026-05-18
chore: added de-duplication and pre-filtering for YT tags update

umar8hassan:
https://github.com/mitodl/hq/issues/8719

Optimizes the `update_youtube_tags` management command to add pre-filtering and deduplicated tasks in Kubernetes due to POD restarts and releases.

**Automatic scheduling via Celery:** When `--schedule` is passed and the video count exceeds `--threshold` (default: 150), the command dispatches `update_youtube_tags_batch` Celery tasks with staggered `countdown` delays (24h apart) so work is spread across days automatically. A `--weekends-only` flag restricts scheduled tasks to Saturday/Sunday slots.

**Pre-filtering before scheduling:** Before dispatching any Celery tasks, the command fetches current YouTube tags for all candidate videos and uses `compute_merged_tags` to identify which videos actually need updating. Videos already up to date are excluded from the schedule entirely, saving both task slots and future execution quota. The cost is only ~1 quota unit per 50 videos at dispatch time.

**Celery task deduplication (`update_youtube_tags_batch`):**
- Removed `acks_late=True`: the root cause of duplicate task delivery in Kubernetes. Without it, the broker acknowledges the message on receipt so pod restarts no longer redeliver parked ETA tasks.
- Added `@single_task(timeout=30min, raise_block=False)`: enforces a cluster-wide Redis mutex so only one batch task runs at a time, even if duplicate messages do arrive simultaneously.

**Command-line options (no env vars needed):**
- `--schedule` — enable Celery dispatch when above threshold
- `--threshold N` — video count that triggers scheduling (default: 150)
- `--daily-quota N` — quota budget per day for scheduling math (default: 9000)
- `--weekends-only` — restrict tasks to weekends
- `--quota-limit N` — stop processing before exceeding this many units
- `--add-course-tag` — add course URL slug as a tag
- `--out FILE` — export results to CSV

### How can this be tested?
1. **Verify `@single_task` deduplication (Django shell):**

Dispatch the same batch task twice with a short ETA so both messages land before either runs. Only the first to acquire the Redis lock should execute; the second should log a skip and return immediately.

```python
# python manage.py shell
from videos.tasks import update_youtube_tags_batch

# Use a real YouTube ID that exists in your DB, or any string — the lock check
# happens before any API calls so a fake ID is fine for testing the mutex.
fake_ids = ["test-yt-id-1", "test-yt-id-2"]

# Dispatch two copies of the same batch with a 5-second delay
update_youtube_tags_batch.apply_async(args=[fake_ids], countdown=5)
update_youtube_tags_batch.apply_async(args=[fake_ids], countdown=5)

print("Two tasks dispatched. Watch the Celery worker logs.")
```

Expected worker logs (one worker, ~5 seconds later):
```
# First message — lock acquired, task runs:
[INFO] YouTube tag batch complete: 0 updated, 0 skipped, 0 errors

# Second message — lock already held by first:
[INFO] update_youtube_tags_batch skipped: another instance is already running ...
```
If both messages arrive after the first has already finished, the second will acquire the lock and run (correctly re-processing — tags unchanged → 0 updates, 0 quota wasted on `videos.update`). Either outcome confirms the mutex is working.

### DIFF
--- a/videos/constants.py
+++ b/videos/constants.py
@@ -23,6 +23,11 @@ ARCHIVE_URL_FILESIZE_TASK_RATE_LIMIT = "0.1/s"
 S3_FILESIZE_TASK_RATE_LIMIT = "5/s"
 
 
+# Lock TTL for the update_youtube_tags_batch single_task mutex.
+# Long enough for a full batch to complete; short enough to self-heal after a crash.
+YTAGS_BATCH_LOCK_TTL = 30 * 60  # 30 minutes
+
+
 class VideoStatus:
     """Simple class for possible Video statuses"""
 
--- a/videos/management/commands/update_youtube_tags.py
+++ b/videos/management/commands/update_youtube_tags.py
@@ -381,7 +381,32 @@ class Command(WebsiteFilterCommand):
             next_saturday += timedelta(days=7)
         return int((next_saturday - now).total_seconds())
 
-    def _schedule_celery_tasks(  # noqa: C901, PLR0912, PLR0913, PLR0915
+    def _prefilter_videos(self, all_youtube_ids, yt_id_to_resources, add_course_tag):
+        """
+        Fetch current YouTube tags and return only the IDs that need updating.
+
+        Calls videos.list in batches (~1 quota unit per 50 IDs) and excludes
+        videos whose merged tags would be identical to their current YouTube
+        tags.  Returns (filtered_ids, list_quota_used).
+        """
+        youtube = YouTubeApi()
+        snippets = fetch_youtube_snippets(youtube, all_youtube_ids)
+        list_quota_used = (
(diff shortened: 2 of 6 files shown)


videos/constants.py
```diff
@@ -23,6 +23,11 @@
 S3_FILESIZE_TASK_RATE_LIMIT = "5/s"
 
 
+# Lock TTL for the update_youtube_tags_batch single_task mutex.
+# Long enough for a full batch to complete; short enough to self-heal after a crash.
+YTAGS_BATCH_LOCK_TTL = 30 * 60  # 30 minutes
```
zawan-ila, line 28: [nit] The name can be improved e.g YOUTUBE_TAGS_LOCK_TTL

zawan-ila (approved):

PR #3050, opened 2026-05-25
use keycloak for authentication

zawan-ila:
https://github.com/mitodl/hq/issues/11110

This PR adds support for using keycloak as an auth provider. It also drops supports for Touchstone SAML, which we intend to replace with keycloak.

### How can this be tested?
1. Follow the instructions at https://github.com/mitodl/ocw-studio/blob/bdd06963a1d31bc5280b2b7a8d8b3ab984095131/README.md#testing-keycloak-login-locally.
3. Especially, try logging in with the default keycloak account `admin@odl.local` and password `admin`. Verify that studio logs in successfully and the User object in django admin is populated with the email and name from keycloak.
4. For your local admin account, create an associated account in keycloak with same email. Then log in through keycloak. Verify that you are logged into your existing account, rather than creating a new one.


docker-compose.yml
```diff
@@ -160,6 +160,55 @@ services:
...
+      POSTGRES_PASSWORD: postgres # pragma: allowlist secret
+    ports:
+    - "5432"
+    networks:
+      default-network:
+        aliases:
+        - keycloak-db
+
+    volumes:
+    - kc-db-data:/var/lib/postgresql/18/docker
```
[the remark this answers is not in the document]
zawan-ila: The data directory switched in recent images of postgresql

main/settings.py
```diff
@@ -927,6 +927,38 @@
...
+    default=None,
+    description="Keycloak authorization URL for social auth",
+    required=False,
+)
+SOCIAL_AUTH_KEYCLOAK_ACCESS_TOKEN_URL = get_string(
+    name="SOCIAL_AUTH_KEYCLOAK_ACCESS_TOKEN_URL",
+    default=None,
+    description="Keycloak access token URL for social auth",
+    required=False,
+)
```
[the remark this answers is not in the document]
zawan-ila: The auth is working fine without these. The official [docs](https://python-social-auth.readthedocs.io/en/latest/backends/keycloak.html) dont mention these either

README.md
```diff
@@ -51,48 +51,28 @@ The `ocw-www` starter is meant for creating a home page, aka the "root website."
...
-  - Submit the form
-- In an incognito browser window, go to `http://MYNAME.ocw-studio.odl.local:8043/login/saml/?next=%2F&idp=default`
-- You should be redirected to SSOCircle, fill out the captcha and click `Continue SAML Single Sign On`
-- You should be redirected back to the /sites/ pages, and be logged in.
-- Log out & back in as a superuser and to go the Users admin page.
-  - There should be a new user with the same email address and name that you used to register with SSOCircle.
+- Navigate to the app and click **Login with MIT Keycloak**. You will be redirected to the Keycloak login page.
+- After a successful login you will be redirected back to the app and authenticated.
+- The development realm includes a default test user: **`admin@odl.local`** / **`admin`**.
+- To administer the local Keycloak instance (create users, inspect clients, etc.), log in to the Keycloak admin console at `http://localhost:<KEYCLOAK_PORT>` with the admin credentials defined in `docker-compose.yml`.
```
umar8hassan, line 78: without the `SOCIAL_AUTH_KEYCLOAK_PUBLIC_KEY` key in place the auth flow breaks with error. and i don't see proper steps to get public key. Maybe try something like [this](https://github.com/mitodl/odl-video-service/pull/1456) and place the keys beforehand for local dev. Also, add the steps as below:

````suggestion
- To administer the local Keycloak instance (create users, inspect clients, etc.), log in to the Keycloak admin console at `http://localhost:<KEYCLOAK_PORT>` with the admin credentials defined in `docker-compose.yml`.

#### Get `SOCIAL_AUTH_KEYCLOAK_PUBLIC_KEY`

`SOCIAL_AUTH_KEYCLOAK_PUBLIC_KEY` is the realm's **RS256 signing** public key, base64-encoded with no PEM headers or line breaks. The realm has two RSA keys (one for signing, one for encryption) — you need the signing one.

To retrieve it from any Keycloak instance:

1. Log in to the Keycloak admin console
2. Select the correct realm
3. Go to **Realm Settings → Keys**
4. Find the row where **Algorithm = RS256** and **Use = SIG**
5. Click the **Public key** button
6. Copy the displayed value into `SOCIAL_AUTH_KEYCLOAK_PUBLIC_KEY`

The local dev value is already set correctly in `.env.example`.
````

**Review by umar8hassan — CHANGES_REQUESTED** [review]
Mostly looks good. I was able to sign in after placing the public key.

Also, `LogoutView` only clears Django's session, the Keycloak SSO session stays alive. So, the next login redirects to Keycloak and it silently re-authenticates without prompting for credentials.

**Comment by zawan-ila** [comment]
@umar8hassan We log out from keycloak now too. I have also updated the docs.

File: README.md [hunk 4]
```diff
@@ -67,6 +67,7 @@ Keycloak runs automatically as part of the local Docker Compose stack. A pre-con
   SOCIAL_AUTH_KEYCLOAK_AUTHORIZATION_URL=http://localhost:<KEYCLOAK_PORT>/realms/<realm>/protocol/openid-connect/auth
   SOCIAL_AUTH_KEYCLOAK_ACCESS_TOKEN_URL=http://localhost:<KEYCLOAK_PORT>/realms/<realm>/protocol/openid-connect/token
   ```
+- For local development, values in `.env.example` can be used as defaults for all of the variables above except `SOCIAL_AUTH_KEYCLOAK_PUBLIC_KEY`. Retrieve `SOCIAL_AUTH_KEYCLOAK_PUBLIC_KEY` from `http://kc.odl.local:7080/admin/realms/ocw-studio`.
```
umar8hassan: it didn't work for me. Adding steps to retrieve the Public key and pre-emptively adding it into .env.example is the safest bet in opinion.

Image

umar8hassan (changes requested):

README.md
zawan-ila: There was a typo. I have fixed it now. Please try again

config/keycloak/realms/ocw-studio-realm.json
```diff
@@ -0,0 +1,2597 @@
...
+      "directAccessGrantsEnabled": false,
+      "serviceAccountsEnabled": false,
+      "publicClient": false,
+      "frontchannelLogout": true,
+      "protocol": "openid-connect",
+      "attributes": {
+        "request.object.signature.alg": "RS256",
+        "user.info.response.signature.alg": "RS256",
+        "frontchannel.logout.session.required": "true",
+        "post.logout.redirect.uris": "http://localhost:8043##http://ocw-studio.odl.local:8043/",
```
umar8hassan: uri with `localhost` is missing trailling `/` which causes logout to fail with `invalid redirect uri`
````suggestion
        "post.logout.redirect.uris": "http://localhost:8043/##http://ocw-studio.odl.local:8043/",

````

File: config/keycloak/realms/ocw-studio-realm.json [hunk 6]
```diff
@@ -0,0 +1,2597 @@
...
+      "baseUrl": "",
+      "surrogateAuthRequired": false,
+      "enabled": true,
+      "alwaysDisplayInConsole": true,
+      "clientAuthenticatorType": "client-secret",
+      "secret": "Aq0wmEvbUpZlhKPIEwi1RnuHtdeis77l",
+      "redirectUris": [
+        "http://ocw-studio.odl.local:8043/",
+        "http://localhost:8043/auth/complete/keycloak/",
+        "http://localhost:8043",
```
umar8hassan: uri  is missing a trailling `/` which causes logout to fail with `invalid redirect uri`
````suggestion
        "http://localhost:8043/",
````

umar8hassan (approved):
Looks Good 👍 

i have made suggestions to fix the redirect uri in local realm json which causes `invalid redirect uri` error on logout. Rest of the changes look good.

Image

config/keycloak/realms/ocw-studio-realm.json
zawan-ila: Fixed: added trailing `/` to `http://localhost:8043` in the `post.logout.redirect.uris` attribute — it now reads `http://localhost:8043/##http://ocw-studio.odl.local:8043/`.

config/keycloak/realms/ocw-studio-realm.json
zawan-ila: Fixed: added trailing `/` to `http://localhost:8043` in the `redirectUris` array — it now reads `http://localhost:8043/`.

PR #3075, opened 2026-06-19
Add reconcile_open_textbook_lrt management command

zawan-ila:
https://github.com/mitodl/hq/issues/11882

Adds a management command, `reconcile_open_textbook_lrt`, that restricts the **"Open Textbooks"** Learning Resource Type (LRT) to complete-textbook PDF resources, so open-textbook search results link to complete texts rather than chapter excerpts/fragments.

The `--textbook-uuids` list is the single source of truth. The command is declarative (strip-everything-then-add-back):

1. **Strip** `"Open Textbooks"` from every `WebsiteContent` and every `Website` that has it.
2. **Add** it to the complete-textbook resources and their courses' `sitemetadata` content. `Website.metadata` is **strip-only** — never re-added (the published/indexed course metadata is built from the `sitemetadata` content, not `Website.metadata`).
3. **Flag** every impacted site `has_unpublished_draft/live = True` and **log** their names so they can be selectively republished afterward (the command does not auto-publish).

Net effect: only the complete-textbook resources (and the course-level metadata of courses that own one) keep the tag; pages, excerpts, and textbook-less courses lose it.

Input is validated up front (no data is mutated on bad input): it errors if a UUID matches no resource, or if a UUID resolves to resources in more than one site (`text_id` is unique per-site, not globally). Duplicate/whitespace UUIDs are deduped.

Intentional trade-offs: dirty-flagging is best-effort (a few already-correct textbook sites may be flagged), and re-runs are not idempotent (textbook sites are re-stripped/re-added to the same final state and re-flagged each run) — acceptable for a one-off cleanup.

### How can this be tested?
2. In a local site, tag a page and two resources with "Open Textbooks" (one of the resources representing a complete-textbook PDF), and a couple of courses' sitemetadata.
3. Run `./manage.py reconcile_open_textbook_lrt --textbook-uuids <textbook_resource_text_id>`.
4. Verify: the page and the non-textbook resource lose the tag; the textbook resource keeps/gains it; the owning course's `sitemetadata` content gains the tag while its `Website.metadata` is stripped; a textbook-less course is fully stripped; and the impacted site is logged and marked `has_unpublished_live/draft`.

### DIFF
--- a/websites/management/commands/reconcile_open_textbook_lrt.py
+++ b/websites/management/commands/reconcile_open_textbook_lrt.py
@@ -17,8 +17,6 @@ log = logging.getLogger(__name__)
 
 OPEN_TEXTBOOK_LRT = "Open Textbooks"
 LRT_FIELD = "learning_resource_types"
-# Postgres JSONField containment lookup: matches rows whose
-# metadata["learning_resource_types"] list currently contains OPEN_TEXTBOOK_LRT.
 HAS_TAG_FILTER = {f"metadata__{LRT_FIELD}__contains": OPEN_TEXTBOOK_LRT}
 BULK_UPDATE_BATCH_SIZE = 100
 
--- a/websites/management/commands/reconcile_open_textbook_lrt.py
+++ b/websites/management/commands/reconcile_open_textbook_lrt.py
@@ -52,6 +52,8 @@ def add_open_textbook(objects) -> list:
     for obj in objects:
         metadata = obj.metadata if isinstance(obj.metadata, dict) else {}
         learning_resource_types = metadata.get(LRT_FIELD) or []
+        if not isinstance(learning_resource_types, list):
+            learning_resource_types = []
         if OPEN_TEXTBOOK_LRT not in learning_resource_types:
             obj.metadata = {
                 **metadata,
--- a/websites/management/commands/reconcile_open_textbook_lrt.py
+++ b/websites/management/commands/reconcile_open_textbook_lrt.py
@@ -1,0 +1,189 @@
+"""Reconcile the "Open Textbooks" learning resource type to complete-textbook resources."""  # noqa: INP001, E501
+
+import logging
+from collections import Counter
+
+from django.core.management.base import BaseCommand, CommandError
+from django.db import transaction
+
+from websites.constants import (
+    CONTENT_TYPE_METADATA,
+    CONTENT_TYPE_PAGE,
+    CONTENT_TYPE_RESOURCE,
+)
+from websites.models import Website, WebsiteContent
+
+log = logging.getLogger(__name__)
+
+OPEN_TEXTBOOK_LRT = "Open Textbooks"
+LRT_FIELD = "learning_resource_types"
+# Postgres JSONField containment lookup: matches rows whose
+# metadata["learning_resource_types"] list currently contains OPEN_TEXTBOOK_LRT.
+HAS_TAG_FILTER = {f"metadata__{LRT_FIELD}__contains": OPEN_TEXTBOOK_LRT}
+BULK_UPDATE_BATCH_SIZE = 100
+
+
+def strip_open_textbook(objects) -> list:
+    """
+    Remove OPEN_TEXTBOOK_LRT from each object's ``metadata[LRT_FIELD]`` list.
+
+    ``objects`` is any iterable of objects exposing a ``.metadata`` dict (both
+    ``WebsiteContent`` and ``Website``). Returns the objects as a list, ready for
+    ``bulk_update(..., ["metadata"])`` and for collecting impacted website ids.
+    """
+    objects = list(objects)
+    for obj in objects:
+        learning_resource_types = obj.metadata.get(LRT_FIELD) or []
+        obj.metadata = {
+            **obj.metadata,
+            LRT_FIELD: [
+                lrt for lrt in learning_resource_types if lrt != OPEN_TEXTBOOK_LRT
+            ],
+        }
+    return objects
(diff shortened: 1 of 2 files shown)


websites/management/commands/reconcile_open_textbook_lrt.py
```diff
@@ -0,0 +1,189 @@
...
+    objects = list(objects)
+    for obj in objects:
+        learning_resource_types = obj.metadata.get(LRT_FIELD) or []
+        obj.metadata = {
+            **obj.metadata,
+            LRT_FIELD: [
+                lrt for lrt in learning_resource_types if lrt != OPEN_TEXTBOOK_LRT
+            ],
+        }
+    return objects
```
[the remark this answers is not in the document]
zawan-ila: This func is only called for objects that have open textbooks in the lrt so metadata will not be None.

websites/management/commands/reconcile_open_textbook_lrt.py
```diff
@@ -0,0 +1,189 @@
...
+            msg = (
+                f"{len(textbook_uuids)} textbook UUID(s) matched {match_count} "
+                "resources — a UUID resolves to more than one resource "
+                "(text_id is unique per site, not globally). Resolve the ambiguity "
+                "and retry."
+            )
+            raise CommandError(msg)
+        textbook_website_ids = set(
+            textbook_resources.values_list("website_id", flat=True)
+        )
```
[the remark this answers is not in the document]
zawan-ila: I don't think this optimization is necessary. We don't have enough objects for this to be worthwhile.

pdpinch:
Would we remove the code for this management command after running it successfully?

umar8hassan (changes requested):
This should be a migration unless needed time and again. Is there any specific requirement/scenario for adding a mangement command for this one time change?

zawan-ila:
> This should be a migration unless needed time and again. Is there any specific requirement/scenario for adding a management command for this one time change?

@umar8hassan I do not think this change makes sense in a migration. It is not as if we are removing the Open Textbooks LRT entirely, or replacing it with another LRT everywhere else. Migrations generally make sense when the change corresponds to a deterministic, reproducible transformation that applies consistenly across any deployment (think of someone running a separate instance of ocw-studio) Here, we are removing the lrt from all but a handful of resources. We'd have to hardcode the uuids for these resources in the migration file itself, which is brittle and not generalizable.

zawan-ila:
> Would we remove the code for this management command after running it successfully?

@pdpinch yes, we'll remove the code after running the command.

Another option is to skip merging this PR entirely and instead run the management command as a one-off script directly against the application pods. That saves us from the hassle of merge, deploy, run, removal PR, merge, deploy again cycle.

pdpinch:
Ok. I leave it to your judgement.

websites/management/commands/reconcile_open_textbook_lrt.py
```diff
@@ -0,0 +1,189 @@
...
+                "these resources and their courses' sitemetadata."
+            ),
+        )
+
+    def handle(self, *args, **options):  # noqa: ARG002
+        """Reconcile the tag to the textbook set and mark impacted sites dirty."""
+        textbook_uuids = {
+            uuid.strip()
+            for uuid in (options["textbook_uuids"] or "").split(",")
+            if uuid.strip()
```
umar8hassan, line 89: [nit] It's a bit redundant
```suggestion
            if uuid
```

websites/management/commands/reconcile_open_textbook_lrt.py
```diff
@@ -0,0 +1,189 @@
...
+        # Validate the keep-list up front so a bad input never mutates data.
+        textbook_resources = WebsiteContent.objects.filter(
+            type=CONTENT_TYPE_RESOURCE, text_id__in=textbook_uuids
+        )
+        missing = sorted(
+            textbook_uuids - set(textbook_resources.values_list("text_id", flat=True))
+        )
+        if missing:
+            msg = f"No resource found for textbook UUID(s): {', '.join(missing)}"
+            raise CommandError(msg)
```
umar8hassan, line 101: This could be a log instead of an error to allow execution for valid resources

websites/management/commands/reconcile_open_textbook_lrt.py
```diff
@@ -0,0 +1,189 @@
...
+    """
+    Remove OPEN_TEXTBOOK_LRT from each object's ``metadata[LRT_FIELD]`` list.
+
+    ``objects`` is any iterable of objects exposing a ``.metadata`` dict (both
+    ``WebsiteContent`` and ``Website``). Returns the objects as a list, ready for
+    ``bulk_update(..., ["metadata"])`` and for collecting impacted website ids.
+    """
+    objects = list(objects)
+    for obj in objects:
+        learning_resource_types = obj.metadata.get(LRT_FIELD) or []
```
umar8hassan, line 34:
```suggestion
        learning_resource_types = obj.metadata.get(LRT_FIELD) or []
        if not isinstance(learning_resource_types, list):
            learning_resource_types = []
```

websites/management/commands/reconcile_open_textbook_lrt.py
umar8hassan: [nit] But if we must gate it, then we could use the walrus operator:
```
            _uuid
            for uuid in (options["textbook_uuids"] or "").split(",")
            if (_uuid := uuid.strip())
```

umar8hassan (approved):
Looks Good 👍

PR #3078, opened 2026-06-25
feat: adopt mitodl/ol-python-base:3.13, granian

blarghmatey:
## Summary

- Replaces the hand-rolled `FROM python:3.13.7-slim` multi-stage build with `FROM mitodl/ol-python-base:3.13`; app-specific extras (`libmagic1`, `libxmlsec1-openssl`) installed inline with BuildKit apt cache
- Switches the WSGI server from uwsgi → granian (`--interface wsgi`) in docker-compose and CMD
- Updates `config/nginx.conf` (dev) from `uwsgi_pass` to `proxy_pass http://` in the app server block; S3 preview server blocks (ports 8044–8046) already used `proxy_pass` and are unchanged
- Adds BuildKit `--mount=type=cache` on `uv sync` for faster layer rebuilds
- Removes `uwsgi` and `uwsgitop` from `pyproject.toml` and regenerates `uv.lock`

## Test plan

- [x] `docker build --target runtime .` succeeds
- [x] `uv.lock` regenerated — uwsgi v2.0.31 and uwsgitop v0.12 removed
- [x] nginx.conf uses `proxy_pass http://` for the app server block
- [x] docker-compose `web` command runs granian on port 8041

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # syntax=docker/dockerfile:1
 # hadolint global ignore=DL3008
 
-FROM mitodl/ol-python-base:3.13 AS base
+FROM mitodl/ol-python-base:3.14 AS base
 LABEL maintainer="ODL DevOps <mitx-devops@mit.edu>"
 
 # App-specific apt extras; common-core packages (git, curl, libjpeg-dev,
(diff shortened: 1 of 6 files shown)


feoh:
@blarghmatey Is this still valid given @umar8hassan's 3.14 PR?

feoh (approved):

PR #3084, opened 2026-07-01
Add opt-in django-aqueduct typed settings

blarghmatey:
## Summary

Adds a parallel, **opt-in** [django-aqueduct](https://github.com/mitodl/django-aqueduct)-based settings module for this app, on **django-aqueduct 0.9.0 (codegen v2)**. `main/settings.py` and `main/envs.py` remain the default settings everywhere; opting in is the only behavior change.

- `main/aqueduct_settings.py`: a typed `AqueductSettings` pydantic-settings model generated by codegen v2 (static AST discovery — the settings module is never imported) plus a hand-written preserved region holding the ported validators/derivations, and a `DevAqueductSettings` subclass that layers Vault under env vars for local dev.
- `main/settings_aqueduct.py` / `main/settings_aqueduct_dev.py`: thin shims that call `configure_django_settings(...)`. Sentry initializes through the `pre_configure` hook (before settings injection), matching `main/settings.py`'s "before anything else" ordering. Opt in via `DJANGO_SETTINGS_MODULE=main.settings_aqueduct` / `..._dev`.
- `pyproject.toml`: `django-aqueduct[vault,mitol,derivations]>=0.9.0`; a `[tool.aqueduct]` table drives generation and parity; `main/aqueduct_settings.py` is excluded from ruff (documented fallback — see below). `django_aqueduct` is added to `INSTALLED_APPS` for the management commands.
- `docs/aqueduct.md`: documents the v2 workflow (managed regions, `[tool.aqueduct]`, env-var-driven Vault, parity).

### 0.9.0 rework — workarounds shed

0.9.0 (and 0.8.1) fix upstream the three issues earlier revisions worked around by hand; each removal is re-tested:

- **URL types**: 0.9.0 makes `str`→`AnyUrl` promotion opt-in (`[tool.aqueduct] enrich_url_types`), only promotes a field whose own default validates as an absolute URL, denylists Django's relative-URL settings, and pairs every promotion with a `str` `field_serializer`. **The 26-field AnyUrl revert block from 0.8.0 is deleted** — the `*_URL` fields keep the plain `str` static discovery gives them. `enrich_url_types` is left **off**: enabling it was tested and promotes 17 fields, but it is still wrong here — several URL fields feed in-model derivations doing string ops (`dv.first_url(REDISCLOUD_URL, …)` calls `.strip()`; `CACHES` `LOCATION` is built from `CELERY_BROKER_URL`), and `AnyUrl` objects have no `.strip()`; `AnyUrl` also normalizes a trailing slash onto values like `POSTHOG_API_HOST`, changing the value legacy stores.
- **Parity**: 0.9.0 subset-compares dict settings, so **`DATABASES` and `DEFAULT_DATABASE_CONFIG` are dropped from `parity_ignore`** — Django's `ConnectionHandler`-injected keys (`ATOMIC_REQUESTS`/`AUTOCOMMIT`/`TIME_ZONE`/`TEST`) are legacy-extra, and the model's SQLite-safe empty `OPTIONS` is a subset of legacy's stray `sslmode`. `parity_ignore` is now 8 keys (was 10); the rest are `STORAGES`, `NPLUSONE_LOGGER`, and the inline-only DB primitives the model declares as real fields.
- **Static discovery** now recognizes `main/envs.py get_list_of_str`, so `OCW_TEST_SITE_SLUGS` is generated as `list[str]` and handled by the generated container decoder; its hand override (and literal-parser entry) are removed. The two `OCW_EXTRA_THEMES_*` dicts use `get_dict_of_str` (still `Any`) and keep their typed `NoDecode` overrides. `OPEN_CATALOG_URLS` keeps a one-line default-only override (0.9.0's decoder maps `""`→`[]`, but legacy `get_delimited_list(default="")` yields `[""]`).
- **`extend-exclude` is kept** (documented fallback): 0.8.1 made output `ruff format`-stable and dropped the blanket `# ruff: noqa` (also removed from this file's header), but ocw-studio selects rule groups beyond the ~9 validated against — the managed-region override pattern re-declares generated fields, unavoidably tripping `PIE794`/`F811` (duplicate class field); generated TODO markers trip `TD002`/`TD003`/`FIX002`; ported derivations trip `C901`/`PLR0912`/`PLR0915`.

### Type precision

The original 0.6 scaffold carried **52 `Any`-typed fields behind 55 `# TODO: refine type` markers** (which forced a per-file ruff suppression). v2's typed IR + static discovery types nearly all of them; effective `Any` fields are the two opaque `NPLUSONE_*` loggers. Shared boilerplate (`DATABASES` / Celery–Redis fallback / `CACHES` / `FEATURES`) uses `django_aqueduct.derivations`.

### Review-flagged bugs fixed

- **LOGGING snapshot (HIGH)**: rebuilt in a validator driven by `LOG_HOST`/`LOG_HOST_PORT`/`LOG_LEVEL`; `HOSTNAME` derives from `platform.node()` (the baked `"ikoria"`/hardcoded syslog is gone).
- **Parity divergences (HIGH)**: `ENVIRONMENT` restored to legacy default `"dev"`; `OPEN_CATALOG_URLS` kept `[""]` (what legacy actually produces).
- **Idempotent key transform**: `GITHUB_APP_PRIVATE_KEY` skips already-`bytes` values on re-validation.
- **Vault KeyError (MEDIUM)**: `DevAqueductSettings` uses `vault_source_from_env` (`VAULT_*`), runs Vault-less when unset.
- **STORAGES**: a declared field (default = Django's built-in), not validator-injected via `extra="allow"`.

The pre-existing `MIDDLEWARE` list-append bug in `main/settings.py` is reproduced verbatim for parity.

## Test plan

- [x] `uv lock` / `uv sync` resolve `django-aqueduct==0.9.0` from PyPI
- [x] `AqueductSettings()` instantiates; URL fields are `str`, derivations run, all bug fixes verified
- [x] `manage.py check` passes under `main.settings_aqueduct`, `main.settings_aqueduct_dev`, and default `main.settings`
- [x] `manage.py generate_aqueduct_settings --check` — no drift
- [x] `manage.py check_aqueduct_settings` — in parity (8 ignored)
- [x] `ruff check` / `ruff format --check --force-exclude` clean on changed files
- [ ] DB-bound settings tests skipped (no live Postgres in sandbox)

### Review feedback resolution

All open sentry/gemini review comments were triaged against the current 0.9.0-reworked code:

- **GITHUB_APP_PRIVATE_KEY re-validation `AttributeError`** (sentry/gemini, LOW/HIGH) — **resolved**: the transform is guarded by `isinstance(self.GITHUB_APP_PRIVATE_KEY, str)`, so re-validation of an already-`bytes` value is a no-op.
- **LOGGING hardcoded syslog `("localhost", 514)` / level / `"ikoria"` hostname** (sentry, HIGH) — **resolved**: `LOGGING` is rebuilt in the derive validator from `LOG_HOST`/`LOG_HOST_PORT`/`LOG_LEVEL` and `HOSTNAME = platform.node()`.
- **ENVIRONMENT made hard-required** (sentry, HIGH) — **resolved**: keeps its legacy default `"dev"`.
- **`VAULT_ADDR` raw `KeyError`** (gemini, MEDIUM) — **resolved**: `DevAqueductSettings` builds Vault via `vault_source_from_env` (`VAULT_*`), degrading gracefully when unset.
- **Hardcoded local `django-aqueduct` path in `pyproject.toml`** (gemini, MEDIUM) — **resolved**: depends on the published `django-aqueduct>=0.9.0` from PyPI; no `[tool.uv.sources]` override remains.
- **`CELERY_BROKER_URL=None` → `dv.redis_cache()` may crash at load** (sentry, HIGH) — **not a defect**: `dv.redis_cache(None)` returns `{… "LOCATION": None}` without raising, and Django instantiates cache backends lazily, so an unused None-location cache never crashes at startup. This exactly mirrors legacy `main/settings.py` (`CACHES["redis"]["LOCATION"] = CELERY_BROKER_URL`, also `None` when unset); guarding it would diverge from the legacy settings the model reproduces. Documented inline.
- **`OPEN_CATALOG_URLS` empty default → `[""]` vs `[]`** (sentry, HIGH) — **intended parity**: legacy `get_delimited_list(default="")` yields `[""]` (verified); the model pins that default while the generated decoder parses real env values. The `[""]` behavior is pre-existing in `main/settings.py` and out of scope for a parity-preserving model.


blarghmatey:
django-aqueduct 0.6.0 is now published on PyPI, so I've removed the `[tool.uv.sources]` local-path override in `pyproject.toml` and re-ran `uv lock`. The lockfile now resolves `django-aqueduct` 0.6.0 from PyPI (registry source, sdist/wheel from files.pythonhosted.org) instead of the local worktree path. No longer blocked on the TODO.

Verified with `uv run python manage.py check` against both `main.settings_aqueduct` (passes cleanly) and `main.settings_aqueduct_dev` (fails only at the expected live Vault network call, no Python/import errors).

blarghmatey:
Closing for now — we'll revisit and recreate this later.

PR #3134, opened 2026-08-03
Harden GH Actions supply chain: zizmor scanning + 7-day dependency cool-down

blarghmatey:
N/A

Hardens supply-chain security around this repo's GitHub Actions and Python dependency resolution, in line with the pattern already in use in mitodl/mitxonline.

1. **Adds a zizmor GitHub Actions static-analysis workflow** (`.github/workflows/actions-static-analysis.yml`). It runs on push/PR whenever anything under `.github/workflows/**` changes, checks out the repo with `persist-credentials: false`, and runs `zizmorcore/zizmor-action` (pinned by commit SHA) against `.github/workflows/`, failing on findings of `high` severity / `medium` confidence or above. The workflow itself has `permissions: {}` at the top level with only `contents: read` and `actions: read` granted to the job, following least-privilege practice.

2. **Adds zizmor as a local pre-commit hook** so the same checks run before a workflow change is even pushed. Added to `.pre-commit-config.yaml`:
   ```yaml
   - repo: https://github.com/zizmorcore/zizmor-pre-commit
     rev: v1.29.0
     hooks:
       - id: zizmor
   ```

3. **Adds a 7-day cool-down on new dependency versions.** uv 0.11+ supports `exclude-newer` as a rolling relative duration (not just a fixed date). Added `exclude-newer = "7d"` to the existing `[tool.uv]` table in `pyproject.toml`, then regenerated `uv.lock` with `uv lock`. This means `uv lock`/`uv sync` will never resolve a package version published in the last 7 days, giving the community time to catch and yank newly-introduced malicious or vulnerable releases (mitigating npm/PyPI-style supply-chain attacks where a compromised package is published and pulled in within hours). The lockfile now records this as a rolling window (`exclude-newer-span = "P7D"` under a new `[options]` block), not a fixed cutoff date, so it self-updates on every future `uv lock` run rather than needing periodic manual bumps.

No application code changes. No dependency versions were shifted by regenerating the lockfile — the only diff in `uv.lock` is the new `[options]` block recording the exclude-newer span.

N/A - no UI changes

### How can this be tested?
- Confirm the new workflow parses as valid YAML: `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/actions-static-analysis.yml'))"`.
- Confirm `.pre-commit-config.yaml` still parses as valid YAML and pre-commit accepts the config: `pre-commit run zizmor --files .github/workflows/actions-static-analysis.yml` — this was run locally against this branch and **passed**, confirming the hook installs and executes correctly (zizmor found no high-severity/medium-confidence issues in the new workflow file).
- Confirm `pyproject.toml` has `exclude-newer = "7d"` inside the existing `[tool.uv]` table (no duplicate `[tool.uv]` header), then run `uv lock` — it should resolve successfully and report something like `Resolving despite existing lockfile due to addition of global exclude newer ...`. Verify `uv.lock` now contains:
  ```
  [options]
  exclude-newer = "0001-01-01T00:00:00Z" # This has no effect and is included for backwards compatibility when using relative exclude-newer values.
  exclude-newer-span = "P7D"
  ```
- On merge, reviewers can also verify the new CI job triggers correctly by pushing a trivial change under `.github/workflows/` and confirming the "GitHub Actions Static Analysis" check appears and passes.

### Additional Context
Reference pattern: mitodl/mitxonline already uses this same zizmor CI workflow shape.

zizmor docs: https://woodruffw.github.io/zizmor/
uv `exclude-newer` docs: https://docs.astral.sh/uv/reference/settings/#exclude-newer


blarghmatey:
Follow-up: added \`[tool.uv.exclude-newer-package]\` with \`= "0d"\` overrides for our in-house packages (mitodl/ol-concourse, mitodl/django-aqueduct, mitodl/open-edx-plugins, mitodl/ol-django). The global \`exclude-newer = "7d"\` is meant to guard against newly-published, potentially malicious/vulnerable *third-party* packages — it shouldn't also delay picking up our own releases, which we iterate on quickly. \`"0d"\` is a zero-length relative duration, so these packages always resolve to latest, same self-updating mechanism as the global setting (not a fixed date).

Ran \`uv lock\` to regenerate the lockfile — it succeeded and only added the new \`[options.exclude-newer-package]\` metadata block; no package versions changed.

pyproject.toml
```diff
@@ -85,6 +85,7 @@ dev = [
 package = false
 default-groups = "all"
 no-binary-package = ["lxml"]
+exclude-newer = "7d"
```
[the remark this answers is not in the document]
blarghmatey: Fixed in 229765d3: added required-version = ">=0.9.17" to [tool.uv] — the exact uv release that introduced relative exclude-newer duration support (https://github.com/astral-sh/uv/releases/tag/0.9.17). An older uv now fails loudly with a clear version-mismatch error instead of silently mishandling the setting.

.github/workflows/ci.yml
```diff
@@ -44,7 +46,7 @@ jobs:
     - name: Install uv
       uses: astral-sh/setup-uv@d31148d669074a8d0a63714ba94f3201e7020bc3 # v8.3.0
       with:
-        enable-cache: true
```
zawan-ila, line 47: Why do we make this change?

pyproject.toml
```diff
@@ -85,6 +85,55 @@ dev = [
...
+required-version = ">=0.9.17"
+
+[tool.uv.exclude-newer-package]
+ol-concourse = "0d"
+django-aqueduct = "0d"
+edx-sysadmin = "0d"
+edx-username-changer = "0d"
+ol-openedx-ai-static-translations = "0d"
```
zawan-ila: Should these entries should be relevant to the project? ocw does not care about e.g ol-openedx-ai-static-translations and many of the packages here

.github/workflows/production.yml
zawan-ila: Can we remove production.yml and release-candidate.yml? They relate to heroku deployment and we shouldn't need that anymore?

zawan-ila (changes requested):

.github/workflows/actions-static-analysis.yml
```diff
@@ -0,0 +1,33 @@
+---
+name: GitHub Actions Static Analysis
+
+on:
+  push:
+    paths:
+    - ".github/workflows/**"
+  pull_request:
+    paths:
+    - ".github/workflows/**"
```
ChristopherChudzicki, line 12: `push` + `pull` is redundant, right?

**Suggestion:** Just do `push`.

🤖 Posted by [Claude Code](https://claude.com/claude-code)

blarghmatey:
Dropped the `pull_request` trigger — good catch, it was firing zizmor twice per PR commit (once from `push`, once from `pull_request`) since both were scoped to the same paths with no branch filter. `push` alone still covers PR branch commits.

.github/workflows/production.yml
```diff
@@ -16,9 +19,10 @@ jobs:
       github.event_name == 'workflow_dispatch' ||
       github.event.workflow_run.conclusion == 'success'
     steps:
-    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
```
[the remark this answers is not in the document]
blarghmatey: Moot — removed the file in fd97bb01. Confirmed with tmacey: ocw-studio no longer deploys to Heroku.

.github/workflows/production.yml
blarghmatey: Done in fd97bb01 — removed both production.yml and release-candiate.yml. Confirmed with tmacey: ocw-studio no longer deploys to Heroku, so their workflow_run dangerous-triggers findings are moot too.

.github/workflows/ci.yml
blarghmatey: It's zizmor's cache-poisoning fix, not a mistake. This step's cache was flagged because the workflow runs on `push` (not just `pull_request`) — for a push-triggered job, a compromised commit could seed a poisoned dependency cache that a later, more-trusted run then reuses. `astral-sh/setup-uv`'s cache only saves ~10-20s of resolve time here, so disabling it was the safer trade. Same finding category resurfaced on the `setup-node` step after the rebase — flagging that one separately as it's from an already-merged renovate PR, not this PR's diff.

pyproject.toml
blarghmatey: Trimmed in 6c4443f7 — kept only the 6 packages ocw-studio actually locks (mitol-django-authentication, mitol-django-common, mitol-django-mail, mitol-django-transcoding, mitol-drf-lint, ol-concourse), dropped the rest. Same fix applied across all 8 repos with this allowlist (mit-learn's reviewer flagged the identical issue independently).

zawan-ila (approved):

PR #3137, opened 2026-08-04
feat: language and locale override for resources

umar8hassan:
https://github.com/mitodl/hq/issues/12685

Paired with mitodl/ocw-hugo-projects#400, which declares the two form fields. The two can merge in either order. Until the starter ships, this branch finds no metadata value and behaves exactly as today.

A caption or transcript's language is currently derived only from its uploaded filename, which has to follow the `_captions-<lang>[-<locale>]` convention. A file that does not follow it cannot be labeled, and correcting a language means renaming the file. This lets an editor set the language from the resource form instead.

- `videos/utils.py`: new `resolve_language_locale`, the single source of precedence (explicit metadata, then the parsed filename, then `en`). Language and locale resolve independently, so an explicit language does not inherit a region parsed from a filename it no longer agrees with. Non-string or whitespace-only metadata is treated as unset rather than raising, since this runs in the publish path where an exception fails a whole site build.
- `videos/utils.py`: `resource_file_paths` now goes through the resolver, which is what makes an override reach published frontmatter.
- `websites/serializers.py`: `WebsiteContentDetailSerializer.to_representation` fills both keys from the resolver, so the form always shows the value that will publish and saving materializes it. An incorrect detection is then visible and correctable.

No change is needed in ocw-hugo-themes. Its partials already consume `{file, language, locale}`.

### How can this be tested?

**Load the starter from the linked branch.** `import_website_starters` refuses to sync unless the commit you pass matches the head of `GITHUB_WEBHOOK_BRANCH`, and it prints "Successfully updated" either way, so both have to be set or the import silently does nothing.

```bash
BRANCH=umar/12685-resource-language-locale-fields
SHA=$(git ls-remote https://github.com/mitodl/ocw-hugo-projects "refs/heads/$BRANCH" | cut -f1)
docker compose run --rm -e GITHUB_WEBHOOK_BRANCH="$BRANCH" web \
  python manage.py import_website_starters https://github.com/mitodl/ocw-hugo-projects -c "$SHA"
```

Confirm the output lists the config files rather than "No ocw-studio.yaml files found", then check the fields actually landed:

```bash
docker compose run --rm web python manage.py shell -c "
from websites.models import WebsiteStarter
from websites.site_config_api import SiteConfig
item = SiteConfig(WebsiteStarter.objects.get(slug='ocw-course-v2').config).find_item_by_name('resource')
print([f['name'] for f in item.fields if f['name'] in ('language', 'locale')])
"
```

Expected: `['language', 'locale']`.

**Then in the editor:**

1. Open a caption resource whose filename carries a language, for example `..._captions-fr.vtt`. Language is pre-selected as French rather than blank. A file with no language suffix shows English.
2. Change it to Spanish and save. The published frontmatter for the video that links it shows `language: es` rather than the `fr` in the filename.
3. Confirm a resource on a starter without these fields is unaffected, with no `language` key added to its metadata.

### DIFF
--- a/websites/serializers_test.py
+++ b/websites/serializers_test.py
@@ -16,6 +16,7 @@ from main.constants import ISO_8601_FORMAT
 from users.factories import UserFactory
 from users.models import User
 from videos.constants import YT_THUMBNAIL_IMG
+from videos.utils import resource_file_paths
 from websites.constants import (
     CONTENT_TYPE_INSTRUCTOR,
     CONTENT_TYPE_METADATA,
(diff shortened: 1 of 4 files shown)


websites/serializers_test.py
```diff
@@ -934,3 +935,199 @@ def test_update_page_url_on_title_change_legacy_index_behavior(
...
+    assert "language" not in data["metadata"]
+
+
+@pytest.mark.django_db
+def test_detail_serializer_does_not_mutate_stored_metadata():
+    """Serializing must not write the derived value onto the instance.
+
+    to_representation's result can alias instance.metadata. Mutating it in
+    place would let a derived guess be persisted by any later save, which is
+    the failure mode fixed in full_metadata under hq#12635.
```
zawan-ila: remove reference to ticket

zawan-ila (approved):
I think that it is worth thinking about scoping these new fields only to captions and transcripts. Having them show on all resources, like e.g Images, is just unnecessary cognitive strain for the course authors.

websites/serializers_test.py
umar8hassan: Fixed in https://github.com/mitodl/ocw-studio/pull/3137/commits/2712498e - dropped the ticket reference, kept the reason the assertion exists. Checked the rest of the diff and this was the only one.

PR #3139, opened 2026-08-07
Update CKEditor to 41.4.2 and remove community type definitions

### LINKED ISSUE #2054: Update CKEditor & remove deprecated webpack integration
# Description/Context
We should:
- [ ] Update CKEditor to a newer version, probably the latest
    - Studio uses v35 and community provided typescript definitions. CKEditor has first-class support for Typescript starting in v38, so we should remove the community provided type definitions. 
- [ ] Stop using [`@ckeditor/ckeditor5-dev-webpack-plugin`](https://www.npmjs.com/package/@ckeditor/ckeditor5-dev-webpack-plugin), which is deprecated. 

# Plan/Design
For updating the webpack integration, the CKEditor5 guide [Integrating from source using webpack](https://ckeditor.com/docs/ckeditor5/latest/installation/advanced/alternative-setups/integrating-from-source-webpack.html) describes what to do. See also implementation in MIT Open.

One issue is that starting in v39, CKEditor includes some small branding. See [Managing the “Powered by CKEditor” logo](https://ckeditor.com/docs/ckeditor5/latest/support/licensing/managing-ckeditor-logo.html). Here's [one way](https://github.com/mitodl/mit-open/blob/ce2ecbfcff13d6bfa58fda9f94aa86e07f3aaf1d/frontends/ol-ckeditor/src/styles.scss#L3) to remove it. *Or we could leave it, since Studio is just internal anyways.*

umar8hassan:
## What are the relevant tickets?

Closes #2054

## Description

Moves CKEditor from 35.4.0 to 41.4.2 and deletes the hand-written type shim at `static/js/types/ckeditor.d.ts`. CKEditor has shipped first-class TypeScript since v38, so the shim was suppressing real type information on every editor import.

41.4.2 is the highest version that keeps the current architecture intact. Above it the markdown turndown singleton disappears, deep `src/` imports are removed, a license key becomes mandatory, and the math plugin has no compatible build. The list feature is pinned to `LegacyList` because v41 swapped the implementation behind the old import path.

The deprecated `@ckeditor/ckeditor5-dev-webpack-plugin` was already gone from this repo. This branch keeps it that way and leaves the from-source webpack integration unchanged.

Deleting the shim surfaced latent bugs the `any` types were hiding. Fixed here:

- `ResourceLinkUI` dereferenced a null actions view. LinkUI builds it lazily at 41.4.2, and an empty balloon also reports a null visible view, so the two nulls compared equal and the handler ran with no link balloon open.
- Collapsed-selection inserts in `CustomLink`, `ResourceLink` and `ResourceEmbed` passed a possibly-null position straight into the model.
- `MarkdownConfigPlugin` defaults omitted `allowedHtml`, which could reach Turndown as undefined.
- `ResourcePicker` threw on click when the host app supplied no callback.
- `imageStyle:full` had not been a valid style name since around v29 and rendered no button. Restored as `imageStyle:block`.

Also adds characterization tests that boot all three real editor configs and assert their toolbar items resolve and that no unexpected warnings are logged. Nothing did that before, so a plugin failing to load was only catchable by hand.

## How can this be tested?

Behaviour should be identical to before the upgrade. This editor has no source-editing mode, so everything below goes through the toolbar. Verify each by saving, reloading the page, and confirming the content comes back unchanged. Round-tripping is the thing most likely to break, not first render.

Keep the browser console open throughout. It should stay clean, in particular of `ckeditor-duplicated-modules`.

### Math

Use the `math` toolbar button, which opens a dialog with a live preview. Each sample targets a different escaping rule in the markdown conversion:

| Enter this TeX | What it checks |
| --- | --- |
| `x^2 + y^2 = z^2` | baseline inline math |
| `\frac{1}{2}` | backslash handling, must not double up on reload |
| `90 \%` | percent escaping |
| `a_1 + b_2` | underscore escaping, must not turn into italics |
| `x * y` | asterisk escaping, must not turn into italics |
| `1 + 1 < 3` | angle brackets are not treated as HTML |

Try each as both inline and display math. Stored markdown is `\\(...\\)` for inline and `\\[...\\]` for display. The failure mode to watch for is backslashes accumulating or disappearing across a save and reload cycle.

### Links

1. **Resource link.** Click `Add link`, pick an existing resource, give it link text. Stored markdown should be `{{% resource_link "<uuid>" "link text" %}}`.
2. **Resource link with a heading anchor.** Same, targeting a heading. Stored markdown gains a third argument, `{{% resource_link "<uuid>" "link text" "#some-heading-id" %}}`.
3. **Link balloon.** Click an existing resource link. The balloon should show the preview button, and the Edit button should be hidden. This is `ResourceLinkUI`, which this PR changed.
4. **External URL.** Only if `OCW_STUDIO_CUSTOM_LINKUI_ENABLE` is on for your user. Select text, use the standard link button, paste `https://example.com`. It should create an external-resource record and rewrite the href to a resource link.

### Lists

The list feature is pinned to LegacyList, so this is the highest-risk area in the PR. Verify against the raw markdown in the Django admin.

1. Paste this into the Markdown field in admin and save:
```
- aardvark
- **bold** and *italic* and a [link](https://example.com/)
- chickadee
    - nested bullet
    - another nested bullet

1. first
2. second
    - nested under ordered
    - another nested
3. third
```

2. Open the same content in the Studio editor. It should render properly

3. Save from the Studio editor without changing anything. Reopen the record in admin. The Markdown field must be identical to the raw from step 1.

Exercise separately. Add this item and save:
```
- beaver
    second paragraph in the same item
```

It must keep both the bullet and the second paragraph. The stored form uses trailing spaces and a four-space indent rather than a blank line. This shape gains a non-breaking space on every save, a pre-existing bug present identically on master, so ignore the length drift and confirm the structure survives.

### Other

- **Tables.** Insert a table with a header row, put a bulleted list in one cell and bold text in another, save and reload.
- **Images.** Select an existing image and confirm the toolbar offers both style buttons, block and side, and that switching between them works in both directions.
- **Toolbar icons.** Confirm every button renders an icon rather than a blank square, which would mean the webpack icon loader broke.
- **Logo.** Confirm that no "Powered by CKEditor" balloon appears in the bottom-right of the editing area.


static/js/lib/ckeditor/CKEditor.ts
```diff
@@ -114,12 +114,7 @@ export const FullEditorConfig = {
     ],
   },
   image: {
-    toolbar: [
-      "imageStyle:full",
-      "imageStyle:side",
-      "|",
-      "imageTextAlternative",
-    ],
+    toolbar: ["imageStyle:side", "|", "imageTextAlternative"],
```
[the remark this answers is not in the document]
umar8hassan: Fixed in https://github.com/mitodl/ocw-studio/pull/3139/commits/afe7cfb3 - added `imageStyle:block`, which is what `full` was renamed to.

One correction on the reasoning. This was not a regression introduced here. `imageStyle:full` was already invalid at 35.4.0, where the valid names are `inline`, `block`, `side`, `alignLeft`, `alignRight`, `alignCenter`, `alignBlockLeft` and `alignBlockRight`. Booting `FullEditorConfig` on the parent commit logs `toolbarview-item-unavailable { item: 'imageStyle:full' }`, so that button never rendered and there was no two-way control to lose. The new characterization test in `CKEditor.test.ts` asserts the exact set of boot warnings per config, which is how it surfaced.

The suggested change is still worth making, since it restores the control the config clearly intended, so it is in. That same test now passes with an empty expected-warning list, which confirms `imageStyle:block` resolves to a real toolbar component rather than repeating the problem.

On `docs/ckeditor/plugin.md:673-679`, that block is a tutorial snippet that copied the same stale config, including `imageUpload` and a YouTube embed that are not in the real editor config, so it documents the bug rather than the intent. Worth a separate cleanup.

package.json
```diff
@@ -7,33 +7,32 @@
...
+    "@ckeditor/ckeditor5-paragraph": "^41.4.2",
+    "@ckeditor/ckeditor5-react": "^7.0.0",
+    "@ckeditor/ckeditor5-table": "^41.4.2",
+    "@ckeditor/ckeditor5-theme-lark": "^41.4.2",
+    "@ckeditor/ckeditor5-ui": "^41.4.2",
+    "@ckeditor/ckeditor5-utils": "^41.4.2",
+    "@ckeditor/ckeditor5-widget": "^41.4.2",
     "@dnd-kit/core": "^3.1.1",
     "@dnd-kit/sortable": "^4.0.0",
+    "@isaul32/ckeditor5-math": "41.4.2",
```
ChristopherChudzicki, line 35: @umar8hassan This one is pinned exactly, the others have `^41.4.2`.

**Suggestion:** Let's pin them all exactly. It doesn't matter a lot, since yarn.lock pins the exact versions used, but might help avoid drift.

static/js/lib/ckeditor/CKEditor.ts
```diff
@@ -114,8 +125,11 @@ export const FullEditorConfig = {
     ],
   },
   image: {
+    // `imageStyle:block` is what CKEditor renamed the old `full` option to.
+    // Pairing it with `side` keeps the two-way style control this toolbar has
+    // always meant to offer.
     toolbar: [
-      "imageStyle:full",
+      "imageStyle:block",
       "imageStyle:side",
```
ChristopherChudzicki: I'm not sure the comment above is accurate.

I think we only support markdown image syntax `alt text` for legacy OCW content authored before the days of Studio. But I don't think we do support the styling at all. I mean, CkEditor might show the shortcuts, but Hugo doesn't do anything with it... We on't persist the styles to markdown, afaik.

I'd remove the comment, and potentially the plugins. But I guess the plugins were pre-existing.

ChristopherChudzicki (dismissed):
Approving, but a couple minor changes requested below. Happy to look again if you want, but they are simple.


> [!NOTE]
> Warning: I ran into something weird (see below), **but as far as I can tell, it is not related to your branch.**

I kept getting this alert locally, even when i hadn't made any edits. I would just opent he drawer, close it, alert would appear. No eidt.
- This happened consistently on local `master`, `release-candiate`, and your branch
- Confusingly, I could not get it to trigger on RC.

So I don't think it's new to your branch, but I am a bit confused about it. Have you run into this?

Image

ChristopherChudzicki:
@umar8hassan One other thing...
> - Closes https://github.com/mitodl/ocw-studio/issues/2054

IMO, it's worth leaving that issue open or starting a new one. We still have 7 major versions to go. Left a comment on https://github.com/mitodl/ocw-studio/issues/2054#issuecomment-5247282490 re math.

package.json
umar8hassan: Fixed in https://github.com/mitodl/ocw-studio/pull/3139/commits/12381a09 - pinned the whole 41.4.2 set exactly.

static/js/lib/ckeditor/CKEditor.ts
umar8hassan: Fixed in https://github.com/mitodl/ocw-studio/pull/3139/commits/12381a09 - the comment was wrong. Rewrote it.
The buttons only change how the image looks inside the editor, and nothing reaches Hugo. The comment now says that outright instead of claiming it restores a real control.

I left the plugins in place. Removing them is a genuine cleanup, which predates this branch and is not something to fold into an upgrade. I'll open a ticket for it.

ChristopherChudzicki (approved):

PR #3160, opened 2026-08-19
fix: store relation blanks in the widget's shape so new content stays editable

umar8hassan:
https://github.com/mitodl/hq/issues/12926

`SiteConfig.generate_item_metadata()` builds blank metadata for a new item from the site config schema, but its leaf-value logic didn't know the relation widget's own shape. It returned a bare `[]`/`""` for a relation field instead of `{"content": [] or "", "website": <site name>}`. The content form's relation schema requires the `content` key, so the bare value failed as a type error and displayed as a false "is a required field" message, blocking every save on the resource until the value was replaced.

- `websites/site_config_api.py`: `generate_item_metadata()` gives relation fields their own blank-value branch, respecting `multiple`/`sortable` and a field-level `website` override, and accepts a `website_name` to fill in.
- `gdrive_sync/api.py`, `backfill_orphaned_caption_transcript_files.py`: pass the site name through, since both create resources via this generator.

### How can this be tested?
1. Check out `umar/12926-empty-captions-transcript-field-cant-be-saved`
2. Sync a video from GDrive with no caption/transcript files present. Confirm the resulting resource's edit drawer saves cleanly with no error.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -76,6 +76,21 @@ def omnibus_config(settings, db):  # noqa: ARG001
     return SiteConfig(parsed_config)
 
 
+@pytest.fixture
+def ocw_course_config(settings):
+    """Returns the parsed OCW course site config.
+
+    WebsiteStarterFactory's default config has no relation widgets, so tests
+    that care about relation fields build a starter from this instead.
+    """  # noqa: D401
+    with open(  # noqa: PTH123
+        os.path.join(  # noqa: PTH118
+            settings.BASE_DIR, "localdev/configs/ocw-course-site-config.yml"
+        )
+    ) as f:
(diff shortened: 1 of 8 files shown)


zawan-ila (approved):
Looks good. One unrelated thing I noted is that ocw-www is still using video_captions_file and video_transcript_file for its video resources. Is that expected?

umar8hassan:
> ... ocw-www is still using video_captions_file and video_transcript_file for its video resources. Is that expected?

I never found any data requiring multi-language transcripts/captions. A few videos that we have on RC don't have any transcripts/captions to begin with. I'll leave a comment on the parent ticket and will open a PR for ocw-www if needed for consistency.

PR #3182, opened 2026-09-01
fix: clarify youtube tags help text is per-tag not total

umar8hassan:
Closes mitodl/hq#13052

Paired with mitodl/ocw-hugo-projects#404, which holds the actual production config for this field.

The Youtube Tags label read as a 100 character limit total, but YouTube allows 100 characters per tag and 500 total across all tags. This confused the team during a retrospective.

- `localdev/configs/ocw-course-site-config.yml`: added a `help` line stating both limits, keeping the label itself short, matching the companion PR.

### How can this be tested?
1. Checkout `umar/13052-studio-youtube-tags-help-text`, run `python manage.py override_site_config`, and confirm the Youtube Tags field under Video Metadata shows the new help text.

### DIFF
--- a/localdev/configs/ocw-course-site-config.yml
+++ b/localdev/configs/ocw-course-site-config.yml
@@ -157,8 +157,7 @@ collections:
     - label: Youtube Speakers
       name: video_speakers
       widget: string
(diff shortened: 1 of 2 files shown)


pdpinch (approved):
lgtm!

PR #3199, opened 2026-09-09
Purge the MIT Learn Fastly distribution on live builds

zawan-ila:

Part of https://github.com/mitodl/hq/issues/13251

`ClearCdnCacheStep` purged only the OCW Fastly distribution. ocw-course-v3 content is
also served through the MIT Learn distribution, which no pipeline ever purged, so a live
v3 publish could leave stale content behind it.

Purging is now done with the [Fastly resource](https://github.com/mitodl/ol-concourse/tree/main/resources/fastly)
from `ol-concourse` rather than curl calls to the Fastly API

- `ClearCdnCacheStep` is a `PutStep` with `mode: surrogate_key` instead of a curl
  `TaskStep`.
- Live builds purge MIT Learn as a second put against a second resource. Draft and
  test builds are unchanged — only live content reaches that distribution.
- Services are identified by **domain**, which the resource resolves to a service ID at
  runtime. The domains already exist as settings per environment
  (`OCW_STUDIO_{DRAFT,LIVE}_URL`, `STATIC_API_BASE_URL_TEST`,
  `COURSE_V3_CANONICAL_DOMAIN`), so **no service IDs and no new credential need to be
  provisioned in Concourse.** The shared `((fastly.api_token))` is used, via a new
  `CONCOURSE_FASTLY_API_TOKEN_VAR` setting so that renaming the credential — or the
  field inside it — needs no code change.
- `CONCOURSE_HARD_PURGE` still applies, as the resource's `soft` param. Retries stay on
  the step, since the resource has no retry of its own.

Also deletes `mass-build-sites.yml`, a stale rendered pipeline copy referenced by no
code that still contained a **hand-written curl purge**.

### How can this be tested?

The purge testing will be performed on RC.
The Surrogate-Key plumbing this depends on merged in
https://github.com/mitodl/ol-infrastructure/pull/5800.

1. Warm two different courses until both report `x-cache: HIT`:
   `curl -sI https://rc.learn.mit.edu/courses/o/<course-slug>/ | grep -i x-cache`
2. Republish one in ocw-studio so that the build runs and the fastly cache is cleared.
3. Re-request both. The republished course must come back `x-cache: MISS` and the other must stay `HIT`.

Locally: upsert all affected pipelines (mass build, site build, theme-assets, site-removal, e2e tests) and verify that they run as expected.

### Checklist:
- [ ] Re-upsert pipelines after deploy (`upsert_mass_build_pipeline`, `upsert_theme_assets_pipeline`, `upsert_site_removal_pipeline`, `backpopulate_pipelines`) — each pipeline now declares a Fastly resource and resource type

### DIFF
--- a/content_sync/pipelines/definitions/concourse/common/steps.py
+++ b/content_sync/pipelines/definitions/concourse/common/steps.py
@@ -19,6 +19,7 @@ from ol_concourse.lib.models.pipeline import (
     TryStep,
 )
 
+from content_sync.constants import VERSION_LIVE
 from content_sync.pipelines.definitions.concourse.common.identifiers import (
     OCW_STUDIO_WEBHOOK_CURL_STEP_IDENTIFIER,
     OCW_STUDIO_WEBHOOK_RESOURCE_TYPE_IDENTIFIER,
@@ -31,6 +32,41 @@ from content_sync.pipelines.definitions.concourse.common.image_resources import 
     CURL_REGISTRY_IMAGE,
 )
 from content_sync.utils import get_ocw_studio_api_url
+
+# The Concourse var holding Fastly credentials for the MIT Learn distribution,
+# which serves ocw-course-v3 content.
+LEARN_FASTLY_VAR = "fastly_learn"
+# Only live content reaches the MIT Learn distribution, so it is purged alongside
+# this var and no other. Keying off the Fastly var rather than the pipeline name is
+# deliberate: the e2e test pipeline runs under pipeline_name "live" but purges
+# fastly_test, and it must not touch the learn distribution.
+LIVE_FASTLY_VAR = f"fastly_{VERSION_LIVE}"
+
+
+def fastly_purge_args(fastly_var: str, site_name: str) -> list[str]:
+    """
+    Build the curl arguments for a single Fastly surrogate key purge.
+
+    Args:
+        fastly_var(str): The name of the var to pull Fastly properties from
+        site_name(str): The surrogate key to purge from the cache
+
+    Returns:
+        list[str]: curl arguments for one purge request
+    """
+    args = [
+        "-f",
+        "-X",
+        "POST",
+        "-H",
+        f"Fastly-Key: (({fastly_var}.api_token))",
+    ]
+    if not settings.CONCOURSE_HARD_PURGE:
+        args.extend(["-H", "Fastly-Soft-Purge: 1"])
+    args.append(
+        f"https://api.fastly.com/service/(({fastly_var}.service_id))/purge/{site_name}"
+    )
+    return args
 
 
 def add_error_handling(  # noqa: PLR0913, PLR0917
@@ -172,8 +208,17 @@ class SlackAlertStep(TryStep):
 
 class ClearCdnCacheStep(TaskStep):
     """
-    A TaskStep using the curlimages/curl Docker image that sends an
-    API request to Fastly to clear the cache for a given URL
+    A TaskStep using the curlimages/curl Docker image that sends an API request to
+    Fastly to clear the cache for a given URL.
+
+    When purging the live distribution, the MIT Learn distribution that serves
+    ocw-course-v3 content is purged as well. Draft and test builds are not served
+    through it, so they render a single purge exactly as before.
+
+    The two purges run as separate transfers of one curl invocation, separated by
+    --next so that each carries its own Fastly-Key header. --fail-early is required:
+    without it curl exits 0 when an earlier transfer fails but a later one succeeds,
+    which would report a failed purge as a success.
 
     Args:
         name(str): The name to use as the Identifier for the task argument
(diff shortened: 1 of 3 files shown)


content_sync/pipelines/definitions/concourse/common/steps_test.py
```diff
@@ -147,31 +149,79 @@ def test_site_content_git_task_step(
...
+    """
+    Assert that a live ClearCdnCacheStep purges both the live distribution and the
+    MIT Learn distribution in a single curl invocation.
+    """
+    site_name = "test_site"
+    rendered_args = _rendered_curl_args(LIVE_FASTLY_VAR, site_name)
+    # Concourse passes args straight to exec, so shell quoting must never appear
+    for arg in rendered_args:
+        assert "'" not in arg
+    _assert_purges(rendered_args, LIVE_FASTLY_VAR, site_name)
```
pt2302: It would be good to assert that both transfers include `-f`. Without it, HTTP errors can return exit code 0 despite `--fail-early` (i.e., split on `--next` and check each transfer as a unit).

pt2302 (approved):
Looks good to me! One suggestion on improving a test.

content_sync/pipelines/definitions/concourse/common/steps_test.py
zawan-ila: Added in https://github.com/mitodl/ocw-studio/pull/3199/changes/f1a4a0d66dca5629d77fee24222c08c92eddb054

content_sync/pipelines/definitions/concourse/common/steps.py
```diff
@@ -32,6 +33,41 @@
 )
 from content_sync.utils import get_ocw_studio_api_url
 
+# The Concourse var holding Fastly credentials for the MIT Learn distribution,
+# which serves ocw-course-v3 content.
+LEARN_FASTLY_VAR = "fastly_learn"
```
blarghmatey, line 38: We currently already have a globally shared fastly token that is available under the var `fastly` so that you can use `((fastly.api_token))`, but it would also be good to make this an env-configurable value instead of hardcoding it in the event that we ever need to rename it.

content_sync/pipelines/definitions/concourse/common/steps.py
```diff
@@ -32,6 +33,41 @@
...
+# Only live content reaches the MIT Learn distribution, so it is purged alongside
+# this var and no other. Keying off the Fastly var rather than the pipeline name is
+# deliberate: the e2e test pipeline runs under pipeline_name "live" but purges
+# fastly_test, and it must not touch the learn distribution.
+LIVE_FASTLY_VAR = f"fastly_{VERSION_LIVE}"
```
blarghmatey, line 43: As above, we can use the shared global `fastly.api_token` here and the var name should be env-configurable.

blarghmatey (changes requested):

content_sync/pipelines/definitions/concourse/common/steps.py
feoh: That shared credential was duplicated into QA in https://github.com/mitodl/ol-infrastructure/pull/5809/changes

zawan-ila: @blarghmatey @feoh We also need access to the appropriate fastly service ids. How do you suggest I access those? Currently for the ocw services we have them available [here](https://github.com/mitodl/ol-infrastructure/blob/main/src/bridge/secrets/concourse/operations.production.yaml#L92-L100), but I don't see any place we can access the learn service id. The learn ids do appears to be present [here](https://github.com/mitodl/ol-infrastructure/blob/main/src/bridge/secrets/concourse/operations.production.yaml#L51-L53) but I do not think ocw pipelines can read that variable. Can we perhaps move those under shared/fastly?

zawan-ila: I'd like to have the service id available as fastly_learn.service_id so that I don't have to do too much surgery in this PR :-)

content_sync/pipelines/definitions/concourse/common/steps.py
zawan-ila: This variable also contains the appropriate service ids, which the shared `fastly` one does not.

content_sync/pipelines/definitions/concourse/common/steps.py
blarghmatey: For performing the fastly purges I suggest removing the logic embedded in these pipelines and instead using our Fastly resource - https://github.com/mitodl/ol-concourse/tree/main/resources/fastly

That handles the purging logic as well as being able to retrieve the service ID from the domain so that you don't have to incorporate it into this code. There is already library code in the `ol_concourse.lib.resources` module - https://github.com/mitodl/ol-concourse/blob/main/pipeline_lib/src/ol_concourse/lib/resources.py#L654-L706

You can refer to prior art here: https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_concourse/pipelines/infrastructure/k8s_apps/pipeline.py

That avoids having to provision any extra configuration in Concourse and reduces the complexity of your code here.

zawan-ila: Chris pointed me to that resource. However, I felt the other way about "reducing the complexity of the code here" :-)

It does seem to have the nice quality of being immune to any service id changes. I'll work on getting that wired up.

blarghmatey (approved):

PR #3200, opened 2026-09-10
feat: reference image gallery images by their resource uuid

zawan-ila:

https://github.com/mitodl/hq/issues/13086

Adds a `markdown_cleanup` rule (alias `image_gallery_item_uuid`) that gives each `image-gallery-item` a `uuid` param naming the image resource it points at.

```
{{< image-gallery-item href="c3e2...eff_diagram.png" text="Fig 1." >}}
{{< image-gallery-item uuid="c3e28341-74a4-2a89-c56c-3a1a5bcc0eff" href="c3e2...eff_diagram.png" text="Fig 1." >}}
```

- Additive only. `href`, `text` and `data-ngdesc` stay, because v2 galleries still render from `href`.
- An item is rewritten only once its uuid resolves to a live `type=resource` / `resourcetype=Image` record **in the same website** as the page.
- Anything that fails a check is left untouched and the reason recorded in the run's CSV `outcome` column, so a `--out` run reports every problem instead of aborting on the first.

### How can this be tested?
1. Ensure you have existing image gallery content in your local setup. If not, use [this script](https://gist.github.com/zawan-ila/24564b09614e64e718f66c18c996720c) to add some image gallery markdown to your local instance.
2. Dry-Run the cleanup `./manage.py markdown_cleanup image_gallery_item_uuid --out output.csv`
3. Inspect output.csv and verify that the dry run details are as expected.
4. Run the cleanup `./manage.py markdown_cleanup image_gallery_item_uuid --commit --out output.csv --skip-sync`
5. Verify that the image-gallery-items in the markdown now have uuid param that points to the image resource itself.

### DIFF
--- a/websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid.py
+++ b/websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid.py
@@ -112,10 +112,9 @@ class ImageGalleryItemUuidRule(PyparsingRule):
         if resource is None:
             return "cross_site" if self._exists_elsewhere(uuid) else "uuid_not_found"
 
+        metadata = resource.metadata if isinstance(resource.metadata, dict) else {}
         notes.resolved_type = resource.type
-        notes.resolved_resourcetype = (resource.metadata or {}).get(
-            "resourcetype"
-        ) or ""
+        notes.resolved_resourcetype = metadata.get("resourcetype") or ""
 
         if resource.deleted is not None:
             return "target_deleted"
--- a/websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid.py
+++ b/websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid.py
@@ -1,0 +1,153 @@
+"""Add resource uuids to image gallery items. See mitodl/hq#13086."""
+
+import posixpath
+import re
+from dataclasses import dataclass
+from uuid import UUID
+
+from websites.constants import CONTENT_TYPE_RESOURCE, RESOURCE_TYPE_IMAGE
+from websites.management.commands.markdown_cleaning.cleanup_rule import PyparsingRule
+from websites.management.commands.markdown_cleaning.parsing_utils import ShortcodeParam
+from websites.management.commands.markdown_cleaning.shortcode_parser import (
+    ShortcodeParser,
+    ShortcodeParseResult,
+)
+from websites.models import WebsiteContent
+
+ITEM_SHORTCODE = "image-gallery-item"
+
+# Gallery items reference their image by the legacy OCW filename, which is the
+# resource's uuid (dashes stripped) followed by the original filename.
+UUID_PREFIX_REGEX = re.compile(r"^(?P<hex>[0-9a-fA-F]{32})_.+$")
+
+
+class ImageGalleryItemUuidRule(PyparsingRule):
+    """
+    Add a `uuid` param naming the image resource an image-gallery-item points at.
+
+    Purely additive: `href`, `text` and `data-ngdesc` are left in place because
+    v2 galleries still render from `href`, and removing them would break those.
+
+        {{< image-gallery-item href="c3e2...eff_diagram.png" text="Fig 1." >}}
+
+    becomes
+
+        {{< image-gallery-item uuid="c3e28341-...-3a1a5bcc0eff"
+            href="c3e2...eff_diagram.png" text="Fig 1." >}}
+
+    An item is only rewritten once the uuid is confirmed to name a live image
+    resource in the same website as the page. Everything else is left exactly as
+    it was and the reason recorded in `outcome`, so a `--out` run reports the
+    problems rather than aborting on the first one.
+    """
+
+    alias = "image_gallery_item_uuid"
+
+    Parser = ShortcodeParser
+
+    @dataclass
+    class ReplacementNotes:
+        is_gallery_item: bool = False
+        outcome: str = ""
+        href: str = ""
+        uuid: str = ""
+        resolved_type: str = ""
+        resolved_resourcetype: str = ""
+
+    def should_parse(self, text: str):
+        """Only a few hundred pages site-wide contain a gallery."""
+        return ITEM_SHORTCODE in text
+
+    def replace_match(
+        self,
+        s: str,  # noqa: ARG002
+        l: int,  # noqa: ARG002, E741
+        toks: ShortcodeParseResult,
+        website_content: WebsiteContent,
+    ):
+        shortcode = toks.shortcode
+        original_text = toks.original_text
+
+        if shortcode.name != ITEM_SHORTCODE:
+            return original_text, self.ReplacementNotes()
+
+        notes = self.ReplacementNotes(is_gallery_item=True)
+        uuid = self._uuid_from_href(shortcode, notes)
+        if uuid is not None:
+            notes.outcome = self._check_target(uuid, website_content.website_id, notes)
+        if notes.outcome != "ok":
+            return original_text, notes
+
+        notes.uuid = str(uuid)
+        shortcode.params.insert(0, ShortcodeParam(name="uuid", value=str(uuid)))
+        return shortcode.to_hugo(), notes
+
+    def _uuid_from_href(self, shortcode, notes):
+        """
+        Extract the referenced uuid from the item's href.
+
+        Returns None and records why on `notes` when there is nothing to work
+        with.
+        """
+        if shortcode.get("uuid") is not None:
+            notes.outcome = "already_has_uuid"
+            return None
+
+        href = shortcode.get("href")
+        notes.href = href or ""
+        if not href:
+            notes.outcome = "no_href"
+            return None
+
+        match = UUID_PREFIX_REGEX.match(posixpath.basename(href.strip()))
+        if not match:
+            notes.outcome = "href_no_uuid_prefix"
+            return None
+
+        return UUID(match.group("hex"))
+
+    def _check_target(self, uuid: UUID, website_id, notes) -> str:
+        """Return the outcome of resolving `uuid` to a live image resource."""
+        resource = self._find_in_website(uuid, website_id)
+        if resource is None:
+            return "cross_site" if self._exists_elsewhere(uuid) else "uuid_not_found"
+
+        notes.resolved_type = resource.type
+        notes.resolved_resourcetype = (resource.metadata or {}).get(
+            "resourcetype"
+        ) or ""
+
+        if resource.deleted is not None:
+            return "target_deleted"
+        if resource.type != CONTENT_TYPE_RESOURCE:
+            return "wrong_content_type"
+        if notes.resolved_resourcetype != RESOURCE_TYPE_IMAGE:
+            return "wrong_resourcetype"
+        return "ok"
+
+    @staticmethod
+    def _text_id_spellings(uuid: UUID):
+        """text_id is stored dashed, but do not assume every legacy import was."""
+        return [str(uuid), uuid.hex]
+
+    def _find_in_website(self, uuid: UUID, website_id):
+        """
+        Find the referenced content within `website_id`.
+
+        Scoped to the one website deliberately: text_id is unique per
+        (website, text_id) and not globally, so a match in another site is not
+        safe to treat as the intended target.
+        """
+        return (
+            WebsiteContent.all_objects.filter(
+                website_id=website_id, text_id__in=self._text_id_spellings(uuid)
+            )
+            .order_by("id")
+            .first()
+        )
+
+    def _exists_elsewhere(self, uuid: UUID):
+        """Distinguish a dangling uuid from one belonging to a different site."""
+        return WebsiteContent.all_objects.filter(
+            text_id__in=self._text_id_spellings(uuid)
+        ).exists()
--- a/websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid.py
+++ b/websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid.py
@@ -124,11 +124,6 @@ class ImageGalleryItemUuidRule(PyparsingRule):
             return "wrong_resourcetype"
         return "ok"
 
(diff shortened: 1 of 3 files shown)


websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid.py
```diff
@@ -0,0 +1,153 @@
...
+
+        if shortcode.name != ITEM_SHORTCODE:
+            return original_text, self.ReplacementNotes()
+
+        notes = self.ReplacementNotes(is_gallery_item=True)
+        uuid = self._uuid_from_href(shortcode, notes)
+        if uuid is not None:
+            notes.outcome = self._check_target(uuid, website_content.website_id, notes)
+        if notes.outcome != "ok":
+            return original_text, notes
```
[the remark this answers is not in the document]
zawan-ila: Correct, and thanks — the defect was in the invocation I documented, not the rule.

Every non-`ok` outcome returns `original_text` unchanged, so `has_changed` is False and `cleaner.py:163-165` drops the row under `-ch`. The dry run I advertised would have reported 3205 changes and silently hidden all 3 skips, plus any future `uuid_not_found` / `cross_site` / invalid-target result — exactly the outcomes the rule exists to surface.

Fixed by dropping `-ch` from the documented run rather than changing the rule. I measured the cost on the real corpus: without the flag the 330 gallery-bearing pages yield **5973** CSV rows rather than 3205 — 3208 items, 660 container tags, 861 `resource_link`, and ~1200 table/anchor shortcodes. That is small enough to read, and the `is_gallery_item` and `outcome` columns filter it to the 3208 gallery rows in one step. A gallery-item-only parser would also work, but it would mean a custom grammar to save ~2700 rows, which does not seem worth the extra surface.

PR body updated with the corrected command.

websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid.py
```diff
@@ -0,0 +1,153 @@
...
+    def _check_target(self, uuid: UUID, website_id, notes) -> str:
+        """Return the outcome of resolving `uuid` to a live image resource."""
+        resource = self._find_in_website(uuid, website_id)
+        if resource is None:
+            return "cross_site" if self._exists_elsewhere(uuid) else "uuid_not_found"
+
+        notes.resolved_type = resource.type
+        notes.resolved_resourcetype = (resource.metadata or {}).get(
+            "resourcetype"
+        ) or ""
```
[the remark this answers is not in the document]
zawan-ila: Valid — the field is an unrestricted JSONField and I confirmed it accepts a non-dict: writing a list committed to the row (a pre-existing post_save signal then raised the same AttributeError, which is why such rows are hard to produce through normal paths). `(resource.metadata or {})` only rescues None and falsy values, so a truthy non-dict fell through to `.get()` and would have aborted the whole run — contradicting what this rule promises, which is to record the reason and continue.

Addressed by the autofix in 7f67e8c4, now normalising via `isinstance(resource.metadata, dict)`. Note the "also appears on line 141" part does not apply: there is only one metadata read in the file, and it is this one.

websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid.py
```diff
@@ -0,0 +1,152 @@
...
+        Scoped to the one website deliberately: text_id is unique per
+        (website, text_id) and not globally, so a match in another site is not
+        safe to treat as the intended target.
+        """
+        return (
+            WebsiteContent.all_objects.filter(
+                website_id=website_id, text_id__in=self._text_id_spellings(uuid)
+            )
+            .order_by("id")
+            .first()
```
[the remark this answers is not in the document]
zawan-ila: Valid mechanism, fixed in 93aa80dc — though by removing the ambiguity rather than detecting it.

The undashed spelling was speculative on my part and the data says it is dead code: all 3205 gallery items that resolve in production point at a 36-character dashed text_id, and none is stored undashed, so `uuid.hex` never matched anything. Detecting the ambiguity would have kept that dead branch and added a second one reachable only through it.

`_find_in_website` now filters on `text_id=str(uuid)` exactly. Since `unique_text_id` covers the exact (website, text_id) pair, that returns at most one row, so `.first()` can no longer choose between candidates and the `order_by("id")` tiebreak is gone. `_text_id_spellings` is deleted.

If an undashed record ever did exist, the rule now reports `uuid_not_found` rather than resolving it — a visible skip in the CSV instead of a silent rewrite naming the wrong resource, which is the failure direction this rule should prefer.

websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid.py
```diff
@@ -0,0 +1,148 @@
...
+        notes = self.ReplacementNotes(is_gallery_item=True)
+        uuid = self._uuid_from_href(shortcode, notes)
+        if uuid is not None:
+            notes.outcome = self._check_target(uuid, website_content.website_id, notes)
+        if notes.outcome != "ok":
+            return original_text, notes
+
+        notes.uuid = str(uuid)
+        shortcode.params.insert(0, ShortcodeParam(name="uuid", value=str(uuid)))
+        return shortcode.to_hugo(), notes
```
umar8hassan, line 83: [nit] `to_hugo()` rewrites the whole tag, not just the param we add. Anything the parser cannot round-trip gets silently rewritten and still reports `outcome=ok`.

Reproducible locally. Passed `text="Line one\nline two"`. Got `text="Line" "one" "line" "two"`. Caption gone, `outcome=ok`. Studio's editor can produce that shape, and this rule may get re-run later.

Optional guard: compare `to_hugo()` to `toks.original_text` before inserting. Skip if they differ.

websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def _uuid_from_href(self, shortcode, notes):
+        """
+        Extract the referenced uuid from the item's href.
+
+        Returns None and records why on `notes` when there is nothing to work
+        with.
+        """
+        if shortcode.get("uuid") is not None:
```
umar8hassan: This only looks for a *named* uuid, so `{{< image-gallery-item "positional" href="..." >}}` falls through and line 82 inserts `uuid=` ahead of the positional param. Hugo's docs are explicit that named and positional args can't be mixed in a single call, so that tag stops building. Worth an `any(p.name is None for p in shortcode.params)` guard with its own outcome — the fidelity check above won't catch this one, since the unmodified tag round-trips faithfully.

websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid.py
```diff
@@ -0,0 +1,148 @@
...
+        is_gallery_item: bool = False
+        outcome: str = ""
+        href: str = ""
+        uuid: str = ""
+        resolved_type: str = ""
+        resolved_resourcetype: str = ""
+
+    def should_parse(self, text: str):
+        """Only a few hundred pages site-wide contain a gallery."""
+        return ITEM_SHORTCODE in text
```
umar8hassan, line 59: [nit] This has no record on RC. This opts gallery pages into full shortcode parsing. The parser raises a bare `ValueError` on shapes it cannot handle. Unquoted `text={{< sub 2 >}}` gives `Unexpected shortcode nesting.` Nothing catches it before the `for wc in page` loop. The CSV is written after that loop. One bad page loses the whole run's report, including pages that were fine.

Reproduced locally. 2 gallery pages gave a CSV with 6 rows. The same 2 plus one bad page gave a ValueError and no CSV at all. RC has 0 parse failures across all 330 gallery pages. 79 items do contain nested shortcodes inside the tag, but all are quoted and parse fine.

websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid.py
```diff
@@ -0,0 +1,148 @@
...
+
+        href = shortcode.get("href")
+        notes.href = href or ""
+        if not href:
+            notes.outcome = "no_href"
+            return None
+
+        match = UUID_PREFIX_REGEX.match(posixpath.basename(href.strip()))
+        if not match:
+            notes.outcome = "href_no_uuid_prefix"
```
umar8hassan: Single-quoted params arrive here with their quotes intact — `hugo_unescape_shortcode_param_value` forces double-quote handling, so `href='<hex>_x.jpg'` keeps the apostrophes, the regex misses on the leading `'`, and we file it as `href_no_uuid_prefix`. Skipping is the right call, but the reason is misleading: that href does carry a usable uuid prefix. Either strip surrounding quotes before matching, or give it a distinct outcome so nobody triages it as a malformed filename.

websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid.py
```diff
@@ -0,0 +1,148 @@
...
+            return None
+
+        return UUID(match.group("hex"))
+
+    def _check_target(self, uuid: UUID, website_id, notes) -> str:
+        """Return the outcome of resolving `uuid` to a live image resource."""
+        resource = self._find_in_website(uuid, website_id)
+        if resource is None:
+            return "cross_site" if self._exists_elsewhere(uuid) else "uuid_not_found"
```
umar8hassan, line 113: [nit] Never fires on current data. 0 `cross_site` across 3208 RC items.

websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid.py
```diff
@@ -0,0 +1,148 @@
...
+
+        metadata = resource.metadata if isinstance(resource.metadata, dict) else {}
```
umar8hassan: nit: `(resource.metadata or {})` is the idiom we use for this field elsewhere (`websites/api.py:244`, `:338`). The isinstance form only differs for a truthy non-dict, which the schema won't produce.

websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid.py
```diff
@@ -0,0 +1,148 @@
...
+        Matched on the canonical dashed spelling alone. Every text_id in
+        production carries dashes, and `unique_text_id` covers the exact pair,
+        so this returns at most one row. Accepting an undashed spelling as well
+        would allow two rows in one website to normalise to the same uuid, and
+        the lookup would then have to guess which of them the emitted dashed
+        uuid actually names.
+        """
+        return WebsiteContent.all_objects.filter(
+            website_id=website_id, text_id=str(uuid)
+        ).first()
```
umar8hassan: This is one query per gallery item, plus a second from line 148 for anything unresolved. The other rules that resolve content — `baseurl`, `link_resolveuid`, `link_wrapped_images` — build a `ContentLookup` once in `__init__` and answer from memory for the whole run.

`ContentLookup.by_uuid` is keyed globally so it won't serve the site-scoped lookup you want here, but adding a `(website_id, uuid)` index to it would keep this rule on the same path as the rest and leave the next rule with a reusable lookup instead of hand-rolled queries.

websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid_test.py
```diff
@@ -0,0 +1,295 @@
...
+        params.append(f'href="{href}"')
+    if ngdesc is not None:
+        params.append(f'data-ngdesc="{ngdesc}"')
+    if text is not None:
+        params.append(f'text="{text}"')
+    return "{{< image-gallery-item " + " ".join(params) + " >}}"
+
+
+def gallery(*items):
+    return "\n".join([GALLERY_OPEN, *items, GALLERY_CLOSE])
```
umar8hassan: Every test passes exactly one item. Nothing exercises a page with several. Worth adding a test with more than one item to reflect actual scenarios.

drf_lint_baseline.json
```diff
@@ -1,4 +1,19 @@
...
+  "websites/serializers.py:234:15:ORM006",
+  "websites/serializers.py:242:15:ORM003",
+  "websites/serializers.py:317:19:ORM004",
+  "websites/serializers.py:322:15:ORM004",
+  "websites/serializers.py:326:15:ORM004",
+  "websites/serializers.py:761:15:ORM006",
+  "websites/serializers.py:774:33:ORM006",
+  "websites/serializers.py:806:39:ORM006",
+  "websites/serializers.py:837:25:ORM004",
+  "websites/serializers.py:990:38:ORM006"
```
umar8hassan: 15 new suppressions landing with the drf-lint bump. The drf-lint, renovate, and pre-commit commits are 6 of the 9 files here and are unrelated to galleries. It'll be easier to review and revert, if needed, in a separate PR.

renovate.json
```diff
@@ -1,6 +1,9 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": ["local>mitodl/.github:renovate-config"],
+  "pre-commit": {
+    "enabled": true
```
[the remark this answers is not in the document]
umar8hassan: worth considering

umar8hassan (approved):
Looks Good 👍 

Although Claude suggested a few improvements to avoid data corruption which edge cases but I have verified those cases do not exist on RC and are harmless. I have suggested a few nits. Let's dry run and analyze output on RC before applying fixes.

websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid_test.py
zawan-ila: Added a test `test_rewrites_every_gallery_on_a_page` with multiple galleries

websites/management/commands/markdown_cleaning/rules/image_gallery_item_uuid.py
zawan-ila: Ignoring this and other related nits since they do not pose any problems for us as our production and RC databases do not contain any of the problematic data

PR #3205, opened 2026-09-15
Add a minimal editor variant with sub/superscript

zawan-ila:

https://github.com/mitodl/hq/issues/13087

Adds a third minimal editor variant, `minimal: with-subsup`, following the same shape as #2869 which introduced `with-math`. Fields that opt into it get working subscript and superscript; nothing else changes.

**Why a separate variant** rather than adding sub/sup to the existing minimal configs: it keeps every field currently on `minimal: true` byte-identical, and matches the precedent already set for `with-math`. The trade-off is that `minimal` becomes a three-valued enum of feature combinations, so a future field wanting both math and sub/sup would need a fourth variant.

### How can this be tested?
Needs to be tested together with https://github.com/mitodl/ocw-hugo-projects/pull/409
1. Pull in the starter configs from https://github.com/mitodl/ocw-hugo-projects/pull/409 `GITHUB_WEBHOOK_BRANCH=copilot/add-allowed-html-to-image-metadata ./manage.py import_website_starters https://github.com/mitodl/ocw-hugo-projects`
2. Load any image resource in studio
3. Verify that the sup and sub toolbar icons show for the captions field.
4. Add a caption containing sup and sub and save the resource
5. Reload the page and verify that the superscripts and subscripts render as expected.

### Additional Context

1. sub/sup currently exist as Hugo shortcodes (`{{< sub 2 >}}`) in image captions fields. Converting them to \<sup>/\<sub> will require separate data fixes.


static/js/components/widgets/MarkdownEditor.test.tsx
```diff
@@ -110,6 +112,12 @@ describe("MarkdownEditor", () => {
       expectedConfig: MinimalWithMathEditorConfig,
       configLabel: "MinimalWithMathEditorConfig",
     },
+    {
```
umar8hassan: Neither new case asserts that sub/sup survive. This one deep-equals the exported config object, and the `CKEditor.test.ts` case resolves toolbar items. Both pass against a config that drops the tags.

Booted `MinimalWithSubSupEditorConfig` with `allowedHtml: []`. Passed `H<sub>2</sub>O`. Got `H2O`. That is the exact failure this PR fixes, and nothing added here would catch it coming back.

`Markdown.test.ts` already has the harness. One `markdownTest` case asserting `H<sub>2</sub>O` round-trips under this config would cover it.

Separately, the `otherProps: { allowedHtml: ["sub", "sup"] }` on this case is load-bearing rather than decoration. Without it `toolbarItemsFilter` strips the two toolbar items and the deep-equal fails.

static/js/lib/ckeditor/CKEditor.ts
```diff
@@ -233,3 +233,43 @@ export const MinimalWithMathEditorConfig = {
...
+    ListPlugin,
+    ParagraphPlugin,
+    ResourceEmbed,
+    ResourcePicker,
+    ResourceLink,
+    SubscriptPlugin,
+    SuperscriptPlugin,
+    MarkdownListSyntax,
+    Markdown,
+    LegacyShortcodes,
```
umar8hassan: [nit] `LegacyShortcodes` is registered after `Markdown`, so its turndown rules never reach the data processor.

Passed `Water H{{< sub 2 >}}O` through this config. Got `Water H{{\< sub 2 >}}O`. Hugo stops recognising the delimiter, so the published page renders the literal shortcode text.

Not a regression. `MinimalEditorConfig` and `MinimalWithMathEditorConfig` already have this order, and I confirmed a caption on the current config is corrupted the same way. Raising it here because this variant exists for the fields that hold `{{< sub >}}` today, and the description notes those still need a data fix. Worth knowing that ordinary edits are already mangling the content that fix will run against.

`FullEditorConfig` orders these the other way and round-trips to `{{< sub "2" >}}`. Moving `LegacyShortcodes` above `Markdown` in this array gives the same result. Verified both orders.

static/js/lib/ckeditor/CKEditor.ts
```diff
@@ -233,3 +233,43 @@ export const MinimalWithMathEditorConfig = {
...
+    CodePlugin,
+    BlockQuotePlugin,
+    LinkPlugin,
+    ListPlugin,
+    ParagraphPlugin,
+    ResourceEmbed,
+    ResourcePicker,
+    ResourceLink,
+    SubscriptPlugin,
+    SuperscriptPlugin,
```
umar8hassan: [nit] Both plugins load unconditionally, but `toolbarItemsFilter` gates the two buttons on `allowed_html` independently. A field can end up with one and not the other.

Passed `x<sup>2</sup>` through this config with `allowed_html: ["sub"]`. Got `x2`. The superscript command is still registered, the button is hidden, and existing superscript content in that field is destroyed on the next save with nothing shown to the author.

static/js/lib/ckeditor/CKEditor.ts
```diff
@@ -233,3 +233,43 @@ export const MinimalWithMathEditorConfig = {
   },
   language: "en",
 }
+
+export const MinimalWithSubSupEditorConfig = {
```
umar8hassan: [nit] Fourth near-identical config. This block differs from `MinimalEditorConfig` by two plugin entries and two toolbar entries.

The description already calls out that a field wanting math and sub/sup needs a fifth. A factory taking a feature list would collapse the three minimal variants into one plugin list, and would put the `LegacyShortcodes` ordering in one place instead of three.

static/js/components/widgets/MarkdownEditor.tsx
```diff
@@ -147,9 +149,11 @@ export default function MarkdownEditor(props: Props): JSX.Element {
     const baseConfig =
       minimal === MINIMAL_WITH_MATH
         ? MinimalWithMathEditorConfig
-        : minimal
-          ? MinimalEditorConfig
-          : FullEditorConfig
+        : minimal === MINIMAL_WITH_SUBSUP
```
umar8hassan: [nit] Selecting this variant without `allowed_html: ["sub", "sup"]` is a silent no-op.

Passed `H<sub>2</sub>O` with `minimal: "with-subsup"` and `allowedHtml: []`. Got `H2O`. Both buttons are filtered out and existing tags are stripped on save. Nothing surfaces in the UI or the console.

The reverse mistake is equally quiet. Passed the same input with `minimal: true` and `allowedHtml: ["sub", "sup"]`. Got `H2O`.

The companion PR sets both, so this is not a live problem today. A `console.warn` when the variant is set without the matching `allowed_html` would stop the next field getting it wrong.

static/js/components/widgets/MarkdownEditor.tsx
```diff
@@ -147,9 +149,11 @@ export default function MarkdownEditor(props: Props): JSX.Element {
...
+          ? MinimalWithSubSupEditorConfig
+          : minimal
```
umar8hassan: [nit] The truthiness fallback absorbs any unrecognised string. `minimal: "with-subsup "` with a trailing space lands on `MinimalEditorConfig`, and neither TypeScript nor this branch objects. The yamale enum is the only thing that catches it.

websites/config_schema/site-config-schema.yml
```diff
@@ -25,7 +25,7 @@ field:
     'object', 'relation', 'menu', 'hierarchical-select', 'website-collection')
   label: str()
   name: str()
-  minimal: any(bool(), enum('with-math'), required=False)
+  minimal: any(bool(), enum('with-math', 'with-subsup'), required=False)
```
umar8hassan: Confirmed the deploy ordering. Validated the caption block from mitodl/ocw-hugo-projects#409 against both schemas. The master schema rejects it with `'with-subsup' not in ('with-math',)`. This schema accepts it. So this has to be deployed before that config is imported.

static/js/components/widgets/MarkdownEditor.test.tsx
umar8hassan: Neither new case asserts that sub/sup survive. This one deep-equals the exported config object, and the `CKEditor.test.ts` case resolves toolbar items. Both pass against a config that drops the tags.

Booted `MinimalWithSubSupEditorConfig` with `allowedHtml: []`. Passed `H<sub>2</sub>O`. Got `H2O`. That is the exact failure this PR fixes, and nothing added here would catch it coming back.

`Markdown.test.ts` already has the harness. One `markdownTest` case asserting `H<sub>2</sub>O` round-trips under this config would cover it.

Separately, the `otherProps: { allowedHtml: ["sub", "sup"] }` on this case is load-bearing rather than decoration. Without it `toolbarItemsFilter` strips the two toolbar items and the deep-equal fails.

static/js/lib/ckeditor/CKEditor.ts
umar8hassan, line 265: [nit] `LegacyShortcodes` is registered after `Markdown`, so its turndown rules never reach the data processor.

Passed `Water H{{< sub 2 >}}O` through this config. Got `Water H{{\< sub 2 >}}O`. Hugo stops recognising the delimiter, so the published page renders the literal shortcode text.

Not a regression. `MinimalEditorConfig` and `MinimalWithMathEditorConfig` already have this order, and a caption on the current config is corrupted the same way.

`FullEditorConfig` orders these the other way. Moving `LegacyShortcodes` above `Markdown` in this array gives the same result. Verified both orders.

static/js/lib/ckeditor/CKEditor.ts
umar8hassan: [nit] Both plugins load unconditionally, but `toolbarItemsFilter` gates the two buttons on `allowed_html` independently. A field can end up with one and not the other.

Passed `x<sup>2</sup>` through this config with `allowed_html: ["sub"]`. Got `x2`. The superscript command is still registered, the button is hidden, and existing superscript content in that field is destroyed on the next save with nothing shown to the author.

static/js/lib/ckeditor/CKEditor.ts
umar8hassan, line 237: [nit] Fourth near-identical config. This block differs from `MinimalEditorConfig` by two plugin entries and two toolbar entries.

The description already calls out that a field wanting math and sub/sup needs a fifth. A factory taking a feature list would collapse the three minimal variants into one plugin list, and would put the `LegacyShortcodes` ordering in one place instead of three.

static/js/components/widgets/MarkdownEditor.tsx
umar8hassan: [nit] Selecting this variant without `allowed_html: ["sub", "sup"]` is a silent no-op.

Passed `H<sub>2</sub>O` with `minimal: "with-subsup"` and `allowedHtml: []`. Got `H2O`. Both buttons are filtered out and existing tags are stripped on save. Nothing surfaces in the UI or the console.

The reverse mistake is equally quiet. Passed the same input with `minimal: true` and `allowedHtml: ["sub", "sup"]`. Got `H2O`.

The companion PR sets both, so this is not a live problem today. A `console.warn` when the variant is set without the matching `allowed_html` would stop the next field getting it wrong.

static/js/components/widgets/MarkdownEditor.tsx
umar8hassan: [nit] The truthiness fallback absorbs any unrecognised string. `minimal: "with-subsup "` with a trailing space lands on `MinimalEditorConfig`, and neither TypeScript nor this branch objects. The yamale enum is the only thing that catches it.

websites/config_schema/site-config-schema.yml
umar8hassan, line 28: Confirmed the deploy ordering. Validated the caption block from mitodl/ocw-hugo-projects#409 against both schemas. The master schema rejects it with `'with-subsup' not in ('with-math',)`. This schema accepts it. So this has to be deployed before that config is imported.

umar8hassan (approved):
Looks Good 👍

static/js/lib/ckeditor/CKEditor.ts
zawan-ila: Swapped the order in 40e32e5b6513d4e648afb3da7bac2283b02a8399

PR #3207, opened 2026-09-15
feat: make the local S3 and pipeline API addresses configurable

gumaerc:

Closes mitodl/hq#13381 (together with mitodl/ol-infrastructure#5882)

Makes the docker compose network addresses configurable so ocw-studio can run somewhere other than compose. Nothing changes for compose, which keeps the same addresses as fallbacks.

Three were hardcoded:

- `main/s3_utils.py` forced `http://10.1.0.100:9000` whenever the environment is dev. That is the compose Minio container's static IP, so every boto3 call hangs elsewhere. Now honours `AWS_S3_ENDPOINT_URL`, which django-storages reads too. `AWS_S3_CUSTOM_DOMAIN` comes along with it, because the address a pod uses and the one a browser can reach are not always the same.
- `get_cli_endpoint_url()` put the same IP into every `aws` call in a pipeline task. Now uses the same setting, since the object store is at one address for both the app and the task containers.
- `get_ocw_studio_api_url()` returned the compose nginx container's IP in dev, used by pipelines to report build status back. It cannot just fall back to `SITE_BASE_URL`, which is a localhost URL under compose and would point a task container at itself, so `OCW_STUDIO_PIPELINE_API_URL` is the override.

The S3 endpoint was spelled out in four separate places, so fixing them one at a time meant each pipeline failed in turn. They now share `get_s3_endpoint_url()`, with a test asserting no accessor emits a `10.1.0.x` address once configured.

### How can this be tested?

`docker compose run --rm web pytest content_sync main/s3_utils_test.py` passes (2947 in content_sync, plus the new s3_utils cases). Behaviour under compose is unchanged: with none of the new settings set, every accessor returns exactly what it did before.

Exercised for real against the k8s local-dev stack in mitodl/ol-infrastructure#5882: theme assets build and upload, a site publish triggered from the CMS, and the pipeline's status callback to Studio.

### Additional Context

`OCW_STUDIO_PIPELINE_API_URL` is new and unused under compose. The other two settings have no effect unless set.


content_sync/pipelines/definitions/concourse/common/resources.py
```diff
@@ -194,10 +193,11 @@ def __init__(self, bucket: str, branch, **kwargs):
             },
             **kwargs,
         )
-        if is_dev():
+        endpoint_url = get_s3_endpoint_url()
+        if endpoint_url:
```
umar8hassan: This gate used to be `is_dev()`. Now it fires in any environment where `AWS_S3_ENDPOINT_URL` is set, and `main/s3_utils_test.py` treats that as a supported combination (`ENVIRONMENT = "production"` plus an endpoint). Two ways it bites:

- Keys populated: the pipeline config Concourse stores now holds the real secret key in plain text, readable by anyone with `fly get-pipeline` or UI access.
- Keys empty, which is the EKS case with an instance profile: it writes `access_key_id: ""` and `secret_access_key: ""`. Per the comment right above, the dev overrides take precedence over `enable_aws_creds_provider`, so this quietly undoes #3206.

It also splits from the `aws` CLI tasks in `theme_assets_pipeline.py` and `s3_bucket_sync_pipeline.py`, which still gate keys on `is_dev()`. In that same config one side gets `--endpoint-url` with no keys and 403s.

Separating the two would keep both honest:

```python
endpoint_url = get_s3_endpoint_url()
if endpoint_url:
    self.source["endpoint"] = endpoint_url
if is_dev():
    self.source.update(
        {
            "access_key_id": (settings.AWS_ACCESS_KEY_ID or ""),
            "secret_access_key": (settings.AWS_SECRET_ACCESS_KEY or ""),
        }
    )
```

content_sync/utils.py
```diff
@@ -149,14 +150,39 @@ def get_common_pipeline_vars():
...
-    """Get the ocw-studio URL, which points to Docker for local dev"""
-    return "http://10.1.0.102:8043" if is_dev() else settings.SITE_BASE_URL
+    """Get the ocw-studio URL as a pipeline task container sees it.
+
+    Not SITE_BASE_URL in dev: under docker-compose that is a localhost URL,
+    which means the task container itself rather than the app. The fallback is
+    the compose nginx container's static IP; OCW_STUDIO_PIPELINE_API_URL is the
+    escape hatch for any other local setup, where neither address resolves.
+    """
+    if is_dev():
```
umar8hassan: `OCW_STUDIO_PIPELINE_API_URL` only does anything when `is_dev()`. Anywhere else it is read and dropped, with nothing in the logs to say so.

That is the opposite of `AWS_S3_ENDPOINT_URL` two functions up, where the docstring says it "wins wherever it is set, in any environment", and the example in `.env.example` is `https://studio.ocw.example.com`, which reads like a real deployment rather than a laptop.

So picture someone on RC setting it to an in-cluster address because `SITE_BASE_URL` is not reachable from the workers. They restart, re-upsert the pipelines, and every status callback still goes to `SITE_BASE_URL`. Nothing errors, so they go digging in the wrong layer.

Either honour it everywhere:

```python
return settings.OCW_STUDIO_PIPELINE_API_URL or (
    DEV_OCW_STUDIO_API_URL if is_dev() else settings.SITE_BASE_URL
)
```

or say "only when `OCW_STUDIO_ENVIRONMENT=dev`" in the setting description and in `.env.example`, and use an example value that looks local.

content_sync/utils.py
```diff
@@ -149,14 +150,39 @@ def get_common_pipeline_vars():
...
-    return f" --endpoint-url {DEV_ENDPOINT_URL}" if is_dev() else ""
+    """Get the S3 endpoint-url flag for `aws` calls inside pipeline tasks."""
+    endpoint_url = get_s3_endpoint_url()
+    return f" --endpoint-url {endpoint_url}" if endpoint_url else ""
 
 
 def get_ocw_studio_api_url():
-    """Get the ocw-studio URL, which points to Docker for local dev"""
-    return "http://10.1.0.102:8043" if is_dev() else settings.SITE_BASE_URL
+    """Get the ocw-studio URL as a pipeline task container sees it.
```
umar8hassan: Worth deciding now what a base URL with a path prefix should do, since this one is described as a "Base URL" and that invites someone to put a prefix on it.

The two callers disagree. `common/resources.py:113` joins with `urljoin(base, "api/...")`, `common/steps.py:337` joins with `base.rstrip("/") + "/api/..."`, and `urljoin` drops the prefix unless the base ends in a slash:

| base | urljoin | rstrip |
| --- | --- | --- |
| `https://example.edu/studio` | `https://example.edu/api/...` | `https://example.edu/studio/api/...` |
| `https://example.edu/studio/` | `https://example.edu/studio/api/...` | `https://example.edu/studio/api/...` |

Pre-existing for `SITE_BASE_URL`, but newly reachable. The result is the webhook resource 404ing while the curl step works, which is a miserable thing to debug. Normalising here, always returning with a trailing slash, would settle it in one place.

content_sync/utils.py
```diff
@@ -149,14 +150,39 @@ def get_common_pipeline_vars():
...
+    """Get the S3 endpoint pipelines should use, or None for real AWS.
+
+    AWS_S3_ENDPOINT_URL wins wherever it is set, in any environment. The
+    DEV_ENDPOINT_URL fallback is the docker-compose Minio container's static
+    IP, which only exists in that setup -- anywhere else running a local S3
+    emulator has to say where it is, or pipeline steps and resource checks
+    hang against an address that never answers.
+
+    Everything that needs the endpoint goes through here. It was previously
+    spelled out separately in four places, and fixing them one at a time meant
```
umar8hassan: Five, unless I miscounted. `e2e_test_site_pipeline.py:64` still builds its own on this branch:

```python
CLI_ENDPOINT_URL = f" --endpoint-url {DEV_ENDPOINT_URL}" if is_dev() else ""
```

Harmless today, since nothing reads that constant and the real `aws` calls in that file already go through `get_cli_endpoint_url()`. But it looks live, so it will send the next person down the wrong path. Worth deleting along with its `DEV_ENDPOINT_URL` import.

content_sync/utils.py
```diff
@@ -149,14 +150,39 @@ def get_common_pipeline_vars():
...
+    """
+    return settings.AWS_S3_ENDPOINT_URL or (DEV_ENDPOINT_URL if is_dev() else None)
+
+
 def get_cli_endpoint_url():
-    """Get the S3 endpoint-url, which points to Minio for local dev"""
-    return f" --endpoint-url {DEV_ENDPOINT_URL}" if is_dev() else ""
+    """Get the S3 endpoint-url flag for `aws` calls inside pipeline tasks."""
+    endpoint_url = get_s3_endpoint_url()
+    return f" --endpoint-url {endpoint_url}" if endpoint_url else ""
```
umar8hassan: Nit: this goes straight into `sh -exc` bodies. The value is ours so I am not worried about injection, but a trailing newline or space, which is easy to pick up from a k8s `stringData:` block, splits the `aws` command line and surfaces as a baffling shell error inside a Concourse task. A `.strip()` in `get_s3_endpoint_url()` would cover it.

content_sync/utils_test.py
```diff
@@ -409,3 +430,29 @@ def test_is_extra_theme(settings, theme_slug, extra_themes, expected):
     """is_extra_theme should return True only if theme_slug is in OCW_EXTRA_COURSE_THEMES"""
     settings.OCW_EXTRA_COURSE_THEMES = extra_themes
     assert is_extra_theme(theme_slug) == expected
+
+
+def test_no_compose_addresses_when_configured(settings, mocker):
```
umar8hassan: The docstring says this stops a new pipeline quietly reintroducing a compose address, but the test never renders a pipeline. It calls the same three accessors that are already parametrised twenty lines up and checks their overrides won.

The proof is that `e2e_test_site_pipeline.py:64` still hardcodes `DEV_ENDPOINT_URL` on this branch and this test passes.

The `get_common_pipeline_vars()` part is inert too. The autouse `required_concourse_settings` fixture already sets every URL it reads, and `RESOURCE_BASE_URL_DRAFT`/`_LIVE` default to `""`.

To do what it says, build the real definitions (`ThemeAssetsPipelineDefinition`, `SitePipelineDefinition`, `UnpublishedSiteRemovalPipelineDefinition`, `S3BucketSyncPipelineDefinition`), dump them to JSON and assert `"10.1.0." not in rendered`. Otherwise I would drop it, since a green test that cannot fail is worse than no test at all.

.env.example
```diff
@@ -48,6 +48,22 @@ OCW_STUDIO_USE_S3=true
 MINIO_ROOT_USER=changeme
 MINIO_ROOT_PASSWORD=changeme
 
+# S3 endpoint override. Leave unset for docker compose: with
+# OCW_STUDIO_ENVIRONMENT=dev the app falls back to the compose Minio
+# container's static IP. Set it when S3 lives somewhere else, e.g. the
+# k8s local-dev stack:
+# AWS_S3_ENDPOINT_URL=http://rustfs.local-infra.svc.cluster.local:9000
```
umar8hassan: Leaving this here because `pyproject.toml` is not in the diff, but this is the line that triggers it.

The `[tool.pytest.ini_options] env` list blanks `AWS_ENDPOINT_URL`, `DRIVE_SHARED_ID` and friends precisely so `.env` values cannot reach tests, and compose gives `web` an `env_file: .env`. Neither new setting was added to it.

So someone uncomments this line, runs the command from the PR description, and the module-level `CLI_ENDPOINT_URL` constants pick the value up at import time. `theme_assets_pipeline_test.py:82` and `site_pipeline_test.py:158/218/833` assert the literal `DEV_ENDPOINT_URL` and fail. CI stays green because CI has no `.env`, so it only ever hits people locally.

Two lines fix it:

```
"AWS_S3_ENDPOINT_URL=",
"OCW_STUDIO_PIPELINE_API_URL=",
```

The new tests here already set both to `None` by hand, which I read as the same thing being noticed and patched one spot at a time.

.env.example
```diff
@@ -48,6 +48,22 @@ OCW_STUDIO_USE_S3=true
...
+# AWS_S3_CUSTOM_DOMAIN is what django-storages builds media URLs from, for
+# when the browser cannot reach AWS_S3_ENDPOINT_URL directly:
+# AWS_S3_CUSTOM_DOMAIN=s3.example.com/ol-ocw-studio-app-local
+# The same endpoint is used for `aws` calls inside Concourse task
+# containers, so it must resolve from there too.
+#
+# Base URL for this app as a Concourse task container reaches it. Unset
+# under docker compose, where the pipeline falls back to the nginx
+# container's static IP:
+# OCW_STUDIO_PIPELINE_API_URL=https://studio.ocw.example.com
```
umar8hassan: If this setting stays dev only, this example works against it. A public https URL reads like a production address, so someone will try it in one. An address from the k8s local-dev stack would make the scope obvious. See the note on `get_ocw_studio_api_url`.

content_sync/pipelines/definitions/concourse/theme_assets_pipeline.py
```diff
@@ -34,10 +34,11 @@
     SlackAlertStep,
     clear_cdn_cache_steps,
 )
+from content_sync.utils import get_cli_endpoint_url
 from main.utils import is_dev
 from websites.constants import OCW_HUGO_THEMES_GIT
 
-CLI_ENDPOINT_URL = f" --endpoint-url {DEV_ENDPOINT_URL}" if is_dev() else ""
+CLI_ENDPOINT_URL = get_cli_endpoint_url()
```
umar8hassan: This and `s3_bucket_sync_pipeline.py:19` call the helper at import time, which freezes the setting before any test can override it. That is the mechanism behind the `.env` leak I flagged on `.env.example`.

`remove_unpublished_sites.py` and `site_pipeline.py` already call it inside `__init__`. Moving these two there makes them overridable and removes the whole class of problem.

content_sync/pipelines/definitions/concourse/remove_unpublished_sites.py
```diff
@@ -48,7 +48,7 @@
 )
 from main.utils import is_dev
 
-CLI_ENDPOINT_URL = f" --endpoint-url {DEV_ENDPOINT_URL}" if is_dev() else ""
+CLI_ENDPOINT_URL = get_cli_endpoint_url()
```
umar8hassan: Nit: nothing reads this one. `__init__` calls `get_cli_endpoint_url()` again at line 95 and the task bodies use that local. Safe to drop, along with the import if it goes unused.

main/settings.py
```diff
@@ -443,6 +443,33 @@
...
+OCW_STUDIO_PIPELINE_API_URL = get_string(
+    name="OCW_STUDIO_PIPELINE_API_URL",
+    default=None,
+    description=(
+        "Base URL for this app as a Concourse task container reaches it, for "
+        "local setups where neither SITE_BASE_URL nor the docker-compose "
+        "address resolves from inside a pipeline task."
+    ),
+)
+AWS_S3_CUSTOM_DOMAIN = get_string(
```
umar8hassan: django-storages builds these URLs as `{AWS_S3_URL_PROTOCOL or "https:"}//{domain}/{key}`, and `AWS_S3_URL_PROTOCOL` is not declared here. Same reason the other two needed declaring: Django never reads it, so it cannot be set from the environment.

On a local k8s stack serving plain HTTP, every media URL comes out `https://` and images fail on a TLS error, with no setting available to fix it. Either declare it alongside these two, or say in `.env.example` that the custom domain has to speak https.

main/settings.py
```diff
@@ -443,6 +443,33 @@
     default=False,
     description="Enables querystring auth for S3 urls",
 )
+AWS_S3_ENDPOINT_URL = get_string(
```
umar8hassan: Nit: `AWS_S3_ENDPOINT_URL=rustfs:9000`, with the scheme left off, is accepted here and only fails at the first boto3 call as `ValueError: Invalid endpoint`. A scheme check next to the `OCW_STUDIO_USE_S3` block below would catch the typo at startup. Empty string is already handled, it falls through to the dev fallback.

main/s3_utils.py
```diff
@@ -18,7 +18,13 @@ def get_boto3_options(extra_options=None):
...
         "aws_secret_access_key": settings.AWS_SECRET_ACCESS_KEY,
     }
-    if settings.ENVIRONMENT == "dev":
+    # AWS_S3_ENDPOINT_URL wins wherever it is set. The hardcoded address below
+    # is the docker-compose Minio container's static IP, which only exists in
+    # that setup -- anywhere else (the k8s local-dev stack, say) needs to say
+    # where its own S3 endpoint is.
+    if settings.AWS_S3_ENDPOINT_URL:
+        options.update({"endpoint_url": settings.AWS_S3_ENDPOINT_URL})
+    elif settings.ENVIRONMENT == "dev":
```
umar8hassan: Nit: the address on the next line is the same one as `content_sync/constants.py:DEV_ENDPOINT_URL`. I realise the import cannot go the other way, since `content_sync.utils` imports this module, but `main/constants.py` exists and `content_sync/constants.py` could re-export from there. One place to change instead of two, or four counting both test files.

The function docstring above also still says "connecting to Minio if the environment is dev", which is now only half the story.

content_sync/pipelines/definitions/concourse/common/resources.py
umar8hassan: This gate used to be `is_dev()`. Now it fires in any environment where `AWS_S3_ENDPOINT_URL` is set, and `main/s3_utils_test.py` treats that as a supported combination (`ENVIRONMENT = "production"` plus an endpoint). Two ways it bites:

- Keys populated: the pipeline config Concourse stores now holds the real secret key in plain text, readable by anyone with `fly get-pipeline` or UI access.
- Keys empty, which is the EKS case with an instance profile: it writes `access_key_id: ""` and `secret_access_key: ""`. Per the comment right above, the dev overrides take precedence over `enable_aws_creds_provider`, so this quietly undoes #3206.

It also splits from the `aws` CLI tasks in `theme_assets_pipeline.py` and `s3_bucket_sync_pipeline.py`, which still gate keys on `is_dev()`. In that same config one side gets `--endpoint-url` with no keys and 403s.

Separating the two would keep both honest:

```python
endpoint_url = get_s3_endpoint_url()
if endpoint_url:
    self.source["endpoint"] = endpoint_url
if is_dev():
    self.source.update(
        {
            "access_key_id": (settings.AWS_ACCESS_KEY_ID or ""),
            "secret_access_key": (settings.AWS_SECRET_ACCESS_KEY or ""),
        }
    )
```

content_sync/utils.py
umar8hassan: `OCW_STUDIO_PIPELINE_API_URL` only does anything when `is_dev()`. Anywhere else it is read and dropped, with nothing in the logs to say so.

That is the opposite of `AWS_S3_ENDPOINT_URL` two functions up, where the docstring says it "wins wherever it is set, in any environment", and the example in `.env.example` is `https://studio.ocw.example.com`, which reads like a real deployment rather than a laptop.

So picture someone on RC setting it to an in-cluster address because `SITE_BASE_URL` is not reachable from the workers. They restart, re-upsert the pipelines, and every status callback still goes to `SITE_BASE_URL`. Nothing errors, so they go digging in the wrong layer.

Either honour it everywhere:

```python
return settings.OCW_STUDIO_PIPELINE_API_URL or (
    DEV_OCW_STUDIO_API_URL if is_dev() else settings.SITE_BASE_URL
)
```

or say "only when `OCW_STUDIO_ENVIRONMENT=dev`" in the setting description and in `.env.example`, and use an example value that looks local.

content_sync/utils.py
umar8hassan: Worth deciding now what a base URL with a path prefix should do, since this one is described as a "Base URL" and that invites someone to put a prefix on it.

The two callers disagree. `common/resources.py:113` joins with `urljoin(base, "api/...")`, `common/steps.py:337` joins with `base.rstrip("/") + "/api/..."`, and `urljoin` drops the prefix unless the base ends in a slash:

| base | urljoin | rstrip |
| --- | --- | --- |
| `https://example.edu/studio` | `https://example.edu/api/...` | `https://example.edu/studio/api/...` |
| `https://example.edu/studio/` | `https://example.edu/studio/api/...` | `https://example.edu/studio/api/...` |

Pre-existing for `SITE_BASE_URL`, but newly reachable. The result is the webhook resource 404ing while the curl step works, which is a miserable thing to debug. Normalising here, always returning with a trailing slash, would settle it in one place.

content_sync/utils.py
umar8hassan: Five, unless I miscounted. `e2e_test_site_pipeline.py:64` still builds its own on this branch:

```python
CLI_ENDPOINT_URL = f" --endpoint-url {DEV_ENDPOINT_URL}" if is_dev() else ""
```

Harmless today, since nothing reads that constant and the real `aws` calls in that file already go through `get_cli_endpoint_url()`. But it looks live, so it will send the next person down the wrong path. Worth deleting along with its `DEV_ENDPOINT_URL` import.

content_sync/utils.py
umar8hassan: Nit: this goes straight into `sh -exc` bodies. The value is ours so I am not worried about injection, but a trailing newline or space, which is easy to pick up from a k8s `stringData:` block, splits the `aws` command line and surfaces as a baffling shell error inside a Concourse task. A `.strip()` in `get_s3_endpoint_url()` would cover it.

content_sync/utils_test.py
umar8hassan: The test never renders a pipeline. It calls the same three accessors that are already parametrised a few lines up, and checks their overrides won. `e2e_test_site_pipeline.py:64` still hardcodes `DEV_ENDPOINT_URL` on this branch, and this test passes.

The `get_common_pipeline_vars()` part is inert too. The autouse `required_concourse_settings` fixture already sets every URL it reads, and `RESOURCE_BASE_URL_DRAFT`/`_LIVE` default to `""`.

I'd suggest building the definitions (`ThemeAssetsPipelineDefinition`, `SitePipelineDefinition`, `UnpublishedSiteRemovalPipelineDefinition`, `S3BucketSyncPipelineDefinition`), dumping them to JSON, and asserting that `"10.1.0." not in rendered`.

.env.example
umar8hassan, line 55: [Nit] pytest `env` list in `pyproject.toml` already blanks `AWS_ENDPOINT_URL`. Without it, the tests fail in the compose stack. We should add the two lines to avoid failures:

```
"AWS_S3_ENDPOINT_URL=",
"OCW_STUDIO_PIPELINE_API_URL=",
```

.env.example
umar8hassan: If this setting stays dev only, this example works against it. A public https URL reads like a production address, so someone will try it in one. An address from the k8s local-dev stack would make the scope obvious. See the note on `get_ocw_studio_api_url`.

content_sync/pipelines/definitions/concourse/theme_assets_pipeline.py
umar8hassan: This and `s3_bucket_sync_pipeline.py:19` call the helper at import time, which freezes the setting before any test can override it. That is the mechanism behind the `.env` leak I flagged on `.env.example`.

`remove_unpublished_sites.py` and `site_pipeline.py` already call it inside `__init__`. Moving these two there makes them overridable and removes the whole class of problem.

content_sync/pipelines/definitions/concourse/remove_unpublished_sites.py
umar8hassan: [nit] unused variable. `__init__` calls `get_cli_endpoint_url()` again at line 95 and the task bodies use that local.

main/settings.py
umar8hassan, line 470: django-storages builds these as `{AWS_S3_URL_PROTOCOL or "https:"}//{domain}/{key}`, and `AWS_S3_URL_PROTOCOL` is not declared, so Django never reads it. On a local stack serving plain HTTP every media URL comes out `https://` with no setting to fix it. Declare it alongside these, or note the https requirement in `.env.example`.

main/settings.py
umar8hassan: Nit: `AWS_S3_ENDPOINT_URL=rustfs:9000`, with the scheme left off, is accepted here and only fails at the first boto3 call as `ValueError: Invalid endpoint`. A scheme check next to the `OCW_STUDIO_USE_S3` block below would catch the typo at startup. Empty string is already handled, it falls through to the dev fallback.

main/s3_utils.py
umar8hassan: Nit: the address on the next line is the same one as `content_sync/constants.py:DEV_ENDPOINT_URL`. I realise the import cannot go the other way, since `content_sync.utils` imports this module, but `main/constants.py` exists and `content_sync/constants.py` could re-export from there. One place to change instead of two, or four counting both test files.

The function docstring above also still says "connecting to Minio if the environment is dev", which is now only half the story.

umar8hassan (approved):
Mostly looks good 👍. Left a few comments to look at before merging.

also, `e2e_test_site_pipeline.py:64` still builds its own cli endpoint url:
```python
CLI_ENDPOINT_URL = f" --endpoint-url {DEV_ENDPOINT_URL}" if is_dev() else ""
```

gumaerc:
@umar8hassan Thanks for the review, this is ready for another look.

gumaerc:
@umar8hassan I see now that you approved this and just wanted those changes made, so I'll merge it shortly.

main/settings.py
```diff
@@ -443,6 +443,48 @@
...
+    get_string(
+        name="AWS_S3_ENDPOINT_URL",
+        default=None,
+        description=(
+            "S3 API endpoint to use instead of AWS, for local S3 emulation "
+            "(Minio, RustFS). Also read by django-storages for the default "
+            "file storage backend."
+        ),
+    )
+    or None
```
umar8hassan: `or None` catches `""` but not `"   "`, and whitespace is the likelier typo of the two since it comes from a stray newline in a k8s `stringData:` block rather than from someone typing nothing.

It renders a command whose subcommand gets eaten:

```
whitespace only: get_s3_endpoint_url() = '   '
       rendered: 'aws s3 --endpoint-url     sync s3://a/ s3://b/'
     shell argv: ['aws', 's3', '--endpoint-url', 'sync', 's3://a/', 's3://b/']
```

So `--endpoint-url` swallows `sync` and the job dies with an `aws` usage error nowhere near the actual cause.

`(... or "").strip() or None` covers both. While you are there, a scheme check would turn the other quiet one into a startup error, `rustfs:9000` is accepted today and only surfaces as `ValueError: Invalid endpoint: rustfs:9000` at the first boto3 call. Trailing slashes are genuinely fine, boto3 takes them.

content_sync/utils_test.py
```diff
@@ -409,3 +438,83 @@ def test_is_extra_theme(settings, theme_slug, extra_themes, expected):
...
+        test_bucket="test_bucket",
+        ocw_hugo_themes_branch="main",
+    )
+    bucket_sync = S3BucketSyncPipelineDefinition(
+        import_bucket="import_bucket",
+        storage_bucket="storage_bucket",
+    )
+
+    assert "objectstore.test:9000" in theme_assets.json()
+    assert "objectstore.test:9000" in bucket_sync.json()
```
umar8hassan, line 520: We should add another line for `remove_unpublished`:

```suggestion
    assert "objectstore.test:9000" in bucket_sync.json()

    remove_unpublished = UnpublishedSiteRemovalPipelineDefinition()
    assert "objectstore.test:9000" in remove_unpublished.json()
```

umar8hassan (approved):
still looks good 👍

